PR #245, opened 2014-01-22
Retrieve and use tagged filters

nicoschoenmaker:
Allows bundles to add their own filters.

Using tags for the filters feels nicer in the Symfony ecosystem. It also allows setter injection of application specific dependencies that one might want to use in a filter.

The author of a bundle can say whether a filter should be enabled by default
```
   my_bundle.valid_date_filter:
        class: Hostnet\MyBundle\Filter\ValidDateFilter
        arguments: ["@doctrine.orm.entity_manager"]
        tags:
            - { name: doctrine.filter, filter_name: page_valid_date, enabled: true }
```

The user of the bundle can, of course, ignore the authors suggestion and do something different in ```config.yml```:
```
                filters:
                    page_valid_date:
                        enabled: false
```

The current implementation requires some adjustment in Doctrine\ORM as well
- ```Doctrine\ORM\EntityManager``` gets a setFilters method.
- In ```Doctrine\ORM\Query\FilterCollection``` $enabledFilters and $filtersState become protected.

Adding filters as is done right now is still supported. Tips & tricks to improve this PR are appreciated :)

### DIFF
--- a/DependencyInjection/Compiler/FilterConfigurationPass.php
+++ b/DependencyInjection/Compiler/FilterConfigurationPass.php
@@ -7,6 +7,7 @@
  * view the LICENSE file that was distributed with this source code.
  */
 namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
+
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Reference;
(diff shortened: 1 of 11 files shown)


kimhemsoe:
Is this still alive or should i close it?

stof:
I'm clearly not a fan of this implementation, because it goes against the way the ORM manages filter instances (the ORM instantiates them itself, it does not allow injecting instances configured externally). Hacking the ORM to use DI for filters will likely create hard-to-debug issues and will hurt maintenance of the bundle (because we will do things unsupported by the ORM team)

kimhemsoe:
Closing then

PR #275, opened 2014-02-28
added stacktraces for queries

dmecke:
I've added a stacktrace to the query profiler. This can help to find out why a specific query has been executed. This one depends on pull request https://github.com/doctrine/dbal/pull/535 in doctrine/dbal as it fetches the stacktrace information.

### DIFF
--- a/Controller/ProfilerController.php
+++ b/Controller/ProfilerController.php
@@ -65,33 +65,4 @@ class ProfilerController extends ContainerAware
             'query' => $query,
         ));
     }
-
-    /**
-     * Renders the stacktrace for the given token and query.
-     *
-     * @param string $token
-     * @param string $connectionName
-     * @param integer $query
-     *
-     * @return Response
-     */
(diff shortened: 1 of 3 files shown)


stof:
closing as the DBAL PR has been rejected

sustmi:
Just for reference, this feature is already available in Doctrine bundle: https://symfony.com/blog/new-in-doctrinebundle-1-11-0#store-backtraces-when-profiling-queries (PR: https://github.com/doctrine/DoctrineBundle/pull/954) .

You just need to enable it by setting `doctrine.dbal.profiling_collect_backtrace: true` (I set it to `'%kernel.debug%'` instead because it is the default value of `profiling` setting that this feature depends on).

Web search got me here and from this conversation it looks like it may not be available yet.

stof:
Well, this conversation is from 2015. There are other PRs newer than that.

sustmi:
I know that it's an old conversation but googling "

PR #285, opened 2014-04-29
Ability to move generated Entities to Entity subdirectory

insekticid:
there is no way to move generated entities to subdirectory 
e.g. from Acme/DemoBundle/Entity to Acme/DemoBundle/Entity/Database1


kimhemsoe:
I think we should avoid exposing this option, it may give the user more troubles then it is worth.

PR #288, opened 2014-05-05
Fix loading of cache providers

zerrvox:
Fix a typo and prefixed cache services with the entity manager name which is needed if you have more than one.

I'm not really sure why the cache_provider needs to be prefixed with doctrine_cache.providers. instead of just using it as it is. This would allow to user other Cache adapter services that implement the Doctrine Cache interface.


guilhermeblanco:
Already implemented as I could see in latest master.
Closing as this PR is already in.

PR #290, opened 2014-05-19
Adds profiling ability for SQL Server

flip111:
in case of a SELECT the query is actually executed again but with a lot more information. In case of a non-SELECT query only the execution plan is made available (no costs and additional info)

### DIFF
--- a/Controller/ProfilerController.php
+++ b/Controller/ProfilerController.php
@@ -57,7 +57,7 @@ class ProfilerController extends ContainerAware
         try {
             if ($connection->getDatabasePlatform() instanceof SQLServerPlatform) {
                 if (strtoupper(substr($query['sql'], 0, 6)) === 'SELECT') {
-                    $sql = 'SET STATISTICS PROFILE ON; ' . $query['sql'] . ' SET STATISTICS PROFILE OFF;';
+                    $sql = 'SET STATISTICS PROFILE ON; ' . $query['sql'] . '; SET STATISTICS PROFILE OFF;';
                 } else {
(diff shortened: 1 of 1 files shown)


Controller/ProfilerController.php
```diff
@@ -54,8 +55,19 @@ public function explainAction($token, $connectionName, $query)
         /** @var $connection \Doctrine\DBAL\Connection */
         $connection = $this->container->get('doctrine')->getConnection($connectionName);
         try {
-            $results = $connection->executeQuery('EXPLAIN '.$query['sql'], $query['params'], $query['types'])
-                ->fetchAll(\PDO::FETCH_ASSOC);
+            if ($connection->getDatabasePlatform() instanceof SQLServerPlatform) {
```
stof: @flip111 can you actually at least move this to a private method ?

flip111: Sure that is a minor change, but will that be enough for a merge? If it's not enough i will wait until there are more comments.

Controller/ProfilerController.php
```diff
@@ -54,8 +55,19 @@ public function explainAction($token, $connectionName, $query)
...
+                if (strtoupper(substr($query['sql'], 0, 6)) === 'SELECT') {
```
deeky666: Please use `stripos($query['sql'], 'SELECT') === 0`

stof:
It took time, but here we go, this is in now. Thank you very much @flip111.

flip111:
Thanks for merge

PR #293, opened 2014-05-30
db profiller template fix

covus-filipe-garcia:
fixing the error 
Key "default" for array with keys "" does not exist in @Doctrine/Collector/db.html.twig at line 209


stof:
Great to finally have an information about the case where this happens: when a manager does not have any mapped entities (actually, any loaded mapped entity). Nobody was giving this information yet when reporting the issue.

stof:
Thanks for fixing this bug @covus-filipe-garcia.

Issue #223, opened 2013-10-24
Introduce Container aware entity listener resolver in order to use entity listener as service


egeloen:
Hey guys!

I'm currently moving some logic inside the new entity listeners introduced in Doctrine 2.4. The bundles supports this feature by allowing to define our own entity listener resolver service in the configuration. The issue about that is you have to redefine all the logic in order to be able to use entity listener as service.

For now, to be able to use entity listener as service, I have created a `ContainerAwareEntityListenerResolver` which maps class names to service id through a compiler pass. When it resolves the entity listener, if the class names exists, I just return the service else it fallbacks on the Default entity listener resolver behavior.

My question is: do you think this feature can be natively supported by the bundle? If yes, I can share my work in a PR.

pkruithof:
I've implemented this in #443

PR #298, opened 2014-06-13
Added support for tagging entity listeners, without the need to map the listener

kimhemsoe:
Adds support for tagging entity listeners, defining entity and method. 

Every listener service is created when the entitymanager is created for now. A solution could be marking the services lazy as suggested by @Ocramius and @Stof in https://github.com/doctrine/DoctrineBundle/issues/223

### DIFF
--- a/DependencyInjection/Compiler/EntityListenerPass.php
+++ b/DependencyInjection/Compiler/EntityListenerPass.php
@@ -48,7 +48,7 @@ class EntityListenerPass implements CompilerPassInterface
                     continue;
                 }
 
-                if (isset($attributes['entity'])) {
+                if (isset($attributes['entity']) && isset($attributes['type'])) {
                     $this->attachToListener($container, $name, $id, $attributes);
                 }
 
--- a/DependencyInjection/Compiler/EntityListenerPass.php
+++ b/DependencyInjection/Compiler/EntityListenerPass.php
@@ -71,7 +71,7 @@ class EntityListenerPass implements CompilerPassInterface
         $serviceDef = $container->getDefinition($id);
 
         $args = array(
-	        $attributes['entity'],
+            $attributes['entity'],
             $serviceDef->getClass(),
         );
 
--- a/DependencyInjection/Compiler/EntityListenerPass.php
+++ b/DependencyInjection/Compiler/EntityListenerPass.php
@@ -48,8 +48,41 @@ class EntityListenerPass implements CompilerPassInterface
                     continue;
                 }
 
+                if (isset($attributes['entity'])) {
+                    $this->attachToListener($container, $name, $id, $attributes);
+                }
+
                 $container->getDefinition($resolver)->addMethodCall('register', array(new Reference($id)));
             }
         }
+    }
+
+    private function attachToListener(ContainerBuilder $container, $name, $id, array $attributes)
+    {
+        $listenerId   = sprintf('doctrine.orm.%s_listeners.attach_entity_listeners', $name);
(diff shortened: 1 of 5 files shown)


austinh:
Would love to be able to use this feature!

hacfi:
:+1: Trying to implement this feature myself I came across this PR. Would love to see this get merged.

kimhemsoe:
@hacfi I never really tested this in a real app. It should work but im not 100% sure. Have you or anyone else tried it ?

hacfi:
@kimhemsoe Debugging this at the moment..looks like the entity listener isn’t added properly to the class metadata.

kimhemsoe:
@hacfi I have done a rebase to make it bit easier.

DependencyInjection/Compiler/EntityListenerPass.php
```diff
@@ -48,8 +48,41 @@ public function process(ContainerBuilder $container)
...
+
+        if (!$container->hasDefinition($listenerId)) {
+            return;
+        }
+
+        $serviceDef = $container->getDefinition($id);
+
+        $args = array(
+            $serviceDef->getClass(),
+            $attributes['entity'],
```
hacfi: this should be

```php
            $attributes['entity'],
            $serviceDef->getClass(),
```

See https://github.com/doctrine/doctrine2/blob/master/docs/en/changelog/migration_2_5.rst#events-add-api-to-programatically-add-event-listeners-to-entity

kimhemsoe: Fixed

DependencyInjection/DoctrineExtension.php
```diff
@@ -429,6 +429,12 @@ protected function loadOrmEntityManager(array $entityManager, ContainerBuilder $
             ));
         }
 
+	    if (version_compare(Version::VERSION, "2.5.0-DEV") < 0) {
+		    $listenerId = sprintf('doctrine.orm.%s_listeners.attach_entity_listeners', $entityManager['name']);
+		    $listenerDef = $container->setDefinition($listenerId, new Definition('%doctrine.orm.listeners.attach_entity_listeners.class%'));
+            $listenerDef->addTag('doctrine.event_listener', array('event' => 'loadClassMetadata'));
+        }
```
hacfi: I don’t understand this part..using ORM 2.5.0 the listener is missing in the `attachToListener` method.

hacfi: Thanks for the quick update @kimhemsoe !

I think this should be 

```php
if (version_compare(Version::VERSION, "2.5.0-DEV") < 0) {
```
instead. `doctrine.orm.listeners.attach_entity_listeners.class` isn’t available before 2.5 !?

kimhemsoe: You are correct, i was a bit to fast in the rebase.

hacfi:
@kimhemsoe Perfect...works as expected now.

As @chasen suggested the name `event` is more consistent that `type`. What do you think?

kimhemsoe:
The name "type" is from doctrine ORM, but "event" would be fine with me.

hacfi:
Great! Thanks so much for this @kimhemsoe...I almost started writing a custom annotation for this :stuck_out_tongue_winking_eye:

kimhemsoe:
You welcome @hacfi and thanks for the help. I will merge when travis and Scrutinizer returns green.

hacfi:
@kimhemsoe Updating my current project to get the latest master..so useful I don’t want to wait :)

PR #310, opened 2014-07-09
Correcting the output message

mikeSimonson:
It used to take the database name for the connection name which can be very confusing

### DIFF
--- a/Command/CreateDatabaseDoctrineCommand.php
+++ b/Command/CreateDatabaseDoctrineCommand.php
@@ -56,7 +56,12 @@ EOT
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $connection = $this->getDoctrineConnection($input->getOption('connection'));
+        $connectionName = $input->getOption('connection');
+        if (empty($connectionName) === true) {
+            $connectionName = $this->getContainer()->get('doctrine')->getDefaultConnectionName();
+        }
+        $connection = $this->getDoctrineConnection($connectionName);
+
         $ifNotExists = $input->getOption('if-not-exists');
 
         $params = $connection->getParams();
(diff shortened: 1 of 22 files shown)


Command/CreateDatabaseDoctrineCommand.php
```diff
@@ -55,7 +55,13 @@ protected function configure()
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $connection = $this->getDoctrineConnection($input->getOption('connection'));
+        if (empty($input->getOption('connection'))) {
+            $connectionName = $this->getContainer()->get('doctrine')->getDefaultManagerName();
```
stof: Well, there is an issue currently indeed. But your fix is wrong. Connection names are manager names are not guaranteed to match (even if the shortcut notation uses ``default`` as name for both of them).

The issue is that currently, the message says `` Created database for connection named `symfony_test2` `` but the name it displays is the DB name, not the connection name

mikeSimonson:
I think, it makes more sense now.

Command/CreateDatabaseDoctrineCommand.php
```diff
@@ -56,7 +56,15 @@ protected function configure()
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $connection = $this->getDoctrineConnection($input->getOption('connection'));
+        if (empty($input->getOption('connection'))) {
```
stof: this does not work on PHP 5.3 and 5.4. Only variables can be passed to ``empty``

mikeSimonson: I added a test case that would have catch that issue I think.

mikeSimonson:
Any idea what the issue is with the prefer lowest ?

mikeSimonson:
@stof gentle ping :gift:

mikeSimonson:
@stof ping

xabbuh:
The test is run with symfony/console `2.0.4`, but you need at least `2.0.5` which fixed the Composer autoloading (see symfony/symfony@fc97472f64cd5494f1ec481cec762817cf638464).

Tests/Command/CreateDatabaseDoctrineTest.php
```diff
@@ -0,0 +1,84 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: mike
+ * Date: 30/07/14
+ * Time: 16:48
+ */
```
stof: this is the wrong header

Tests/Command/CreateDatabaseDoctrineTest.php
```diff
@@ -0,0 +1,84 @@
...
+
+        $command = $application->find('doctrine:database:create');
+        $command->setContainer($this->getMockContainer($connectionName, $params));
+
+        $commandTester = new CommandTester($command);
+        $commandTester->execute(
+            array_merge(array('command' => $command->getName()))
+        );
+
+        $this->assertRegExp("/Created database \"$dbName\" for connection named $connectionName/", $commandTester->getDisplay());
```
stof: you could use ``assertContains`` instead of usign a regex. It will be simpler (and the failure message will be more understandable in case of failure)

Tests/Command/CreateDatabaseDoctrineTest.php
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    protected function getMockContainer($connectionName, $params=null)
```
stof: should be private

mikeSimonson:
@stof Can someone tell me why it keeps installing phpunit dev-master instead of 4.6 ?

guilhermeblanco:
@mikeSimonson that's because the dependency is `~4.6` which means `at least 4.6 or higher`... and also because the minimum-stability is dev, it does install dev-master

PR #312, opened 2014-07-11
Options to simplify the life of those who use provisioning systems

mikeSimonson:
One just make sure that the database is not created if it doesn't exist
and the other make sure to delete it before to recreate it.


mikeSimonson:
close because the feature has been inserted by another commit

Issue #60, opened 2012-04-18
A way to enable auto_mapping with multipe Entity Manager


goetas:
Currently i have 50 active bundles in my project.

10 bundles do not require database interaction.
39 bundles uses the default entity manager, that uses the default dbal connection.
1 bundle use the 'custom' entity manager, that uses the 'custom' dbal connection.

to do this, i need to disable auto_mapping and specify all mappings.

<pre>
doctrine:
    dbal:
        default_connection: default
        connections:
                default:
                    driver:   %database_driver%
                   ....
                custom:
                    driver:   pdo_pgsql
                    ...
    orm:
        entity_managers:
            default:
                mappings:
                    Bundle1: ~
                    Bundle2: ~
                    Bundle3: ~
                    Bundle4: ~
                    Bundle4: ~
                    ........
            superdpi:
                mappings: 
                    SuperDpiCoreBundle: ~
                connection: custom     
</pre>

this is very annoying... 
currently i've resolved the problem in this way:
1)

<pre>
doctrine:
    orm:
        entity_managers:
            default:
                auto_mapping : true
            superdpi:
                mappings: 
                    SuperDpiCoreBundle: ~
                connection: custom     
</pre>

2)
removed auto_mapping check inside `DoctrineExtension::loadOrmEntityManager()`
3)
removed auto mapping expansion inside `AbstractDoctrineExtension::loadMappingInformation()`
4) added `DoctrineExtension::collectAutoMappings()` 

<pre>
protected function collectAutoMappings (array $entityManagerConfigs, ContainerBuilder $container){
    $defindedBundles = array_keys($container->getParameter('kernel.bundles'));
    foreach ($entityManagerConfigs as $name => $entityManager) {
        if(isset($entityManager['auto_mapping']) && $entityManager['auto_mapping']){
            foreach ($defindedBundles as $bundle){
                foreach ($entityManagerConfigs as $name2 => $entityManager2) {
                    if($entityManager2['mappings'][$bundle]){
                        continue 2;
                    }
                }
                $entityManagerConfigs[$name]['mappings'][$bundle]['mapping']=true;
            }
        }
    }
    return $entityManagerConfigs;
}
</pre>

5) added 

<pre>
$config['entity_managers'] = $this->collectAutoMappings($config['entity_managers'], $container);
</pre>

inside `DoctrineExtension::ormLoad()` just before

<pre>
foreach ($config['entity_managers'] as $name => $entityManager) {
[...]

PR #321, opened 2014-08-11
Added a way to enable auto_mapping option using multipe entity managers

goetas:
Finally, I fond some time to create this PR. This should close https://github.com/doctrine/DoctrineBundle/issues/60

I've also added  a small test. Feedbacks are welcome!

(if merged, this will require some lines on SF documentation...)


phackwer:
People, I can't find how to activate that for Symfony use. Any suggestions? Is there any new docs about how to do it?

PR #340, opened 2014-10-16
Fix Twig error when mappingErrors is unpopulated

Kaivosukeltaja:
Undefined collector.mappingErrors['default'] could result in Twig error 'Key "default" does not exist as the array is empty'.

### DIFF
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -206,7 +206,7 @@
                     <tr>
                         <td>{{ class }}</td>
                         <td>
-                            {% if collector.mappingErrors[manager][class] is defined %}
+                            {% if collector.mappingErrors[manager] is defined and collector.mappingErrors[manager][class] is defined %}
                                 <ul>
                                     {% for error in collector.mappingErrors[manager][class] %}
                                         <li>{{ error }}</li>


Th3Mouk:
Hello @stof,

 I've the same problem, i think it's probably a configuration mistake from me can u or an other contributor help us.

Thanks !

stof:
@Th3Mouk do you have the Twig C extension enabled or no ?

Th3Mouk:
@stof no at all ^^ I just discovered its existence pretty cool

stof:
@Th3Mouk OK, one possible reason eliminated (the code triggering this exception is precisely the piece of code reimplemented in C in Twig).

Can you tell me which version of Twig you are using ? ``composer show -i twig/twig`` will show you this info.

Th3Mouk:
``` bash
name     : twig/twig
descrip. : Twig, the flexible, fast, and secure template language for PHP
keywords : templating
versions : * v1.18.0
type     : library
license  : BSD-3-Clause
source   : [git] https://github.com/twigphp/Twig.git 4cf7464348e7f9893a93f7096a90b73722be99cf
dist     : [zip] https://api.github.com/repos/twigphp/Twig/zipball/4cf7464348e7f9893a93f7096a90b73722be99cf 4cf7464348e7f9893a93f7096a90b73722be99cf
names    : twig/twig

autoload
psr-0
Twig_ => lib/

requires
php >=5.2.4

```

stof:
Closing as I merged #293 instead (it was the first submission of the fix), after finally getting some information to reproduce the issue

PR #347, opened 2014-10-28
mappingErrors key checking in db.html.twig

pae23:
This error was returned without the patch:
'Key "default" does not exist as the array is empty in @Doctrine/Collector/db.html.twig at line 209'


Th3Mouk:
#340

stof:
Closing as duplicate

PR #350, opened 2014-11-28
Fix minor typo

goabonga:
Fix minor typo


stof:
Good catch, thanks @goabonga.

PR #360, opened 2014-12-30
Replaced configuration documentation with a link to the up to date config documentatio...

eriwin:
...n in the Symfony docs

There are two versions of the DoctrineBundle documentation with different information.

https://github.com/symfony/symfony-docs/blob/master/reference/configuration/doctrine.rst
https://github.com/doctrine/DoctrineBundle/blob/master/Resources/doc/configuration.rst

The one in the symfony-docs is kept up to date so I added a link to them. 

Please see for more information the issue I made in the symfony-docs repo: https://github.com/symfony/symfony-docs/issues/4037


weaverryan:
+1 from me. Even though DoctrineBundle is its own entity, we've always had a reference section on Symfony.com, and I don't see a reason to change this (we also have a book chapter, cookbook entries, etc).

Ocramius:
I'm strongly -1 on this. This kind of change makes versioning of documentation a mess, and also makes it harder to keep track of configuration changes in docs (while we can currently enforce the author of a PR to also update the docs).

If the symfony-docs team wants to help maintaining this documentation, then fine, but the docs should not be moved outside THIS repository.

Ocramius:
Cases such as symfony/symfony-docs#4719 actually support what I just said btw.

weaverryan:
Hey @Ocramius!

That makes sense. But we should also solve some practical issues.

1) This repo gets less attention from people than the symfony-docs repo. So, docs here don't naturally get as much "love". You can see that when comparing the entity listeners docs in this repo (https://github.com/doctrine/DoctrineBundle/blob/master/Resources/doc/entity-listeners.rst) with the one from symfony-docs https://github.com/symfony/symfony-docs/blob/master/cookbook/doctrine/event_listeners_subscribers.rst.

2) The reference section is not rendered on symfony.com. If we decide to store the reference section here, we'll need to fix that (or link to the GitHub rendered version of the configuration from here: http://symfony.com/doc/current/reference/index.html - probably a decent option).

3) If we keep the reference section here, then logically, we could move the book and cookbook entry here too. But then we wouldn't be able to integrate them like they are now on symfony.com (i.e. as book and cookbook entries). Of course, we could just keep the configuration section here and other stuff there, I'm not married to being consistent, but wanted to point this out :).

From all of this, I would propose keeping the reference section in this repo, but only that. It would give you a lot of the pros you want (e.g. enforcing docs changes to the reference section) and would be easy to do on symfony.com. It would minimize the negatives of (1) and (3).

Thanks!

Ocramius:
> 1) This repo gets less attention from people than the symfony-docs repo. So, docs here don't naturally get as much "love". You can see that when comparing the entity listeners docs in this repo (https://github.com/doctrine/DoctrineBundle/blob/master/Resources/doc/entity-listeners.rst) with the one from symfony-docs https://github.com/symfony/symfony-docs/blob/master/cookbook/doctrine/event_listeners_subscribers.rst.

Agree that this repo is less visible than the symfony-docs repo, but upgrades in here are still to be kept in sync with the bundled documentation, therefore (assuming that maintainers follow good common sense) the docs will maybe not be as nice as the symfony ones, but at least always upgraded and historically consistent.

 > 2) The reference section is not rendered on symfony.com. If we decide to store the reference section here, we'll need to fix that (or link to the GitHub rendered version of the configuration from here: http://symfony.com/doc/current/reference/index.html - probably a decent option

Fair enough. You could probably also just import files from there into here, assuming RST docs are compatible

 > 3) If we keep the reference section here, then logically, we could move the book and cookbook entry here too. But then we wouldn't be able to integrate them like they are now on symfony.com (i.e. as book and cookbook entries). Of course, we could just keep the configuration section here and other stuff there, I'm not married to being consistent, but wanted to point this out :).

cookbook is more about integration rather than pure documentation. Yes, there is some redundancy, but I'm fine with leaving the cookbook on symfony's docs (because symfony changes also affect the cookbooks then).

A base reference of what is in this repo should stay in here. How to use it properly can live wherever it wants on the interwebs IMO.

weaverryan:
Great, thanks Marco - I'll close the reference issue (well, I'll first make sure others agree) on symfony-docs and get a plan for linking over here or importing the RST.

Cheers!

Ocramius:
@weaverryan awesome! Thanks for your understanding :-)

PR #362, opened 2014-12-31
Delete quote for database name

Prims47:
I delete quote for $name because in_array($name, $connection->getSchemaManager()->listDatabases()) is always false.


kimhemsoe:
Old and no response.

PR #364, opened 2015-01-02
Make data collector test compatible with different ORM versions

deeky666:
This is the first step in making the test suite go green again. It makes the `Doctrine\Bundle\DoctrineBundle\Tests\DataCollector\DoctrineDataCollectorTest` compatible with ORM < 2.5.0-DEV again.
The test suite is still failing because of recent Symfony deprecations. Also composer is still installing ORM `2.4.x-dev` while it should install `dev-master` IMO.
Both issues should be dealt with separately though.

### DIFF
--- a/Tests/DataCollector/DoctrineDataCollectorTest.php
+++ b/Tests/DataCollector/DoctrineDataCollectorTest.php
@@ -39,9 +39,11 @@ class DoctrineDataCollectorTest extends \PHPUnit_Framework_TestCase
             ->method('getConfiguration')
             ->will($this->returnValue($config));
 
-        $config->expects($this->once())
-            ->method('isSecondLevelCacheEnabled')
-            ->will($this->returnValue(false));
+        if (method_exists($config, 'isSecondLevelCacheEnabled')) {
+            $config->expects($this->once())
+                ->method('isSecondLevelCacheEnabled')
+                ->will($this->returnValue(false));
+        }
 
         $metadatas = array(
             $this->createEntityMetadata(self::FIRST_ENTITY),
@@ -68,6 +70,7 @@ class DoctrineDataCollectorTest extends \PHPUnit_Framework_TestCase
     {
         $metadata = new ClassMetadataInfo($entityFQCN);
         $metadata->name = $entityFQCN;
+        $metadata->reflClass = new \ReflectionClass('stdClass');
 
         return $metadata;
     }
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,6 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <phpunit colors="true" bootstrap="vendor/autoload.php">
+    <php>
+        <!-- Disable E_USER_DEPRECATED until reaching compatibility with Symfony 3.0 -->
+        <!-- php -r 'echo -1 & ~E_USER_DEPRECATED;' -->
+        <ini name="error_reporting" value="-16385"/>
+    </php>
+
     <testsuites>
         <testsuite name="DoctrineBundle for the Symfony Framework">
             <directory>./Tests</directory>


deeky666:
Concerning the Symfony deprecations, see: https://github.com/symfony/symfony/commit/bd8531d2d893d154f2f639657afa828254c52d97

Tests/DataCollector/DoctrineDataCollectorTest.php
```diff
@@ -68,6 +74,7 @@ private function createEntityMetadata($entityFQCN)
     {
         $metadata = new ClassMetadataInfo($entityFQCN);
         $metadata->name = $entityFQCN;
+        $metadata->reflClass = new \ReflectionClass('\stdClass');
```
Ocramius: Don't prepend with `\` when in strings

Tests/DataCollector/DoctrineDataCollectorTest.php
```diff
@@ -39,9 +40,14 @@ public function testCollectEntities()
             ->method('getConfiguration')
             ->will($this->returnValue($config));
 
-        $config->expects($this->once())
-            ->method('isSecondLevelCacheEnabled')
-            ->will($this->returnValue(false));
+        if (version_compare(Version::VERSION, '2.5.0-DEV') < 0) {
```
Ocramius: Consider checking if a particular method exists on the ORM configuration class

stof:
I know about this. But I don't want to switch to it yet, because I prefer supporting all current stable versions of Symfony rather than avoiding deprecation notices in the not yet released version of Symfony (especially given the warnings are not finalized yet either). I think we should run tests on the stable branch of Symfony (what about running it with stable releases everywhere, to get the same thing than most of our users ?)

Regarding the ORM 2.4 vs master, this is because the ORM 2.4 allows DBAL master, while the ORM master only allows DBAL 2.5. Composer cannot use the latest version of both packages, so it had to downgrade one of them.

deeky666:
@stof if you want to support all current stable versions of Symfony, we will have to maintain legacy XML service definition files (which are affected by the deprecations) and have them being loaded conditionally for different Symfony versions. Not sure if that is a desired option. Otherwise we'll have to bump requirements as soon as 2.7 goes stable.
Concerning testing against stable Symfony versions, isn't that counterproductive as we won't be able to see dev changes that might affect DoctrineBundle that easily anymore before it's actually going stable?

stof:
@deeky666 once 2.7 goes stable (i.e. in 4 months), I will indeed bump the requirement. But until then, I don't want to drop the support of the current LTS just to be able to ignore deprecation warnings in the not-released next LTS.

the solution might b to change the error_reporting level in the phpunit.xml file if we want to keep using dev versions in our tests.

deeky666:
@stof like this? Excluded `E_USER_DEPRECATED` for now. Mergable?

phpunit.xml.dist
```diff
@@ -1,6 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <phpunit colors="true" bootstrap="vendor/autoload.php">
+    <php>
+        <!-- Disable E_USER_DEPRECATED until reaching compatibility with Symfony 3.0 -->
+        <!-- php -r 'echo -1 & ~E_USER_DEPRECATED;' -->
+        <ini name="error_reporting" value="-16385"/>
```
Ocramius: Please let's not do this, really, as we will also be unable to find deprecations in other software.
Instead, lock onto <2.7 until 2.7 is out.

stof: IMO, we should not lock <2.7 in the composer.json, otherwise people wanting to try Symfony 2.7 already (ignoring the deprecation warnings for now) would be unable to use the latest version of the bundle (they will be forced to use the 1.3.0 release, and they will then complain about bugs already fixed in newer versions of the bundle once we merge some bug fixes).
Otherwise, we should exclude 2.7 only in Travis (or in dev requirements)

Ocramius: Ooooooooooooor fix symfony 2.7 :P

deeky666: Okay I won't change anything in this PR until we agree on one solution. Thinking about the possibilities again I vote for locking on `"symfony/dependency-injection": "~2.2,<2.7"` in `dev` only. This way no notices are silenced (also possible future ones), users can still use DoctrineBundle with latest Symfony version and tests pass. This option is the best compromise IMO.

PR #366, opened 2015-01-04
Close connections and clear managers on shutdown

stof:
When shutting down the kernel, the connections should be closed to avoid leaking connections during tests, which could reach the connection limit. The entity managers should also cleared to give more chances to entities to be garbage collected.
Closes https://github.com/Behat/Symfony2Extension/issues/52

### DIFF
--- a/DoctrineBundle.php
+++ b/DoctrineBundle.php
@@ -23,6 +23,7 @@ use Doctrine\ORM\Proxy\Autoloader;
 use Symfony\Component\Console\Application;
 use Symfony\Component\DependencyInjection\Compiler\PassConfig;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\IntrospectableContainerInterface;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 use Symfony\Bridge\Doctrine\DependencyInjection\CompilerPass\DoctrineValidationPass;
 use Symfony\Bridge\Doctrine\DependencyInjection\CompilerPass\RegisterEventListenersAndSubscribersPass;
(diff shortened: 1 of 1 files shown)


Ocramius:
Is this just a matter of tests?
Can a small integration test be written?

What if I want to re-use a connection from a kernel after the kernel has been shut down (for example for late batch-processing of a 201 or 202 request?)

stof:
@ocramius in a web context, the kernel is never shut down. A kernel.terminate action happens after handling the request and sending the response, but not after shutdown

The main action of shutting down the kernel is removing the reference to the container, so that booting again recreates a new container (your existing container would ultimately be garbage collected but not immediately because of circular object graphs).

The goal of the shutdown is to reset your kernel instance entirely. the only use case I see is testing (to isolate your tests while keeping the same kernel reference).

driesvints:
Just a FYI but I also encountered this when running unit tests. I'm not using Symfony but I'm using Laravel 4.2 + Doctrine (DBAL 2.5.1). A simple connection `close` and entity manager `clear`, like @stof suggested above solved the problem for me as well.

oleg-andreyev:
:+1: need this.

stof:
I'm going to merge this one. Writing an integration test for the shutdown would be a pain

peterrehm:
This change creates issues on my side as per the following simplified test.

````php
    public function testCloseRoute()
    {
        $task = $this->createTask();
        $this->callRoute('task_close', array('id' => $task->getId()));
        $this->assertEquals(302, $this->client->getResponse()->getStatusCode());
        $this->client->followRedirect();
        $this->assertEquals(200, $this->client->getResponse()->getStatusCode());
        $this->removeEntity($task);
    }
````

After this change I am getting the error:

````
1) TaskBundle\Tests\Controller\TaskTest::testCloseRoute
Doctrine\ORM\ORMInvalidArgumentException: A detached entity was found during removed TaskBundle\Entity\Task@000000000d592bdf00000001251a1486
````

A $em->merge() is not working as the entity is stated as not found. What would you recommend as workaround?

PR #367, opened 2015-01-09
[WIP] Bump Symfony requirement to 2.6

deeky666:
Following https://github.com/doctrine/DoctrineBundle/pull/364 this PR is intended to reach compatibility with Symfony >= 2.6.
It makes DoctrineBundle use the new service factory definition as the current one is deprecated. It also bumps the Symfony requirements to `~2.6` and allows `E_USER_DEPRECATED` to trigger errors in the test suite again.
I marked it as WIP until the next Symfony LTS 2.7 is released as we won't bump the Symfony requirements before.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,16 +25,16 @@
     ],
     "require": {
         "php": ">=5.3.2",
-        "symfony/framework-bundle": "~2.2",
+        "symfony/framework-bundle": "~2.6",
         "doctrine/dbal": "~2.3",
         "jdorn/sql-formatter": "~1.1",
-        "symfony/doctrine-bridge": "~2.2",
+        "symfony/doctrine-bridge": "~2.6",
         "doctrine/doctrine-cache-bundle": "~1.0"
     },
     "require-dev": {
         "doctrine/orm": "~2.3",
-        "symfony/yaml": "~2.2",
-        "symfony/validator": "~2.2",
+        "symfony/yaml": "~2.6",
+        "symfony/validator": "~2.6",
         "twig/twig": "~1",
         "satooshi/php-coveralls": "~0.6.1",
         "phpunit/phpunit": "~3.7",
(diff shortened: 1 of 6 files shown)


flip111:
I think this is a bad idea because symfony 2.3 has LTS support until 2016. It would be possible if DoctrineBundle worked with different branches to support symfony 2.3 as well.

kimhemsoe:
@deeky666 Maybe we should wait with this until at least some time after ORM 2.5.

stof:
I agree that dropping the support for the LTS now is not the best idea. It would force us to continue maintaining an older branch for bugfixes (not new features) for people using the LTS. I suggest delaying this a bit

composer.json
```diff
@@ -25,16 +25,16 @@
     ],
     "require": {
         "php": ">=5.3.2",
-        "symfony/framework-bundle": "~2.2",
+        "symfony/framework-bundle": "~2.6",
         "doctrine/dbal": "~2.3",
         "jdorn/sql-formatter": "~1.1",
-        "symfony/doctrine-bridge": "~2.2",
+        "symfony/doctrine-bridge": "~2.6",
```
stof: you don't have anything depending on symfony/doctrine-bridge 2.6. You have things depending on symfony/dependency-injection 2.6

danez:
Maybe fixing it like in the AsseticBundle, would be an alternative? https://github.com/fabpot/AsseticBundle/commit/5be3b6fff6b5ac9078d09dfde4250d1853ea2988

xabbuh:
@danez see #381

deeky666:
So I guess this PR is not required anymore then, right?

fabpot:
This one should be closed now as #381 does the same but also keep BC with 2.3.

deeky666:
thx @fabpot !

PR #371, opened 2015-01-16
on line 696 printf is changed to sprintf

nimasdj:
on line 696 printf() is changed to sprintf(). printf() causes to output `doctrine_cache.providers.metadata_cache` as string on screen.

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -693,7 +693,7 @@ class DoctrineExtension extends AbstractDoctrineExtension
     {
         if (!empty($driverMap['cache_provider'])) {
             $aliasId = $this->getObjectManagerElementName($driverName);
-            $serviceId = printf('doctrine_cache.providers.%s', $driverMap['cache_provider']);
+            $serviceId = sprintf('doctrine_cache.providers.%s', $driverMap['cache_provider']);
 
             $container->setAlias($aliasId, new Alias($serviceId, false));
 


deeky666:
Could need a test to verify and avoid regressions.

nimasdj:
Yes already tested and confirmed.

nimasdj:
See https://github.com/doctrine/DoctrineCacheBundle/issues/38

pkruithof:
I think he meant a unit test.

deeky666:
yep :)

nimasdj:
You can do that?

pkruithof:
Something like this:

```php
$this->assertTrue($container->hasAlias('doctrine.orm.metadata_cache'));
$alias = $container->getAlias('doctrine.orm.metadata_cache');
$this->assertEquals('doctrine_cache.providers.metadata_cache', (string) $alias);
```

nimasdj:
I am not familiar with test unit. Can you please give some advices how the complete test class should be and how to run it?

nimasdj:
@pkruithof I am reading phpunit doc to learn it. What the class method of the snippet you suggested above should be?

nimasdj:
And how to inject $container object into test class?

stof:
@nimasdj we already have lots of tests for the DependencyInjection layer in https://github.com/doctrine/DoctrineBundle/tree/master/Tests/DependencyInjection.
What you need to do is adding a test in DoctrineExtensionTest to cover the case of the cache_provider option

nimasdj:
@stof
Just by adding the snippet of pkruithof to
testDependencyInjectionConfigurationDefaults() method?

nimasdj:
@stof @pkruithof 
I added this to 
testDependencyInjectionConfigurationDefaults() method, 
```php
$this->assertTrue($container->hasAlias('doctrine.orm.metadata_cache'));
$alias = $container->getAlias('doctrine.orm.metadata_ca');
$this->assertEquals('doctrine_cache.providers.metadata_ca', (string) $alias);
```
and I expected to get error for typos in 'cache' word, I used:
```
php phpunit.phar --strict --debug --process-isolation DoctrineExtensionTest
```
but I got:
```
Time: 2.41 seconds, Memory: 4.25Mb
OK (15 tests, 0 assertions)
```

what wrong I did?

pkruithof:
I think you're not executing that test, when I add those rules to that `testDependencyInjectionConfigurationDefaults`, I get the following error:

```
$ ./vendor/bin/phpunit --filter testDependencyInjectionConfigurationDefaults
PHPUnit 3.7.38 by Sebastian Bergmann.

Configuration read from /Users/peter/projects/DoctrineBundle/phpunit.xml.dist

F

Time: 139 ms, Memory: 8.75Mb

There was 1 failure:

1) Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection\DoctrineExtensionTest::testDependencyInjectionConfigurationDefaults
Failed asserting that false is true.
```

nimasdj:
@pkruithof 
I globaly installed phpunit.phar in /usr/local/bin/phpunit
then under doctrine-bundle where the xml file is located I use:
```
$ phpunit --filter testDependencyInjectionConfigurationDefaults
```
but I get error for cannot find /vendor/autoload.php file. please advice how should I use that?

nimasdj:
That snippet you gave what does actually test? test aliases are set or not?
and how to test that sprintf() too?

pkruithof:
Install the dependencies with composer. The dev dependencies contain phpunit, which creates an executable in the `vendors/bin` folder.

@stof: I looked at the other issue (the service alias that is not set, causing [this error](https://github.com/doctrine/DoctrineCacheBundle/issues/38#issuecomment-69631811)), and it seems to me that either the alias name should be changed from `doctrine.orm.<cache_provider>` to `doctrine.orm.<entity_manager>_<cache_provider>`, or that this alias should be added, besides the current alias. I prefer the first one as it creates 1 working alias, but I'm not sure if it will break things when `doctrine.orm.<cache_provider>` is no longer available. Your thoughts please?

pkruithof:
@nimasdj that snippet tests that a specific alias is created. It does not specifically test the `sprintf` call, but rather the result it yields. Currently that alias is not set (due to `printf` instead of `sprintf`), which causes this test to fail. When changing the call to `sprintf` the test passes as the alias is created.

pkruithof:
I took the liberty to fix both problems myself, so you can see what I mean @nimasdj:

https://github.com/doctrine/DoctrineBundle/pull/372

nimasdj:
@pkruithof I did change it to sprint() but with your snippet I am still getting error. can you show your full test method?

guilhermeblanco:
Merged in #372. Closing this one

PR #372, opened 2015-01-16
Fixed cache provider alias in extension

pkruithof:
Replacement for doctrine/DoctrineBundle#371

Regarding the error in [this post](https://github.com/doctrine/DoctrineCacheBundle/issues/38#issuecomment-69631811), it seems that the extension currently has 2 problems:

1. A `printf` call that should be `sprintf`
2. An alias with the wrong name

It seems to me that either the alias name should be changed from `doctrine.orm.<cache_provider>` to `doctrine.orm.<entity_manager>_<cache_provider>`, or that this alias should be added, besides the current alias. I prefer the first one as it creates 1 working alias, but I'm not sure if it will break things when `doctrine.orm.<cache_provider>` is no longer available.

This PR uses the first solution, if it should be solved differently let me know.

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -692,8 +692,8 @@ class DoctrineExtension extends AbstractDoctrineExtension
     protected function loadCacheDriver($driverName, $entityManagerName, array $driverMap, ContainerBuilder $container)
     {
         if (!empty($driverMap['cache_provider'])) {
-            $aliasId = $this->getObjectManagerElementName($driverName);
-            $serviceId = printf('doctrine_cache.providers.%s', $driverMap['cache_provider']);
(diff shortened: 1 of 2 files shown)


nimasdj:
@pkruithof with your two changes applied then how should this configuration be?

```yaml
doctrine_cache:
  providers:
    metadata_cache:
      aliases: [doctrine.orm.default_metadata_cache]
      file_system:
        directory: "%kernel.cache_dir%/doctrine/metadata"

  orm:
    metadata_cache_driver:
      cache_provider: metadata_cache
```

pkruithof:
Like that, but without the need for the `aliases:` part.

nimasdj:
@pkruithof it seems another bug in symfony is found: https://github.com/symfony/AsseticBundle/issues/340

PR #377, opened 2015-01-19
Fixed the reStructuredText syntax errors of the docs

javiereguiluz:
We're integrating the DoctrineBundle into symfony.com and we're facing some issues because of the following RST syntax errors:

```
doctrine-bundle/en/master/installation.rst:20: ERROR: Error in "code-block" directive:
1 argument(s) required, 0 supplied.

doctrine-bundle/en/master/installation.rst:36: ERROR: Error in "code-block" directive:
1 argument(s) required, 0 supplied.
```

In addition, an exception was triggered because there was no doctree in the docs. We've created the `index.rst` to solve this issue.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -1,6 +1,15 @@
 DoctrineBundle
 ==============
 
+Doctrine DBAL & ORM Bundle for the Symfony Framework.
+
+Symfony does not force or suggest a specific persistence solution on its users.
+That's why this bundle was removed from the core of the Symfony framework and
+moved into an independent repository.
+
+Since Doctrine2 will still be a major player in the Symfony world, the bundle is
+actively maintained by developers in the Doctrine and Symfony communities.
+
 .. toctree::
 
     installation
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -1,14 +1,8 @@
 DoctrineBundle
 ==============
 
-Doctrine DBAL & ORM Bundle for the Symfony Framework.
-
-Symfony does not force or suggest a specific persistence solution on its users.
-That's why this bundle was removed from the core of the Symfony framework and
-moved into an independent repository.
-
-Since Doctrine2 will still be a major player in the Symfony world, the bundle is
-actively maintained by developers in the Doctrine and Symfony communities.
+Integrates Doctrine's ORM and DBAL projects into Symfony applications. It provides
+configuration options, console commands and even a web debug toolbar collector.
 
 .. toctree::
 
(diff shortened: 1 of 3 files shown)


Resources/doc/installation.rst
```diff
@@ -17,7 +19,7 @@ Installation
 
 Add the following snippets to "deps" files:
 
-.. code-block::
+.. code-block:: ini
```
stof: please drop the deps file instructions instead of fixing them. they make no sense anymore (the only reason they are still there is because this doc was not used anywhere for real)

Resources/doc/installation.rst
```diff
@@ -1,14 +1,16 @@
...
+Because Symfony does not want to force or suggest a specific persistence
+solutions on the users this bundle was removed from the core of the Symfony
+framework. Doctrine2 will still be a major player in the Symfony world and the
+bundle is maintained by developers in the Doctrine and Symfony communities.
 
-IMPORTANT: This bundle is developed for Symfony 2.1 and up. For Symfony 2.0 applications the DoctrineBundle
-is still shipped with the core Symfony repository.
+IMPORTANT: This bundle is developed for Symfony 2.1 and up. For Symfony 2.0
+applications the DoctrineBundle is still shipped with the core Symfony
+repository.
```
stof: this warning should be removed:

- Symfony 2.0 does not exist anymore
- the latest versions are not compatible with 2.1 anymore actually. They require 2.3+.
- handling compatibility between versions is best left to Composer anyway

javiereguiluz:
Installation instructions have been updated to comply with modern Symfony practices.

Resources/doc/index.rst
```diff
@@ -0,0 +1,17 @@
+DoctrineBundle
+==============
+
+Doctrine DBAL & ORM Bundle for the Symfony Framework.
+
+Symfony does not force or suggest a specific persistence solution on its users.
+That's why this bundle was removed from the core of the Symfony framework and
```
stof: This sentence should probably be removed, because probably only Symfony early adopters will know that this bundle used to be in the Symfony namespace (the renaming in symfony 2.1 is already old history)

stof:
Thank you @javiereguiluz.

Resources/doc/installation.rst
```diff
@@ -1,43 +1,43 @@
...
-Installation
-------------
+Open a command console, enter your project directory and execute the following
+command to download the latest stable version of this bundle:
 
-1. Old deps and bin/vendors way
+.. code-block:: bash
 
-Add the following snippets to "deps" files:
+    $ composer require doctrine/doctrine-bundle
```
WouterJ: you forgot the version

javiereguiluz: Yes, but the version is no longer required in Composer, right?

PR #378, opened 2015-01-19
Ensure DoctrineDataCollector manages ClassMetadataInfo

farconada:
A bug in Doctrine 2.5-dev could load transient classes so the DataCollector fails. A simple check could avoid it

### DIFF
--- a/DataCollector/DoctrineDataCollector.php
+++ b/DataCollector/DoctrineDataCollector.php
@@ -70,6 +70,9 @@ class DoctrineDataCollector extends BaseCollector
 
             /** @var $class \Doctrine\ORM\Mapping\ClassMetadataInfo */
             foreach ($factory->getLoadedMetadata() as $class) {
+                if (!$class instanceof ClassMetadataInfo) {
+                    continue;
+                }
                 if (!isset($entities[$name][$class->getName()])) {
                     $classErrors = $validator->validateClass($class);
                     $entities[$name][$class->getName()] = $class->getName();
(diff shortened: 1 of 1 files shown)


xabbuh:
Is it not possible to fix the `getLoadedMetadata()` method instead?

farconada:
sure and it has to, but it's slower than it. And it's a condition that should be matched

stof:
Please report the bug to the Doctrine ORM, as it is something which must be fixed before the 2.5 stable release

DataCollector/DoctrineDataCollector.php
```diff
@@ -70,6 +70,9 @@ public function collect(Request $request, Response $response, \Exception $except
 
             /** @var $class \Doctrine\ORM\Mapping\ClassMetadataInfo */
             foreach ($factory->getLoadedMetadata() as $class) {
+                if (!$class instanceof \Doctrine\ORM\Mapping\ClassMetadataInfo) {
```
stof: please add a use statement instead of using the FQCN inline

guilhermeblanco:
Fixed in doctrine/common as of https://github.com/doctrine/common/commit/56360cc13bfd13893731c6c4d1d82005a61d7694

Ocramius:
@guilhermeblanco re-opening: please also reference tests for that hotfix.

stof:
Closing this PR as there is nothing to do in DoctrineBundle

PR #379, opened 2015-01-20
Fixes and tweaks for the documentation

javiereguiluz:
This all started as a minor fix of the syntax issue found on lines 336 and 337. But then, I added some more fixes:

  * Option listings use the recommended definition list format instead of the regular bullet list.
  * Some values in the YAML code listings have been wrapped in quotes. Otherwise, they completely disable YAML syntax highlighting.


stof:
Thank you @javiereguiluz.

PR #381, opened 2015-01-25
made the factories compatible with all Symfony versions (2.3 to 2.7)

fabpot:
This PR removes the deprecation notices when using DoctrineBundle with the upcoming Symfony 2.7.

ping @Ocramius

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,21 @@ php:
   - 5.6
   - hhvm
 
-before_script: 
-    - composer install --dev --prefer-source
+matrix:
+  include:
+    - php: 5.3
+      env: deps=low
+
+env:
+  global:
+    - deps=no
+
+before_install:
+  - composer self-update
+
+install:
+  - if [ "$deps" = "no" ]; then composer --prefer-source install; fi;
+  - if [ "$deps" = "low" ]; then composer --prefer-source --prefer-lowest --prefer-stable update; fi;
 
 script:
   - ./vendor/bin/phpunit -v --coverage-clover ./build/logs/clover.xml
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
     ],
     "require": {
         "php": ">=5.3.2",
-        "symfony/framework-bundle": "~2.2",
+        "symfony/framework-bundle": "~2.3",
         "doctrine/dbal": "~2.3",
         "jdorn/sql-formatter": "~1.1",
         "symfony/doctrine-bridge": "~2.2",
@@ -35,7 +35,7 @@
         "doctrine/orm": "~2.3",
         "symfony/yaml": "~2.2",
         "symfony/validator": "~2.2",
-        "twig/twig": "~1",
+        "twig/twig": "~1.10",
         "satooshi/php-coveralls": "~0.6.1",
         "phpunit/phpunit": "~3.7",
         "phpunit/php-code-coverage": "~1.2",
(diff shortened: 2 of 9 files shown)


fabpot:
By the way, I've removed the error_reporting setting for PHPUnit as it's not needed anymore.

fabpot:
I've also updated the `.travis.yml` file to test the bundle with the lowest deps and the highest deps.

I've fixed the `composer.json` dep versions to make tests pass.
Twig needs to be in 1.4 as this is the first version with a valid `composer.json` file.
Symfony cannot be set to 2.0 as the `composer.json` file was not valid back then.
`Doctrine\ORM\EntityManagerInterface` cannot be used in tests as it was introduced in Doctrine 2.4.

.travis.yml
```diff
@@ -1,14 +1,28 @@
...
+    global:
+        - deps=no
+
+before_install:
+    - composer self-update
+
+install:
+    - if [ "$deps" = "no" ]; then composer --prefer-source install; fi;
+    - if [ "$deps" = "low" ]; then composer --prefer-source --prefer-lowest --prefer-stable update; fi;
+    - if [ "$deps" = "high" ]; then composer --prefer-source update; fi;
```
danez: As far as i can see $deps = "no" and $deps = "high" make no difference as the library has no composer.lock anyway. So both will install a mix of 2.7.x-dev and dev-master currently.

stof: deps=no or deps=high is doing exactly the same here, given there is no lock file. There is no need to duplicate things IMO

composer.json
```diff
@@ -35,7 +35,7 @@
         "doctrine/orm": "~2.3",
         "symfony/yaml": "~2.2",
         "symfony/validator": "~2.2",
-        "twig/twig": "~1",
+        "twig/twig": "~1.4",
```
stof: technically, we need at least 1.10 as we use the namespaced notation when registering the profiler template in the service tag

.travis.yml
```diff
@@ -1,14 +1,28 @@
...
-  - 5.4
-  - 5.5
-  - 5.6
-  - hhvm
+matrix:
+    include:
+        - php: 5.3
+        - php: 5.4
+        - php: 5.5
+        - php: 5.6
```
stof: I suggest keeping these builds in the ``php`` section instead of including them explicitly in the matrix here. It would be easier to understand for people reading this file by matching the common cases (Symfony does it differently only because @nicolas-grekas tried to run the component builds first at one point, which required building the whole matrix manually)

stof:
@fabpot it looks like you pushed a similar commit for the .travis.ml file to the master branch of the bundle, thus creating conflicts here. Can you rebase ?

stof:
Actually, I will rever the commit in master as it breaks the Travis setup, and I will let this PR provide the working config at the end.

.travis.yml
fabpot: fixed

.travis.yml
fabpot: fixed

composer.json
fabpot: fixed

fabpot:
Comments taken into account, tests green.

Ocramius:
Thanks, @fabpot!

fabpot:
@Ocramius Can you release a new version of DoctrineBundle? That would help us a lot by removing some deprecation notices.

Ocramius:
That's up to @kimhemsoe

PR #383, opened 2015-01-29
New feature: Execute query on db profiler

sergio-toro:
My coworker (@AMoQu) and I have written a little feature to execute logged queries on the db profiler.

### DIFF
--- a/Controller/ProfilerController.php
+++ b/Controller/ProfilerController.php
@@ -57,7 +57,7 @@ class ProfilerController extends ContainerAware
         $connection = $this->container->get('doctrine')->getConnection($connectionName);
 
         try {
-            $results =  $connection->executeQuery($query['sql'], $query['params'], $query['types'])
+            $results = $connection->executeQuery($query['sql'], $query['params'], $query['types'])
                 ->fetchAll(\PDO::FETCH_ASSOC);
         } catch (\Exception $e) {
             return new Response('This query cannot be executed.');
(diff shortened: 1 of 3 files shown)


Controller/ProfilerController.php
```diff
@@ -35,6 +35,49 @@ class ProfilerController extends ContainerAware
...
+        $profile = $profiler->loadProfile($token);
+        $queries = $profile->getCollector('db')->getQueries();
+
+        if (!isset($queries[$connectionName][$query])) {
+            return new Response('This query does not exist.');
+        }
+
+        $query = $queries[$connectionName][$query];
+        if (!$query['explainable']) {
+            return new Response('This query cannot be explained.');
```
stof: the message is wrong, as this is not about explaining

stof:
this allows to replay INSERT, DELETE and UPDATE queries, which is an issue

Controller/ProfilerController.php
AMoQu: Do you think we should remove the if condition or just change message text to something like 'This query cannot be executed.'

AMoQu:
@stof We have patched it to only allow SELECT queries. It is okay now?

stof:
The issue is that this does not allow all select queries anymore (not all select queries have the SELECT at the beginning, for instance when using CTEs in PostgreSQL)

stof:
I'm -1 on this. Making it possible to re-execute queries is dangerous as it is quite impossible to figure out whether the query is safe or no to re-execute.

sergio-toro:
@stof Yep, I know it could be dangerous, but for our use case it just saves tons of work for debugging queries because we have more than 10 different connections.
Please just give it a chance and think about it.

guilhermeblanco:
-1 on this... closing

Issue #384, opened 2015-02-02
Doctrine2 create database throws exception even if --if-not-exists option is passed


ggirtsou:
Command
```
$  php app/console --env=test doctrine:database:create --if-not-exists
```

Exception:
```
Could not create database for connection named `testSymfony`
An exception occurred while executing 'CREATE DATABASE `testSymfony`':

SQLSTATE[HY000]: General error: 1007 Can't create database 'testSymfony'; database exists
```

```
$ composer show -i

doctrine/annotations                     v1.2.3             Docblock Annotations Parser
doctrine/cache                           v1.4.0             Caching library offering an object-oriented API for many cache backends
doctrine/collections                     v1.2               Collections Abstraction library
doctrine/common                          v2.4.2             Common Library for Doctrine projects
doctrine/data-fixtures                   v1.0.0             Data Fixtures for all Doctrine Object Managers
doctrine/dbal                            v2.5.1             Database Abstraction Layer
doctrine/doctrine-bundle                 dev-master 698b10b Symfony DoctrineBundle
doctrine/doctrine-cache-bundle           v1.0.1             Symfony2 Bundle for Doctrine Cache
doctrine/doctrine-fixtures-bundle        v2.2.0             Symfony DoctrineFixturesBundle
doctrine/doctrine-migrations-bundle      dev-master 81575a4 Symfony DoctrineMigrationsBundle
doctrine/inflector                       v1.0.1             Common String Manipulations with regard to casing and singular/plural rules.
doctrine/instantiator                    1.0.4              A small, lightweight utility to instantiate objects in PHP without invoking their constructors
doctrine/lexer                           v1.0.1             Base library for a lexer that can be used in Top-Down, Recursive Descent Parsers.
doctrine/migrations                      dev-master 1ac14fa Database Schema migrations using Doctrine DBAL
doctrine/orm                             v2.4.7             Object-Relational-Mapper for PHP

```

Problem is that database name is quoted because there is no path so this check:

```php
$shouldNotCreateDatabase = $ifNotExists && in_array($name, $tmpConnection->getSchemaManager()->listDatabases());
```
fails. `$ifNotExists` is `true` but `testSymfony` (with backticks) does not exist in array. `testSymfony` (without backticks) exists though.

There is also a second bug, `sprintf()` is missing the database name:

```php
$output->writeln(sprintf('<info>Database for connection named <comment>%s</comment> already exists. Skipped.</info>'));
```

should be:
```php
[...]

PR #385, opened 2015-02-02
#384 Removes backtick from db name to get --if-not-exists option working

ghost:
Also adds missing argument to `sprintf()`

### DIFF
--- a/Command/CreateDatabaseDoctrineCommand.php
+++ b/Command/CreateDatabaseDoctrineCommand.php
@@ -79,10 +79,11 @@ EOT
 
         $error = false;
         try {
-            $shouldNotCreateDatabase = $ifNotExists && in_array($name, $tmpConnection->getSchemaManager()->listDatabases());
+            $cleanName = str_replace('`', '', $name);
+            $shouldNotCreateDatabase = $ifNotExists && in_array($cleanName, $tmpConnection->getSchemaManager()->listDatabases());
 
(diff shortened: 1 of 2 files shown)


xabbuh:
@ggirtsou Looks like you need to rebase here.

Command/CreateDatabaseDoctrineCommand.php
```diff
@@ -79,10 +79,11 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         $error = false;
         try {
-            $shouldNotCreateDatabase = $ifNotExists && in_array($name, $tmpConnection->getSchemaManager()->listDatabases());
+            $cleanName = str_replace('`', '', $name);
```
deeky666: TBH I would rather consider not explicitly quoting the database identifier at all. IMO this is DBAL/user responsibility anyways. Some platforms even REQUIRE unquoted database identifiers like SQL Anywhere.

deeky666: Also quoting the identifier [here](https://github.com/ggirtsou/DoctrineBundle/blob/master/Command/CreateDatabaseDoctrineCommand.php#L77) does not always enclose the database name in backticks. This is only working on MySQL but not for other platforms!

guilhermeblanco:
Closing as this is a duty of DBAL to do, not the bundle

PR #386, opened 2015-02-04
Add support to set PostgreSQL application_name

davividal:
This PR is dependant upon https://github.com/doctrine/dbal/pull/798 .
PostgreSQL allows the user to set the application_name is connecting
to database. It is useful for monitoring purposes.
Currently one could just manually add 'application_name=foo' to DSN
by appending it to dbname, like:

> dbname:   %database_name% application_name=foo

It works but having a parameter eases setting it.

### DIFF
--- a/Resources/config/schema/doctrine-1.0.xsd
+++ b/Resources/config/schema/doctrine-1.0.xsd
@@ -47,6 +47,7 @@
         <xsd:attribute name="port" type="xsd:string" />
         <xsd:attribute name="user" type="xsd:string" />
         <xsd:attribute name="password" type="xsd:string" />
+        <xsd:attribute name="application_name" type="xsd:string" />
         <xsd:attribute name="path" type="xsd:string" />
         <xsd:attribute name="unix-socket" type="xsd:string" />
         <xsd:attribute name="memory" type="xsd:string" />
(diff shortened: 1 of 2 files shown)


Resources/config/schema/doctrine-1.0.xsd
```diff
@@ -47,6 +47,7 @@
         <xsd:attribute name="port" type="xsd:string" />
         <xsd:attribute name="user" type="xsd:string" />
         <xsd:attribute name="password" type="xsd:string" />
+        <xsd:attribute name="application_name" type="xsd:string" />
```
xabbuh: Should probably be `application-name` to be consistent with the other options.

davividal: I thought about that, but the GUC variable and the DSN part is called `application_name`. Wouldn't it make more sense for the users?

stof: the XML format always uses dashes in places where YAML uses underscores (it is normalized by the config system). What makes sense for users is to respect the XML convention (they will get autocompletion based on the XSD in their IDE anyway)

davividal:
@xabbuh  @stof what about now?

stof:
Well, it cannot be merged here until the DBAL team merges the feature in DBAL itself

kimhemsoe:
Thanks @davividal

PR #404, opened 2015-03-11
Enabling profiling should not automatically enable logging

tucksaun:
This can be particularly problematic in consumers as it gets Stopwatch injected.
Loggers can usually be flushed and so you can avoid leaking memory, but with Stopwatch it's not possible.


lyrixx:
:+1:

stof:
Good catch, thanks @tucksaun.

PR #406, opened 2015-03-18
Changed default naming strategy to underscore

hason:
This PR partially fixes the problem with case-sensitive identifiers http://dev.mysql.com/doc/refman/5.0/en/identifier-case-sensitivity.html


stof:
this is a BC break as it will change name for all existing projects

stof:
What we could do though is to change it in the standard-edition for new Symfony projects

guilhermeblanco:
Closing as it will never be merged due to BC break. =)

PR #409, opened 2015-04-22
Improve the rendering of the query explanation for Postgres

stof:
Postgres does not return a table for the explanation but a list of lines displaying the explanation in a formatted text.

The switch between the formats is based on whether each result row contains several columns (table formatting) or a single column (text formatting)


stof:
This is best reviewed by looking at the diff without space changes, as I changed the indentation: https://github.com/doctrine/DoctrineBundle/pull/409/files?w=1

PR #413, opened 2015-04-24
Fix html syntax

### LINKED ISSUE #261: Invalid html in profiler: <code><pre>
block level element `<pre>` must not be within inline `<code>` element. But the profiler generates such html code in https://github.com/doctrine/DoctrineBundle/blob/master/Resources/views/Collector/db.html.twig#L103

Also it uses invalid `target-data-id` html attribute. Why not use `data-target-id` which would be an allowed data-attribute.

Tobion:
Fixes #261 and some cleanup

### DIFF
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -65,7 +65,19 @@
             'query': app.request.query.get('query')
         })) }}
     {% else %}
-        {{ block('queries') }}
+        <div>
+            <style scoped="scoped">
+                h3 {
+                    margin-bottom: 0px;
+                }
+
+                pre {
+                    padding: 5px;
+                }
+            </style>
+
+            {{ block('queries') }}
+        </div>
     {% endif %}
 {% endblock %}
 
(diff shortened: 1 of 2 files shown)


Resources/views/Collector/db.html.twig
```diff
@@ -367,18 +379,4 @@
         }
 
     //]]></script>
-
-    <style>
```
Tobion: style not allowed in this context

Twig/DoctrineExtension.php
```diff
@@ -38,8 +38,8 @@ public function getFilters()
     {
         return array(
             new \Twig_SimpleFilter('doctrine_minify_query', array($this, 'minifyQuery')),
```
Tobion: this doesnt return html

Twig/DoctrineExtension.php
```diff
@@ -229,9 +229,6 @@ public function minifyQuery($query)
             $result = $this->composeMiniQuery($query, $keywords, $required);
         }
 
-        // Remove unneeded boilerplate HTML
-        $result = str_replace(array("<pre style='background:white;'", "</pre>"), array("<span", "</span>"), $result);
```
Tobion: leftover useless code

stof:
Thank you @Tobion.

PR #419, opened 2015-05-05
remove old and unnecessary preamble from the README

jrobeson:
I'm pretty sure information about Symfony 2.0 and 2.1 is unnecessary anymore. Nor do i think we need to explain why the DoctrineBundle was pulled out of core.


stof:
Thank you @jrobeson.

PR #421, opened 2015-05-08
Fix DisconnectedMetadataFactory::findNamespaceAndPathForMetadata

ptlis:
In cases where the class did not already exist the FQCN was being returned, it should now always return the namespace of the class.

This was causing me problems with a custom mapping driver & file locator I'm building - hopefully this is the correct fix, all tests are passing at least. 

I couldn't figure out a workable way to introduce a new test for these behaviours - any suggestions would be welcome.

### DIFF
--- a/Mapping/DisconnectedMetadataFactory.php
+++ b/Mapping/DisconnectedMetadataFactory.php
@@ -122,12 +122,20 @@ class DisconnectedMetadataFactory
         if (class_exists($all[0]->name)) {
             $r = new \ReflectionClass($all[0]->name);
             $path = $this->getBasePathForClass($r->getName(), $r->getNamespaceName(), dirname($r->getFilename()));
-        } elseif (!$path) {
+            $ns = $r->getNamespaceName();
+
+        } elseif ($path) {
+            // Get namespace by removing the last component of the FQCN
+            $nsParts = explode('\\', $all[0]->name);
+            array_pop($nsParts);
+            $ns = implode('\\', $nsParts);
+
+        } else {
             throw new \RuntimeException(sprintf('Unable to determine where to save the "%s" class (use the --path option).', $all[0]->name));
         }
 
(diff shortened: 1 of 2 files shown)


Mapping/DisconnectedMetadataFactory.php
```diff
@@ -126,8 +126,19 @@ public function findNamespaceAndPathForMetadata(ClassMetadataCollection $metadat
             throw new \RuntimeException(sprintf('Unable to determine where to save the "%s" class (use the --path option).', $all[0]->name));
         }
 
+        // When class exists use it's namespace
+        if (isset($r)) {
+            $ns = $r->getNamespaceName();
```
stof: this should be done inside the existing ``if (class_exists($all[0]->name)) {`` instead of checking whether the ``$r`` variable is set

ptlis: Thanks for the feedback - I've made the requested requested and rebased from master.

stof:
@ptlis it looks like you did a mess with a rebase here. You rebased some commits from master into your own branch, which is totally wrong.

ptlis:
Oops, should be fixed now!

stof:
@ptlis is it possible to add a test covering this ?

ptlis:
@stof yep! Is there anything else I can do to help get this PR merged?

stof:
Thanks for fixing this bug @ptlis.

ptlis:
@stof Thanks, but it looks like the tests weren't merged - was this intentional?

stof:
they have been merged. but commits were squashed together in https://github.com/doctrine/DoctrineBundle/commit/0eae6e885451c642e14444c95facafd34cf3c6fc (which is also why the PR appears as closed rather than merged). the message of the merge commit explains that the PR got squashed when merging

PR #423, opened 2015-05-22
Fix caching drivers example

gonzalovilaseca:


stof:
This is wrong. The short syntax is the existing documentation, and the expanded syntax requires an extra ``entity_managers`` level

gonzalovilaseca:
The I wonder why I'm getting ``Unrecognized option "metadata_cache_driver" under "doctrine.orm"`` when I add this to the configuration:
``
doctrine:
    orm:
        metadata_cache_driver: apc
``

PR #424, opened 2015-05-27
Fix lowsest dep

mikeSimonson:


stof:
why is there many changes not related to the title of the PR here ?

mikeSimonson:
Thanks @guilhermeblanco . Do you know why composer keeps installing phpunit master ? even when I set ~4.6 like here https://github.com/doctrine/DoctrineBundle/pull/310

mikeSimonson:
The last commit was just to make the tests pass and the tests failed only because they revealed that the composer requirement weren't good.

Is there something else that is a problem ? @stof

stof:
But all the first commits in this PR have nothing to do with the title: ``Fix lowsest dep``

mikeSimonson:
sorry I mixed up the PR.

The last commit was to fix #310 .
PR that I messed up a bit anyway.

PR #427, opened 2015-05-27
use variable name instead of hardcode name

OskarStark:


stof:
Thank you @OskarStark.

PR #430, opened 2015-05-27
Tying to fix the issue with hhvm

mikeSimonson:


mikeSimonson:
@guilhermeblanco To fix the build with hhvm, we can either way use this PR or remove the stability dev .

Is that right ? What's the best solution ?
@stof

guilhermeblanco:
@mikeSimonson if you `~4.6` does it break?

mikeSimonson:
yes because it's stil getting the code-coverage dev-master

mikeSimonson:
I start to think that the best way would be to remove the stability dev otherwise we are going to end up having to pin all the dependency it uses

guilhermeblanco:
Give it a shot and remove the `minimum-stability`... =)
IIRC, it was added to allow some specifics on doctrine/common a few months ago while we had deployed ORM and DBAL, but not Common

PR #438, opened 2015-06-19
Add entity annotation sample

davidfuhr:
Took me a while to figure out you have to define the entity listener as
annotation as well.

### DIFF
--- a/Resources/doc/entity-listeners.rst
+++ b/Resources/doc/entity-listeners.rst
@@ -31,3 +31,24 @@ entity manager it should be registered with. Example:
                 </service>
             </services>
         </container>
+
+But remember you have to register the entity listener in your entity as well:
+
+.. code-block:: php
+
+    <?php
+    // User.php
+
+    use Doctrine\ORM\Mapping as ORM;
+
+    /**
+     * @ORM\Entity
+     * @ORM\EntityListeners({"UserListener"})
+     */
+    class User
+    {
+        // ....
+    }
+
+See also http://doctrine-orm.readthedocs.org/en/latest/reference/events.html#entity-listeners for more info on entity listeners.
+


xabbuh:
Afaik this is not what you should have to do. Registering the listener should be enough.

davidfuhr:
What do you call "registering"? The annotation in the entity or the service definition?

xabbuh:
The service definition with the right tag (`doctrine.entity_listener`).

pkruithof:
I agree, the [Doctrine documentation][1] also states this. If people read about this feature on the bundle documentation it will not work when just the example is copied.

[1]: http://doctrine-orm.readthedocs.org/en/latest/reference/events.html#entity-listeners

xabbuh:
Could it be that your installed version of the DoctrineBundle simply does not contain this compiler pass (unknown tags don't trigger errors, but will silently be ignored)?

pkruithof:
No the service does gets registered with the resolver, that is not the problem. But without the annotation, there is no link between the entity class and the listener, so Doctrine cannot know where the listener can be applied to.

xabbuh:
Don't you configure that [with the entity attribute](https://github.com/doctrine/DoctrineBundle/blob/master/DependencyInjection/Compiler/EntityListenerPass.php#L51-53) of the tag?

pkruithof:
Hmm, I have not tried that, didn't know it existed (so basically, another issue). It looks like you have to define a tag for each event this way (like the `event_listener` tag), while when you tag the entity, you can use annotations like `@PrePersist` on the listener.

Anyway, the documentation could be extended with all of these options, there are a bunch of ways to do this apparently.

davidfuhr:
If I add the `entity` *and* the `event` attribute (only `entity` does not work) the `EntityListenerPass` enters the `attachToListener()` method but [fails to retrieve the listener from the conainer](https://github.com/doctrine/DoctrineBundle/blob/master/DependencyInjection/Compiler/EntityListenerPass.php#L64):

```yaml
services:
    user_listener:
        class: \UserListener
        tags:
            - { name: doctrine.orm.entity_listener, event: postLoad, entity: 'AppBundle\Entity\User' }
```

Even if this would work it only allows listening to one event. The `@EntityListeners` annotation allows listening to all entity events.

xabbuh:
Well, but if the listeners does not work as expected, it should rather be fixed than providing some documentation that just works around this.

meyerbaptiste:
Works fine since https://github.com/doctrine/DoctrineBundle/pull/298 but `doctrine/orm` must be >= 2.5.0. You can see the restriction here: [DoctrineExtension.php#L434-L438](https://github.com/doctrine/DoctrineBundle/blob/master/DependencyInjection/DoctrineExtension.php#L434-L438).

Check your `doctrine/orm` version @davidfuhr.

kimhemsoe:
Should this be merged and if yes is it ready?

mikeSimonson:
@davidfuhr Thanks merged in #576

PR #442, opened 2015-07-01
[Fix] Doctrine profiler "Expand all queries" action.

mishunika:
The "Expand all queries" button is expanding the queries from all the connections that are present in the application. However, there is a button for each connection.

This pull request fixes the issue by separating the action for each button.

### DIFF
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -258,12 +258,12 @@
 
         function expandAllQueries(button) {
             var conn = button.getAttribute('data-connection'),
-                queries = document.getElementById('queries-' + conn).getElementsByClassName('query-section'),
+                queries = document.querySelectorAll('#queries-' + conn + ' .query-section'),
                 i = queries.length,
                 action = button.getAttribute('data-action');
 
             if (action == 'expand') {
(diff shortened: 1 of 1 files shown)


Resources/views/Collector/db.html.twig
```diff
@@ -257,7 +257,8 @@
         }
 
         function expandAllQueries(button) {
-            var queries = document.getElementsByClassName('query-section'),
+            var conn = button.getAttribute('data-connection'),
+                queries = document.getElementById('queries-' + conn).getElementsByClassName('query-section'),
```
xabbuh: `getElementsByClassName()` does not work with IE before 9.0. Not sure what policy is followed for the DoctrineBundle, but probably it's better to choose an alternative solution here.

xabbuh: FYI, in the Symfony core we use `querySelectorAll()` instead (see also http://tanalin.com/en/blog/2014/03/get-elements-by-class-name-ie8/)

mishunika: Before my change, it was also `getElementsByClassName()`, and there are a few more calls. That's why I didn't bother to verify the compatibility.

Thanks, I'll rewrite it  using the best practices ;)

stof:
Thanks for fixing this bug @mishunika.

PR #443, opened 2015-07-06
Added lazy-loading entity listeners

pkruithof:
Following up on #223:

This implements lazy loading entity listeners, much like the lazy loading event listeners in Doctrine. By implementing a container-aware resolver, only the service-id's are registered instead of the actual listeners. When resolving, the service is fetched from the container.

This behaviour is opt-in by adding `lazy="true"` to the `entity_listener` tag, so existing tags won't be affected. The resolver and its test are largely copied from the one in `doctrine/orm`. Extending the default resolver was not really an option because of the private properties.

Todo/questions:
- [ ] The compiler pass depends on the resolver having a `registerService` method, which is not in the interface. We could either create a separate interface for this, as part of the bundle (since it's Symfony-specific), or check if the resolver exposes this method.

### DIFF
--- a/DependencyInjection/Compiler/EntityListenerPass.php
+++ b/DependencyInjection/Compiler/EntityListenerPass.php
@@ -52,7 +52,12 @@ class EntityListenerPass implements CompilerPassInterface
                     $this->attachToListener($container, $name, $id, $attributes);
                 }
 
-                $container->getDefinition($resolver)->addMethodCall('register', array(new Reference($id)));
+                if (isset($attributes['lazy']) && $attributes['lazy']) {
+                    $listener = $container->getDefinition($id);
+                    $container->getDefinition($resolver)->addMethodCall('registerService', array($listener->getClass(), $id));
+                } else {
+                    $container->getDefinition($resolver)->addMethodCall('register', array(new Reference($id)));
+                }
             }
         }
     }
--- a/DependencyInjection/Compiler/EntityListenerPass.php
+++ b/DependencyInjection/Compiler/EntityListenerPass.php
@@ -54,6 +54,14 @@ class EntityListenerPass implements CompilerPassInterface
 
                 if (isset($attributes['lazy']) && $attributes['lazy']) {
                     $listener = $container->getDefinition($id);
+                    if (!$listener->isPublic()) {
+                        throw new \InvalidArgumentException(sprintf('The service "%s" must be public as this entity listener is lazy-loaded.', $id));
+                    }
+
+                    if ($listener->isAbstract()) {
+                        throw new \InvalidArgumentException(sprintf('The service "%s" must not be abstract as this entity listener is lazy-loaded.', $id));
+                    }
+
                     $container->getDefinition($resolver)->addMethodCall('registerService', array($listener->getClass(), $id));
                 } else {
                     $container->getDefinition($resolver)->addMethodCall('register', array(new Reference($id)));
--- a/DependencyInjection/Compiler/EntityListenerPass.php
+++ b/DependencyInjection/Compiler/EntityListenerPass.php
@@ -14,6 +14,7 @@ namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
 
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
 use Symfony\Component\DependencyInjection\Reference;
 
 /**
@@ -40,15 +41,12 @@ class EntityListenerPass implements CompilerPassInterface
                 }
 
                 $resolverId = sprintf('doctrine.orm.%s_entity_listener_resolver', $name);
-                if ($container->hasAlias($resolverId)) {
-                    $resolverId = (string) $container->getAlias($resolverId);
-                }
 
-                if (!$container->hasDefinition($resolverId)) {
+                if (!$container->has($resolverId)) {
                     continue;
                 }
 
-                $resolver = $container->getDefinition($resolverId);
+                $resolver = $container->findDefinition($resolverId);
 
                 if (isset($attributes['entity']) && isset($attributes['event'])) {
                     $this->attachToListener($container, $name, $id, $attributes);
@@ -58,11 +56,11 @@ class EntityListenerPass implements CompilerPassInterface
                     $listener = $container->findDefinition($id);
 
                     if (!$listener->isPublic()) {
-                        throw new \InvalidArgumentException(sprintf('The service "%s" must be public as this entity listener is lazy-loaded.', $id));
+                        throw new InvalidArgumentException(sprintf('The service "%s" must be public as this entity listener is lazy-loaded.', $id));
                     }
 
                     if ($listener->isAbstract()) {
-                        throw new \InvalidArgumentException(sprintf('The service "%s" must not be abstract as this entity listener is lazy-loaded.', $id));
+                        throw new InvalidArgumentException(sprintf('The service "%s" must not be abstract as this entity listener is lazy-loaded.', $id));
                     }
 
                     $interface = 'Doctrine\\Bundle\\DoctrineBundle\\Mapping\\EntityListenerServiceResolver';
@@ -70,11 +68,11 @@ class EntityListenerPass implements CompilerPassInterface
 
                     if (substr($class, 0, 1) === '%') {
                         // resolve container parameter first
-                        $class = $container->getParameter(trim($class, '%'));
+                        $class = $container->getParameterBag()->resolveValue($resolver->getClass());
                     }
 
                     if (!is_a($class, $interface, true)) {
-                        throw new \InvalidArgumentException(
+                        throw new InvalidArgumentException(
                             sprintf('Lazy-loaded entity listeners can only be resolved by a resolver implementing %s.', $interface)
                         );
                     }
--- a/DependencyInjection/Compiler/EntityListenerPass.php
+++ b/DependencyInjection/Compiler/EntityListenerPass.php
@@ -39,21 +39,24 @@ class EntityListenerPass implements CompilerPassInterface
                     continue;
                 }
 
-                $resolver = sprintf('doctrine.orm.%s_entity_listener_resolver', $name);
-                if ($container->hasAlias($resolver)) {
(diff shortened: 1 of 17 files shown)


pkruithof:
ping @stof @beberlei

pkruithof:
ping, again @stof @beberlei

If this is not something you want, could you notify me so I can close this PR?

DependencyInjection/Compiler/EntityListenerPass.php
```diff
@@ -52,7 +52,12 @@ public function process(ContainerBuilder $container)
                     $this->attachToListener($container, $name, $id, $attributes);
                 }
 
-                $container->getDefinition($resolver)->addMethodCall('register', array(new Reference($id)));
+                if (isset($attributes['lazy']) && $attributes['lazy']) {
+                    $listener = $container->getDefinition($id);
+                    $container->getDefinition($resolver)->addMethodCall('registerService', array($listener->getClass(), $id));
```
kimhemsoe: This method is not in the interface so it will break for people tho changes the resolver.

xabbuh:
For lazily registered event listeners, you have to ensure that they are not private. Otherwise, your application can break when the resolver tries to retrieve a service from the container which has been inlined during container compilation (thus leading to an error).

pkruithof:
True, the same goes for the existing lazy loading event listeners. Is there a problem with the PR regarding this, or was this just a side comment?

DependencyInjection/Compiler/EntityListenerPass.php
pkruithof: Good point. It could be added to the interface, however I'm not sure if this would count as a BC break (classes implementing this outside of this bundle would have to be updated). 

We could also check for the existence of the method and throw an exception if the configured resolver does not. It's up to the maintainers, I think.

xabbuh:
Which other entity listeners do you mean? And yes, I would throw an exception if the service is configured as private (see [here](https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/EventDispatcher/DependencyInjection/RegisterListenersPass.php#L61-63) how this is done in Symfony).

pkruithof:
The existing *event* listeners, not entity listeners :wink: exactly what you linked to just now.

But yes, it seems the same solution is applied there, I will include it here too.

cleentfaar:
@xabbuh Any feedback on the last commit or is this PR not likely to be merged? If it is, what are your thoughts about the proposed solution for the compiler pass mentioned in the description:

>The compiler pass depends on the resolver having a registerService method, which is not in the interface. We could either create a separate interface for this, as part of the bundle (since it's Symfony-specific), or check if the resolver exposes this method.

xabbuh:
@cleentfaar I cannot say anything about how likely this is to be merged as I am not a maintainer of this bundle.

hacfi:
Would be quite useful! Any reasons not to merge this?

Mapping/ContainerAwareEntityListenerResolver.php
```diff
@@ -0,0 +1,115 @@
...
+        $className = $this->normalizeClassName(get_class($object));
+
+        $this->instances[$className] = $object;
+    }
+
+    /**
+     * @param string $className
+     * @param string $serviceId
+     */
+    public function registerService($className, $serviceId)
```
kimhemsoe: Maybe add this to a new interface and check in compiler pass. What do you think?

DependencyInjection/Compiler/EntityListenerPass.php
```diff
@@ -52,7 +52,20 @@ public function process(ContainerBuilder $container)
                     $this->attachToListener($container, $name, $id, $attributes);
                 }
 
-                $container->getDefinition($resolver)->addMethodCall('register', array(new Reference($id)));
+                if (isset($attributes['lazy']) && $attributes['lazy']) {
+                    $listener = $container->getDefinition($id);
```
xabbuh: You may want to use `findDefinition()` here.

DependencyInjection/Compiler/EntityListenerPass.php
```diff
@@ -52,7 +52,20 @@ public function process(ContainerBuilder $container)
...
+                    if (!$listener->isPublic()) {
+                        throw new \InvalidArgumentException(sprintf('The service "%s" must be public as this entity listener is lazy-loaded.', $id));
+                    }
+
+                    if ($listener->isAbstract()) {
+                        throw new \InvalidArgumentException(sprintf('The service "%s" must not be abstract as this entity listener is lazy-loaded.', $id));
+                    }
+
+                    $container->getDefinition($resolver)->addMethodCall('registerService', array($listener->getClass(), $id));
```
xabbuh: This won't work if the user defined their own value for the `doctrine.orm.entity_listener_resolver.class` parameter that isn't your custom custom container aware entity resolver.

yellow1912:
I would like to have this merged as well, so I'm also waiting on the maintainers of this to comment on this issue :)

kimhemsoe:
It seems that PR is orphaned and someone need to adopt it.

pkruithof:
No, I'm still paying attention to it, however I've not seen any comments from one of the Doctrine people in the ~1.5 years since I created this. And if it's not gonna get in, I don't want to waste my time on it.

Happy to work to get it in, if I see willingness for it :)

kimhemsoe:
@pkruithof Ohh sorry, thought it was dead since questions and comments was going answered :D

pkruithof:
No worries, I see you are a repo member. I'll look into the remaining comments!

pkruithof:
@kimhemsoe @xabbuh I applied your feedback, could you have a look?

kimhemsoe:
@pkruithof Could you rebase please? So we know it will work with master.

DependencyInjection/Compiler/EntityListenerPass.php
```diff
@@ -39,20 +39,50 @@ public function process(ContainerBuilder $container)
...
+                if ($container->hasAlias($resolverId)) {
+                    $resolverId = (string) $container->getAlias($resolverId);
                 }
 
-                if (!$container->hasDefinition($resolver)) {
+                if (!$container->hasDefinition($resolverId)) {
                     continue;
                 }
 
+                $resolver = $container->getDefinition($resolverId);
```
stof: all this could be simplified using ``$resolver = $container->findDefinition($resolverId);`` to let the container resolve aliases by itself instead of doing it manually (and using ``!$container->has($resolverId)`` in the continue block, to account for aliases)

DependencyInjection/Compiler/EntityListenerPass.php
```diff
@@ -39,20 +39,50 @@ public function process(ContainerBuilder $container)
...
                 if (isset($attributes['entity']) && isset($attributes['event'])) {
                     $this->attachToListener($container, $name, $id, $attributes);
                 }
 
-                $container->getDefinition($resolver)->addMethodCall('register', array(new Reference($id)));
+                if (isset($attributes['lazy']) && $attributes['lazy']) {
+                    $listener = $container->findDefinition($id);
+
+                    if (!$listener->isPublic()) {
+                        throw new \InvalidArgumentException(sprintf('The service "%s" must be public as this entity listener is lazy-loaded.', $id));
```
stof: I suggest using the InvalidArgumentException of the DI component instead, to be consistent with Symfony

DependencyInjection/Compiler/EntityListenerPass.php
```diff
@@ -39,20 +39,50 @@ public function process(ContainerBuilder $container)
...
+                    if ($listener->isAbstract()) {
+                        throw new \InvalidArgumentException(sprintf('The service "%s" must not be abstract as this entity listener is lazy-loaded.', $id));
+                    }
+
+                    $interface = 'Doctrine\\Bundle\\DoctrineBundle\\Mapping\\EntityListenerServiceResolver';
+                    $class = $resolver->getClass();
+
+                    if (substr($class, 0, 1) === '%') {
+                        // resolve container parameter first
+                        $class = $container->getParameter(trim($class, '%'));
```
stof: this is wrong, as the resolution logic is more complex than that. Use ``$class = $container->getParameterBag()->resolveValue($resolver->getClass());`` instead

DependencyInjection/Compiler/EntityListenerPass.php
```diff
@@ -39,20 +39,50 @@ public function process(ContainerBuilder $container)
...
+                    }
+
+                    if (!is_a($class, $interface, true)) {
+                        throw new \InvalidArgumentException(
+                            sprintf('Lazy-loaded entity listeners can only be resolved by a resolver implementing %s.', $interface)
```
stof: remaining question here: should we register it as non-lazy instead of failing the container building ?

pkruithof: I think not: if you mark a listener as lazy you specifically opt-in to certain behaviour. Silently falling back on a different solution will not only change the expected behaviour but most likely trigger a fatal error since most of the time you'll want to use this with a service that has a constructor with arguments.

pkruithof:
@kimhemsoe done

@stof I applied your feedback

yellow1912:
@pkruithof thank you for pursuing this, I think this is a much needed feature :)

mikeSimonson:
If @stof or @kimhemsoe don't mention any other problem within a week this PR I will merge it as it.
Thanks @pkruithof

yellow1912:
@pkruithof By lazy loading entity listener, we can avoid the circular reference issue when we inject services that also make use of doctrine repository into the listeners?

@mikeSimonson Seem like there is no objection from @stof and @kimhemsoe, can we have this soon please?

kimhemsoe:
@pkruithof Thanks

goetas:
great feature, but it took me one hour to realize that the `lazy` attribute has to be put on the tag, and not on the service definition! Is this feature documented? Cold not find that much about it.

SenseException:
@goetas Thank you for pointing this out. There is no documentation yet, that covers this feature.
https://www.doctrine-project.org/projects/doctrine-bundle/en/latest/entity-listeners.html
https://symfony.com/doc/master/bundles/DoctrineBundle/entity-listeners.html

Do you like to create a PR to add this feature to the documentation?

alcaeus:
Maybe this got lost due to a late merge-up or some branch problems in the Symfony docs, but there's https://symfony.com/doc/current/bundles/DoctrineBundle/entity-listeners.html#lazy-entity-listeners that clearly documents this feature ;)

PR #445, opened 2015-07-10
Update collector template for the new toolbar design

javiereguiluz:
Symfony is redesigning its toolbar for 2.8+ versions (see https://github.com/symfony/symfony/pull/15160). We need to update Doctrine toolbar panel too. @stof told me to not change the current HTML markup, so  that's why I include two different contents depending on the Symfony version.


javiereguiluz:
I'm about to update the collectors of other bundles. Therefore, it would be great if someone could validate that this strategy is valid to maintain BC. Thanks!

Resources/views/Collector/db.html.twig
```diff
@@ -2,22 +2,40 @@
...
-        <span class="sf-toolbar-status{% if 50 < collector.querycount %} sf-toolbar-status-yellow{% endif %}">{{ collector.querycount }}</span>
-        {% if collector.querycount > 0 %}
-            <span class="sf-toolbar-info-piece-additional-detail">in {{ '%0.2f'|format(collector.time * 1000) }} ms</span>
-        {% endif %}
-        {% if collector.invalidEntityCount > 0 %}
-            <span class="sf-toolbar-info-piece-additional sf-toolbar-status sf-toolbar-status-red">{{ collector.invalidEntityCount }}</span>
-        {% endif %}
-        {% if collector.cacheEnabled %}
-            <span class="sf-toolbar-info-piece-additional sf-toolbar-status sf-toolbar-status-green" title="Second level cache enabled">2l cache</span>
+        {% set is_symfony_2_8_or_higher = constant('\\Symfony\\Component\\HttpKernel\\Kernel::VERSION_ID')|default(0) >= 20800 %}
```
stof: it would be better to have a way to detect the version of the WebProfilerBundle

Resources/views/Collector/db.html.twig
```diff
@@ -25,7 +43,7 @@
         </div>
         <div class="sf-toolbar-info-piece">
             <b>Invalid entities</b>
-            <span class="sf-toolbar-status sf-toolbar-status-{{ collector.invalidEntityCount > 0 ? 'red' : 'green' }}">{{ collector.invalidEntityCount }}</span>
+            <span class="sf-toolbar-status sf-toolbar-status-{{ collector.invalidEntityCount > 0 ? 'red' }}">{{ collector.invalidEntityCount }}</span>
```
stof: ``sf-toolbar-status-`` should move inside the if then

Resources/views/Collector/db.html.twig
javiereguiluz: You are right, but I don't know how to do that :cry:

Resources/views/Collector/db.html.twig
javiereguiluz: I've reverted this change. Thanks.

Resources/views/Collector/db.html.twig
```diff
@@ -2,22 +2,44 @@
...
+
+                <span class="sf-toolbar-value">{{ collector.querycount }}</span>
+                <span class="sf-toolbar-info-piece-additional-detail">
+                    <span class="sf-toolbar-label">in</span>
+                    <span class="sf-toolbar-value">{{ '%0.2f'|format(collector.time * 1000) }}</span>
+                    <span class="sf-toolbar-label">ms</span>
+                </span>
+            {% endif %}
+        {% else %}
+            <img width="20" height="28" alt="Database" src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABQAAAAcCAYAAABh2p9gAAAAGXRFWHRTb2Z0d2FyZQBBZG9iZSBJbWFnZVJlYWR5ccllPAAAAQRJREFUeNpi/P//PwM1ARMDlcGogZQDlpMnT7pxc3NbA9nhQKxOpL5rQLwJiPeBsI6Ozl+YBOOOHTv+AOllQNwtLS39F2owKYZ/gRq8G4i3ggxEToggWzvc3d2Pk+1lNL4fFAs6ODi8JzdS7mMRVyDVoAMHDsANdAPiOCC+jCQvQKqBQB/BDbwBxK5AHA3E/kB8nKJkA8TMQBwLxaBIKQbi70AvTADSBiSadwFXpCikpKQU8PDwkGTaly9fHFigkaKIJid4584dkiMFFI6jkTJII0WVmpHCAixZQEXWYhDeuXMnyLsVlEQKI45qFBQZ8eRECi4DBaAlDqle/8A48ip6gAADANdQY88Uc0oGAAAAAElFTkSuQmCC" />
```
stof: I suggest using the svg icon on older versions too (Symfony 2.7 already uses SVG)

Resources/views/Collector/db.html.twig
```diff
@@ -2,22 +2,44 @@
...
-            <span class="sf-toolbar-info-piece-additional sf-toolbar-status sf-toolbar-status-red">{{ collector.invalidEntityCount }}</span>
-        {% endif %}
-        {% if collector.cacheEnabled %}
-            <span class="sf-toolbar-info-piece-additional sf-toolbar-status sf-toolbar-status-green" title="Second level cache enabled">2l cache</span>
+        {% set is_symfony_2_8_or_higher = constant('\\Symfony\\Component\\HttpKernel\\Kernel::VERSION_ID')|default(0) >= 20800 %}
+
+        {# the HTML markup changed in Symfony 2.8, but the original one is used for previous Symfony versions #}
+        {% if is_symfony_2_8_or_higher %}
+            {% if collector.querycount > 0 %}
+                {% set status = collector.invalidEntityCount > 0 ? 'red' : collector.querycount > 50 ? 'yellow' : '' %}
```
stof: I don't see where this variable is used

Resources/views/Collector/db.html.twig
stof: a solution would be to make the WebProfilerBundle pass a variable to this template to tell it should display the Sf 3 design. If the variable is not there, the template would assume it is the old bundle

stof: this could even work in the future for further design changes by changing the value of the variable btw

Resources/views/Collector/db.html.twig
```diff
@@ -2,22 +2,44 @@
...
+                    <path fill="#AAAAAA" d="M5,8h14c1.7,0,3-1.3,3-3s-1.3-3-3-3H5C3.3,2,2,3.3,2,5S3.3,8,5,8z M18,3.6c0.8,0,1.5,0.7,1.5,1.5S18.8,6.6,18,6.6s-1.5-0.7-1.5-1.5S17.2,3.6,18,3.6z M19,9H5c-1.7,0-3,1.3-3,3s1.3,3,3,3h14c1.7,0,3-1.3,3-3S20.7,9,19,9z M18,13.6
+                    c-0.8,0-1.5-0.7-1.5-1.5s0.7-1.5,1.5-1.5s1.5,0.7,1.5,1.5S18.8,13.6,18,13.6z M19,16H5c-1.7,0-3,1.3-3,3s1.3,3,3,3h14c1.7,0,3-1.3,3-3S20.7,16,19,16z M18,20.6c-0.8,0-1.5-0.7-1.5-1.5s0.7-1.5,1.5-1.5s1.5,0.7,1.5,1.5S18.8,20.6,18,20.6z"/>
+                </svg>
+
+                <span class="sf-toolbar-value">{{ collector.querycount }}</span>
+                <span class="sf-toolbar-info-piece-additional-detail">
+                    <span class="sf-toolbar-label">in</span>
+                    <span class="sf-toolbar-value">{{ '%0.2f'|format(collector.time * 1000) }}</span>
+                    <span class="sf-toolbar-label">ms</span>
+                </span>
```
stof: We don't display the number of mapping errors anymore. Is it expected ? If there are mapping errors, we want the user to know about it

Resources/views/Collector/db.html.twig
```diff
@@ -107,18 +147,18 @@
...
                             </div>
-                            <code id="code-{{ i }}-{{ loop.parent.loop.index }}">
-                                {{ query.sql|doctrine_pretty_query(i, loop.parent.loop.index)|raw }}
-                            </code>
+                            <div id="code-{{ i }}-{{ loop.parent.loop.index }}" style="display: none;">
+                                {{ query.sql|doctrine_pretty_query(i, loop.parent.loop.index) }}
+                            </div>
                             <span id="original-query-{{ i }}-{{ loop.parent.loop.index }}" style="display: none;">
-                                {{ query.sql|doctrine_replace_query_parameters(query.params)|raw }}
+                                {{ query.sql|doctrine_replace_query_parameters(query.params) }}
```
stof: your diff seems to include many changes which should not be in this PR (they are already in master). Can you rebase your branch to fix this ?

javiereguiluz:
@stof I like your last proposal. I'm trying to implement it in https://github.com/javiereguiluz/symfony/commit/44f101fac55cae40a19db21c34bd4119733c91ea and https://github.com/javiereguiluz/DoctrineBundle/commit/83d04a1614e442c96a148e3d241db514b66f5c86

Resources/views/Collector/db.html.twig
javiereguiluz: We can't because the SVG icon has change its style slightly (all the icons showed in the new toolbar have been slightly redesigned).

Resources/views/Collector/db.html.twig
javiereguiluz: In the new toolbar, that information is conveyed via the background color of the panel. If there is a mapping error, the Doctrine panel is displayed in a red background. And if you perform a lot of queries, the background is yellow.

Resources/views/Collector/db.html.twig
stof: but would it look bad ?

Resources/views/Collector/db.html.twig
```diff
@@ -2,22 +2,38 @@
...
-        <span class="sf-toolbar-status{% if 50 < collector.querycount %} sf-toolbar-status-yellow{% endif %}">{{ collector.querycount }}</span>
-        {% if collector.querycount > 0 %}
-            <span class="sf-toolbar-info-piece-additional-detail">in {{ '%0.2f'|format(collector.time * 1000) }} ms</span>
-        {% endif %}
-        {% if collector.invalidEntityCount > 0 %}
-            <span class="sf-toolbar-info-piece-additional sf-toolbar-status sf-toolbar-status-red">{{ collector.invalidEntityCount }}</span>
-        {% endif %}
-        {% if collector.cacheEnabled %}
-            <span class="sf-toolbar-info-piece-additional sf-toolbar-status sf-toolbar-status-green" title="Second level cache enabled">2l cache</span>
+        {% if toolbar_version|default(-1) == constant('\\Symfony\\Bundle\\WebProfilerBundle\\Controller\\ProfilerController::TOOLBAR_VERSION_2')%}
```
stof: this will break for older versions of Symfony where the constant is not available

javiereguiluz: I propose to revert to the previous way of detecting the toolbar version to use.

Resources/views/Collector/db.html.twig
stof: which change have you reverted ? You are still putting a ``sf-toolbar-status-`` class when there is no mapping errors

Resources/views/Collector/db.html.twig
stof: which previous way ? Checking the kernel version id here ? This was broken (even though your implementation of ``toolbar_version`` is currently broken in the same way in the Symfony PR)

javiereguiluz:
I've reverted all the recent changes made to this PR. Now we're back to the original situation:

```twig
{% set is_symfony_2_8_or_higher = constant('Symfony\\Component\\HttpKernel\\Kernel::VERSION_ID')|default(0) >= 20800 %}

{% if is_symfony_2_8_or_higher %}
    ...
{% else %}
    ...
{% endif %}
```

If you release a new version of this bundle compatible with Symfony 2.3+, that code should be enough to display the appropriate toolbar depending on the Symfony version.

Resources/views/Collector/db.html.twig
```diff
@@ -1,23 +1,45 @@
 {% extends app.request.isXmlHttpRequest ? 'WebProfilerBundle:Profiler:ajax_layout.html.twig' : 'WebProfilerBundle:Profiler:layout.html.twig' %}
 
+{% set profiler_markup_version = profiler_markup_version|default(1) %}
```
stof: this needs to be done inside the ``toolbar`` block too, because the rendering of the toolbar does not render the full template but only the toolbar block

stof: and actually, it is needed only inside the toolbar block for now, as the profiler itself does not yet have a new markup and so we don't need this variable yet there

javiereguiluz: Updated it. Thanks.

Resources/views/Collector/db.html.twig
stof: the change is still not reverted

stof:
Before merging this, I will wait until the Symfony PR gets merged, to be sure that no other changes to the template are needed based on some review

Resources/views/Collector/db.html.twig
javiereguiluz: I think I finally did it here: https://github.com/doctrine/DoctrineBundle/commit/461e2e35669d793cb91fe9f39ff28c33800e5fb1

Resources/views/Collector/db.html.twig
```diff
@@ -40,9 +66,15 @@
                 <b>Cache puts</b>
                 <span class="sf-toolbar-status sf-toolbar-status-{{ collector.cachePutsCount > 0 ? 'yellow' : 'green' }}">{{ collector.cachePutsCount }}</span>
             </div>
+        {% else %}
+            <div class="sf-toolbar-info-piece">
+                <b>Second Level Cache</b>
+                <span class="sf-toolbar-status sf-toolbar-status-red">disabled</span>
```
stof: we should avoid displaying a red box for people using Doctrine ORM 2.4 as they cannot enable the SCL if it does not exist.

Resources/views/Collector/db.html.twig
```diff
@@ -25,7 +47,11 @@
         </div>
         <div class="sf-toolbar-info-piece">
             <b>Invalid entities</b>
-            <span class="sf-toolbar-status sf-toolbar-status-{{ collector.invalidEntityCount > 0 ? 'red' : 'green' }}">{{ collector.invalidEntityCount }}</span>
+            {% if profiler_markup_version == 1 %}
+                <span class="sf-toolbar-status sf-toolbar-status-{{ collector.invalidEntityCount > 0 ? 'red' : 'green' }}">{{ collector.invalidEntityCount }}</span>
+            {% else %}
+                <span class="sf-toolbar-status sf-toolbar-status-{{ collector.invalidEntityCount > 0 ? 'red' }}">{{ collector.invalidEntityCount }}</span>
```
stof: you are still displaying a ``sf-toolbar-status-`` class here when the condition is false, which is what I'm askign you to change since the very beginning, to produce a clean markup.

I don't care about the fact that the green background is removed in the old markup, as it is not very important here. What is important is the red background on error.
Btw, I don't think the green background for this setting is worth the duplication. I think we can live without it.

Resources/views/Collector/db.html.twig
```diff
@@ -40,9 +66,15 @@
                 <b>Cache puts</b>
                 <span class="sf-toolbar-status sf-toolbar-status-{{ collector.cachePutsCount > 0 ? 'yellow' : 'green' }}">{{ collector.cachePutsCount }}</span>
```
stof: you are keeping the green background here when the value is 0 while removing it for invalid entities. What is the reasonning about when a green background should be used in info pieces ?

Resources/views/Collector/db.html.twig
```diff
@@ -1,23 +1,45 @@
...
+
+                <svg version="1.1"xmlns="http://www.w3.org/2000/svg" x="0px" y="0px" height="24" viewBox="0 0 24 24" enable-background="new 0 0 24 24" xml:space="preserve">
+                    <path fill="#AAAAAA" d="M5,8h14c1.7,0,3-1.3,3-3s-1.3-3-3-3H5C3.3,2,2,3.3,2,5S3.3,8,5,8z M18,3.6c0.8,0,1.5,0.7,1.5,1.5S18.8,6.6,18,6.6s-1.5-0.7-1.5-1.5S17.2,3.6,18,3.6z M19,9H5c-1.7,0-3,1.3-3,3s1.3,3,3,3h14c1.7,0,3-1.3,3-3S20.7,9,19,9z M18,13.6
+                    c-0.8,0-1.5-0.7-1.5-1.5s0.7-1.5,1.5-1.5s1.5,0.7,1.5,1.5S18.8,13.6,18,13.6z M19,16H5c-1.7,0-3,1.3-3,3s1.3,3,3,3h14c1.7,0,3-1.3,3-3S20.7,16,19,16z M18,20.6c-0.8,0-1.5-0.7-1.5-1.5s0.7-1.5,1.5-1.5s1.5,0.7,1.5,1.5S18.8,20.6,18,20.6z"/>
+                </svg>
+
+                <span class="sf-toolbar-value">{{ collector.querycount }}</span>
+                <span class="sf-toolbar-info-piece-additional-detail">
+                    <span class="sf-toolbar-label">in</span>
+                    <span class="sf-toolbar-value">{{ '%0.2f'|format(collector.time * 1000) }}</span>
```
stof: shouldn't we hide the execution time when there is no query, as done previously ? the time does not make sense if there is no query (granted it is not the common case as soon as your users are loaded from the DB)

Resources/views/Collector/db.html.twig
javiereguiluz: OK. Removed.

Resources/views/Collector/db.html.twig
javiereguiluz: You are right. The green background is no longer displayed.

Resources/views/Collector/db.html.twig
javiereguiluz: I've made this change too. Thanks.

Resources/views/Collector/db.html.twig
javiereguiluz: I've tweaked the display of information a bit:

1) If there are no Doctrine errors and no queries, you see nothing:

no_queries_no_errors

2) If there are queries but no errors, you see num of queries + time:

queries_but_no_errors

3) If there are no queries but some errors, you clearly see that:

no_queries_but_errors

4) If there are both queries and errors, you see the num of queries + time in a red background:

queries_and_errors

The expanded information contains details about the errors:

queries_and_errors_expanded

Resources/views/Collector/db.html.twig
```diff
@@ -1,23 +1,50 @@
...
+                {% if collector.invalidEntityCount > 0 %}
+                    <span class="sf-toolbar-info-piece-additional sf-toolbar-status sf-toolbar-status-red">{{ collector.invalidEntityCount }}</span>
+                {% endif %}
+
+        {% else %}
+
+            {% if collector.querycount > 0 or collector.invalidEntityCount > 0 %}
+                {% set status = collector.invalidEntityCount > 0 ? 'red' : collector.querycount > 50 ? 'yellow' : '' %}
+
+                <svg version="1.1"xmlns="http://www.w3.org/2000/svg" x="0px" y="0px" height="24" viewBox="0 0 24 24" enable-background="new 0 0 24 24" xml:space="preserve">
```
stof: I suggest extracting the icon to a ``.svg`` file. This way, it is easier to reuse it for the menu panel, and Github is able to display a preview.
As we have a single icon here, it may be placed in the Collector folder rather than creating a separate folder.

javiereguiluz: I've done this change. However, the new icon is not used yet in the profiler. The reason is that the new SVG icon has been designed for the new toolbar. We can't use it in the current profiler because the icon's style doesn't match the other icons. This will be done when we redesign the profiler and change all the icons at the same time. Thanks!

stof: yeah, I know that the panel is not yet redesigned

Resources/views/Collector/db.html.twig
```diff
@@ -1,23 +1,47 @@
...
+                {% if collector.invalidEntityCount > 0 %}
+                    <span class="sf-toolbar-info-piece-additional sf-toolbar-status sf-toolbar-status-red">{{ collector.invalidEntityCount }}</span>
+                {% endif %}
+
+        {% else %}
+
+            {% if collector.querycount > 0 or collector.invalidEntityCount > 0 %}
+                {% set status = collector.invalidEntityCount > 0 ? 'red' : collector.querycount > 50 ? 'yellow' : '' %}
+
+                {{ include('@DoctrineBundle/Collector/icon.svg') }}
```
stof: wrong template name. The Twig namespace for the bundle is ``@Doctrine``, not ``@DoctrineBundle``

javiereguiluz: Fixed. Thanks.

stof:
Thank you @javiereguiluz.

PR #447, opened 2015-07-17
Removed non-working example

magnusnordlander:
The dir parameter does not take arrays.


stof:
Good catch, thanks @magnusnordlander.

PR #451, opened 2015-08-13
Add a Configuration Option to Use Savepoints

### LINKED ISSUE #450: No Configuration Option to Use Savepoints for Nested Transactions
[`Connection::setNestTransactionsWithSavepoints`](https://github.com/doctrine/dbal/blob/3a698bd6fedbf684675ec0ecc471d07f77feea63/lib/Doctrine/DBAL/Connection.php#L1122) can be called to enable savepoints for nested transactions, but there's no easy way to do this in the bundle. I ended up having to [create a listener](https://gist.github.com/chrisguitarguy/99382991aaa2c14652d7) to solve this issue today.

Would be nice to have an option to just enable savepoints.

chrisguitarguy:
Closes #450

This adds the `use_savepoints` option to connection configuration which
turns on savepoints for nested transactions.

This is a bit uglier than it could be since [`setNestTransactionsWithSavepoints` throws](https://github.com/doctrine/dbal/blob/3a698bd6fedbf684675ec0ecc471d07f77feea63/lib/Doctrine/DBAL/Connection.php#L1128-L1130) when called on database platforms without savepoint support regardless of whether savepoints are being enabled.

I may not have done the tests correctly here.

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -232,7 +232,7 @@ class DoctrineExtension extends AbstractDoctrineExtension
         ;
 
         if (!empty($connection['use_savepoints'])) {
-            $def->addMethodCall('setNestTransactionsWithSavepoints', array(true));
+            $def->addMethodCall('setNestTransactionsWithSavepoints', array($connection['use_savepoints']));
         }
     }
 
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -221,7 +221,7 @@ class DoctrineExtension extends AbstractDoctrineExtension
 
         $options = $this->getConnectionOptions($connection);
 
-        $container
+        $def = $container
             ->setDefinition(sprintf('doctrine.dbal.%s_connection', $name), new DefinitionDecorator('doctrine.dbal.connection'))
             ->setArguments(array(
                 $options,
@@ -230,6 +230,10 @@ class DoctrineExtension extends AbstractDoctrineExtension
                 $connection['mapping_types'],
             ))
         ;
+
+        if (!empty($connection['use_savepoints'])) {
+            $def->addMethodCall('setNestTransactionsWithSavepoints', array(true));
+        }
     }
 
     protected function getConnectionOptions($connection)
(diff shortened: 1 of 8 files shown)


DependencyInjection/DoctrineExtension.php
```diff
@@ -230,6 +230,10 @@ protected function loadDbalConnection($name, array $connection, ContainerBuilder
                 $connection['mapping_types'],
             ))
         ;
+
+        if (!empty($connection['use_savepoints'])) {
+            $def->addMethodCall('setNestTransactionsWithSavepoints', array(true));
```
stof: Please use the value of ``$connection['use_savepoints']`` to set the argument, to make the code less fragile.

stof:
you need to update the XSD file to allow using this in XMl config files too. and you also need to update the doc

chrisguitarguy:
Thank's @stof changes made. Not sure if I did the XSD stuff correctly. I also added some configuration tests for the savepoints stuff.

Resources/config/schema/doctrine-1.0.xsd
```diff
@@ -37,6 +37,7 @@
         <xsd:attribute name="logging" type="xsd:string" default="false" />
         <xsd:attribute name="profiling" type="xsd:string" default="false" />
         <xsd:attribute name="server-version" type="xsd:string" />
+        <xsd:attribute name="use-savepoints" type="xsd:boolean" />
```
xabbuh: This must be added to the `driver-config` attribute group.

chrisguitarguy: Why? Drivers don't have anything related to save point configuration in them. The savepoint stuff is in the connection object itself.

stollr:
I have a question. Why is this option not set to `true` by default?

chrisguitarguy:
> I have a question. Why is this option not set to true by default?

Not sure it was explicitly set to `true` -- there's no `defaultTrue` in the config definition anyway.

hellomedia:
Should this be affected by https://github.com/doctrine/dbal/pull/5401 ?

dmaicher:
FYI this will be enabled by default in the recipe now: https://github.com/symfony/recipes/pull/1290#event-11742863967

When dropping DBAL 3 support we might want to deprecate this option then.

PR #452, opened 2015-08-13
Updated the profiler panel for the new version of the Symfony Profiler

javiereguiluz:
Refs https://github.com/symfony/symfony/pull/15523

### DIFF
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -1,23 +1,47 @@
 {% extends app.request.isXmlHttpRequest ? 'WebProfilerBundle:Profiler:ajax_layout.html.twig' : 'WebProfilerBundle:Profiler:layout.html.twig' %}
 
 {% block toolbar %}
+    {% set profiler_markup_version = profiler_markup_version|default(1) %}
+
     {% set icon %}
-        <img width="20" height="28" alt="Database" src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABQAAAAcCAYAAABh2p9gAAAAGXRFWHRTb2Z0d2FyZQBBZG9iZSBJbWFnZVJlYWR5ccllPAAAAQRJREFUeNpi/P//PwM1ARMDlcGogZQDlpMnT7pxc3NbA9nhQKxOpL5rQLwJiPeBsI6Ozl+YBOOOHTv+AOllQNwtLS39F2owKYZ/gRq8G4i3ggxEToggWzvc3d2Pk+1lNL4fFAs6ODi8JzdS7mMRVyDVoAMHDsANdAPiOCC+jCQvQKqBQB/BDbwBxK5AHA3E/kB8nKJkA8TMQBwLxaBIKQbi70AvTADSBiSadwFXpCikpKQU8PDwkGTaly9fHFigkaKIJid4584dkiMFFI6jkTJII0WVmpHCAixZQEXWYhDeuXMnyLsVlEQKI45qFBQZ8eRECi4DBaAlDqle/8A48ip6gAADANdQY88Uc0oGAAAAAElFTkSuQmCC" />
-        <span class="sf-toolbar-status{% if 50 < collector.querycount %} sf-toolbar-status-yellow{% endif %}">{{ collector.querycount }}</span>
-        {% if collector.querycount > 0 %}
-            <span class="sf-toolbar-info-piece-additional-detail">in {{ '%0.2f'|format(collector.time * 1000) }} ms</span>
-        {% endif %}
-        {% if collector.invalidEntityCount > 0 %}
-            <span class="sf-toolbar-info-piece-additional sf-toolbar-status sf-toolbar-status-red">{{ collector.invalidEntityCount }}</span>
-        {% endif %}
-        {% if collector.cacheEnabled %}
-            <span class="sf-toolbar-info-piece-additional sf-toolbar-status sf-toolbar-status-green" title="Second level cache enabled">2l cache</span>
+        {% if profiler_markup_version == 1 %}
+
+            <img width="20" height="28" alt="Database" src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABQAAAAcCAYAAABh2p9gAAAAGXRFWHRTb2Z0d2FyZQBBZG9iZSBJbWFnZVJlYWR5ccllPAAAAQRJREFUeNpi/P//PwM1ARMDlcGogZQDlpMnT7pxc3NbA9nhQKxOpL5rQLwJiPeBsI6Ozl+YBOOOHTv+AOllQNwtLS39F2owKYZ/gRq8G4i3ggxEToggWzvc3d2Pk+1lNL4fFAs6ODi8JzdS7mMRVyDVoAMHDsANdAPiOCC+jCQvQKqBQB/BDbwBxK5AHA3E/kB8nKJkA8TMQBwLxaBIKQbi70AvTADSBiSadwFXpCikpKQU8PDwkGTaly9fHFigkaKIJid4584dkiMFFI6jkTJII0WVmpHCAixZQEXWYhDeuXMnyLsVlEQKI45qFBQZ8eRECi4DBaAlDqle/8A48ip6gAADANdQY88Uc0oGAAAAAElFTkSuQmCC" />
+                <span class="sf-toolbar-value sf-toolbar-status {% if collector.querycount > 50 %}sf-toolbar-status-yellow{% endif %}">{{ collector.querycount }}</span>
+                {% if collector.querycount > 0 %}
+                    <span class="sf-toolbar-info-piece-additional-detail">in {{ '%0.2f'|format(collector.time * 1000) }} ms</span>
+                {% endif %}
+                {% if collector.invalidEntityCount > 0 %}
+                    <span class="sf-toolbar-info-piece-additional sf-toolbar-status sf-toolbar-status-red">{{ collector.invalidEntityCount }}</span>
+                {% endif %}
+
+        {% else %}
+
+            {% if collector.querycount > 0 or collector.invalidEntityCount > 0 %}
+                {% set status = collector.invalidEntityCount > 0 ? 'red' : collector.querycount > 50 ? 'yellow' : '' %}
+
+                {{ include('@Doctrine/Collector/icon.svg') }}
+
+                {% if collector.querycount == 0 and collector.invalidEntityCount > 0 %}
+                    <span class="sf-toolbar-value">{{ collector.invalidEntityCount }}</span>
+                    <span class="sf-toolbar-label">errors</span>
+                {% else %}
+                    <span class="sf-toolbar-value">{{ collector.querycount }}</span>
+                    <span class="sf-toolbar-info-piece-additional-detail">
+                        <span class="sf-toolbar-label">in</span>
+                        <span class="sf-toolbar-value">{{ '%0.2f'|format(collector.time * 1000) }}</span>
+                        <span class="sf-toolbar-label">ms</span>
+                    </span>
+                {% endif %}
+
+            {% endif %}
         {% endif %}
     {% endset %}
+
     {% set text %}
         <div class="sf-toolbar-info-piece">
-            <b>DB Queries</b>
-            <span>{{ collector.querycount }}</span>
+            <b>Database Queries</b>
+            <span class="sf-toolbar-status">{{ collector.querycount }}</span>
         </div>
         <div class="sf-toolbar-info-piece">
             <b>Query time</b>
@@ -25,7 +49,7 @@
         </div>
         <div class="sf-toolbar-info-piece">
             <b>Invalid entities</b>
-            <span class="sf-toolbar-status sf-toolbar-status-{{ collector.invalidEntityCount > 0 ? 'red' : 'green' }}">{{ collector.invalidEntityCount }}</span>
+            <span class="sf-toolbar-status {{ collector.invalidEntityCount > 0 ? 'sf-toolbar-status-red' : '' }}">{{ collector.invalidEntityCount }}</span>
         </div>
         {% if collector.cacheEnabled %}
             <div class="sf-toolbar-info-piece">
@@ -34,15 +58,21 @@
             </div>
             <div class="sf-toolbar-info-piece">
                 <b>Cache misses</b>
-                <span class="sf-toolbar-status sf-toolbar-status-{{ collector.cacheMissesCount > 0 ? 'yellow' : 'green' }}">{{ collector.cacheMissesCount }}</span>
+                <span class="sf-toolbar-status {{ collector.cacheMissesCount > 0 ? 'sf-toolbar-status-yellow' : '' }}">{{ collector.cacheMissesCount }}</span>
             </div>
             <div class="sf-toolbar-info-piece">
                 <b>Cache puts</b>
-                <span class="sf-toolbar-status sf-toolbar-status-{{ collector.cachePutsCount > 0 ? 'yellow' : 'green' }}">{{ collector.cachePutsCount }}</span>
+                <span class="sf-toolbar-status {{ collector.cachePutsCount > 0 ? 'sf-toolbar-status-yellow' : '' }}">{{ collector.cachePutsCount }}</span>
+            </div>
+        {% else %}
+            <div class="sf-toolbar-info-piece">
+                <b>Second Level Cache</b>
+                <span class="sf-toolbar-status">disabled</span>
             </div>
         {% endif %}
     {% endset %}
-    {% include 'WebProfilerBundle:Profiler:toolbar_item.html.twig' with { 'link': profiler_url } %}
+
+    {{ include('@WebProfiler/Profiler/toolbar_item.html.twig', { link: profiler_url, status: status|default('') }) }}
 {% endblock %}
 
 {% block menu %}
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -249,7 +249,7 @@
                 </thead>
                 <tbody>
                 {% for class in classes %}
-                    {% set contains_errors = (collector.mappingErrors[manager] is defined) and (collector.mappingErrors[manager][class] is defined) %}
+                    {% set contains_errors = collector.mappingErrors[manager] is defined and collector.mappingErrors[manager][class] is defined %}
                     <tr>
                         <td>{{ class }}</td>
                         <td class="font-normal {{ contains_errors ? 'status-error' }}">
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -1,7 +1,5 @@
 {% extends app.request.isXmlHttpRequest ? 'WebProfilerBundle:Profiler:ajax_layout.html.twig' : 'WebProfilerBundle:Profiler:layout.html.twig' %}
 
-{% block page_title 'Doctrine' %}
-
 {% block toolbar %}
     {% set profiler_markup_version = profiler_markup_version|default(1) %}
 
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -107,7 +107,7 @@
 {% block panel %}
     {% set profiler_markup_version = profiler_markup_version|default(1) %}
 
-    {% if profiler_markup_version == 1 %}
+    {% if profiler_markup_version == 1 and page != 'explain' %}
         <style>
             .hidden { display: none; }
             .queries-table td, .queries-table th { vertical-align: top; }
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -432,7 +432,7 @@
             }
 
             header.setAttribute('data-sort-direction', (-1*direction).toString());
-            header.children[0].innerHTML = direction > 0 ? '<span>&#9650;</span>' : '<span>&#9660;</span>';
+            header.children[0].innerHTML = direction > 0 ? '<span class="text-muted">&#9650;</span>' : '<span class="text-muted">&#9660;</span>';
 
             items.sort(function(a, b) {
                 return direction * (parseFloat(a.children[column].innerHTML) - parseFloat(b.children[column].innerHTML));
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -18,7 +18,7 @@
         {% else %}
 
             {% if collector.querycount > 0 or collector.invalidEntityCount > 0 %}
-                {% set status = collector.invalidEntityCount > 0 ? 'red' : collector.querycount > 50 ? 'yellow' : '' %}
+                {% set status = collector.invalidEntityCount > 0 ? 'red' : collector.querycount > 50 ? 'yellow' %}
 
                 {{ include('@Doctrine/Collector/icon.svg') }}
 
@@ -33,8 +33,8 @@
                         <span class="sf-toolbar-label">ms</span>
                     </span>
                 {% endif %}
-
             {% endif %}
+
         {% endif %}
     {% endset %}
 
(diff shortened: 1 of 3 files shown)


Resources/views/Collector/db.html.twig
```diff
@@ -76,38 +76,61 @@
...
+            <strong>Doctrine</strong>
+            <span class="count">
+                <span>{{ collector.querycount }}</span>
+                <span>{{ '%0.0f'|format(collector.time * 1000) }} ms</span>
+            </span>
+        </span>
+
+    {% else %}
+
+        <span class="label {{ collector.querycount > 0 ? '' : 'disabled' }}">
```
stof: this looks weird to me. You can have data in the panel even without queries (mapping issues for instance if you loaded the metadata without performing any query, for instance because of a SLC providing all data you need)

Resources/views/Collector/db.html.twig
```diff
@@ -411,3 +434,238 @@
 
     //]]></script>
 {% endblock %}
+
+{% block queries_v2 %}
```
stof: Wouldn't it be possible to avoiding duplicating the whole panel ? Having the whole panel duplicated to show the info makes the maintenance much harder, and makes it very likely to forget updating both of them in the future.

Note that making some changes to the way the panel looks like in older versions is OK, as long as it looks good.
for instance, if you can get rid of the query shrinking in the old panel too while keeping things looking good, it is fine with me (the shrinking logic is insane anyway).
Looking at the refactoring in both Symfony 2.7 and 2.8 may make the work a bit harder for you, but it will make the future maintenance a lot easier for me (and it may help documenting the necessary changes for the migration process as other bundles would also benefit from sharing as much markup as possible).

Resources/views/Collector/db.html.twig
```diff
@@ -411,3 +434,238 @@
...
+                {% endfor %}
+                </tbody>
+            </table>
+        {% endif %}
+    {% endfor %}
+
+    <h2>Database Connections</h2>
+
+    {% if collector.connections %}
+        {{ include('WebProfilerBundle:Profiler:table.html.twig', { data: collector.connections, labels: ['Name', 'Service'] }, with_context = false ) }}
```
stof: I suggest switching to ``@WebProfiler/...`` template names for consistency with WebProfilerBundle recommendations (even though DoctrineBundle is not usable outside Symfony to display a Doctrine profiler currently)

Resources/views/Collector/db.html.twig
stof: @javiereguiluz please take this comment into account. I won't merge it until it is done, and I will also delay the merge of the new design in Symfony until this is done, because I want to know how the upgrade path looks like for bundles before the feature is merged (we may need to change a bit the available CSS classes in the new design to make the upgrade path easier)

javiereguiluz:
In 6333063 I've removed most of the duplicated code.

Resources/views/Collector/db.html.twig
```diff
@@ -186,55 +162,83 @@
...
-        {% include 'WebProfilerBundle:Profiler:table.html.twig' with {data: collector.connections} only %}
+    {% if not collector.connections %}
+        <div class="empty">
+            <p>There are no configured database connections.</p>
+        </div>
     {% else %}
-        <p>
-            <em>No connections.</em>
-        </p>
+        {{ include('WebProfilerBundle:Profiler:table.html.twig', { data: collector.connections, labels: ['Name', 'Service'] }, with_context = false ) }}
```
stof: please use the ``@WebProfiler/`` naming instead to reference WebProfilerBundle templates

Resources/views/Collector/db.html.twig
```diff
@@ -186,55 +162,83 @@
...
     {% endif %}
 
     <h2>Second Level Cache</h2>
 
-    {% if collector.cacheCounts %}
-        {% include 'WebProfilerBundle:Profiler:table.html.twig' with {data: collector.cacheCounts} only %}
+    {% if not collector.cacheEnabled %}
+    <div class="empty">
+        <p>Second Level Cache is not enabled.</p>
+    </div>
```
stof: should be indented

Resources/views/Collector/db.html.twig
```diff
@@ -186,55 +162,83 @@
...
+    {% else %}
+        {% if collector.cacheCounts %}
```
stof: you should add an info saying that profiling was not enabled for the SLC in the else clause (in the current version of the code, this case also shows ``No cache.`` for that case, which was crappy (it was not checking ``collector.cacheEnabled``)

Resources/views/Collector/db.html.twig
```diff
@@ -186,55 +162,83 @@
...
-            <em>No cache.</em>
-        </p>
     {% endif %}
 
-    <h2>Mapping</h2>
+    <h2>Entities Mapping</h2>
 
     {% for manager, classes in collector.entities %}
-        <h3>Manager <em>{{ manager }}</em></h3>
+        <h3>{{ manager }} <small>entity manager</small></h3>
```
stof: should we also remove the title in case of a single manager, as done for connection queries ?

Resources/views/Collector/db.html.twig
```diff
@@ -245,17 +249,18 @@
                 </thead>
                 <tbody>
                 {% for class in classes %}
+                    {% set contains_errors = (collector.mappingErrors[manager] is defined) and (collector.mappingErrors[manager][class] is defined) %}
```
stof: parenthesis are not needed

Resources/views/Collector/db.html.twig
javiereguiluz: I tried ... but I don't know how to obtain the complete list of managers to check if there is only one.

Resources/views/Collector/db.html.twig
```diff
@@ -266,58 +271,54 @@
...
         function explain(link) {
             "use strict";
 
             var imgs = link.children,
-                target = link.getAttribute('data-target-id');
+                    target = link.getAttribute('data-target-id');
 
             Sfjs.toggle(target, imgs[0], imgs[1])
-                .load(
+                    .load(
```
stof: you should not change the indentation here IMO. It makes it inconsistent

Resources/views/Collector/db.html.twig
```diff
@@ -266,58 +271,54 @@
...
+    {% endif %}
 
         function expandQuery(link) {
             var sections = link.children,
-                target = link.getAttribute('data-target-id'),
-                targetId = target.replace('code', ''),
-                queriesParameters = document.getElementById('original-query' + targetId);
+                    target = link.getAttribute('data-target-id'),
+                    targetId = target.replace('code', ''),
+                    queriesParameters = document.getElementById('original-query' + targetId);
```
stof: I don't think the indentation should change

Resources/views/Collector/db.html.twig
```diff
@@ -266,58 +271,54 @@
...
-                );
+            );
 
             return false;
         }
+    {% else %}
+        function explain(link) {
+            "use strict";
 
-        function expandAllQueries(button) {
```
stof: is it intended to remove the ``expandAllQueries`` method ?

Resources/views/Collector/db.html.twig
```diff
@@ -115,70 +132,29 @@
...
-            </p>
+            <div class="empty">
+                <p>No database queries were performed.</p>
+            </div>
         {% else %}
-            <p>
-                <button type="button" class="sf-button" onclick="expandAllQueries(this);" data-action="expand" data-connection="{{ loop.index }}">
-                    <span class="border-l">
-                        <span class="border-r">
-                            <span class="btn-bg">Expand all queries</span>
```
stof: I think this should be kept in the old markup display, given queries are kept shrinked there

Resources/views/Collector/db.html.twig
```diff
@@ -409,5 +410,81 @@
...
+            </td>
+        </tr>
+    {% endfor %}
+{% endblock %}
+
+{% block query_table_v2 %}
+    {% for i, query in queries %}
+        <tr id="queryNo-{{ i }}-{{ loop.parent.loop.index }}">
+            <td>{{ loop.index }}</td>
+            <td class="nowrap">{{ '%0.2f'|format(query.executionMS * 1000) }} ms</td>
```
stof: ``&nbsp;`` was used in the previous version. It is intended to use a normal space now ?

Resources/views/Collector/db.html.twig
javiereguiluz: I agree. this was unintended and made by mistake. I'm sorry.

Twig/DoctrineExtension.php
```diff
@@ -314,6 +315,33 @@ function ($matches) use ($parameters, &$i) {
...
 
     /**
+     * Highlights the syntax of the given SQL statement.
+     *
+     * @param  string $sql
+     * @return string
+     */
+    public function highlightSqlSyntax($sql)
+    {
+        $highlighter = new \SqlFormatter();
```
stof: no need to create an instance. SqlFormatter is a procedural API wrapped in a class anyway, and you are still changing the global state even if you access static params through the instance rather than the class name

Resources/views/Collector/db.html.twig
javiereguiluz: My guess is that the explicit `&nbsp;` was intended to solve the same problem that we now solve with the `nowrap` CSS class.

Twig/DoctrineExtension.php
```diff
@@ -40,6 +40,7 @@ public function getFilters()
             new \Twig_SimpleFilter('doctrine_minify_query', array($this, 'minifyQuery')),
             new \Twig_SimpleFilter('doctrine_pretty_query', 'SqlFormatter::format', array('is_safe' => array('html'))),
             new \Twig_SimpleFilter('doctrine_replace_query_parameters', array($this, 'replaceQueryParameters'), array('is_safe' => array('html'))),
```
stof: This method is actually not safe when highlighting is disabled.
I think it might be worth using a dedicated filter to highlight all the time

Resources/views/Collector/db.html.twig
```diff
@@ -409,5 +410,81 @@
...
+                    <a href="#" id="expandParams-{{ i }}-{{ loop.parent.loop.index }}" class="sf-toggle link-inverse" data-toggle-selector="#original-query-{{ i }}-{{ loop.parent.loop.index }}" data-toggle-alt-content="Hide runnable query">View runnable query</a>
+
+                    {% if query.explainable %}
+                        &nbsp;&nbsp;
+                        <a class="link-inverse" href="{{ path('_profiler', { panel: 'db', token: token, page: 'explain', connection: connection, query: i }) }}" onclick="return explain(this);" data-target-id="explain-{{ i }}-{{ loop.parent.loop.index }}">Explain query</a>
+                    {% endif %}
+                </div>
+
+                <div id="original-query-{{ i }}-{{ loop.parent.loop.index }}" class="sql-runnable hidden">
+                    {{ (query.sql ~ ';')|doctrine_replace_query_parameters(query.params, highlight = false)|doctrine_highlight_sql }}
```
stof: why adding a semicolon ?

Resources/views/Collector/db.html.twig
```diff
@@ -266,18 +283,42 @@
     {% endfor %}
 
     <script type="text/javascript">//<![CDATA[
+    {% if profiler_markup_version == 1 %}
+        function expandAllQueries(button) {
```
stof: can you keep it after the ``explain`` function to avoid unnecessary diffs ?

Resources/views/Collector/db.html.twig
javiereguiluz: In some apps, when you don't add the `;` you cannot run the query. In contrast, adding it doesn't prevent any app to run the query (as far as I know).

javiereguiluz:
Regarding this comment:

```
>      {% endif %}
>  
>      <h2>Second Level Cache</h2>
>  
> -    {% if collector.cacheCounts %}
> -        {% include 'WebProfilerBundle:Profiler:table.html.twig' with {data: collector.cacheCounts} only %}
> +    {% if not collector.cacheEnabled %}
> +    <div class="empty">
> +        <p>Second Level Cache is not enabled.</p>
> +    </div>
> +    {% else %}
> +        {% if collector.cacheCounts %}
you should add an info saying that profiling was not enabled for the SLC in the else clause (in the current version of the code, this case also shows No cache. for that case, which was crappy (it was not checking collector.cacheEnabled)
```

I'm afraid don't know which of the following two comments are you referring:

```
Second Level Cache is not enabled.
Second level cache information is not available.
```

stof:
hmm, I missed the message ``Second level cache information is not available.`` when reviewing

Twig/DoctrineExtension.php
javiereguiluz: In 9aa72e2 I've made the proposed changes in the Doctrine Twig extension.

stof:
@javiereguiluz do you actually need to keep the query shrinking for the old design if we drop it in the new design ? Or could we drop it everywhere ? The more we keep content similar between versions the better IMO.

javiereguiluz:
In cee5319 I've deduplicated all code. In fact, this pull request now removes more than it adds:

removed_added_lines

The panel keeps looking the same in the new profiler, but in the old profiler it now looks and behaves as follows:

old_profiler_doctrine_panel

kimhemsoe:
What is the status for this ?

stof:
@javiereguiluz can you show how it looks like for very long queries ?

Resources/views/Collector/db.html.twig
```diff
@@ -76,38 +76,64 @@
...
+        </span>
+
+    {% endif %}
 {% endblock %}
 
 {% block panel %}
+    {% set profiler_markup_version = profiler_markup_version|default(1) %}
+
+    {% if profiler_markup_version == 1 %}
+        <style>
```
stof: this should not be done for the ``explain`` page, right ?

Twig/DoctrineExtension.php
```diff
@@ -305,15 +305,35 @@ function ($matches) use ($parameters, &$i) {
...
+        \SqlFormatter::$reserved_attributes = 'class="keyword"';
+        \SqlFormatter::$boundary_attributes = 'class="symbol"';
+        \SqlFormatter::$number_attributes = 'class="number"';
+        \SqlFormatter::$word_attributes = 'class="word"';
+        \SqlFormatter::$error_attributes = 'class="error"';
+        \SqlFormatter::$comment_attributes = 'class="comment"';
+        \SqlFormatter::$variable_attributes = 'class="variable"';
+
+        $html = \SqlFormatter::highlight($sql);
+        $html = preg_replace('/<pre class="(.*)">(.*)<\/pre>/Us', '<div class="\1"><pre>\2</pre></div>', $html);
```
stof: ``(.*)`` should be ``([^"]*+)`` to catching more than necessary here (in case there is nested tags inside the pre with their own attributes). The ``+`` in my proposal is optimizing the regex by making the quantifier possessive, which is possible with my proposal (as the quantifier cannot match the ``"`` following it)

Resources/views/Collector/explain.html.twig
```diff
@@ -1,8 +1,6 @@
-<strong>Explanation:</strong>
-
 {% if data[0]|length > 1 %}
     {# The platform returns a table for the explanation (e.g. MySQL), display all columns #}
-    <table style="margin: 5px 0;">
+    <table class="sql-explanation" style="margin: 5px 0;">
```
stof: there is no such class in the profiler styles

Twig/DoctrineExtension.php
```diff
@@ -39,7 +39,8 @@ public function getFilters()
         return array(
             new \Twig_SimpleFilter('doctrine_minify_query', array($this, 'minifyQuery')),
```
stof: given that the panel does not minify queries anymore, this filter should be marked as deprecated (Twig has a feature for that)

Twig/DoctrineExtension.php
```diff
@@ -39,7 +39,8 @@ public function getFilters()
...
             new \Twig_SimpleFilter('doctrine_pretty_query', 'SqlFormatter::format', array('is_safe' => array('html'))),
```
stof: This filter should be deprecated in favor of ``doctrine_highlight_sql``, or the existing filter name should be kept.

stof:
@javiereguiluz I like the way the panel looks like in the Symfony 2.7 profiler now. Getting rid of the query shrinking makes it much easier to identify queries, especially for complex ones where the shrinking was not behaving well.

javiereguiluz:
I don't know if this query is long enough, but here it is a more real example:

long_queries

guilhermeblanco:
@javiereguiluz I'm more fan if we could line break between certain keywords, like (FROM, LEFT [OUTER] JOIN, RIGHT [OUTER] JOIN, ON, WHERE, AND, OR, ORDER BY, GROUP BY, HAVING, LIMIT, OFFSET, VALUES, SET).
That should be easy by adding a line break if it matches space + keyword + space. Nothing else should have that or any RDBMS engine would crash.
Here is an example how the same query you posted would look like:

```sql
SELECT b0_.id AS id0, b9_.title AS title1, b0_.body AS body2, b0_.html_body AS html_body3, b0_.excerpt AS excerpt4, b0_.allow_comments AS allow_comments5, b0_.published AS published6, b0_.published_at AS published_at7, b0_.created_at AS created_at8, b0_.updated_at AS updated_at9, b0_.slug AS slug10, b0_.format AS format11, s1_.id AS id12, s1_.uuid AS uuid13, s1_.name AS name14, s1_.username AS username15, s1_.email AS email16, s1_.access_token AS access_token17, s1_.created_at AS created_at18, s1_.updated_at AS updated_at19, s1_.commenter_badge AS commenter_badge20, b2_.id AS id21, b2_.name AS name22, b2_.slug AS slug23, b0_.user_id AS user_id24, b0_.author_id AS author_id25
FROM blog_post b0_
LEFT JOIN user s1_
ON b0_.user_id = s1_.id
LEFT JOIN blog_post_category b3_
ON b0_.id = b3_.post_id
LEFT JOIN blog_category b2_
ON b2_.id = b3_.category_id
WHERE b0_.published = ?
AND b0_.published_at <= ?
ORDER BY b0_.published_at DESC
LIMIT 3
```

Also with some initial spacing, we could even make it similar to this:

```sql
SELECT b0_.id AS id0, b9_.title AS title1, b0_.body AS body2, b0_.html_body AS html_body3, b0_.excerpt AS excerpt4, b0_.allow_comments AS allow_comments5, b0_.published AS published6, b0_.published_at AS published_at7, b0_.created_at AS created_at8, b0_.updated_at AS updated_at9, b0_.slug AS slug10, b0_.format AS format11, s1_.id AS id12, s1_.uuid AS uuid13, s1_.name AS name14, s1_.username AS username15, s1_.email AS email16, s1_.access_token AS access_token17, s1_.created_at AS created_at18, s1_.updated_at AS updated_at19, s1_.commenter_badge AS commenter_badge20, b2_.id AS id21, b2_.name AS name22, b2_.slug AS slug23, b0_.user_id AS user_id24, b0_.author_id AS author_id25
  FROM blog_post b0_
  LEFT JOIN user s1_
       ON b0_.user_id = s1_.id
  LEFT JOIN blog_post_category b3_
       ON b0_.id = b3_.post_id
  LEFT JOIN blog_category b2_
       ON b2_.id = b3_.category_id
 WHERE b0_.published = ?
   AND b0_.published_at <= ?
 ORDER BY b0_.published_at DESC
 LIMIT 3
```

If we remove ON, AND and OR from line break list, we'd have:

```sql
SELECT b0_.id AS id0, b9_.title AS title1, b0_.body AS body2, b0_.html_body AS html_body3, b0_.excerpt AS excerpt4, b0_.allow_comments AS allow_comments5, b0_.published AS published6, b0_.published_at AS published_at7, b0_.created_at AS created_at8, b0_.updated_at AS updated_at9, b0_.slug AS slug10, b0_.format AS format11, s1_.id AS id12, s1_.uuid AS uuid13, s1_.name AS name14, s1_.username AS username15, s1_.email AS email16, s1_.access_token AS access_token17, s1_.created_at AS created_at18, s1_.updated_at AS updated_at19, s1_.commenter_badge AS commenter_badge20, b2_.id AS id21, b2_.name AS name22, b2_.slug AS slug23, b0_.user_id AS user_id24, b0_.author_id AS author_id25
  FROM blog_post b0_
  LEFT JOIN user s1_ ON b0_.user_id = s1_.id
  LEFT JOIN blog_post_category b3_ ON b0_.id = b3_.post_id
  LEFT JOIN blog_category b2_ ON b2_.id = b3_.category_id
 WHERE b0_.published = ? AND b0_.published_at <= ?
 ORDER BY b0_.published_at DESC
 LIMIT 3
```

stof:
@guilhermeblanco how would this work when you have a query using CTEs or subqueries ? These are cases where the query display is currently really weird in the profiler (even worse with query shrinking)

javiereguiluz:
I agree with most of what you are saying. I have two questions:

1) Can someone paste here a really long and complex (but real) SQL query so we can use it as edge case for the design? Thanks!

2) I like @guilhermeblanco's ideas about formatting long queries. Before doing those changes, does anyone know a good example of formatting SQL queries for readability? Maybe you use a desktop application that solves this problem right and you can paste a screenshot here. Thanks!

stof:
@javiereguiluz I can give you some queries coming from my work project if you want (they rely on Postgres-specific features, but you may not need to execute them)

javiereguiluz:
@stof thanks! That would be great. If they cannot be displayed in public, please send them to javier.eguiluz@sensiolabs.com and tell me that so I don't paste a screenshot of them. Thanks!

stof:
email sent

javiereguiluz:
@stof has kindly provided me some long and complex queries and after playing a bit with them, I'd like to propose the following change.

This is how the profiler shows these complex queries:

profiler-long-query

I propose to add a new link action called `View formatted query` which displays the result of applying the 
SqlFormatter filter to the original query. Example for this very complex query:

profiler-long-query-formatted

And the runnable query would remain the same because you just want to copy+paste it to execute it:

profiler-long-query-runnable

And this is how it works when you combine everything:

profiler-queries

javiereguiluz:
I've just fixed all the issues reported by @stof, so before merging this PR, we just need to decide if the new "View formatted query" is the way to go or not. Thanks!

Resources/views/Collector/db.html.twig
```diff
@@ -76,38 +76,64 @@
...
+            {% endif %}
+        </span>
+
+    {% endif %}
 {% endblock %}
 
 {% block panel %}
+    {% set profiler_markup_version = profiler_markup_version|default(1) %}
+
+    {% if profiler_markup_version == 1 and page != 'explain' %}
```
stof: rather than checking for ``explains`` here, move it inside the ``queries`` block

javiereguiluz: Much better. Done. Thanks.

Twig/DoctrineExtension.php
```diff
@@ -37,9 +37,10 @@ class DoctrineExtension extends \Twig_Extension
     public function getFilters()
     {
         return array(
-            new \Twig_SimpleFilter('doctrine_minify_query', array($this, 'minifyQuery')),
+            new \Twig_SimpleFilter('doctrine_minify_query', array($this, 'minifyQuery'), array('deprecated' => true)),
             new \Twig_SimpleFilter('doctrine_pretty_query', 'SqlFormatter::format', array('is_safe' => array('html'))),
```
stof: is ``doctrine_pretty_query`` still used anywhere ? We don't need 2 different filers to highlight SQL.

javiereguiluz: it wasn't used ... until this morning where we introduced the "View formatted query" action: https://github.com/doctrine/DoctrineBundle/pull/452/files#diff-3fa34991d339dfb273791fb46a6cff1cR186

stof: well, this method does exactly the same than the new ``doctrine_highlight_sql``, except that one is using the profiler styles and the other one uses the raw SqlFormatter colors. This looks weird to me.

fabpot:
@javiereguiluz Can you update the first line to read like this:

```jinja
{% extends app.request.isXmlHttpRequest ? '@WebProfiler/Profiler/ajax_layout.html.twig' : '@WebProfiler/Profiler/layout.html.twig' %}
```

stof:
this needs to be rebased as there is a conflict

Twig/DoctrineExtension.php
javiereguiluz: What do you propose to solve this issue?

fabpot:
I'd love to have this one merged ASAP... and then having a release in a couple of days would be awesome as well.

javiereguiluz:
Rebased and ready to be merged. @stof I propose you to solve the issue with the duplicated Twig filter in another pull request. That way we can merge this ASAP. Thanks.

stof:
It took time, but here we go, this is in now. Thank you very much @javiereguiluz.

javiereguiluz:
@stof thanks for your review. Superb as always!

PR #455, opened 2015-08-17
[WIP] Add `--shard` option to doctrine commands

vincentchalamon:
I need to use shard connections in my project, but it seems that some code is not up to date with sharding. This PR add some compatibility & options with shard connections.

- [x] Add `--shard` option to `doctrine:database` commands
- [x] Switch to correct shard connection if exists, else select _global_ one
- [x] Update [PoolingShardConnection](http://www.doctrine-project.org/api/dbal/2.4/class-Doctrine.DBAL.Sharding.PoolingShardConnection.html) => https://github.com/doctrine/dbal/pull/896
- [x] Create dynamic service `doctrine.<connection name>_shard_manager` using [PoolingShardManager](http://www.doctrine-project.org/api/dbal/2.5/class-Doctrine.DBAL.Sharding.PoolingShardManager.html)
- [ ] Add `--shard` option to `doctrine:migrations` commands
- [ ] Add `--shard` option to `doctrine:fixtures` commands

Is there a need to implement this option on `doctrine:fixtures` & `doctrine:migrations` commands?

### DIFF
--- a/Command/CreateDatabaseDoctrineCommand.php
+++ b/Command/CreateDatabaseDoctrineCommand.php
@@ -35,6 +35,7 @@ class CreateDatabaseDoctrineCommand extends DoctrineCommand
         $this
             ->setName('doctrine:database:create')
             ->setDescription('Creates the configured database')
+            ->addOption('shard', null, InputOption::VALUE_OPTIONAL, 'The shard connection to use for this command')
             ->addOption('connection', null, InputOption::VALUE_OPTIONAL, 'The connection to use for this command')
             ->addOption('if-not-exists', null, InputOption::VALUE_NONE, 'Don\'t trigger an error, when the database already exists')
             ->setHelp(<<<EOT
@@ -65,6 +66,21 @@ EOT
         $params = $connection->getParams();
         if (isset($params['master'])) {
             $params = $params['master'];
+        }
+        if (isset($params['shards'])) {
+            $shards = $params['shards'];
+            // Default select global
+            $params = $params['global'];
+            if ($input->getOption('shard')) {
+                foreach ($shards as $shard) {
+                    if ($shard['id'] === (int)$input->getOption('shard')) {
+                        // Select sharded database
+                        $params = $shard;
+                        unset($params['id']);
+                        break;
+                    }
+                }
+            }
         }
 
         $name = isset($params['path']) ? $params['path'] : (isset($params['dbname']) ? $params['dbname'] : false);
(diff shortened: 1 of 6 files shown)


DependencyInjection/DoctrineExtension.php
```diff
@@ -279,8 +279,16 @@ protected function getConnectionOptions($connection)
                     continue;
                 }
                 $options['master'][$key] = $value;
+                if (!isset($options['master']['driver'])) {
+                    $options['master']['driver'] = $options['driver'];
```
stof: doing this in each iteration of the loop does not make sense.

And why would you copy the driver ? AFAIK, it should not be copied to each connection

vincentchalamon: Originally it was for the sharding configuration: a connection needs a `driver` or a `driverOptions` option. I removed it from _master/slave_ configuration as my PR only concerns sharding.

Command/CreateDatabaseDoctrineCommand.php
```diff
@@ -66,6 +67,21 @@ protected function execute(InputInterface $input, OutputInterface $output)
         if (isset($params['master'])) {
             $params = $params['master'];
         }
+        if (isset($params['shards'])) {
+            $shards = $params['shards'];
+            // Default select global
+            $params = $params['global'];
+            if ($input->getOption('shard')) {
+                foreach ($shards as $shard) {
+                    if ($shard['id'] == $input->getOption('shard')) {
```
dunglas: `===`

Command/DropDatabaseDoctrineCommand.php
```diff
@@ -70,6 +71,21 @@ protected function execute(InputInterface $input, OutputInterface $output)
         if (isset($params['master'])) {
             $params = $params['master'];
         }
+        if (isset($params['shards'])) {
+            $shards = $params['shards'];
+            // Default select global
+            $params = $params['global'];
+            if ($input->getOption('shard')) {
+                foreach ($shards as $shard) {
+                    if ($shard['id'] == $input->getOption('shard')) {
```
dunglas: `===`

Command/CreateDatabaseDoctrineCommand.php
vincentchalamon: Nop cause `$input->getOption('shard')` is interpreted as a string, and `$shard['id']` is an integer

dunglas: So cast and it explicitly. `==` is evil (even not transitive...)

vincentchalamon:
Moved to https://github.com/doctrine/DoctrineBundle/pull/456

PR #456, opened 2015-08-19
Add --shard option to doctrine commands

vincentchalamon:
I need to use shard connections in my project, but it seems that some code is not up to date with sharding. This PR add some compatibility & options with shard connections.

- [x] Add `--shard` option to `doctrine:database` commands
- [x] Switch to correct shard connection if exists, else select _global_ one
- [x] Update [PoolingShardConnection](http://www.doctrine-project.org/api/dbal/2.4/class-Doctrine.DBAL.Sharding.PoolingShardConnection.html) => https://github.com/doctrine/dbal/pull/903 & https://github.com/doctrine/dbal/pull/912
- [x] Create dynamic service `doctrine.<connection name>_shard_manager` using [PoolingShardManager](http://www.doctrine-project.org/api/dbal/2.5/class-Doctrine.DBAL.Sharding.PoolingShardManager.html)
- [x] Add `--shard` option to `doctrine:migrations` commands => https://github.com/doctrine/DoctrineMigrationsBundle/pull/132
- [x] Add `--shard` option to `doctrine:fixtures` commands => https://github.com/doctrine/DoctrineFixturesBundle/pull/145
- [x] Update unit tests

### DIFF
--- a/Command/DoctrineCommand.php
+++ b/Command/DoctrineCommand.php
@@ -14,8 +14,11 @@
 
 namespace Doctrine\Bundle\DoctrineBundle\Command;
 
-use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Sharding\PoolingShardConnection;
+use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\Tools\EntityGenerator;
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
 
 /**
  * Base class for Doctrine console commands to extend from.
@@ -45,13 +48,24 @@ abstract class DoctrineCommand extends ContainerAwareCommand
     /**
      * Get a doctrine entity manager by symfony name.
      *
-     * @param string $name
+     * @param string       $name
+     * @param null|integer $shardId
      *
-     * @return \Doctrine\ORM\EntityManager
+     * @return EntityManager
      */
-    protected function getEntityManager($name)
+    protected function getEntityManager($name, $shardId = null)
     {
-        return $this->getContainer()->get('doctrine')->getManager($name);
+        $manager = $this->getContainer()->get('doctrine')->getManager($name);
+
+        if ($shardId) {
+            if (!$manager->getConnection() instanceof PoolingShardConnection) {
+                throw new \LogicException(sprintf("Connection of EntityManager '%s' must implement shards configuration.", $name));
+            }
+
+            $manager->getConnection()->connect($shardId);
+        }
+
+        return $manager;
     }
 
     /**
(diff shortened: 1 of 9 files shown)


DependencyInjection/DoctrineExtension.php
```diff
@@ -304,10 +313,22 @@ protected function getConnectionOptions($connection)
                 $options['global'][$key] = $value;
                 unset($options[$key]);
             }
+            if (!isset($options['global']['driver'])) {
+                $options['global']['driver'] = $options['driver'];
+            }
+            foreach ($options['shards'] as $i => $shard) {
+                if (!isset($shard['driver'])) {
+                    $options['shards'][$i]['driver'] = $options['driver'];
```
stof: why is that ? the PoolingShardConnection does not need to specify the driver for each shard (it does not make any sense to have a different driver per shard)

vincentchalamon: Cause it will throw an exception to require `driver` or `driverOptions` in connection

Command/DoctrineCommand.php
```diff
@@ -45,24 +48,46 @@ protected function getEntityGenerator()
...
      */
-    protected function getEntityManager($name)
+    protected function getEntityManager($name, $shardId = null)
     {
-        return $this->getContainer()->get('doctrine')->getManager($name);
+        $manager = $this->getContainer()->get('doctrine')->getManager($name);
+
+        if ($shardId) {
+            if (!$manager->getConnection() instanceof PoolingShardConnection) {
+                throw new \LogicException(sprintf("Connection of EntityManager '%s' must implements shards configuration.", $name));
```
guillaumevoisin: "must implement"

guillaumevoisin:
:+1:

MLKiiwy:
:+1:

DependencyInjection/DoctrineExtension.php
stof: where would it throw an exception ?

vincentchalamon: Nowhere now, bug fixed, I remove this code :)

vincentchalamon: For example when executing a `doctrine:database:create` command

stof: well, then ``doctrine:database:create`` should be fixed instead when it builds params in a special way instead of using the configured params.
Adding useless params is the wrong way (the DriverManager will never read ``$params['shard'][$i]['driver']`` but always ``$params['driver']``)

vincentchalamon: :+1:

vincentchalamon:
@iamluc Can you also have a look at it ? Thanks :)

vincentchalamon:
@deeky666 Can you also have a look at it please ? :)

dunglas:
:+1: 

In another PR it would be nice to refactor the create and the drop commands: they share a lot of common code and code duplication is bad :-)

vincentchalamon:
Up ! @stof @deeky666

kimhemsoe:
Thanks @vincentchalamon

PR #462, opened 2015-09-22
fixed hardcoded app/console

fabpot:
Important for Symfony 3.0 where the console is not in `app/` anymore.


Command/Proxy/RunSqlDoctrineCommand.php
```diff
@@ -41,7 +41,7 @@ protected function configure()
 The <info>doctrine:query:sql</info> command executes the given SQL query and
 outputs the results:
 
-<info>php app/console doctrine:query:sql "SELECT * from user"</info>
+<info>php %command.full_name% doctrine:query:sql "SELECT * from user"</info>
```
stof: this is broken. ``%command.full_name%`` replaces ``app/console doctrine:query:sql``, not just ``app/console`` (same for all other places)

PR #465, opened 2015-09-26
Remove the `url` Configuration Parameter in Commands

### LINKED ISSUE #464: Connection Configuration via URL Causes Drop/Create Commands to Fail
In both the Drop and Create database commands, the database name is extracted from the connection parameters then removed before `DriverManager::getConnection` is called. If you configure with a URL however, the URL remains the connection is still attempted with the database name in place which fails if the database doesn't exist.

Example configuration:

```yaml
doctrine:
  dbal:
    url: mysql://root@localhost/dbname
```

Should just need to `unset($params['url'])` in each command.

chrisguitarguy:
When present, `DriverManager::getConnection` parses `url` out into the
appropriate config parameters (like path -> dbname, etc). When the
Create/Drop commands do their thing, they look for `dbname` (or path)
then unset the value. If URL is present, it just goes right back into
`DriverManager::getConnection` and gets re-parsed.

Which means something like the create database command would fail if the
database does not exist (because the connection cannot be made).

Remove the URL, just like what is done with the `dbname` parameter.

Closes #464

Wasn't really sure what to test here. There is only a single command test and very artificial. reproduction this bug relies on `DriverManager::getConnection` having parsed the URL already when the connection is pulled out of the container.

### DIFF
--- a/Command/CreateDatabaseDoctrineCommand.php
+++ b/Command/CreateDatabaseDoctrineCommand.php
@@ -71,6 +71,7 @@ EOT
         if (!$name) {
             throw new \InvalidArgumentException("Connection does not contain a 'path' or 'dbname' parameter and cannot be dropped.");
         }
+        unset($params['url']);
         unset($params['dbname']);
 
         $tmpConnection = DriverManager::getConnection($params);
--- a/Command/DropDatabaseDoctrineCommand.php
+++ b/Command/DropDatabaseDoctrineCommand.php
@@ -75,6 +75,7 @@ EOT
         if (!$name) {
             throw new \InvalidArgumentException("Connection does not contain a 'path' or 'dbname' parameter and cannot be dropped.");
         }
+        unset($params['url']);
         unset($params['dbname']);
 
         if ($input->getOption('force')) {


stof:
if you remove the URL entirely, you remove too much: you will not connect to the same host than the connection normally does. So this makes things even worse than the bug you are trying to fix: you will create the database on a different server (potentially not configured in the project)

chrisguitarguy:
> you will not connect to the same host than the connection normally does. So this makes things even worse than the bug you are trying to fix: you will create the database on a different server (potentially not configured in the project)

Are you 100% on that? The url is parsed out into the params passed into the `Doctrine\DBAL\Connection` constructor. The Create/Drop commands pull the parameters out of the connection itself, not the bundle configuration.

- https://github.com/doctrine/dbal/blob/f44782e91c8c910736272986f543989d214daf1c/lib/Doctrine/DBAL/DriverManager.php#L144
- https://github.com/doctrine/dbal/blob/f44782e91c8c910736272986f543989d214daf1c/lib/Doctrine/DBAL/DriverManager.php#L226
- https://github.com/doctrine/dbal/blob/f44782e91c8c910736272986f543989d214daf1c/lib/Doctrine/DBAL/Connection.php#L205
- https://github.com/doctrine/dbal/blob/f44782e91c8c910736272986f543989d214daf1c/lib/Doctrine/DBAL/Connection.php#L235
- https://github.com/chrisguitarguy/DoctrineBundle/blob/3671281f9ed63d8b6523701e1aae9d74c961b13e/Command/CreateDatabaseDoctrineCommand.php#L65
- https://github.com/chrisguitarguy/DoctrineBundle/blob/3671281f9ed63d8b6523701e1aae9d74c961b13e/Command/DropDatabaseDoctrineCommand.php#L69

kimhemsoe:
Any update on this issue?

chrisguitarguy:
From me or @stof, @kimhemsoe? I don't think removing the URL is too much, but I may misunderstand how everything fits together.

kimhemsoe:
Both :) To me it seems @stof still have concerns about this fix which have not been addressed yet.

kimhemsoe:
If you remove this url how does dbal then know which host is should connect to?

Closing as not a valid fix. You are very welcome to try a different approach with a new PR

chrisguitarguy:
@kimhemsoe read the code in the links above. The Create and Drop database command pull the params out of the connection. `DriverManager::getConnection` parses the url into params specific to doctrine and passes them into the connection object and uses those values to create a new connection.

```php
<?php

use Doctrine\DBAL\DriverManager;
require __DIR__.'/vendor/autoload.php';

$conn = DriverManager::getConnection([
    'url' => 'mysql://root@localhost/test'
]);

$params = $conn->getParams();

echo 'Connection Params', PHP_EOL;
print_r($params);

// unset dbname, but not url
unset($params['dbname']);

$conn2 = DriverManager::getConnection($params);

// `dbname` is back!
echo 'Connection 2 Params', PHP_EOL;
print_r($conn2->getParams());
```

Result:

```
Connection Params
Array
(
    [url] => mysql://root@localhost/test
    [driver] => pdo_mysql
    [host] => localhost
    [user] => root
    [dbname] => test
)
Connection 2 Params
Array
(
    [url] => mysql://root@localhost/test
    [driver] => pdo_mysql
    [host] => localhost
    [user] => root
    [dbname] => test
)
```

This isn't a valid fix because [the fix was already done](https://github.com/doctrine/DoctrineBundle/blob/dd40b0a7fb16658cda9def9786992b8df8a49be7/Command/CreateDatabaseDoctrineCommand.php#L96). It was part of #434.

kimhemsoe:
I did not know that. 
I am trying to clean out dead issues and PR. People tend to create and forget everything about them afterwards.

Anyway problem solved.

chrisguitarguy:
No worries, I closed the issue I had opened for this PR as well. Thanks!

PR #467, opened 2015-09-28
fixed command helps

fabpot:
see @stof comment in #462


Command/ImportMappingDoctrineCommand.php
```diff
@@ -48,22 +48,22 @@ protected function configure()
 The <info>doctrine:mapping:import</info> command imports mapping information
```
stof: Can you also change this to use ``%command.name%`` for consistency ?

fabpot: done

PR #468, opened 2015-09-28
fixed template names

fabpot:
Fixed some template names to use the Twig-compatible-only name (some template references were already using this notation and that makes the bundle compatible with a Twig-only project). It's safe as this notation is supported since Symfony 2.2 and Twig 1.10.

### DIFF
--- a/Controller/ProfilerController.php
+++ b/Controller/ProfilerController.php
@@ -65,7 +65,7 @@ class ProfilerController extends ContainerAware
             return new Response('This query cannot be explained.');
         }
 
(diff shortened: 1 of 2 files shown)


stof:
Thank you @fabpot.

fabpot:
@stof You was too fast, I've just updated the PR, can you merge it again?

stof:
@fabpot the collector template is already updated in the PR updating it for 2.8. Merging this updated PR would create conflicts with #452 while it is almost ready.

fabpot:
Cool, thanks

Issue #441, opened 2015-06-24
Doctrine mapping import ignores filter when checking tables for primary keys


markusgillmeister:
I ran into an issue when importing a mapping from a single table. The database has some tables without primary keys which are not important for my import. So I use the filter parameter to get only the single table. But doctrine does a complete database check and throws a MappingException when scanning the table without the primary key.

> console doctrine:mapping:import --em=customer **--filter="tbl_users"** AcmeHelloBundle xml

>  [Doctrine\ORM\Mapping\MappingException]
>  **Table tbl_abc has no primary key**. Doctrine does not support reverse engineering from tables that don't have a primary key.

I think it would be better if the mapping import only does the check on the mentioned table in the filter parameter und do the import.

mat-m:
I confirm the issue

axeff:
+1

jderusse:
:+1:

axeff:
As workaround I created a new database with just the table I wanted to import.

remmel:
agree

yezongyang:
+1

yangconghou:
agree verymuch!!

mikeSimonson:
If I remember well the table filter works if you use it on the connection directly.
Like this

```yml
doctrine:
    dbal:
        default_connection: default
        connections:
            default:
                driver:   "%database_driver%"
                host:     "%database_host%"
                port:     "%database_port%"
                dbname:   "%database_name%"
                user:     "%database_user%"
                password: "%database_password%"
                charset:  UTF8
            customer:
                driver:   "%database_driver%"
                host:     "%database_host%"
                port:     "%database_port%"
                dbname:   "%database_name%"
                user:     "%database_user%"
                password: "%database_password%"
                charset:  UTF8
                schema_filter: tbl_users
```

And I use the customer connection to create the customer entity manager

```yml
orm:
        auto_generate_proxy_classes: "%kernel.debug%"
        default_entity_manager: default
        entity_managers:
            default:
                connection: default
                mappings:
                            .....
            customer:
                connection: customer #Here comes the already filtered connection
```

Now you can just pass your custom entity manager (customer) with your filtered connection (customer) to the import command.

schematical:
+1

fabianofa:
+1 on this.

@mikeSimonson answer works with, but you have to wrap the value with some delimiters like `$`,  otherwhise you'll get `Warning: preg_match(): Delimiter must not be alphanumeric or backslash`

[...]

PR #473, opened 2015-10-10
fixes #317 #441 Changed filtering method when importing mappings

### LINKED ISSUE #317: Import Mapping Command Filter  Option Still Reads Entire DB
The --filter option for the import mapping command only applies when Doctrine writes the meta data to a file, not when doctrine interprets the DB. This becomes problematic when people try to import specific tables, using --filter, that are totally fine, however other tables throw errors (such as a missing primary key or an unknown field type). Doctrine should not prevent the import of desired tables because undesired tables cause errors. 

It feels like --filter should apply to the database read. Does this sound like a fair request or is there a reason for --filter's current behavior?

federicop91:
The MetadataFilter filtered table names containing any of the —filter arguments. We have to avoid getting all metadata with the Filter Schema Assets Expression which uses regular expressions.

### DIFF
--- a/Command/ImportMappingDoctrineCommand.php
+++ b/Command/ImportMappingDoctrineCommand.php
@@ -105,8 +105,11 @@ EOT
 
         $cmf = new DisconnectedClassMetadataFactory();
         $cmf->setEntityManager($em);
+        if($filters = $input->getOption('filter')){
+        	$regExpFilter = '/'.implode('|',$filters).'/';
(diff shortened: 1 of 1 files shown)


stof:
This is a BC break, because you are now filtering tables while the ``--filter`` option currently filters entity names, which might not be the same (and btw, you don't even change the doc, so the description of the command option is now wrong)

franbenz:
What would be the right thing to do? Should this kind of filter be added as a new --filter-table-name argument?

kimhemsoe:
Closing, BC break

PR #477, opened 2015-10-15
Refactored the doctrine_pretty_query filter to support "highlight only" mode

### LINKED ISSUE #469: Merge the "doctrine_pretty_query" and "doctrine_highlight_sql" Twig filters
As @stof commented in https://github.com/doctrine/DoctrineBundle/pull/452#discussion_r40557258

> well, this method does exactly the same than the new doctrine_highlight_sql, except that one is using the profiler styles and the other one uses the raw SqlFormatter colors. This looks weird to me.

javiereguiluz:
This fixes #469

### DIFF
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -159,12 +159,52 @@
                 </tr>
                 </thead>
                 <tbody id="queries-{{ loop.index }}">
+                {% for i, query in queries %}
+                    <tr id="queryNo-{{ i }}-{{ loop.parent.loop.index }}" class="{{ cycle(['odd', 'even'], i) }}">
+                        <td>{{ loop.index }}</td>
+                        <td>{{ '%0.2f'|format(query.executionMS * 1000) }}&nbsp;ms</td>
+                        <td>
+                            <div class="query-section" data-state="collapsed" onclick="return expandQuery(this);" title="Expand query" data-target-id="code-{{ i }}-{{ loop.parent.loop.index }}" style="cursor: pointer;">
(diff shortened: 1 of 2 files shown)


stof:
@javiereguiluz please clean your branch from all these merge commits. It looks like your local master branch diverged form the upstream one.

stof:
ok, diff got lost now

javiereguiluz:
The PR is now ready to be reviewed. Thanks.

Twig/DoctrineExtension.php
```diff
@@ -327,7 +328,12 @@ public function highlightSqlSyntax($sql)
...
         \SqlFormatter::$variable_attributes = 'class="variable"';
 
-        $html = \SqlFormatter::highlight($sql);
+        if ($highlightOnly) {
+            $html = \SqlFormatter::highlight($sql);
+        } else {
+            $html = \SqlFormatter::format($sql);
+        }
+
         $html = preg_replace('/<pre class="(.*)">([^"]*+)<\/pre>/Us', '<div class="\1"><pre>\2</pre></div>', $html);
```
stof: this could be simplified, by changing ``SqlFormatter::$use_pre`` instead

stof: and when highlighting only, should we really return a ``<pre>`` tag ?

Resources/views/Collector/db.html.twig
```diff
@@ -188,7 +188,7 @@
                                 </div>
 
                                 <div id="original-query-{{ i }}-{{ loop.parent.loop.index }}" class="sql-runnable hidden">
-                                    {{ (query.sql ~ ';')|doctrine_replace_query_parameters(query.params)|doctrine_highlight_sql }}
+                                    {{ (query.sql ~ ';')|doctrine_replace_query_parameters(query.params)|doctrine_pretty_query(highlight_only = true) }}
```
stof: Do we actually need to highlight the runnable query ?

javiereguiluz: Yes, to make it look beautiful on the screen.

Twig/DoctrineExtension.php
javiereguiluz: We can't use `$use_pre` because is a "global state" and we need to output different things depending whther we highlight or not. I've just updated the code.

PR #483, opened 2015-11-05
Fix creating sqlite database

hason:
Fixes #482.

PR #488, opened 2015-11-15
fix quote sqlite path

vvh-empora:


j0k3r:
Already fixed by #483

PR #489, opened 2015-11-24
Allow parameter in entity attribute

sergeyz:


sergeyz:
Works good without this changes, doc are outdated little bit

PR #498, opened 2015-12-04
Marking new types as commented event if no custom type mappings are present.

TiS:
I have found an issue in ConnectionManager, wchich relates marking new custom types as commented with existence of custom type mappings. These two things are unrelated of each other.

Explanation. ConnectionManager is being constructed with $typesConfig argument (passed from bundle configuration), containing new types added by user via config.yml. Then, during connection creation, it calls $this->initializeTypes(); function, which adds these types to Type registry.

Then, something strange happens. If the type has 'commented' parameter set to true, it should have been mark as commented in database platform. But this operation only happened, when in addition of custom type, there was custom mapping type (any) passed to createConnection function.

When user defines a custom Type, it does not have to (and in most cases, he will not) define custom type **MAPPING**, as custom type definition already has a mapping to a standard Doctrine field type. Also, in most cases, he wants to have his new type commented, and relating this action of existence of type mapping makers no sense.

This PR corrects this behavior, relating marking type as commented solely with the existence of marked type.

### DIFF
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -64,7 +64,7 @@ class ConnectionFactory
             }
         }
         if (!empty($this->commentedTypes)) {
-            if (!$platform) {
+            if (empty($platform)) {
                 $platform = $connection->getDatabasePlatform();
             }
             foreach ($this->commentedTypes as $type) {
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -62,6 +62,11 @@ class ConnectionFactory
             foreach ($mappingTypes as $dbType => $doctrineType) {
                 $platform->registerDoctrineTypeMapping($dbType, $doctrineType);
             }
+        }
+        if (!empty($this->commentedTypes)) {
+            if (!$platform) {
+                $platform = $connection->getDatabasePlatform();
+            }
             foreach ($this->commentedTypes as $type) {
                 $platform->markDoctrineTypeCommented(Type::getType($type));
             }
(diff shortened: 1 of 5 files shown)


TiS:
Heh, I have corrected tests. It seems, that incorrect custom type was checked in tests (wrong namespace). Probably error corrected in this PR prevented those test from correct execution...

stof:
I'm actually thinking about removing this entirely, as it breaks lazy-loading of the platform.

There is a way to mark types as commented in the Type class itself (since DBAL 2.2 or 2.3, I don't remember), which is way better as it does not require getting the platform on boot (which breaks things in case the server version needs to be guessed)

TiS:
I agree that this will be a better idea. 
But **before** this happens... :)

stof:
@TiS using the DBAL native feature to mark types as commented is already possible (it works since the day it was implemented in DBAL, i.e. years ago)

TiS:
That's true. But 1 - the library still is broken for people, who doesn't know that and 2 - this PR also corrects some tests, which were incorrectly written, but passed because of this error.

TiS:
Ah, maybe I'll clarify what I mean. Currently it is impossible to configure custom types in config.yml, if you don't add custom type mappings. You can do that programatically, of course. But the idea of config.yml is to not require additional programming to add thisngs. This PR corrects this, besides being not the perfect solution (as you want to rewrite this feature completely).

mikeSimonson:
@stof What about this ? To me it make sense to merge it.

stof:
Closing as #623 was merged, implementing this.

PR #499, opened 2015-12-08
Avoid loading of all manager services on getAliasNamespace function call

chiriacbogdan:
Every time someone writes AcmeDemoBundle:SomeEntity instead of Acme\DemoBundle\Entity\SomeEntity the function getAliasNamespace is called witch loads all entity manager services currently configured which initializes a connection to the database even if it is not necessary.

### DIFF
--- a/Registry.php
+++ b/Registry.php
@@ -144,7 +144,7 @@ class Registry extends ManagerRegistry implements RegistryInterface
      */
     public function getAliasNamespace($alias)
     {
-        foreach (array_keys($this->getManagers()) as $name) {
+        foreach ($this->getManagerNames() as $name => $id) {
             try {
                 return $this->getManager($name)->getConfiguration()->getEntityNamespace($alias);
             } catch (ORMException $e) {


kimhemsoe:
I see the problem, but this is not the way to fix it.
To fix the problem you have i suggest you try specified what version of database you are trying to use. This should fix the issue where dbal connects to autodetect what db version you are running with.

kimhemsoe:
Closing for now.

mihai-stancu:
@kimhemsoe the PR provides a way to avoid an unexpected (unwanted?) side effect of particular action. Is there a reason why this unexpected side effect would in fact be **wanted**?

More clearly: is there a reason why `getAliasNamespace($alias)` **should** trigger the initialization of all managers (and generate connections and autodetect versions etc.)?

If not then IMHO this PR **is** the way to fix the problem. Specifying the mysql version in your configuration seems like a cumbersome workaround.

+1 for this fix.

kimhemsoe:
The PR fixes nothing, it just add a new problem, which is, now we are depending on what order the managers are registered. 
Lets say you have A and B. Now you add C and you start to see this error, that is not really nice and can lead to weird "ohh yeah.. you just need to change the bundle ordering or something.. comments" 
We do not do not want that, we rather want to fail from the beginning so people understand the issue correct the first time.

Some dbal drivers need to connect to the database if someone is asking of the platform. I am guessing that that is the issue we have here. I do not know what is using the platform this early and it may even only be a issue while developing and not with caches enabled.

No, loading everything just to read a string from the config is not perfect. But it is not that easy to fix i believe.

Specifying the db version is not bad idea and it is not much more different then you need to tell doctrine what driver you want to use. It fixes **alot** of other problems too, like creation of database and etc. It also have the nice side effect that it is faster then auto detecting and you may even avoid connecting to the database if you do not do a query.

PR #501, opened 2015-12-10
Fix reference to not existing ContainerAware

### LINKED ISSUE #500: ContainerAware not found
```
Fatal Error: Class 'Symfony\Component\DependencyInjection\ContainerAware' not found
```
occurs when trying to use "Explain query" option via Symfony Profiler -> Doctrine section.
It is caused because `Symfony\Component\DependencyInjection\ContainerAware` was deprecated and removed in favour of `Symfony\Component\DependencyInjection\ContainerAwareTrait`

Trace:
1) https://github.com/symfony/symfony/blob/v3.0.0/composer.json#L79
2) https://github.com/doctrine/DoctrineBundle/blob/v1.4.0/Controller/ProfilerController.php#L27

Found using:
- Symfony 3.0.0
- PHP 7.0

werdlv:
Fix #500

### DIFF
--- a/Controller/ProfilerController.php
+++ b/Controller/ProfilerController.php
@@ -16,7 +16,8 @@ namespace Doctrine\Bundle\DoctrineBundle\Controller;
 
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Platforms\SQLServerPlatform;
-use Symfony\Component\DependencyInjection\ContainerAware;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 use Symfony\Component\HttpFoundation\Response;
 
 /**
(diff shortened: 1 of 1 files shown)


stloyd:
This is "no-go" till this bundle drop support for PHP 5.3. Better would be to implement methods of that deprecated class into that controller.

werdlv:
@stloyd that was my concern. I would vote for dropping support for PHP 5.3(which is not supported for more then year now) - that`s why releases and semantic versioning are for, right? :)

hacfi:
@stof Do you see any other way of solving this besides dropping support?

xabbuh:
You can simply implement the interface and then implement the`setContainer()` method yourself.

stof:
Dropping PHP 5.3 support is indeed a no-go. As 2.8 needs to ship with PHP 5.3 compatible vendors for the next few years, it means we would have to maintain 2 branches of DoctrineBundle to keep a supported version of the bundle included in Symfony 2.8 releases. I don't want to maintain 2 branches just to avoid writing 3 lines of code

werdlv:
@stof makes sense

stof:
Closing as I merged #505 providing the fix with PHP 5.3 compat

PR #502, opened 2015-12-10
Register the PropertyInfo extractor from the Symfony Bridge

dunglas:
Allow to extract metadata about properties coming from Doctrine using the new PropertyInfo component of Symfony 2.8.

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -771,7 +771,7 @@ class DoctrineExtension extends AbstractDoctrineExtension
      * @param string           $entityManagerName
      * @param ContainerBuilder $container
      */
-    protected function loadPropertyInfoExtractor($entityManagerName, ContainerBuilder $container)
+    private function loadPropertyInfoExtractor($entityManagerName, ContainerBuilder $container)
     {
         $metadataFactoryService = sprintf('doctrine.orm.%s_entity_manager.metadata_factory', $entityManagerName);
 
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -396,6 +396,14 @@ class DoctrineExtension extends AbstractDoctrineExtension
                 $def->addTag('doctrine.event_subscriber');
             }
         }
+
+        if (!class_exists('Symfony\Bridge\Doctrine\PropertyInfo\DoctrineExtractor')) {
+            return;
+        }
+
+        foreach ($config['entity_managers'] as $name => $entityManager) {
+            $this->loadPropertyInfoExtractor($name, $container);
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


DependencyInjection/DoctrineExtension.php
```diff
@@ -396,6 +396,14 @@ protected function ormLoad(array $config, ContainerBuilder $container)
                 $def->addTag('doctrine.event_subscriber');
             }
         }
+
+        if (!class_exists('Symfony\Bridge\Doctrine\PropertyInfo\DoctrineExtractor')) {
+            return;
+        }
+
+        foreach ($config['entity_managers'] as $name => $entityManager) {
+            $this->loadPropertyInfoExtractor($name, $container);
```
stof: don't loop a second time. Do it in ``loadOrmEntityManager``

Tests/ContainerTest.php
```diff
@@ -50,6 +50,8 @@ public function testContainer()
         $this->assertInstanceOf('Symfony\Bridge\Doctrine\CacheWarmer\ProxyCacheWarmer', $container->get('doctrine.orm.proxy_cache_warmer'));
         $this->assertInstanceOf('Doctrine\Common\Persistence\ManagerRegistry', $container->get('doctrine'));
         $this->assertInstanceOf('Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntityValidator', $container->get('doctrine.orm.validator.unique'));
+        $this->assertInstanceOf('Doctrine\Common\Persistence\Mapping\ClassMetadataFactory', $container->get('doctrine.orm.default_entity_manager.metadata_factory'));
```
stof: you need to check for the existence of the extractor

DependencyInjection/DoctrineExtension.php
```diff
@@ -396,6 +396,14 @@ protected function ormLoad(array $config, ContainerBuilder $container)
                 $def->addTag('doctrine.event_subscriber');
             }
         }
+
+        if (!class_exists('Symfony\Bridge\Doctrine\PropertyInfo\DoctrineExtractor')) {
```
stof: you should check first for the PropertyInfo component, as the bridge does not force installing PropertyInfo (but we force installing the bridge)

dunglas: If the component is not installed, those services will be removed as they are private and unused.

Tests/ContainerTest.php
dunglas: It's checked the next line.

dunglas:
@stof PR updated

dunglas:
Is it possible to merge this PR?

fabpot:
:+1:

DependencyInjection/DoctrineExtension.php
```diff
@@ -759,6 +766,29 @@ protected function loadOrmCacheDrivers(array $entityManager, ContainerBuilder $c
     }
 
     /**
+     * Loads a property info extractor for each defined entity manager.
+     *
+     * @param string           $entityManagerName
+     * @param ContainerBuilder $container
+     */
+    protected function loadPropertyInfoExtractor($entityManagerName, ContainerBuilder $container)
```
xabbuh: should be `private` imo

composer.json
```diff
@@ -36,6 +36,7 @@
         "doctrine/orm": "~2.3",
         "symfony/yaml": "~2.2|~3.0",
         "symfony/validator": "~2.2|~3.0",
+        "symfony/property-info": "~2.8|~3.0",
```
xabbuh: What about adding this to the `suggest` section too?

dunglas: done

dunglas:
@xabbuh's comments fixed.

composer.json
```diff
@@ -36,14 +36,16 @@
...
         "twig/twig": "~1.10",
         "satooshi/php-coveralls": "~0.6.1",
         "phpunit/phpunit": "~4"
     },
     "suggest": {
-        "symfony/web-profiler-bundle": "to use the data collector",
-        "doctrine/orm": "The Doctrine ORM integration is optional in the bundle."
+        "symfony/web-profiler-bundle": "To use the data collector.",
+        "doctrine/orm": "The Doctrine ORM integration is optional in the bundle.",
+        "symfony/property-info": "To use the PropertyInfo integration."
```
stof: I'm not sure this suggestion makes sense. Saying "install this component to use this component" looks weird. People wanting to use PropertyInfo will install it. And DoctrineBundle does not provide any feature based on the PropertyInfo which would let people wonder whether they want to use PropertyInfo. It only provides the inner parts for PropertyInfo.

kimhemsoe:
Status?

composer.json
dunglas: Finally removed as asked by @stof (I also think it doesn't add any help to suggest this component).

dunglas:
@kimhemsoe finished. Can be merged (according to me).

dunglas:
Sorry to insist, but is it possible to merge this one?

mikeSimonson:
@dunglas Thanks

dunglas:
@mikeSimonson thank you very much for the review and the merge!

PR #503, opened 2015-12-12
Fix ghost item in toolbar when no query

ogizanagi:
Fix https://github.com/symfony/symfony/issues/16976.

However, the item will also be hidden from the toolbar when using the old profiler/toolbar theme. 
I don't think it matters, but tell me if I'm wrong.

### DIFF
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -1,23 +1,24 @@
 {% extends app.request.isXmlHttpRequest ? '@WebProfiler/Profiler/ajax_layout.html.twig' : '@WebProfiler/Profiler/layout.html.twig' %}
 
 {% block toolbar %}
-    {% set profiler_markup_version = profiler_markup_version|default(1) %}
+    {% if collector.querycount > 0 or collector.invalidEntityCount > 0 %}
 
-    {% set icon %}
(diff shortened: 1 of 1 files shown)


dosten:
:+1:

stof:
Good catch, thanks @ogizanagi.

javiereguiluz:
@ogizanagi I'm sorry I didn't see your pull request earlier. I do think that the change in the old behavior matters. People are used to see all toolbar panels when using old Symfony applications. If they now don't see the Doctrine panel, they'll think that something wrong happened. In #508 I solved this issue in a different way to respect the old behavior, just in case someone else thinks this is important.

ogizanagi:
@javiereguiluz : I don't know actually. Retrospectively, I don't think people would expect to have the doctrine toolbar item on a page where there is no query at all.
And if they do, they'll go to the profiler page and understand there is no query. So they'll wonder themselves why the toolbar item is not shown only once, and then benefit from the space cleared.

Of course I agree a change in the behavior is not great in what is supposed to be a bug fix, that's why I asked. ^^

stof:
@javiereguiluz some panels were already hidden in the old toolbar (SwiftmailerBundle for instance). It was just not a convention applied consistently

PR #508, opened 2015-12-22
Hide the toolbar panel properly when there are no queries or errors

javiereguiluz:
### Before

doctrine_before

### After

doctrine_after


javiereguiluz:
Closing it because this was already fixed in a previous PR.

PR #509, opened 2016-01-10
Fixed incorrect link text after clicking on it

pborreli:
Before : 

doctrine-before

After : 

doctrine-after


kimhemsoe:
Thanks @pborreli

PR #513, opened 2016-01-28
update prefix to sf3_

HeahDude:


stof:
I would rather use an example which does not contain the Symfony version, to avoid getting the same issue again

PR #518, opened 2016-02-23
Fix count break on queries list for profiler

blazarecki:
Hi,

This PR fix the count break on the query list for count > 100.

*Before*

screen shot 2016-02-23 at 16 41 22

*After*

screen shot 2016-02-23 at 16 44 38


mikeSimonson:
Thanks @blazarecki

PR #519, opened 2016-03-01
Add "Show Results" link in "Queries" section in Profiler

loalf:
It's been bothering me for a while the fact that sometimes I want to see the results coming back from a query. In those cases, I have to copy the query from the profiler, either use a command line tool or a UI to connect to the database (which usually means copying and pasting the username/password), and finally see the results.

Wouldn't it be great if I could do that from the Web Profiler?! WAIT NO MORE!

captura de pantalla 2016-03-01 a la s 17 14 20

This PR adds a "Show Results" link, when clicked it shows the results.

captura de pantalla 2016-03-01 a la s 17 15 42

It might be the case that there are a lot of columns in the results, in those cases this PR also takes care of this just showing the first six columns and providing buttons to hide/show the columns as you wish. That way, you can focus on the data you are really interested into.

captura de pantalla 2016-03-01 a la s 17 18 24

### DIFF
--- a/Controller/ProfilerController.php
+++ b/Controller/ProfilerController.php
@@ -85,6 +85,35 @@ class ProfilerController implements ContainerAwareInterface
         ));
     }
 
+    public function executeAction($token, $connectionName, $query)
+    {
+        $queryId = $query;
+
+        /** @var $profiler \Symfony\Component\HttpKernel\Profiler\Profiler */
+        $profiler = $this->container->get('profiler');
+        $profiler->disable();
+
+        $profile = $profiler->loadProfile($token);
+        $queries = $profile->getCollector('db')->getQueries();
+
+        if (!isset($queries[$connectionName][$query])) {
+            return new Response('This query does not exist.');
+        }
+
+        /** @var $connection \Doctrine\DBAL\Connection */
+        $connection = $this->container->get('doctrine')->getConnection($connectionName);
+        $query = $queries[$connectionName][$query];
+
+        $results = $connection
+            ->executeQuery($query['sql'], $query['params'], $query['types'])
+            ->fetchAll(\PDO::FETCH_ASSOC);
+
+        return $this->container->get('templating')->renderResponse('@Doctrine/Collector/execute.html.twig', array(
+            'data' => $results,
+            'id'   => $queryId
+        ));
+    }
+
     private function explainSQLServerPlatform(Connection $connection, $query)
     {
         if (stripos($query['sql'], 'SELECT') === 0) {
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -109,8 +109,8 @@
 {% block panel %}
     {% set profiler_markup_version = profiler_markup_version|default(1) %}
 
-    {% if 'explain' == page %}
-        {{ render(controller('DoctrineBundle:Profiler:explain', {
+    {% if ('explain' == page or 'execute' == page) %}
+        {{ render(controller(('DoctrineBundle:Profiler:' ~ page), {
             token: token,
             panel: 'db',
             connectionName: app.request.query.get('connection'),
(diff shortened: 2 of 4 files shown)


Controller/ProfilerController.php
```diff
@@ -85,6 +85,35 @@ public function explainAction($token, $connectionName, $query)
...
+
+        /** @var $profiler \Symfony\Component\HttpKernel\Profiler\Profiler */
+        $profiler = $this->container->get('profiler');
+        $profiler->disable();
+
+        $profile = $profiler->loadProfile($token);
+        $queries = $profile->getCollector('db')->getQueries();
+
+        if (!isset($queries[$connectionName][$query])) {
+            return new Response('This query does not exist.');
```
stof: this should have a 404 status code IMO

Resources/views/Collector/db.html.twig
```diff
@@ -373,24 +362,30 @@
             }
         }
 
-        {% if profiler_markup_version == 1 %}
-            function toggleRunnableQuery(target) {
```
stof: Please don't break the differences between the v1 and v2 profilers. We want to support Symfony 2.8 properly, and ``toggleRunnableQuery`` is necessary only on older versions

Resources/views/Collector/db.html.twig
```diff
@@ -373,24 +362,30 @@
...
+                targetElement.style.display = 'none';
+                target.innerHTML = targetDataAltContent;
             }
-        {% endif %}
+
+            return false;
+        }
 
         //]]></script>
+    {% include '@Doctrine/Collector/_toggle_column.html.twig' %}
```
stof: I don't see any reason to use a separate template for this

Resources/views/Collector/execute.html.twig
```diff
@@ -0,0 +1,43 @@
...
+{% else %}
+    <pre style="margin: 5px 0;">
+        {%- for row in data -%}
+            {{ row|first }}{{ "\n" }}
+        {%- endfor -%}
+    </pre>
+{% endif %}
+
+{% if not app.request.isXmlHttpRequest %}
+    {% include '@Doctrine/Collector/_toggle_column.html.twig' %}
```
stof: This does not make sense. the page never includes a full layout anyway

Resources/views/Collector/execute.html.twig
```diff
@@ -0,0 +1,43 @@
...
+            </tr>
+        {% endfor %}
+        </tbody>
+    </table>
+{% else %}
+    <pre style="margin: 5px 0;">
+        {%- for row in data -%}
+            {{ row|first }}{{ "\n" }}
+        {%- endfor -%}
+    </pre>
```
stof: I don't think this makes sense here. Displaying results as a text is not something which makes sense in the general case (it does only for queries optimized for that, like the postgresql explain command) and it would break things for queries selecting a single column (for instance ``SELECT email FROM users`` to get the list of all used emails)

Resources/views/Collector/db.html.twig
```diff
@@ -193,8 +197,14 @@
                                     {{ (query.sql ~ ';')|doctrine_replace_query_parameters(query.params)|doctrine_pretty_query(highlight_only = true) }}
                                 </div>
 
+                                <div id="execute-{{ i }}-{{ loop.parent.loop.index }}" class="sql-runnable hidden">
```
stof: if the query is not explainable, it will not be runnable either (a non-explainable query means that we lost some info when serializing parameters, and so we cannot execute it again)

Resources/views/Collector/db.html.twig
```diff
@@ -109,8 +109,8 @@
 {% block panel %}
     {% set profiler_markup_version = profiler_markup_version|default(1) %}
 
-    {% if 'explain' == page %}
-        {{ render(controller('DoctrineBundle:Profiler:explain', {
+    {% if ('explain' == page or 'execute' == page) %}
+        {{ render(controller(('DoctrineBundle:Profiler:' ~ page), {
```
stof: braces are useless

Controller/ProfilerController.php
```diff
@@ -85,6 +85,35 @@ public function explainAction($token, $connectionName, $query)
...
+        }
+
```
stof: you should check for ``$query['explainable']`` here (if it is false, you don't have the original params in the storage, and so you cannot run the query again)

kimhemsoe:
@loalf Status?

kimhemsoe:
@loalf Ping

kimhemsoe:
Closing. Author have gone silent.

PR #523, opened 2016-03-11
Close connection after database drop

cepauskas:
Database connection is not closed after database is being dropped. This causes issues when subsequent commands are being executed in certain DBMS (fe: postgresql fails on it, as after database drop connection to database is still active and new database can not be created as resource is still in use).

Provided fix would fixes such scenario:

```php
{
    $application = $this->getApplication()->find('doctrine:database:drop');
    $application->run(new ArrayInput(array(
        'command' => 'doctrine:database:drop',
        '--force'   => true,
        '--if-exists'   => true
    )), $output);

    $application = $this->getApplication()->find('doctrine:database:create');
    $application->run(new ArrayInput(array(
        'command' => 'doctrine:database:create',
        '--if-not-exists'   => true
    )), $output);
}
```

### DIFF
--- a/Command/DropDatabaseDoctrineCommand.php
+++ b/Command/DropDatabaseDoctrineCommand.php
@@ -96,7 +96,9 @@ EOT
                 } else {
                     $output->writeln(sprintf('<info>Database for connection named <comment>%s</comment> doesn\'t exist. Skipped.</info>', $name));
                 }
+                $connection->close();
             } catch (\Exception $e) {
+                $connection->close();
                 $output->writeln(sprintf('<error>Could not drop database for connection named <comment>%s</comment></error>', $name));
                 $output->writeln(sprintf('<error>%s</error>', $e->getMessage()));
 
--- a/Tests/Command/DropDatabaseDoctrineTest.php
+++ b/Tests/Command/DropDatabaseDoctrineTest.php
@@ -1,0 +1,119 @@
+<?php
+
+/*
+ * This file is part of the Doctrine Bundle
+ *
+ * The code was originally distributed inside the Symfony framework.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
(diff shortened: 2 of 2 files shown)


kimhemsoe:
I am not 100% sure about this fix. Maybe dbal should handle this?
Could a solution not be that you ran your commands in separate processes?

cepauskas:
Yes, in case when commands are run in separated processes everything goes OK, but the problem is that we have an command which prepares development environment, it is a good step for automatization (and I suppose not only we are doing it).

As Symfony by it self provides subsequent commands, I would expect that dropping and creating it afterwards should work without issues. As an alternative would be having bash file and execute separated commands, but I suppose commands purpose and was to do such things without having bash files.

 I'm not sure if it is dbal layer issue, as it is logic which comes from postgres, you can't create new database with same name/user till previous connection is open. Connection is not closed in dbal after database drop, because afterwards you might want to execute some other queries within open connection - remove users, permissions, etc. So I would say - this issue can be fixed in bundle instead of dbal?

kimhemsoe:
I would requirement that you use Symfony Process for now to run the drop and create.
Why also create you may ask? Well i believe that dbal postgres is not connected to the correct database when the dbal driver is connecting if the database does not exists. In current stable it will connect to template1 and in master it is connecting to the postgresql database.
Alternative you could go around these 2 commands and run the sql statements manually from your command. 

About your fix i will try ask around about what approach we should take from here.

cepauskas:
It fails when database is being created immediately after it's drop when connection (which dropped database) is not closed (new database can not be created as resource is still in use). 

Anyway, thanks for your investigation.

kimhemsoe:
Closing as wont fix for now. I see the case but there is many ways to work around the issue and rather want that then possible introduce hard to debug issues in the future.

PR #532, opened 2016-03-31
Add SAP Adaptive Server Enterprise support for explain queries

Deltachaos:
This PR adds support for SAP ASE explain queries. It depends on doctrine/dbal#2347.

I hope this addition is appreciated and any feedback/help on the above issues would be welcomed.

### DIFF
--- a/Controller/ProfilerController.php
+++ b/Controller/ProfilerController.php
@@ -108,7 +108,7 @@ class ProfilerController implements ContainerAwareInterface
 
         if ($stmt instanceof ASEStatement) {
             $data = $stmt->getMessages();
-            $data = implode("\n", $data);
+            $data = implode("", $data);
             return array(array($data));
         }
 
--- a/Controller/ProfilerController.php
+++ b/Controller/ProfilerController.php
@@ -15,6 +15,8 @@
 namespace Doctrine\Bundle\DoctrineBundle\Controller;
 
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver\ASE\ASEStatement;
+use Doctrine\DBAL\Platforms\ASEPlatform;
 use Doctrine\DBAL\Platforms\SQLServerPlatform;
 use Symfony\Component\DependencyInjection\ContainerAwareInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
@@ -72,6 +74,8 @@ class ProfilerController implements ContainerAwareInterface
         try {
             if ($connection->getDatabasePlatform() instanceof SQLServerPlatform) {
                 $results = $this->explainSQLServerPlatform($connection, $query);
+            } else if ($connection->getDatabasePlatform() instanceof ASEPlatform) {
+                $results = $this->explainASEServerPlatform($connection, $query);
             } else {
                 $results = $this->explainOtherPlatform($connection, $query);
             }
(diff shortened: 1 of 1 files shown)


mikeSimonson:
@Deltachaos Thanks
I will wait for the doctrine/dbal#2347 PR to be merged before to merge this.

stof:
Rather than making the code even more complex here to deal with each platform, I think we should add a new method in the AbstractPlatform in DBAL: ``getExplainSql($sql)``. This way, all projects integrating Doctrine DBAL could benefit from it instead of duplicating the logic.
What do you think @mikeSimonson @kimhemsoe ?

kimhemsoe:
Ive already aired this idea on IRC a few weeks back. I see to that it gets implemented soonish

Deltachaos:
Would be grate to do it like this!

Deltachaos:
@kimhemsoe i have thought a bit about this topic again an noticed that getting the execution plan for different platforms could be very different.
If you want to get it for ASE the server dont provide the results in a "table like" form, instead it is list of text messages. Something like this:

```
QUERY PLAN FOR STATEMENT 1 (at line 0).


    STEP 1
        The type of query is DECLARE.


QUERY PLAN FOR STATEMENT 2 (at line 1).
Optimized using Parallel Mode


    STEP 1
        The type of query is SELECT.

	5 operator(s) under root

       |ROOT:EMIT Operator (VA = 5)
       |
       |   |RESTRICT Operator (VA = 4)(0)(0)(0)(0)(3)
       |   |
       |   |   |NESTED LOOP JOIN Operator (VA = 3) (Join Type: Inner Join)
       |   |   |
       |   |   |   |RESTRICT Operator (VA = 1)(0)(0)(0)(7)(0)
       |   |   |   |
       |   |   |   |   |SCAN Operator (VA = 0)
       |   |   |   |   |  FROM TABLE
       |   |   |   |   |  cars
       |   |   |   |   |  d0_
       |   |   |   |   |  Index : IDX_84A8793B700047D2
       |   |   |   |   |  Forward Scan.
       |   |   |   |   |  Positioning by key.
       |   |   |   |   |  Keys are:
       |   |   |   |   |    car_id ASC
       |   |   |   |   |  Using I/O Size 2 Kbytes for index leaf pages.
       |   |   |   |   |  With LRU Buffer Replacement Strategy for index leaf pages.
       |   |   |   |   |  Using I/O Size 2 Kbytes for data pages.
       |   |   |   |   |  With LRU Buffer Replacement Strategy for data pages.
       |   |   |
       |   |   |   |SCAN Operator (VA = 2)
       |   |   |   |  FROM TABLE
       |   |   |   |  donat_Ticket
       |   |   |   |  d1_
       |   |   |   |  Using Clustered Index.
       |   |   |   |  Index : car_12277730462
       |   |   |   |  Forward Scan.
       |   |   |   |  Positioning by key.
       |   |   |   |  Index contains all needed columns. Base table will not be read.
       |   |   |   |  Keys are:
       |   |   |   |    id ASC
       |   |   |   |  Using I/O Size 2 Kbytes for index leaf pages.
       |   |   |   |  With LRU Buffer Replacement Strategy for index leaf pages.
```

We should keep this in mind for a proper solution. For this reason i think it would make sens if we add a "explain" method to the ``Driver`` class.
But we also need to unify the return type of this method. Maybe it can return some kind of ``ExecutionPlan`` object wich can return the execution plan in a table like form and as plaintext.
This ``ExecutionPlan`` object could then be implemented by the different drivers in a way which the platform requires it.

stof:
@Deltachaos Postgresql also gives a list of text messages. We handle it properly in the display in the bundle (detecting whether we have 1 column or several ones in the output).

The method in the AbstractPlatform will not return an ExectionPlan object. It will return some SQL (the platform is about generating SQL).

Deltachaos:
@stof the problem here is that for ASE the list of messages is not part of the result set. Therefore just generating the correct SQL in the ``AbstractPlatform`` would not solve the problem. Executing this SQL would not return anything.
For ASE the server produces "messages" additional to result sets. For example the following query

```sql
SELECT foo FROM bar
PRINT 'test'
```
would return in ASE of course a result set with the results from the SELECT but additionally there is a list of messages available which contains the "test" message and maybe other messages produced by the query. Those are not real result sets and therefore not accessible through the default ``Statement`` interface. It is a very platform specific thing.
And the execution plan in ASE is also send to the client as such a message.
This is the reason why i have called ``getMessages`` in line 110 of this PR.

stof:
Oh, they have their custom feature to attach messages to a result set ? Then indeed, this would not fit in a case where we would just want to generate the SQL explaining a query

Deltachaos:
@stof is there a default feature for attaching messages to result sets? Because simply adding them after the result set would change the meaning of the result right?
Or how is this done in postgresql. Dose postgresql also generate messages for normal queries like selects or do the also have such a print command available?

kimhemsoe:
Closing. This won't happen until DBAL get a generic EXPLAIN api.

PR #533, opened 2016-04-01
Highlight query count box according to the db toolbar icon status

bocharsky-bw:
According to the [status](https://github.com/bocharsky-bw/DoctrineBundle/blob/patch-1/Resources/views/Collector/db.html.twig#L22) variable.

### DIFF
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -41,7 +41,7 @@
         {% set text %}
             <div class="sf-toolbar-info-piece">
                 <b>Database Queries</b>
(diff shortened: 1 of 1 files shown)


mikeSimonson:
@bocharsky-bw Did you tested that it indeed change the color in the v2 ?

bocharsky-bw:
@mikeSimonson Yes, I did! It's the same behaviour like for [Invalid entities](https://github.com/bocharsky-bw/DoctrineBundle/blob/patch-1/Resources/views/Collector/db.html.twig#L52), but with the `yellow` colour.

mikeSimonson:
@bocharsky-bw Thanks for the confirmation and the patch

PR #534, opened 2016-04-01
Add global metrics on top of Doctrine page

raziel057:
As the toolbar is not displayed any more on top of the profiler with Symfony 3.x, we don't have direct information about:
- the count of database queries
- the global query time
- the count of invalid entities

I think this information could be displayed on top of the page as global query metrics, like global metrics on other pages in the Web Profiler.

### DIFF
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -140,6 +140,40 @@
         </style>
     {% endif %}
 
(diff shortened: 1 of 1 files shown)


raziel057:
Here is the result:

global-metrics

mikeSimonson:
@raziel057 Thanks

PR #546, opened 2016-06-02
Add doctrine:database:import command to load SQL files

mpdude:
There's already a command for that in Doctrine/DBAL.

### DIFF
--- a/Command/Proxy/ImportDoctrineCommand.php
+++ b/Command/Proxy/ImportDoctrineCommand.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ * This file is part of the Doctrine Bundle
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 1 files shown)


mikeSimonson:
@mpdude Thanks

stof:
This will break things when using an older DBAL version not having the ImportCommand. Please use DelegateCommand as parent class instead, which proxies commands through delegation instead (you will need to overwrite ``isEnabled``, as the DelegateCommand checks the ORM version while this is the DBAL version

mpdude:
The `ImportCommand` was added in DBAL 2.0 and hardy any of the other commands does it that way, so...?

stof:
oh, I thought it was a new feature for which support in the bundle is added now in the bundle. It was in fact an old feature not exposed here

mikeSimonson:
@stof So this can be merged as his ?

mpdude:
*bump*

Any ideas when we might get the next release?

mikeSimonson:
@mpdude thanks

mpdude:
Any idea when you might tag the next minor version?

PR #552, opened 2016-06-26
Add apcu configuration option in docs

jrjohnson:
If I'm reading https://github.com/doctrine/cache/pull/115 correctly apcu is now fully supported.  I updated the listing of provider to indicate this.

Since the default configuration of apcu in PHP 7 no longer provides backwards compatibility with apc API I also changed the example `metadata_cache_driver` to be `apcu` to be copy / paste ready for new deployments.


kimhemsoe:
Thanks @jrjohnson

PR #558, opened 2016-07-21
Fix interpolate query paramters if starting with index 1

Deltachaos:
If paramters start with index 1 (as for sqlserver) the parameters should be interpolated correctly as well.
I am not absolutly shure, but i think they start always with index 1.


stof:
please add a test covering it

Deltachaos:
@stof test is added

xabbuh:
You need to revert the permission changes for the test class.

mikeSimonson:
@Deltachaos Thanks. Merged in #566

PR #559, opened 2016-07-24
Set doctrine.orm.entity_manager.abstract as lazy

nicolas-grekas:
Related to https://github.com/symfony/symfony/pull/19203

### DIFF
--- a/Resources/config/orm.xml
+++ b/Resources/config/orm.xml
@@ -95,7 +95,7 @@
 
         <service id="doctrine.orm.configuration" class="%doctrine.orm.configuration.class%" abstract="true" public="false" />
 
-        <service id="doctrine.orm.entity_manager.abstract" class="%doctrine.orm.entity_manager.class%" abstract="true" />
+        <service id="doctrine.orm.entity_manager.abstract" class="%doctrine.orm.entity_manager.class%" abstract="true" lazy="true" />
 
         <!-- The configurator cannot be a private service -->
         <service id="doctrine.orm.manager_configurator.abstract" class="%doctrine.orm.manager_configurator.class%" abstract="true">
--- a/composer.json
+++ b/composer.json
@@ -27,6 +27,7 @@
         "php": ">=5.3.2",
         "symfony/framework-bundle": "~2.3|~3.0",
         "symfony/console": "~2.3|~3.0",
+        "symfony/dependency-injection": "~2.3|~3.0",
         "doctrine/dbal": "~2.3",
         "jdorn/sql-formatter": "~1.1",
         "symfony/doctrine-bridge": "~2.2|~3.0",


Resources/config/orm.xml
```diff
@@ -95,7 +95,7 @@
 
         <service id="doctrine.orm.configuration" class="%doctrine.orm.configuration.class%" abstract="true" public="false" />
 
-        <service id="doctrine.orm.entity_manager.abstract" class="%doctrine.orm.entity_manager.class%" abstract="true" />
+        <service id="doctrine.orm.entity_manager.abstract" class="%doctrine.orm.entity_manager.class%" abstract="true" lazy="true" />
```
xabbuh: You may want to set the attribute dynamically in the extension class depending on whether or not it is available.

nicolas-grekas: Any idea how you'd do that? This is purely an implementation detail that doesn't leak through the interfaces. Given that next versions are going to need this "lazy", adding the dynamic check would only be a micro optim for soon to be legacy versions. Worth it?

Ocramius: > You may want to set the attribute dynamically in the extension class depending on whether or not it is available.

"lazy" is simply no-op if the required components aren't installed.

stof: I agree with @nicolas-grekas and @Ocramius. there is no need to make it conditional as it is already handled gracefully.

stof:
you need to change the DI min dependency from 2.2 to 2.3 to fix the deps=low tests

Resources/config/orm.xml
xabbuh: I am okay with that too if you then update the min version for the DependencyInjection component instead.

nicolas-grekas:
@stof updated (line for DI added)

kimhemsoe:
Thanks @nicolas-grekas

antanas-arvasevicius:
Hello,
I've noticed that this change will trigger exception in our project:
  [Symfony\Component\DependencyInjection\Exception\ServiceCircularReferenceException]  
  Circular reference detected for service "doctrine.dbal.default_connection",          
   path: "doctrine.dbal.default_connection".  

We are using lazy:true in services.yml. Looks like your change is disabling other services which has lazy=true; Strange. Half of a day I'm searching for a problem, but can't find, maybe you could find why that happens. Could be bug in dependency container?
Thanks.

Test case:

services.yml:
  test:
    class: Test
    arguments: ['@doctrine.orm.default_entity_manager']
    lazy: true
    tags:
      - { name: doctrine.event_listener, event: prePersist }

test.php:
class Test
{
    public function __construct(EntityManager $em)
    {
    }
}

composer.json:
        "symfony/proxy-manager-bridge": "~2.3",
        "ocramius/proxy-manager": "~1.0",
        "symfony/symfony": "~2.8.9",

kimhemsoe:
@antanas-arvasevicius Could you repost in #559 for now. It may/may not be the same error, but for now it is nice to all the information in the same issue until we know more.

antanas-arvasevicius:
Hello, what did you mean by repost in #559 as this thread is #559?

By my example error is clearly known, it's due circular dependency of entity manager in listener service, and that's why I've had made that listener as a lazy service. 
But looks like this commit will break "lazy" support in the system. Or maybe DI won't allow lazy -> lazy -> lazy dependencies.

kimhemsoe:
Sorry i meant #562

PR #560, opened 2016-08-03
Avoid duplication of AttachEntityListeners

aminin:
With multiple connections in `config.yml`

```
doctrine:
    dbal:
        default_connection: default
        connections:
            default:
               ...
            foobar:
                ...
    orm:
        default_entity_manager: default
        entity_managers:
            default:
                connection: default
            foobar:
                connection: foobar
```

each `EntityManager` will have multiple instances of `AttachEntityListeners` subscribed to the same event.

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -474,7 +474,11 @@ class DoctrineExtension extends AbstractDoctrineExtension
         if (version_compare(Version::VERSION, "2.5.0-DEV") >= 0) {
             $listenerId = sprintf('doctrine.orm.%s_listeners.attach_entity_listeners', $entityManager['name']);
             $listenerDef = $container->setDefinition($listenerId, new Definition('%doctrine.orm.listeners.attach_entity_listeners.class%'));
-            $listenerDef->addTag('doctrine.event_listener', array('event' => 'loadClassMetadata'));
+            $listenerTagParams = array('event' => 'loadClassMetadata');
+            if (isset($entityManager['connection'])) {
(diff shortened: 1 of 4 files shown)


mikeSimonson:
@stof Can you confirm that this is the correct behavior ?

trv1:
This patch works fine for me.
I'm waiting for this PR to throw a significant piece of smelly code from my app.

mikeSimonson:
@kimhemsoe Can you confirm that this is the valid behavior ?

mikeSimonson:
ping @kimhemsoe @stof

mikeSimonson:
@nicolas-grekas Can you validate this ?

nicolas-grekas:
I'm sure a test would help here...

kimhemsoe:
@mikeSimonson Sorry for not responding earlier. Tbh i dont know what is correct behavior here.

Doctrinebundle is bit like a magic eightball :)

I agree with @nicolas-grekas about tests here, it is the only way we can try to stop this "What is supposed to happen in this case" madness. So if PR is not correct we should a test so we know next time.

aminin:
@kimhemsoe, good idea :8ball:  :+1:

:8ball:, does DoctrineBundle look like an eightball?



:8ball:, should @kimhemsoe or @stof merge this PR?

kimhemsoe:
@mikeSimonson What do you think? Should we merge and say this is "correct" behavior?

mikeSimonson:
@kimhemsoe I am not merging that without a test.

kimhemsoe:
@mikeSimonson There have been added tests :D

mikeSimonson:
Sorry I missed it.

LGTM

PR #566, opened 2016-09-02
Fix interpolate query paramters if starting with index 1 

mikeSimonson:
#558


romainnorberg:
Thanks @mikeSimonson 
symfony_profiler

PR #569, opened 2016-09-07
Minor tweak for console command description

bocharsky-bw:


mikeSimonson:
@bocharsky-bw Thanks

I just wouldn't change the table name to make is Mysql specific.

bocharsky-bw:
@mikeSimonson Well, these changes are not MySQL specific, I mean:
`users` - valid for all DB providers;
`user` - invalid for MySQL, PostgeSQL... but valid for other.

So the question is: Why do we not want to change the table name to `users` so every DB provider had a benefit but not only a few of them?

PR #574, opened 2016-10-15
Trying to auto add the initialization of oracle

### LINKED ISSUE #217: OracleSessionInit is NOT loaded properly in the Dependency Injection
Hi,
I moved my project from MySQL to Oracle, and I get this error when I try to insert into the database.

An exception occurred while executing 'INSERT INTO arborescence (IDNT_ARBR, NOM_ARBR, DATE_CREATION, IDNT_SERV, IDNT_UTLS) VALUES (?, ?, ?, ?, ?)' with params [7, "B", "2013-10-09 00:00:00", 7, 1]:
ORA-01861: literal does not match format string

I solved this problem by adding the following code to : app/config/config.yml

services:
    oracle.listener:
        class: Doctrine\DBAL\Event\Listeners\OracleSessionInit 
        tags:
            - { name: doctrine.event_listener, event: postConnect }

This seems to be a bug in the DoctrineBundle. If we use the MySQL driver, the corresponding MysqlSessionInit is loaded properly in the Dependency Injection extension. But this doesn't happen with the OracleSessionInit class, if you we the Oracle driver.

Respectfully,
iftah

mikeSimonson:
Would something like this fix #217 ?

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -218,8 +218,19 @@ class DoctrineExtension extends AbstractDoctrineExtension
                 unset($connection['charset']);
             }
         }
-
         $options = $this->getConnectionOptions($connection);
+
+        // oracle session init
+        // the init listener has to be used for oracle.
+        if (isset($connection['driver']) && stripos($connection['driver'], 'oracle') !== false) {
+            $oracleSessionInit = new Definition('%doctrine.dbal.events.oracle_session_init.class%');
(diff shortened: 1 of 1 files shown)


kimhemsoe:
Last ive heard is that we do not want to auto run for BC reasons and not everyone needs it.

mikeSimonson:
@kimhemsoe I understood that the BC break was for doctrine not doctrine bundle.

But if we agree not to do it then I would add a section in the doc of doctrine bundle for it and close this PR and the issue no ?

kimhemsoe:
yeah i think we should do that.

PR #575, opened 2016-10-15
Add a proper exception when Doctrine ORM is not installed

fabpot:
Without this change, and when doctrine/orm is not installed, you get a weird error message about the Version class that does not exist.


kimhemsoe:
Thanks @fabpot

PR #579, opened 2016-10-26
Add ability to group similar queries in Profiler.

kewlar:
DoctrineBundle is great at giving a quick heads-up when the application might benefit from optimizing the database queries.

doctrine-profiler-blip

But sometimes _it isn't easy to find what exactly to optimize,_ as there may be hundreds of queries. Are there a few very slow queries, or is there a bunch of quick ones that just add up? It's not easy to check at a glance which ones are the most expensive.

doctrine-profiler-default

If only the profiler could group similar queries...

doctrine-profiler-group-link

...finding the culprits would be much easier:

doctrine-profiler-group-queries

There, it's much easier to see that 4 different queries take up 80% of execution time.

---

I've considered other alternatives for getting quick insights into which queries to optimize first, when there's a need for that.
1. There's the [sensiolabs/doctrine-query-statistics-bundle](https://github.com/sensiolabs/SensioLabsDoctrineQueryStatisticsBundle), but it's not supported since 2013, and some useful features are missing.
2. I could fork/update the `doctrine-query-statistics-bundle`, but having separate Profiler tab to do _almost_ the same as the DoctrineBundle tab doesn't feel quite right.
3. There are server-side products for insights into application performance, such as [New Relic](https://newrelic.com) or the excellent [Blackfire](https://blackfire.io), but they require separate installation, and raise the barrier to getting quick insights quite a bit higher.

So far, the option of adding this feature into the DoctrineBundle seems the best and most natural fit to me.

---

Though this PR does the job, the functionality and usability (and the code) could be improved further, of course. I'd love to add a quick summary (ex. _"4 queries take up 80% time"_) to encourage the users that stumble upon the Queries page to try to group the queries for a quick insight, but I haven't quite decided how to add it yet. And there may very well be something I missed.

If you think it may be useful but needs some more work, let me know, and I'll be happy to fix it.

### DIFF
--- a/DataCollector/DoctrineDataCollector.php
+++ b/DataCollector/DoctrineDataCollector.php
@@ -192,6 +192,12 @@ class DoctrineDataCollector extends BaseCollector
 
     public function getGroupedQueries()
     {
+        static $groupedQueries = null;
+
+        if ($groupedQueries !== null) {
+            return $groupedQueries;
+        }
+
         $groupedQueries = array();
         $totalExecutionMS = 0;
         foreach ($this->data['queries'] as $connection => $queries) {
@@ -223,5 +229,15 @@ class DoctrineDataCollector extends BaseCollector
         }
 
         return $groupedQueries;
+    }
+
+    public function getGroupedQueryCount()
+    {
+        $count = 0;
+        foreach ($this->getGroupedQueries() as $connectionGroupedQueries) {
+            $count += count($connectionGroupedQueries);
+        }
+
+        return $count;
     }
 }
--- a/DataCollector/DoctrineDataCollector.php
+++ b/DataCollector/DoctrineDataCollector.php
@@ -136,33 +136,10 @@ class DoctrineDataCollector extends BaseCollector
             }
         }
 
-        $groupedQueries = array();
-        foreach ($this->data['queries'] as $connection => $queries) {
-            $connectionGroupedQueries = array();
-            foreach ($queries as $i => $query) {
-                $key = $query['sql'];
-                if (!isset($connectionGroupedQueries[$key])) {
-                    $connectionGroupedQueries[$key] = $query;
-                    $connectionGroupedQueries[$key]['executionMS'] = 0;
-                    $connectionGroupedQueries[$key]['count'] = 0;
-                    $connectionGroupedQueries[$key]['index'] = $i; // "Explain query" relies on query index in 'queries'.
-                }
-                $connectionGroupedQueries[$key]['executionMS'] += $query['executionMS'];
-                $connectionGroupedQueries[$key]['count']++;
-            }
-            usort($connectionGroupedQueries, function ($a, $b) {
-                if ($a['executionMS'] === $b['executionMS']) {
-                    return 0;
-                }
-                return ($a['executionMS'] < $b['executionMS']) ? 1 : -1;
-            });
-            $groupedQueries[$connection] = $connectionGroupedQueries;
-        }
-
         $this->data['entities'] = $entities;
         $this->data['errors'] = $errors;
         $this->data['caches'] = $caches;
-        $this->data['groupedQueries'] = $groupedQueries;
+        $this->data['groupedQueries'] = $this->getGroupedQueries();
     }
 
     public function getEntities()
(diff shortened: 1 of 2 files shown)


DataCollector/DoctrineDataCollector.php
```diff
@@ -136,9 +136,33 @@ public function collect(Request $request, Response $response, \Exception $except
...
+        $groupedQueries = array();
+        foreach ($this->data['queries'] as $connection => $queries) {
+            $connectionGroupedQueries = array();
+            foreach ($queries as $i => $query) {
+                $key = $query['sql'];
+                if (!isset($connectionGroupedQueries[$key])) {
+                    $connectionGroupedQueries[$key] = $query;
+                    $connectionGroupedQueries[$key]['executionMS'] = 0;
+                    $connectionGroupedQueries[$key]['count'] = 0;
+                    $connectionGroupedQueries[$key]['i'] = $i; // "Explain query" relies on query index in 'queries'.
```
stof: I would name it ``index``, not ``i``

Resources/views/Collector/db.html.twig
```diff
@@ -188,19 +188,40 @@
                 <p>No database queries were performed.</p>
             </div>
         {% else %}
+            {% set group_queries = false %}
+            {% if collector.groupedQueries[connection]|length < queries|length %}
+                {% set group_queries = request.query.has('group') %}
```
stof: should be ``request.query.getBoolean('group')``. Otherwise, passing ``?group=false`` would also group them

DataCollector/DoctrineDataCollector.php
```diff
@@ -136,9 +136,33 @@ public function collect(Request $request, Response $response, \Exception $except
             }
         }
 
+        $groupedQueries = array();
+        foreach ($this->data['queries'] as $connection => $queries) {
```
stof: I'm wondering whether we really need to perform this extra work at collection time (which adds time to the page being loaded). It only computes something based purely on already collected data (and not needed by the toolbar, only by the panel), so we could do it on demand in ``getGroupedQueries`` (and ensure that the template calls it only once and then keeps it in a variable)

stof:
The sensiolabs bundle had an extra distinction: same queries (i.e. parameters are the same too) vs similar queries (same SQL but different parameters).

And if we go this way, it may be good to add a number showing the number of similar queries, or something like that.

kewlar:
Good points. Thanks, fixed them.

I'd also like to add a visual indicator for time-per-statement, if it's not too late. What do you think?
screenshot 2016-10-27 23 21 16

As to add a number showing the number of similar queries, I'll get back to that a bit later today.

mikeSimonson:
@kewlar go for it.

kewlar:
I've added some visual indication for how much time _relatively_ to the whole database access time each statement takes – both percentage value, and a progress-like-bar.

I've also took liberty to change word _"queries"_ into _"statements"_ where I'm talking about the grouped statements, in order to lessen the mental confusion between two types of "queries". In my mind, a _statement_ is a SQL string (with placeholders for params, if needed), and a _query_ is the statement executed against the server. My English may be misguided, though, and any feedback from ~~grammar~~ etymology nazis or UI experts is very welcome.

As to adding a number showing the number of similar queries, @stof, did you have in mind something like this:
screenshot 2016-11-10 00 43 58

There's a small issue of query parameters (and, accordingly, "Runnable query") being shown only for the first query in the group. Do you think that:

a) it's a non-issue, or 
b) it should be identified as such (ex. text "showing params only for the 1st query of 166"), or
c) there should be an option to see parameters for all the queries that were ran?

I'd like to think it's a non-issue, because the label may become redundant quite quickly, and listing all the queries is a bit out of scope of what I had in mind – but I'm willing to listen.

I'm done with improvements that I had had planned. Try it out and let me know what you think.

mikeSimonson:
@kewlar The distinction between queries and statement seems appropriate to me.
I don't think that the runnable query part is an issue and if it is someone can add it later without issue.

kewlar:
Yay!

Koc:
@kewlar thank you, it is really useful feature.

But as I understand correctly - queries with `WHERE id IN (:ids)` would considerent as different if `ids` will have different items count?

kewlar:
@koc the profiler will still register different queries, but it should detect the _statements_ as being the same and should group them correctly.

There are still inadequacies, such as `WHERE id IN (?, ?)` and `WHERE id IN (?, ?, ?)` are being treated as different statements, though they should be the treated the same. It's not resolved yet, and at the moment you should have that in mind.

PR #581, opened 2016-11-07
Use VarDumper in web profiler view

alcaeus:


composer.json
```diff
@@ -31,7 +31,8 @@
         "doctrine/dbal": "~2.3",
         "jdorn/sql-formatter": "~1.1",
         "symfony/doctrine-bridge": "~2.2|~3.0",
-        "doctrine/doctrine-cache-bundle": "~1.0"
+        "doctrine/doctrine-cache-bundle": "~1.0",
+        "symfony/web-profiler-bundle": "~2.7|~3.0"
```
Ocramius: Doesn't this also require the web profiler bundle to be enabled?

alcaeus: Yes, which is why I'm not sure whether we need to add this. On the other hand, we should make sure the `profiler_dump` twig function exists. Dropping Symfony 2.3 would probably implicitly ensure it is, but I'd rather have it explicit.

Ocramius: > Dropping Symfony 2.3 would probably implicitly ensure it is

What's blocking this then? :-)

alcaeus: Point taken - might as well remove it in this PR ;)

kimhemsoe:
Requiring 2.7 is a no go as we do not want to maintain more then one version of doctrine bundle. Is there any way to archive this while keeping BC?

Ocramius:
@kimhemsoe we can just drop 2.3 and that's it. There's no need to keep BC with it, unless I'm missing something obvious.

kimhemsoe:
@Ocramius It is a policy set by @stof and i am not going change that without a OK from him atleast. 

From the looks of the it should be pretty simple to support 2.3 and still use VarDumper when it is available.

alcaeus:
We have two choices here:
1. Drop support for 2.3 to at least somewhat ensure the `profiler_dump` function exists
2. Leave it as is and let people fail hard when updating the bundle in their Symfony 2.3 apps.

I'd argue for dropping 2.3 support here. Bugfixing has ended on May 2016, it will only receive security fixes for another 6 months. Should something arise, feel free to ping me and I'll personally deal with the overhead of releasing a security fix version for 1.6 in addition to the regularly scheduled release. Personally, I don't think we need to support 2.3 any longer.

Nyholm:
I agree with dropping 2.3 if it causes pain/issues.

Ocramius:
> Bugfixing has ended on May 2016

That's good enough for me: kill it, I'd say.

We fix security issues wherever possible/sensible anyway, although we don't have an explicit policy about it. DoctrineBundle is a relatively security-safe area, since it's just an integration layer, so I don't expect security fixes to land in here anyway.

kimhemsoe:
Ok, if i do not get any objections in the next few days i will bump the required symfony version.

Ocramius:
@kimhemsoe this PR does that, so merging would be sufficient, I suppose.

alcaeus:
@kimhemsoe that sounds like a plan. I'd also look through the issues and PRs to identify any other things we can add before tagging 1.7.0.

kimhemsoe:
@Ocramius Sorry.. i meant bumping by merging :D

stof:
Now that 2.3 is EOLed, dropping support for it is indeed acceptable

havvg:
What about changing the required version to `2.8` as it's the LTS already? I mean, `3.0` is also allowed. As per release notes `2.8` is the BC variant of `3.0`.

alcaeus:
Symfony 2.7 is a LTS release as well and it's supported until May 2018. As much as I'd like to, I don't think dropping that is an option.

havvg:
Oh, I just took a look on the roadmap, it states 2.8, but the details read 2.7 being LTS, too. Well, that makes my comment obsolete :-)

composer.json
```diff
@@ -25,18 +25,18 @@
     ],
     "require": {
         "php": ">=5.3.2",
```
fabpot: Can be bumped to 5.3.9 as well.

fabpot:
👍

stof:
thanks

PR #582, opened 2016-11-08
Added binary escaping

kucharovic:
Escape binary string according to ODBC standard.
**Before:**
before
**After:**
after

### DIFF
--- a/Tests/Twig/DoctrineExtensionTest.php
+++ b/Tests/Twig/DoctrineExtensionTest.php
@@ -65,4 +65,10 @@ class DoctrineExtensionTest extends \PHPUnit_Framework_TestCase
         $result = $extension->replaceQueryParameters($query, $parameters);
         $this->assertEquals('a=1 OR b=2', $result);
     }
+
+    public function testEscapeBinaryParameter()
+    {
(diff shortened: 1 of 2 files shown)


mikeSimonson:
@kucharovic Can you add a comment in the code on what the preg_match is trying to achieve ?

Thanks

kucharovic:
@mikeSimonson it's the same way as is used in [symfony/var-dumper](https://github.com/symfony/var-dumper/blob/3.1/Caster/CutStub.php#L53). Should I add checking for type `string` first?

```php
case is_string($result) && !preg_match('//u', $result):
    $result = '0x'. strtoupper(bin2hex($result));
    break;
```

mikeSimonson:
@kucharovic Yes I think that you should add the is_string check before otherwise it will error out if an object is passed for instance.
Maybe you can take that opportunity to test the escapeFunction function ?

kucharovic:
@mikeSimonson done. I can cover remaining cases of `DoctrineExtension::escapeFunction` by unit test in another commit.

mikeSimonson:
@kucharovic please do

mikeSimonson:
@kucharovic Thanks a lot

kucharovic:
@mikeSimonson Finally, I've add them to this PR because of merge conflict.

PR #591, opened 2016-12-05
Adding proxy-manager explicitly as a dependency

mikeSimonson:


Ocramius:
Do we really need this? Lazy services are not a requirement, after all... This will basically force 99% of Symfony apps to run with ProxyManager.

I'm not complaining: good marketing for me (j/k), but it seems a bit forced...

kimhemsoe:
@mikeSimonson May i ask why?

mikeSimonson:
@kimhemsoe @Ocramius I thaught that it was necessary because of a class that was requiring it (ManagerRegistry) that I stumble uppon while debugging something else. But that class is in the vendor folder [doctrine-bridge](https://github.com/symfony/doctrine-bridge/blob/master/ManagerRegistry.php#L14) so I need to revert that commit.

Does this mean that the proxy-manager should be required by the doctrine-bridge ?

kimhemsoe:
From a quick climpse of the related code, it does not look required "yet"
to me.

stof:
@mikeSimonson it does not require it. It changes its behavior when the entity manager was generated as lazy. It is fine to avoid using the ProxyManager library if you don't need to reset the entity manager.

Ocramius:
So... revert?

stof:
I would say yes

mikeSimonson:
done

Ocramius:
@mikeSimonson thanks!

PR #595, opened 2016-12-08
Setting doctrine.orm.second_level_cache params via configuration

alertiu5:
Adding a way to customize classes that will be used as doctrine.orm.second_level_cache.* parameters in doctrine configuration.

Was originally done to be able to use our own CacheFactory implementation. Also moved all other parameters in section from xml to Configuration, preserving original defaults.


alertiu5:
Of course, author, replace etc. in composer.json are intended for fork only, in case it won't be possible to merge it.

alertiu5:
decided to make a separate branch without composer.json changes

PR #596, opened 2016-12-08
Enabling second level cache config in doctrine.orm section

alertiu5:
Originally created so we could use our own implementation of CacheFactory via config option. Expanded for all parameters in doctrine.orm.second_level_cache.default_cache_factory to be setup-able.


DependencyInjection/Configuration.php
```diff
@@ -350,10 +350,86 @@ private function addOrmSection(ArrayNodeDefinition $node)
...
+     *
+     * @return \Symfony\Component\Config\Definition\Builder\NodeDefinition
+     */
+    private function getOrmSecondLevelCacheNode()
+    {
+        $treeBuilder = new TreeBuilder();
+        $node = $treeBuilder->root('second_level_cache');
+
+        $node
+            ->info('Global second level cache defaults (for all entity managers).')
```
stof: IMO, we should not have a global config for them. The config should be separate for each entity manager.

Btw, if you want to keep it global, you are missing the code whitelisting the key in the logic handling the shortcut config

DependencyInjection/Configuration.php
```diff
@@ -350,10 +350,86 @@ private function addOrmSection(ArrayNodeDefinition $node)
...
+                    ->children()
+                        ->scalarNode('class')
+                            ->defaultValue('Doctrine\ORM\Cache\DefaultCacheFactory')
+                        ->end()
+                    ->end()
+                ->end()
+                ->arrayNode('default_region')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->scalarNode('class')
```
stof: -1 for making all classes configurable in the config. This is the worst extension point ever (even worse than exposing them as DIC parameters, which are already considered a bad practice today, as it bloats the user-facing semantic config). Replacing only the class of a service is very fragile and likely to break when service definition are altered.

DependencyInjection/DoctrineExtension.php
```diff
@@ -636,7 +653,7 @@ protected function loadOrmEntityManagerMappingInformation(array $entityManager,
      * @param Definition       $ormConfigDef  A Definition instance
      * @param ContainerBuilder $container     A ContainerBuilder instance
      */
-    protected function loadOrmSecondLevelCache(array $entityManager, Definition $ormConfigDef, ContainerBuilder $container)
+    protected function loadEntityManagerSecondLevelCache(array $entityManager, Definition $ormConfigDef, ContainerBuilder $container)
```
stof: Renaming the existing protected method is a BC break, and cannot be done

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -191,16 +191,7 @@ public function testDependencyInjectionConfigurationDefaults()
...
-        $this->assertEquals('Doctrine\ORM\Mapping\Driver\SimplifiedYamlDriver', $container->getParameter('doctrine.orm.metadata.yml.class'));
-
-        // second-level cache
-        $this->assertEquals('Doctrine\ORM\Cache\DefaultCacheFactory', $container->getParameter('doctrine.orm.second_level_cache.default_cache_factory.class'));
-        $this->assertEquals('Doctrine\ORM\Cache\Region\DefaultRegion', $container->getParameter('doctrine.orm.second_level_cache.default_region.class'));
-        $this->assertEquals('Doctrine\ORM\Cache\Region\FileLockRegion', $container->getParameter('doctrine.orm.second_level_cache.filelock_region.class'));
-        $this->assertEquals('Doctrine\ORM\Cache\Logging\CacheLoggerChain', $container->getParameter('doctrine.orm.second_level_cache.logger_chain.class'));
-        $this->assertEquals('Doctrine\ORM\Cache\Logging\StatisticsCacheLogger', $container->getParameter('doctrine.orm.second_level_cache.logger_statistics.class'));
-        $this->assertEquals('Doctrine\ORM\Cache\CacheConfiguration', $container->getParameter('doctrine.orm.second_level_cache.cache_configuration.class'));
```
stof: Removing these assertions is bad. If your code is BC, they should still be passing

stof:
I'm -1 on these changes.

Ocramius:
Closing as per @stof review. @alertiu5 feel free to send another PR if you think that you have an alternate approach.

alertiu5:
Will do soon, I can agree with some of comments, but not everything. A possibility to set up a custom factory without jumping through hoops is needed, I don't think our project is the only one.

stof:
Well, before providing an alternate approach, it would be great to explain the use case exactly.
In some cases, overwriting services is perfectly fine to solve the advanced use cases. Exposing all possible advanced use cases in the user-facing config is not the goal (we have to document all these config settings to explain them to users, which was also missing in this PR btw)

alertiu5:
I will add more detailed description in next PR, of course.

PR #598, opened 2016-12-08
Updating DoctrineExtension.php

alertiu5:
Making entity manager's second_level_cache.factory option work.

Simplified my fork compared to [initial attempt](https://github.com/doctrine/DoctrineBundle/pull/596). Not doing anything extra now, just factory parameter in the old place.

Longer explanation as requested by @stof:

Some pre-history in stackoverflow (no answers received): [SO link](http://stackoverflow.com/questions/40386291/provide-doctrine-with-custom-cache-factory-for-second-level-cache)
We have found no sensible way to set a custom factory via configuration. I may be blind, but I could not find any handling of manager's second_level_cache.factory option in extension class. So I am trying to make it work via changes requested.

I can see a possible issue that it won't work if someone's implementation of CacheFactory will require other parameters for constructor, though.

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -658,9 +658,12 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $regionsDef = $container->setDefinition($regionsId, new Definition('%doctrine.orm.second_level_cache.regions_configuration.class%'));
 
         $slcFactoryId = sprintf('doctrine.orm.%s_second_level_cache.default_cache_factory', $entityManager['name']);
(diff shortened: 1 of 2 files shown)


stof:
Can you write a test covering this ?

alertiu5:
I will try to based on examples, not experienced much in writing tests though. It shall go as another method to Tests/DependencyInjection/DoctrineExtensionTest.php, right?

alertiu5:
Wrote a test, not sure if I got it right though.

alertiu5:
Any feedback? Would be really nice to have a parameter working or have an explanation how to use it if it works as is.

mikeSimonson:
@alertiu5 I did not forget about this PR. I will merge it this weekend. I am just waiting because I refactored the test code quite a bit after this PR.

Thanks

PR #600, opened 2016-12-16
Updating the php supported version tested

mikeSimonson:


stof:
The drawback of this is that Symfony has 5.5.9 as min requirement. And when creating a new project, it ships you dependencies compatible with the min requirement. And DoctrineBundle is included in these dependencies.
So if DoctrineBundle drops support for 5.5.9, new Symfony projects will use an old version of the bundle by default. This means that we would have to support a legacy branch until the end of the Symfony 3.x time
As we don't do any change benefiting from PHP 5.6 features, dropping it actually increases maintenance overhead by locking people to an older version of the bundle by default.

Ocramius:
Fwiw, symfony support doesn't imply support by doctrine. We're likely
dropping 5.x support before the end of this year anyway (ORM, common, etc)

stof:
Well, this bundle is included by default in symfony-standard, so this has an impact on our user base. And then, we would get reports from people about issues which are already solved just because the bootstrapped project shipped with an outdated version due to the PHP constraint.

Ocramius:
Yes, that is not really an issue, as we'd just suggest an upgrade as
resolution path

stof:
but it still means that **all** new symfony projects would be affected by this, thus increasing our maintenance overhead on the bundle, while not bringing as lots of benefits compared to supporting 5.5.9+ in composer.

mikeSimonson:
@stof Are there that many user that will use 5.5 for a new project. It's not supported at all since 6 months...

Issue #602, opened 2016-12-19
Deprecated message "getEntityManager is deprecated since Symfony 2.1"


Mohammadtrabelsi:
I work with symfony "3.1.6" and i user doctrine/doctrine-bundle"1.6.4" and in the production environment, I got this message :

>   getEntityManager is deprecated since Symfony 2.1. Use getManager instead in projet/myprojectname/vendor/doctrine/doctrine-bundle/Registry.php

stof:
Well, audit your codebase to find the place calling the ``getEntityManager`` method of the registry.
The registry is the place telling you about the deprecation (we should update the message btw, as this code was extracted out of Symfony itself so talking about the Symfony version does not make sense)

Mohammadtrabelsi:
lhe call of **getEntityManager** is on 
> vendor/doctrine/doctrine-bundle/Command/ImportMappingDoctrineCommand.php
this is the method :

 

>      /**
>      * {@inheritDoc}
>      */
>     protected function execute(InputInterface $input, OutputInterface $output)
>     {
>         $bundle = $this->getApplication()->getKernel()->getBundle($input->getArgument('bundle'));
> 
>         $destPath = $bundle->getPath();
>         $type = $input->getArgument('mapping-type') ? $input->getArgument('mapping-type') : 'xml';
>         if ('annotation' === $type) {
>             $destPath .= '/Entity';
>         } else {
>             $destPath .= '/Resources/config/doctrine';
>         }
>         if ('yaml' === $type) {
>             $type = 'yml';
>         }
> 
>         $cme = new ClassMetadataExporter();
>         $exporter = $cme->getExporter($type);
>         $exporter->setOverwriteExistingFiles($input->getOption('force'));
> 
>         if ('annotation' === $type) {
>             $entityGenerator = $this->getEntityGenerator();
>             $exporter->setEntityGenerator($entityGenerator);
>         }
> 
>         $em = $this->getEntityManager($input->getOption('em'));
> 
>         $databaseDriver = new DatabaseDriver($em->getConnection()->getSchemaManager());
>         $em->getConfiguration()->setMetadataDriverImpl($databaseDriver);
> 
>         $emName = $input->getOption('em');
>         $emName = $emName ? $emName : 'default';
> 
>         $cmf = new DisconnectedClassMetadataFactory();
>         $cmf->setEntityManager($em);
>         $metadata = $cmf->getAllMetadata();
[...]

PR #603, opened 2016-12-19
Update the deprecation warning convention

stof:
The Symfony convention is to silence the deprecations to avoid issues with error handlers turning PHP errors into exceptions. But these deprecations were written before Symfony decided to use silenced warnings, and were not updated after that.


mikeSimonson:
Thansk @stof. Does this fix #602 ?

stof:
No. #602 is about the project calling the deprecated API, and the user misunderstanding the deprecation warning (the message does not help avoiding this though)

PR #604, opened 2016-12-23
Add documentation on the oracle session environment

### LINKED ISSUE #217: OracleSessionInit is NOT loaded properly in the Dependency Injection

mikeSimonson:
Fix #217 and replace #574

### DIFF
--- a/Resources/doc/configuration.rst
+++ b/Resources/doc/configuration.rst
@@ -722,6 +722,32 @@ the ORM resolves to:
 There are lots of other configuration options that you can use to overwrite
 certain classes, but those are for very advanced use-cases only.
 
+Oracle DB
+=======
+
+If the environment format configured in oracle does not meet doctrine requirements,
+you need to use the OracleSessionInit listener so that doctrine is aware of the format used by Oracle DB.
+
+You can do so easily with
+
+.. code-block:: yaml
+
+    services:
+        oracle.listener:
+            class: Doctrine\DBAL\Event\Listeners\OracleSessionInit
+            tags:
+                - { name: doctrine.event_listener, event: postConnect }
+
+The environment variables that doctrine is going to change in the Oracle DB session are:
+
+.. code-block:: yaml
+
+    NLS_TIME_FORMAT="HH24:MI:SS"
+    NLS_DATE_FORMAT="YYYY-MM-DD HH24:MI:SS"
+    NLS_TIMESTAMP_FORMAT="YYYY-MM-DD HH24:MI:SS"
+    NLS_TIMESTAMP_TZ_FORMAT="YYYY-MM-DD HH24:MI:SS TZH:TZM"
+
+
 Caching Drivers
 ~~~~~~~~~~~~~~~
 


mikeSimonson:
@kimhemsoe Can you validate that explanation ?

Resources/doc/configuration.rst
```diff
@@ -722,6 +722,32 @@ the ORM resolves to:
 There are lots of other configuration options that you can use to overwrite
 certain classes, but those are for very advanced use-cases only.
 
+Oracle
+=======
+
+If the environment format configured in oracle does not meet doctrine requirements,
+you need to use OracleSessionInit listener so that doctrine is aware of the format used by oracle.
```
Ocramius: `use the`

Ocramius: `by Oracle DB`

Resources/doc/configuration.rst
```diff
@@ -722,6 +722,32 @@ the ORM resolves to:
 There are lots of other configuration options that you can use to overwrite
 certain classes, but those are for very advanced use-cases only.
 
+Oracle
```
Ocramius: `Oracle DB`

Resources/doc/configuration.rst
```diff
@@ -722,6 +722,32 @@ the ORM resolves to:
...
+
+.. code-block:: yaml
+
+    services:
+        oracle.listener:
+            class: Doctrine\DBAL\Event\Listeners\OracleSessionInit
+            tags:
+                - { name: doctrine.event_listener, event: postConnect }
+
+The environment variables needed for doctrine and their default format are:
```
Ocramius: These are environment vars for Oracle DB, not Doctrine DBAL: the [listener](https://github.com/doctrine/dbal/blob/b551fcb26a81e230085c0dca88174573b6c36609/lib/Doctrine/DBAL/Event/Listeners/OracleSessionInit.php) doesn't really use the environment, AFAIK.

mikeSimonson:
@Ocramius Thanks

Ocramius:
LGTM! 👍

PR #608, opened 2017-01-12
Dont call addClassesToCompile on PHP 7.0, its deprecated

nicolas-grekas:


fabpot:
👍

mikeSimonson:
@nicolas-grekas Thanks

nicolas-grekas:
It's not deprecated in PHP 7 but in Symfony
See https://github.com/symfony/symfony/issues/20668

PR #610, opened 2017-01-27
Fixed Division by zero when total execution time is 0ms

anddezr:
When you go to Symfony profiler -> Query Metrics and when Query time = 0ms there is an exception:
"An exception has been thrown during the rendering of a template ("Warning: Division by zero")."

[1/2] ContextErrorException: Warning: Division by zero 
in vendor\doctrine\doctrine-bundle\DataCollector\DoctrineDataCollector.php at line 232

[2/2] Twig_Error_Runtime: An exception has been thrown during the rendering of a template ("Warning: Division by zero")
in vendor\doctrine\doctrine-bundle\Resources\views\Collector\db.html.twig at line 157

### DIFF
--- a/DataCollector/DoctrineDataCollector.php
+++ b/DataCollector/DoctrineDataCollector.php
@@ -222,6 +222,9 @@ class DoctrineDataCollector extends BaseCollector
             });
             $groupedQueries[$connection] = $connectionGroupedQueries;
         }
+        if ((float) $totalExecutionMS === 0.0) {
+            $totalExecutionMS = 1;
+        }
         foreach ($groupedQueries as $connection => $queries) {
             foreach ($queries as $i => $query) {
                 $groupedQueries[$connection][$i]['executionPercent'] = $query['executionMS'] / $totalExecutionMS * 100;


mikeSimonson:
Thanks @anddezr.

I think that it's safer to move that if in the percentage calculation than changing the total time of execution for the rest of the code.

anddezr:
$totalExecutionMS won't change inside foreach loop, so I think it's better as is.

DataCollector/DoctrineDataCollector.php
```diff
@@ -222,6 +222,9 @@ public function getGroupedQueries()
             });
             $groupedQueries[$connection] = $connectionGroupedQueries;
         }
+        if ((float) $totalExecutionMS === 0.0) {
```
e-moe: Is it safe to compare float values like this? Maybe we can use very small epsilon value e.g. `if ($totalExecutionMS < $eps)` where `$eps = 1e-5` or something like this

alcaeus: Since the code is there to avoid a division by 0, there's no need for an epsilon check: any non-zero value won't trigger the error condition this check is trying to fix.

kimhemsoe:
@anddezr Can you rebase and add test? Would love to get this issue closed.

mikeSimonson:
fixed by #634

PR #614, opened 2017-02-14
Fix twig requirement constraint

chalasr:
Since https://github.com/doctrine/DoctrineBundle/commit/e2ca576ce35a16a0cf4e300068a4a5a5b54e36c7 the Twig extension makes use of `\Twig_SimpleFilter`, but the class does not exist before Twig 1.12 (not 1.10 nor 1.11).


fabpot:
I would bump it even more, but :+1: as it solves a real dep issue.

stof:
Btw, this tells me that you use an outdated version of Symfony. All uptodate versions of Symfony require a much newer version of Twig...

PR #615, opened 2017-02-15
Add return to ensure prodution settings comand to get exit status of command

bmaziere:


stof:
thanks

stof:
@bmaziere if you have time for that, can you do the same for all other commands ? This is not the only one missing the return statement

PR #616, opened 2017-02-15
Add missing returns to cerate & drop schema commands

bmaziere:
@stof, i only see those two missing but i may miss something.


mikeSimonson:
@bmaziere Thanks

PR #617, opened 2017-02-17
[POC][RFC] New repository factory able to inject dependencies in repositories

magarzon:
Sometimes we need our custom entity (object) repositories to have dependencies on other services and/or configuration parameters.

One way to get this done is to define the repository as a service, and inject the dependencies as any other service:

```
    app.doctrine.user_repository:
        class: AppBundle\Doctrine\Repository\UserRepository
        factory: ["@doctrine.orm.entity_manager", getRepository]
        arguments: [AppBundle\Entity\User]
        calls:
            - [setDependency, ['@example_service_dependency']]
            - [setParameter, ['%configuration_parameter%']]
```

And then inject this repository in another service, like a user manager or whatever.

But there are cases where third party and/or framework bundles use the doctrine service, or the entity manager service, to get the repository, and then is not used this service that we have defined, and the dependencies are not set.

My proposal is to have a tag, i.e. doctrine.repository to mark the repositories exposes as a service, and a repository factory that can create the repository and inject the dependencies.

That way, we only need to add this tag to this service:

```
    app.doctrine.user_repository:
        class: AppBundle\Doctrine\Repository\UserRepository
        factory: ["@doctrine.orm.entity_manager", getRepository]
        arguments: [AppBundle\Entity\User]
        calls:
            - [setDependency, ['@example_service_dependency']]
            - [setParameter, ['%configuration_parameter%']]
        tags:
            - {name: doctrine.repository, entity: AppBundle:User}
```
And the compiler pass that I have added get the calls (could be also other type of dependencies, properties, getter) and add the definitions (that are going to be transformed in actual instances) to the DIRepositoryFactory (yes, the naming could be enhanced).

The DIRepositoryFactory create the repository as de doctrine DefaultRepositoryFactory, but after repository creation, traverse the definitions, and if there is a match between the entity name and the included in the tag, it sets the dependencies in the repository instance.

Obviously, to activate this, the compiler pass should be added in the the DoctrineBundle build method (I haven't added this to not break anything) and the DIRepositoryFactory should replace the default one using doctrine configuration:

```
doctrine:
    orm:
        auto_generate_proxy_classes: "%kernel.debug%"
        naming_strategy: doctrine.orm.naming_strategy.underscore
        auto_mapping: true
        repository_factory: doctrine.orm.di_repository_factory
```

I know that the code can be improved and even there is a little bug (the name of the service is different in the compiler and the config file), but this only a POC, if the idea is good, we should fix that.

### DIFF
--- a/DependencyInjection/Compiler/DoctrineRepositoryPass.php
+++ b/DependencyInjection/Compiler/DoctrineRepositoryPass.php
@@ -1,0 +1,58 @@
+<?php
+
+/*
+ * This file is part of the Doctrine Bundle
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ * (c) Doctrine Project, Benjamin Eberlei <kontakt@beberlei.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * Configure DI Repository Factory with repositories with dependencies
+ *
+ * @author Miguel Angel Garzón <magarzon@gmail.com>
+ */
+class DoctrineRepositoryPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if ($container->hasDefinition('doctrine.di_repository_factory')) {
+            $servicesId = $container->findTaggedServiceIds('doctrine.repository');
+
+            $definitions = [];
+
+            foreach ($servicesId as $serviceId => $parameters) {
+                $definition = $container->getDefinition($serviceId);
+                $definitions[] = $this->createFromServiceDefinition($definition);
+            }
+
+            $factory = $container->getDefinition('doctrine.di_repository_factory');
+            $factory->addArgument($definitions);
+        }
+    }
+
+    /**
+     * @param Definition $definition
+     * @return array
+     */
+    private function createFromServiceDefinition(Definition $definition)
+    {
+        $tag = $definition->getTag('doctrine.repository');
+        $entity = $tag[0]['entity'];
+        //Only setter injection by now
+        $calls = $definition->getMethodCalls();
+
+        return ['entity' => $entity, 'setters' => $calls];
+    }
+}
--- a/Repository/DIRepositoryFactory.php
+++ b/Repository/DIRepositoryFactory.php
@@ -1,0 +1,102 @@
+<?php
+
+/*
+ * This file is part of the Doctrine Bundle
+ *
+ * The code was originally distributed inside the Symfony framework.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ * (c) Doctrine Project, Benjamin Eberlei <kontakt@beberlei.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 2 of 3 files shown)


Ocramius:
Why not just define your own repository services? Note that `getRepository`
is not ready the API for that unless you are subclassing.

magarzon:
@Ocramius that's what I do, but I need something more.

If I explain why I did this, maybe It helps to get this idea better.

I had to develop some authentication, but before getting the user from the database, I needed to do something with the authentication values (username/password), and I needed a service (and a configuration parameter) for this. 

The EntityUserProvider, in the Symfony Doctrine Bridge, helps to get a User from a database for authentication. This EntityUserProvider (as it happens with other services) has a dependency on ManagerRegistry (doctrine.orm.default_entity_manager service), and it uses this service to get the entity repository to get the user, and for this use the DefaultRepositoryFactory, so it couldn't use the repository service that I had defined.

I had to choices: develop my on UserProvider, and make the changes to the parameters there, or develop this new repository factory, that makes use of already defined repository services, instead of creating new repository instances.

As the changes that I had to make to the data was more related with the database (it has something to do with custom doctrine types) that with the authentication, it has sense to me to have this repository factory. With the add-on that the repository service that I have defined can be use anywhere else without any change.

It's a non-BC change, if you want/need to used, you can configure it as the repository factory, and if you don't have any repository service (or for those entities that doesn't have a custom repository), it works just as the default repository factory.

Pierstoval:
I'm not sure that your proposal is really the best implementation for your idea.

Something better might be to just add a simple tag without using the factory in your service definition, like this:

```yml
    app.doctrine.user_repository:
        class: AppBundle\Doctrine\Repository\UserRepository
        arguments: ~ # The tag unfortunately will override all arguments
        calls:
            - [setDependency, ['@example_service_dependency']]
            - [setParameter, ['%configuration_parameter%']]
        tags:
            - { name: doctrine.repository, entity: AppBundle:User } # That's all
```

Then, the compiler pass that manages the `doctrine.repository` tag would add a method call to the RepositoryFactory class, like this:

```php
$repositoryFactoryReference->addMethodCall('createRepository', [$entityManagerReference, $tagArguments['entity']]);
```

And obviously add the correct constructor arguments to it:

```php
$taggedServiceDefinition->setArguments([$tagArguments['entity']]);
```

This way, you can either get the repository by its service, AND by the `$em->getRepository(...)` AND `$repositoryFactory->getRepository(...)` methods.

Adding the tag this way would allow simplifying the code for all doctrine repositories declared as services.

What do you think?

Either I'm not sure if there's gonna be duplicate objects in this case 😕

magarzon:
@Pierstoval I tried to solution but I'm getting an error because seems at compile time doctrine can't resolve short entity names (like AppBundle:Test).

Anyway, with your solution you're creating all tagged repositories in advance, even if you are not going to use them, and you don't get any advantage, because with my solution you already has the ability to get the repository by its service or by the entityManager, and at the end you're defining your service as a factory as well, but hiding it to the user, that IMHO it's not a good thing.

Hanmac:
i often do this in my projects:
```yaml
    abc.base_repository:
        abstract:  true
        class: Doctrine\ORM\EntityRepository
        factory: ['@doctrine.orm.entity_manager', getRepository]

    abc.xyz_repository:
        parent: abc.base_repository
        arguments:
            - AbcBundle\Entity\Xyz
```

and i dont know how that could be optimized.

an interesting idea would be if ORM\Mapping\Entity where repositoryClass is used might get an new flag "repositoryService" where you can just do 'abc.base_repository' and it does the magic?

magarzon:
@Hanmac but why adding a new parameter, when you can get the same using what already exists? That's my point, with my solution you can still use your service definition, but with the advantage that you're going to get that very same service every time that you get the repository, not only when you injected

kimhemsoe:
Closing. I don't believe this belongs in the bundle and there is currently work being done on a solution which may end up provide support for this. See #725

PR #393, opened 2015-02-18
This partially fixes doctrine/DoctrineCacheBundle#46

machinez:
This eradicates  "InvalidConfigurationException in ArrayNode.php line 309.", when defining a Redis database in config.yml

``` yml
doctrine:
    orm:
        metadata_cache_driver: 
            type: redis
            database: 8
            host: 192.168.0.1
        query_cache_driver:
            type: redis
            database: 8
            host: 192.168.0.1
        result_cache_driver:
            type: redis
            database: 8
            host: 192.168.0.1
```

PR #619, opened 2017-02-20
Allow database entry for orm cache driver node config

bmaziere:
This work allow to select database configuration for redis cache driver for instance.
It completes https://github.com/doctrine/DoctrineCacheBundle/issues/46 and "finish" started work https://github.com/doctrine/DoctrineBundle/pull/393

### DIFF
--- a/Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
+++ b/Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
@@ -399,6 +399,22 @@ abstract class AbstractDoctrineExtensionTest extends \PHPUnit_Framework_TestCase
         ));
     }
 
+    public function testEntityManagerRedisMetadataCacheDriverConfigurationWithDatabaseKey()
+    {
+        $container = $this->loadContainer('orm_service_simple_single_entity_manager_redis');
+
+        $definition = $container->getDefinition($container->getAlias('doctrine.orm.default_metadata_cache'));
+        $this->assertDICDefinitionClass($definition, '%doctrine_cache.redis.class%');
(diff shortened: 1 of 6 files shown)


Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -399,6 +399,22 @@ public function testEntityManagerMemcacheMetadataCacheDriverConfiguration()
         ));
     }
 
+    public function testEntityManagerRedisMetadataCacheDriverConfiguration()
```
mikeSimonson: Can you change the name of the method so that it mention the database key of the configuration being used.

mikeSimonson:
@bmaziere Thanks for finiwhing the work of that PR.

Appart from the test method name that isn't clear the rest is good to go to me.

Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
bmaziere: method name changed. No idea why the test that fail on PHP 5.5. I keep investigate

mikeSimonson:
@bmaziere Keep in mind that it only failing on 5.5 with dependencies low.
It probably means that one of the dependency in composer needs to be updated.
Because it currently accept a version of a lib that doesn't have a feature that you use.

bmaziere:
@mikeSimonson, i updated doctrine/doctrine-cache-bundle minimal version to 1.2 to make it work. Does it fit?

mikeSimonson:
I believe so.
Thanks again.

PR #622, opened 2017-02-21
Enable doctrine:database:import command without ORM installed

tucker:
The command is just a proxy to dbal:import and does not depend on Docrine ORM.
So we can include it without ORM installed.


mikeSimonson:
@tucker Thanks

Issue #618, opened 2017-02-20
Schema generator is not adding COMMENT to custom type


fkrauthan:
I am using the DoctrineEnumBundle and have a issue with the enum SQL that gets generated. It does not contain the COMMENT field which is required to identify the enum type correctly.

You can find a lot more information about this in the following two tickets:
* fre5h/DoctrineEnumBundle#87
* doctrine/dbal#2594

The versions in my `composer.json`:

```json
{
    "require": {
        "php": ">=5.5.9",
        "symfony/symfony": "3.2.*",
        "doctrine/orm": "^2.5",
        "doctrine/doctrine-bundle": "^1.6"
    }
}
```

Based on my lock file I am using

* Symfony: 3.2.3
* DoctrineORM: 2.5.6
* DoctrineBundle: 1.6.7

In addition to that my config.yml in regards of the custom type looks like this:

```yml
doctrine:
    dbal:
        #     ...
        types:
            FlowerState: AppBundle\DBAL\Type\FlowerStateType
            FlowerGraphicSetRarity: AppBundle\DBAL\Type\FlowerGraphicSetRarityType
            UserFlowerSort: AppBundle\DBAL\Type\UserFlowerSortType
```

As you can see on my last comment in the dbal ticket I found out that the [ConnectionFactory.php#L60](https://github.com/doctrine/DoctrineBundle/blob/master/ConnectionFactory.php#L60) might be the issue as it correctly detects that types as commented types but it never calls `markDoctrineTypeCommented` for them as `$mappingTypes` is empty.

I testwise moved that for loop that is guarded within the if condition below the if condition and the migration starts adding the comment to my custom enum fields.

Please help as this is very annoying that any DB migration I generate contains entries for my custom enum types to be converted back to string fields on the `down` method which if I forgot to delete would cause potential big issues if I ever have to roll back my migration.

Let me know if you need any more information.

kimhemsoe:
@fkrauthan Is this still an issue?

fkrauthan:
I haven't seen it lately but I also didn't add any new enums in a while. I guess we can close and in case I run into this again I will reopen a ticket.

kirya-dev:
Problem SOLVED.
```yaml
doctrine:
    dbal:
        mapping_types:
            enum: string
        types:
            user_status_enum: App\Enum\UserStatusEnum
 ```
 
 ```php
 use Doctrine\DBAL\Types\Type;
 
 class UserStatusEnum extends Type
{
[...]

PR #623, opened 2017-02-22
Fix an issue were some commented type

mikeSimonson:
They were not marked as such if the mappingType array was
empty.

Sadly I don't see any way to test this right now.

### DIFF
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -62,6 +62,9 @@ class ConnectionFactory
             foreach ($mappingTypes as $dbType => $doctrineType) {
                 $platform->registerDoctrineTypeMapping($dbType, $doctrineType);
             }
+        }
+        if (!empty($this->commentedTypes)) {
+            $platform = $connection->getDatabasePlatform();
             foreach ($this->commentedTypes as $type) {
                 $platform->markDoctrineTypeCommented(Type::getType($type));
             }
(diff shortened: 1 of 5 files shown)


stof:
IMO, we should deprecate this entirely. Types can be marked as commented inside themselves as of DBAL 2.3 or 2.4 (I don't remember anymore), which is much better, because it does not require getting the platform early.

mikeSimonson:
This is related to #618 
Do you think that the issue is somewhere else because apparently the user use a type class that is marked as using a comment.
The issue might be in the EnumBundle but I couldn't reproduce it despite very creative attempts.

mikeSimonson:
@stof It still fixed a few test that were apparently not running for some reason.

mikeSimonson:
@stof Maybe we can add a deprecation when https://github.com/doctrine/DoctrineBundle/blob/master/ConnectionFactory.php#L83 is reached ?

stof:
is there really no failure in DBAL when a non-existent class name is used for a type ?

mikeSimonson:
@stof I didn't really tried understanding how that test passed in past but I am guessing one way or another this part wasn't run as the error is from PHP itself.

mikeSimonson:
@stof For the deprecation notice I suppose we also need to change this default config.
https://github.com/doctrine/DoctrineBundle/blob/master/DependencyInjection/Configuration.php#L100

If I do that I suppose I should wait for a 2.0 ?

PR #624, opened 2017-02-22
Testing on php nightly too

mikeSimonson:


.travis.yml
```diff
@@ -21,6 +22,8 @@ matrix:
       env: SYMFONY_VERSION="2.8.*@dev"
     - php: 5.6
       env: SYMFONY_VERSION="3.0.*@dev"
```
stof: please also remove this one. Testing against an EOLed version does not make sense.

My rule is to list only LTS versions explicitly here (the latest version will be tested by the normal jobs)

mikeSimonson: ok.

Do you mean that the php version is EOL or that the symfony version is EOL ?

stof: I'm talking about Symfony

PR #628, opened 2017-02-23
Type comment deprecation

mikeSimonson:


mikeSimonson:
@stof Follow up of #623

ConnectionFactory.php
```diff
@@ -84,6 +84,9 @@ private function initializeTypes()
                 Type::addType($type, $typeConfig['class']);
             }
             if ($typeConfig['commented']) {
+                @trigger_error(
+                    'The commented option of the type configuration is deprecated since DoctrineBundle 1.7.
+                    Use the feature of the Type class Instead', E_USER_DEPRECATED);
```
Ocramius: Alignment is super weird, and there's a newline in the text too (with whitespace)

DependencyInjection/Configuration.php
```diff
@@ -97,7 +97,7 @@ private function addDbalSection(ArrayNodeDefinition $node)
                             ->end()
                             ->children()
                                 ->scalarNode('class')->isRequired()->end()
-                                ->booleanNode('commented')->defaultTrue()->end()
+                                ->booleanNode('commented')->defaultFalse()->end()
```
Ocramius: This is a BC break...

Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -531,7 +531,7 @@ public function testSetTypes()
         $container = $this->loadContainer('dbal_types');
 
         $this->assertEquals(
-            array('test' => array('class' => TestType::class, 'commented' => true)),
+            array('test' => array('class' => TestType::class, 'commented' => false)),
```
Ocramius: This is a BC Break

DependencyInjection/Configuration.php
mikeSimonson: It's not really a BC break if people can still use the option.
Or should I wait for v2 for this change and v3 to remove the option entirely ?

Ocramius:
It is a bc break if the defaults change and they didn't override them.

DependencyInjection/Configuration.php
stof: I suggest a BC way:

- trigger a deprecation whenever the field is configured explicitly (to allow removal), by using `beforeNormalization`
- if the field is configured as commented (either explicitly or due to the default), the DI extension should check whether the type class overwrites the `requiresSQLCommentHint` method:
  - If no, it should trigger a deprecation saying that types should be marked as commented in the type itself.
  - If yes, it should remove this type from the list of commented types passed to the bundle boot method (actually marking it as not commented in the parameter). Types will still be registered as commented in DBAL due to the type class itself.

stof:
@Ocramius I'm bringing back to a 1.x milestone, as I suggested a BC way to implement it above

kimhemsoe:
@mikeSimonson Ping

DependencyInjection/Configuration.php
weaverryan: For the first part, if it helps, I think this is an example: https://github.com/symfony/symfony/blob/2e47edc4a52c3b61664e512f849c3df5f5c20593/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L714-L721

It's checking to see if `strict_email` is explicitly configured, and it triggers a deprecation warning if it is.

For the second part... I'm not as sure :)

alcaeus:
Superseded by #947.

PR #633, opened 2017-03-01
Add a deprecation notice for this feature

mikeSimonson:
It can be replaced by comment feature of the doctrine
Type object.

### DIFF
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -84,6 +84,9 @@ class ConnectionFactory
                 Type::addType($type, $typeConfig['class']);
             }
             if ($typeConfig['commented']) {
+                @trigger_error(
+                    sprintf('The commented option of the type configuration is deprecated since DoctrineBundle 1.7.' .
+                    'Use the feature of the Type class instead. (in %s)', $type), E_USER_DEPRECATED);
                 $this->commentedTypes[] = $type;
             }
         }


mikeSimonson:
@stof Should I put the env var SYMFONY_DEPRECATIONS_HELPER=weak on every build ?

ConnectionFactory.php
```diff
@@ -84,6 +84,9 @@ private function initializeTypes()
                 Type::addType($type, $typeConfig['class']);
             }
             if ($typeConfig['commented']) {
+                @trigger_error(
+                    'The commented option of the type configuration is deprecated since DoctrineBundle 1.7.
+                    Use the feature of the Type class Instead', E_USER_DEPRECATED);
```
stof: `instead` should not have a capital letter.

And you should not put a newline inside the string itself. If you want to use 2 lines in the source code, use concatenation.

Btw, I would add the `$type` in the message to make it easier to know which type should be updated.

mikeSimonson: @stof Done. What about the tests that fail because of the deprecations errors ?

stof:
@mikeSimonson No. you should add `@group legacy` on the test covering the BC layer (and ensuring other features covered by the same test are also covered by a non-legacy test)

ConnectionFactory.php
```diff
@@ -84,6 +84,9 @@ private function initializeTypes()
                 Type::addType($type, $typeConfig['class']);
             }
             if ($typeConfig['commented']) {
+                @trigger_error(
+                    sprintf('The commented option of the type configuration is deprecated since DoctrineBundle 1.7.' .
+                    'Use the feature of the Type class instead. (in %s)', $type), E_USER_DEPRECATED);
```
stof: I would say `Use the "requiresSQLCommentHint" method of the Type class instead`

alcaeus:
Superseded by #947.

PR #634, opened 2017-03-07
Fix a possible division by zero

mikeSimonson:


DataCollector/DoctrineDataCollector.php
```diff
@@ -222,15 +222,26 @@ public function getGroupedQueries()
             });
             $groupedQueries[$connection] = $connectionGroupedQueries;
         }
+
         foreach ($groupedQueries as $connection => $queries) {
             foreach ($queries as $i => $query) {
-                $groupedQueries[$connection][$i]['executionPercent'] = $query['executionMS'] / $totalExecutionMS * 100;
+                $groupedQueries[$connection][$i]['executionPercent'] =
+                    $this->executionTimePercentage($query['executionMS'], $totalExecutionMS );
```
stof: I would not wrap the line here

DataCollector/DoctrineDataCollector.php
```diff
@@ -222,15 +222,26 @@ public function getGroupedQueries()
...
             }
         }
 
         return $groupedQueries;
     }
 
+    private function executionTimePercentage($executionTimeMS, $totalExecutionTimeMS)
+    {
+        if ($totalExecutionTimeMS === 0.0 || $totalExecutionTimeMS === 0) {
```
e-moe: s it safe to compare float values like this? Maybe we can use very small epsilon value e.g. if ($totalExecutionTimeMS < $eps) where $eps = 1e-5 or something like this

DataCollector/DoctrineDataCollector.php
```diff
@@ -222,15 +222,26 @@ public function getGroupedQueries()
...
+            return 0;
```
greg0ire: Returning 100 would make more sense IMO

DataCollector/DoctrineDataCollector.php
greg0ire: @stof the first parens is after 120 chars, so I think this was the right choice. @mikeSimonson you should remove the space after `$totalExecutionMS`.

kimhemsoe:
Is there any remaining issues which is blocking a merge?

DataCollector/DoctrineDataCollector.php
Majkl578: Yes, this comparison looks dangerous.

PR #640, opened 2017-03-21
Add aliases to help autowiring

nicolas-grekas:


DependencyInjection/DoctrineExtension.php
```diff
@@ -126,6 +126,7 @@ protected function dbalLoad(array $config, ContainerBuilder $container)
 
         $container->setAlias('database_connection', sprintf('doctrine.dbal.%s_connection', $this->defaultConnection));
         $container->setAlias('doctrine.dbal.event_manager', new Alias(sprintf('doctrine.dbal.%s_connection.event_manager', $this->defaultConnection), false));
+        $container->setAlias('Doctrine\DBAL\Driver\Connection', new Alias('database_connection', false));
```
stof: I suggest putting this directly in dbal.xml, as it does not need to be dynamic based on the configuration

DependencyInjection/DoctrineExtension.php
```diff
@@ -373,6 +374,8 @@ protected function ormLoad(array $config, ContainerBuilder $container)
         }
 
         $container->setAlias('doctrine.orm.entity_manager', sprintf('doctrine.orm.%s_entity_manager', $config['default_entity_manager']));
+        $container->setAlias('Doctrine\ORM\EntityManagerInterface', new Alias('doctrine.orm.entity_manager', false));
+        $container->setAlias('Doctrine\Common\Persistence\ObjectManager', new Alias('doctrine.orm.entity_manager', false));
```
stof: same for these (but in orm.xml)

nicolas-grekas: done

Tests/RegistryTest.php
```diff
@@ -116,36 +116,6 @@ public function testGetUnknownEntityManager()
         $registry->getManager('default');
     }
 
-    public function testResetDefaultEntityManager()
```
nicolas-grekas: these two test rely on deprecated behavior, and do not test things in the bundle, be behavior in the doctrine bridge. Since it's not the responsibility of the bundle to test the bridge, they can and should be removed.

nicolas-grekas:
Remaining failure will be fixed once https://github.com/symfony/symfony/pull/22164 is merged up to 2.8.

fabpot:
Thank you @nicolas-grekas.

Issue #638, opened 2017-03-21
Fix possible memory leak break unit tests improvement


LordWeedlle:
A fix was made in https://github.com/doctrine/DoctrineBundle/commit/79bc1e440f8401cc87d33fad45d811ea4ccb8c5d to avoid possible memory leak as descibed in https://github.com/Behat/Symfony2Extension/issues/52.

In my company we used to recycle connections between tests using savepoints to improve unit tests time and avoid reloading database for each test. We first load all data in database and then only use savepoints so we can rollback on initial state before running next test.

Is there a way to specify if we want to close connections?
We could use parameters to check if connections and also entity managers need to be cleared:
```php
// Clear all entity managers to clear references to entities for GC
if ($this->container->hasParameter('doctrine.entity_managers') && !($this->container->hasParameter('doctrine.keep_entity_managers') && $this->container->getParameter('doctrine.keep_entity_managers') === true)) {
    foreach ($this->container->getParameter('doctrine.entity_managers') as $id) {
        if (!$this->container instanceof IntrospectableContainerInterface || $this->container->initialized($id)) {
            $this->container->get($id)->clear();
        }
    }
}

// Close all connections to avoid reaching too many connections in the process when booting again later (tests)
if ($this->container->hasParameter('doctrine.connections') && !($this->container->hasParameter('doctrine.keep_connections') && $this->container->getParameter('doctrine.keep_connections') === true)) {
    foreach ($this->container->getParameter('doctrine.connections') as $id) {
        if (!$this->container instanceof IntrospectableContainerInterface || $this->container->initialized($id)) {
            $this->container->get($id)->close();
        }
    }
}
```

sandvige:
:+1:

dmaicher:
@LordWeedlle I know this issue is quite old but maybe this bundle can help you: https://github.com/dmaicher/doctrine-test-bundle

alcaeus:
I believe the bundle @dmaicher suggested is better suited to help with this. I believe cleaning up on shutdown is desired and should not be worked around in certain environments (e.g. tests). Closing here.

LordWeedlle:
It seems this bundle can help me, I'll try it out!

PR #642, opened 2017-03-22
Adding support for close and clear on shutdown

sandvige:
This pull request attempts to handle this case: https://github.com/doctrine/DoctrineBundle/issues/638


DoctrineBundle.php
```diff
@@ -119,17 +119,21 @@ public function shutdown()
         // Clear all entity managers to clear references to entities for GC
         if ($this->container->hasParameter('doctrine.entity_managers')) {
             foreach ($this->container->getParameter('doctrine.entity_managers') as $id) {
-                if (!$this->container instanceof IntrospectableContainerInterface || $this->container->initialized($id)) {
-                    $this->container->get($id)->clear();
+                if ($this->container->getParameter('doctrine.orm.entity_managers.' . $id . '.clear_on_shutdown') === true) {
```
Ocramius: Can you split this to private methods?

DoctrineBundle.php
```diff
@@ -119,17 +119,21 @@ public function shutdown()
...
                 }
             }
         }
 
         // Close all connections to avoid reaching too many connections in the process when booting again later (tests)
         if ($this->container->hasParameter('doctrine.connections')) {
             foreach ($this->container->getParameter('doctrine.connections') as $id) {
-                if (!$this->container instanceof IntrospectableContainerInterface || $this->container->initialized($id)) {
-                    $this->container->get($id)->close();
+                if ($this->container->getParameter('doctrine.dbal.connections.' . $id . '.close_on_shutdown') === true) {
```
Ocramius: Same as above

DoctrineBundle.php
stof: Putting the whole service id inside the parameter id looks wrong to me.

And your never use the configuration to configure this parameter, so it won't work.

sandvige: Where do you prefer privates methods? At start or end of the file?

sandvige:
@Ocramius @stof We fixed unit tests, but it looks like there's a problem with deprecated calls. Should we investigate the error or this is already done into another PR that is handling it?
```
Resetting a non-lazy manager service is deprecated since Symfony 3.2 and will throw an exception in version 4.0. Set the "doctrine.orm.default_entity_manager" service as lazy and require "symfony/proxy-manager-bridge" in your composer.json file instead
```

sandvige:
Ok everything is fixed after a merge with upstream

sandvige:
@Ocramius ? @stof ?

kimhemsoe:
Is this ready for merge?

mikeSimonson:
@kimhemsoe @sandvige I have the impression that there is not test to verify that close_on_shutdown is called when it's set and not called when it's not. Same for clear_on_shutdown .

kimhemsoe:
ping @sandvige

kimhemsoe:
@sandvige Can you provide the missing tests?

kimhemsoe:
@sandvige Sorry i am closing this for now. If you or someone else have time in the future to add the missing items i will ofc reopen.

PR #647, opened 2017-04-01
Fix division by 0 when total execution ms = 0ms

GeoffreyHervet:
Hello,

When you want to go to the profiler and you total request time is < 0.005ms, total execution time = 0ms.

screen shot 2017-04-01 at 12 00 40
screen shot 2017-04-01 at 11 55 33


mikeSimonson:
@GeoffreyHervet Thanks. This is a duplicate of #634

PR #658, opened 2017-05-06
Custom repository service compiler pass

magnusnordlander:
Added a compiler pass to register custom repositories as services, given they only belong only to one entity manager. 

In accordance with the upcoming Symfony 3.3 standards, and in order to make it easy to use with autowiring, the service name is the class name of the repository.

## Todo

- [x] Write some functional tests for the different scenarios in which repositories are registered or not.

### DIFF
--- a/DependencyInjection/Compiler/RepositoryAliasPass.php
+++ b/DependencyInjection/Compiler/RepositoryAliasPass.php
@@ -8,6 +8,7 @@ use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\HttpKernel\Kernel;
 
 class RepositoryAliasPass implements CompilerPassInterface
 {
@@ -46,19 +47,73 @@ class RepositoryAliasPass implements CompilerPassInterface
                 }
             }
 
+            $rootConflicts = [];
+            foreach ($customRepositories as $repositoryClass => $entities) {
+                $repoConflicts = $this->findConflictingServices($container, $repositoryClass);
+
+                if (count($repoConflicts)) {
+                    $rootConflicts[$repositoryClass] = $repoConflicts;
+                }
+            }
+
             foreach ($customRepositories as $repositoryClass => $entities) {
                 if ($container->has($repositoryClass)) {
                     continue;
                 }
 
-                if (count($entities) === 1) {
-                    $container->register($repositoryClass, $repositoryClass)
-                        ->setFactory([new Reference('doctrine'), 'getRepository'])
-                        ->setArguments($entities[0])
-                        ->setShared(false)
-                    ;
+                if (count($entities) !== 1) {
+                    $this->log($container, "Cannot auto-register repository \"".$repositoryClass."\": Entity belongs to multiple entity managers.");
+                    continue;
                 }
+
+                if (isset($rootConflicts[$repositoryClass])) {
+                    $this->log($container, "Cannot auto-register repository \"".$repositoryClass."\": There are already services for the repository class.");
+                    continue;
+                }
+
+                foreach ($rootConflicts as $conflictingRepo => $conflicts) {
+                    if (is_subclass_of($repositoryClass, $conflictingRepo, true)) {
+                        $this->log($container, "Cannot auto-register repository \"".$repositoryClass."\": There are already services for a superclass of the repository class.");
+                        continue 2;
+                    }
+                }
+
+                $container->register($repositoryClass, $repositoryClass)
+                    ->setFactory([new Reference('doctrine'), 'getRepository'])
+                    ->setArguments($entities[0])
+                    ->setShared(false)
+                    ->setPublic(false)
+                ;
             }
+        }
+    }
+
+    private function findConflictingServices(ContainerBuilder $container, $repositoryClass)
+    {
+        if (Kernel::MAJOR_VERSION >= 4) {
+            return [];
+        }
+
+        $conflictingServices = [];
+        $parameterBag = $container->getParameterBag();
+
+        foreach ($container->getDefinitions() as $id => $definition) {
+            $defClass = $parameterBag->resolveValue($definition->getClass());
+
+            if ($defClass != $repositoryClass) {
+                continue;
+            }
+
+            $conflictingServices[] = $id;
+        }
+
+        return $conflictingServices;
+    }
+
+    private function log(ContainerBuilder $container, $message)
+    {
+        if (method_exists($container, 'log')) {
+            $container->log($this, $message);
         }
     }
 }
--- a/DependencyInjection/Compiler/RepositoryAliasPass.php
+++ b/DependencyInjection/Compiler/RepositoryAliasPass.php
@@ -47,6 +47,10 @@ class RepositoryAliasPass implements CompilerPassInterface
             }
 
             foreach ($customRepositories as $repositoryClass => $entities) {
+                if ($container->has($repositoryClass)) {
+                    continue;
+                }
+
                 if (count($entities) === 1) {
                     $definition = new Definition($repositoryClass);
                     $definition->setFactory(array(
--- a/DependencyInjection/Compiler/RepositoryAliasPass.php
+++ b/DependencyInjection/Compiler/RepositoryAliasPass.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
+
+use Doctrine\Common\Persistence\Mapping\Driver\MappingDriver;
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+
+class RepositoryAliasPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->hasParameter('doctrine.entity_managers')) {
+            return;
+        }
+
+        $entityManagers = $container->getParameter('doctrine.entity_managers');
+        $customRepositories = [];
+
+        foreach ($entityManagers as $name => $serviceName) {
+            $metadataDriverService = sprintf('doctrine.orm.%s_metadata_driver', $name);
+
+            if (!$container->has($metadataDriverService)) {
+                continue;
+            }
+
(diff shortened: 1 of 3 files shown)


DependencyInjection/Compiler/RepositoryAliasPass.php
```diff
@@ -0,0 +1,68 @@
...
+                if (count($entities) === 1) {
+                    $definition = new Definition($repositoryClass);
+                    $definition->setFactory(array(
+                        new Reference('doctrine'),
+                        'getRepository'
+                    ));
+                    $definition->setArguments($entities[0]);
+                    $definition->setShared(false);
+
+                    $container->setDefinition($repositoryClass, $definition);
```
GuilhemN: Why not using `register`?

magnusnordlander: No reason in particular, just habit. I'm not especially attached to using `setDefinition`, though, I'll change it.

DependencyInjection/Compiler/RepositoryAliasPass.php
```diff
@@ -0,0 +1,64 @@
...
+
+        foreach ($entityManagers as $name => $serviceName) {
+            $metadataDriverService = sprintf('doctrine.orm.%s_metadata_driver', $name);
+
+            if (!$container->has($metadataDriverService)) {
+                continue;
+            }
+
+            /** @var MappingDriver $metadataDriver */
+            $metadataDriver = $container->get($metadataDriverService);
```
Koc: Does getting services from non-compiled container not deprecated?

magnusnordlander: It's not marked as deprecated in dev-master at least. I'll look in to it, otherwise we can ping someone who knows for sure.

magnusnordlander: Doesn't seem like it, from what I can find.

Koc: Looks like it was reverted https://github.com/symfony/symfony/pull/18728#discussion_r62531184

DependencyInjection/Compiler/RepositoryAliasPass.php
```diff
@@ -0,0 +1,64 @@
...
+            foreach ($customRepositories as $repositoryClass => $entities) {
+                if ($container->has($repositoryClass)) {
+                    continue;
+                }
+
+                if (count($entities) === 1) {
+                    $container->register($repositoryClass, $repositoryClass)
+                        ->setFactory([new Reference('doctrine'), 'getRepository'])
+                        ->setArguments($entities[0])
+                        ->setShared(false)
```
weaverryan: What about `->setPublic(false)`?

This is a bit controversial, but it's consistent with decisions we've made so far in 3.3 (like adding `public: false` under `_defaults` by default). It also makes this feature more "lean" - if the user doesn't end up using any of these repositories as a service, they'll be removed from the container.

magnusnordlander: I did consider this, and obviously you're more in tune with the new upcoming philosophy, but what made me decide to make the services public is the fact that they're not internal to the bundle, they're intended to be exposed as part of the bundle's "service interface".

Does this make sense?

DependencyInjection/Compiler/RepositoryAliasPass.php
```diff
@@ -0,0 +1,64 @@
...
+                    ];
+                }
+            }
+
+            foreach ($customRepositories as $repositoryClass => $entities) {
+                if ($container->has($repositoryClass)) {
+                    continue;
+                }
+
+                if (count($entities) === 1) {
```
magnusnordlander: Absolutely. I just didn't know that there was a logger facility available in the ContainerBuilder :)

magnusnordlander:
> There is one autowiring consequence: if you already have some of your repositories registered as services, and you are using autowiring, then these new services will be used for autowiring instead of your existing service (because when the service id matches the type exactly, that wins).

Yes... Actually it's even more complicated, because the services registered here uses the manager registry as a factory.

If you:
 * Have repositories registered as a service
 * Are already using autowiring
 * Did not use the class name as the service name
 * Have somehow made ManagerRegistry::getRepository not work (which isn't uncommon)

Then yes, you're gonna have a problem that you need to resolve by either:
 * Renaming or aliasing your current repository service to the class name
 * Implementing a custom repository factory making ManagerRegistry::getRepository work again

DependencyInjection/Compiler/RepositoryAliasPass.php
weaverryan: It's because it's NEW!!! :D The profiler has a new tab for "compilation" stuff... which I don't think many users will use, but it DOES contain answers if you have some weird situation. And for bigger nerds, the new tab is super fun :)

magnusnordlander:
As per OOB conversations with @weaverryan I have now updated this PR with logging, making the services private, and stopping the registration of a repository if there's already a service for it. This is to maintain compatibility in some edge cases with people using autowiring with non-class-name service ids. Since this behaviour is deprecated in 3.3 and dropped in 4.0, the check is only made if the Symfony version is < 4.0.

magnusnordlander:
I've been looking in to writing some functional tests for this, but it's a bit more convoluted to functional test extensions than I had initially guessed, and I've been too busy the past couple of weeks to devote any serious time to it. 

If anyone has a spare bit of time to help by contributing some function tests for this, I'd be very grateful. Does perhaps @Nyholm have some time to spare?

magnusnordlander:
Okay. So, funny story: after having talked about this I got to thinking, and I came up with another way to do the tests that I wanted to. It's still pretty convoluted, but nowhere near as bad as the other approach I had in mind. Found a pretty serious bug too while testing. 

The tests aren't quite as elegant as one could perhaps wish for, but they seem to get the job done. I'm pretty happy with this PR now.

Tests/DependencyInjection/AutoregisterRepositoriesTest.php
```diff
@@ -0,0 +1,165 @@
...
+        $containerBuilder->has(ClassARepository::class)->willReturn(false);
+        if (method_exists(ContainerBuilder::class, 'log')) {
+            $containerBuilder->log($pass, Argument::any())->shouldBeCalled();
+        }
+        $containerBuilder->register(Argument::any(), Argument::any())->shouldNotBeCalled();
+
+        $pass->process($containerBuilder->reveal());
+    }
+
+    protected function prepPropheciesForOneEmAndClassA($containerBuilder, $metadataDriver, $parameterBag)
```
weaverryan: this could be private

DependencyInjection/Compiler/RepositoryAliasPass.php
```diff
@@ -0,0 +1,118 @@
...
+                }
+            }
+        }
+
+        $rootConflicts = [];
+        foreach ($customRepositories as $repositoryClass => $entities) {
+            $repoConflicts = $this->findConflictingServices($container, $repositoryClass);
+
+            if (count($repoConflicts)) {
+                $rootConflicts[$repositoryClass] = $repoConflicts;
```
weaverryan: It looks like the `$repoConflicts` isn't actually used. This could just be set to a boolean (and `findConflictingServices` could also be renamed and return a boolean).

Tests/DependencyInjection/AutoregisterRepositoriesTest.php
```diff
@@ -0,0 +1,165 @@
...
+use Prophecy\Argument;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\RepositoryAliasPass;
+use Doctrine\Common\Persistence\Mapping\Driver\MappingDriver;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\ParameterBag\ParameterBag;
+use Symfony\Component\HttpKernel\Kernel;
+
+
+class AutoregisterRepositoriesTest extends TestCase
```
weaverryan: extra line here - and add the license stuff on top (of both files)

DependencyInjection/Compiler/RepositoryAliasPass.php
```diff
@@ -0,0 +1,118 @@
...
+
+use Doctrine\Common\Persistence\Mapping\Driver\MappingDriver;
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\HttpKernel\Kernel;
+
+class RepositoryAliasPass implements CompilerPassInterface
```
weaverryan: Add some phpdoc here

DependencyInjection/Compiler/RepositoryAliasPass.php
```diff
@@ -0,0 +1,118 @@
...
+            return;
+        }
+
+        $entityManagers = $container->getParameter('doctrine.entity_managers');
+        $customRepositories = [];
+
+        foreach ($entityManagers as $name => $serviceName) {
+            $metadataDriverService = sprintf('doctrine.orm.%s_metadata_driver', $name);
+
+            if (!$container->has($metadataDriverService)) {
```
weaverryan: This is a private service. However, I've just discussed with Nicolas - this is still legal to do in the `ContainerBuilder`.

However, is it possible (and does it make sense) to use `doctrine.orm.default_entity_manager.metadata_factory`? I'm worried about the extra overhead of parsing all the metadata on each rebuild. However, that service might have too many side effects - e.g. caching (and we could be rebuilding the cache from a non-production machine where the cache isn't available). I don't honestly know if it's significant or not - we should probably check that.

magnusnordlander: It is a private service, but, nota bene, it's a private service from this bundle. 

Using `doctrine.orm.default_entity_manager.metadata_factory` does not seem possible to me. That service depends on the entire entity manager being fully constructed. At the very least, it's not possible with the current extension code structure (because the entity manager is constructed using parent services).

magnusnordlander:
Thanks @weaverryan! Fixed.

xabbuh:
You may want to read #660 where @stof have some strong reasons against this feature.

magnusnordlander:
Wow. I didn't think about that. That is a huge bummer. 

I'm not sure that this *should* be a blocker though, given that the exact same issue is already present in the bundle with the entity manager. If the entity manager is closed, any service already having been constructed with the entity manager becomes unusable. If we're concerned about closed entity managers, we should be deprecating those services.

I also suppose the issue could be circumvented though by using a proxy that always grabs a new repository from the registry, but that would have performance implications for each call... I'm gonna do some experimentation with that.

stof:
@magnusnordlander no. As of Symfony 3.2, the EntityManager itself can be injected without breaking the resetting feature (due to using a proxy around the actual entity manager)

kimhemsoe:
Closing. Replaced by #719 and #725

PR #661, opened 2017-05-16
ie11: fix icon.svg width

jglhelias:
`icon.svg` lacks of a width attribute.  
This results (in ie11) the icon is taking too much space.  
I suggest add this explicit attribute as in [twig.svg](https://github.com/symfony/symfony/blob/5129c4cf7e294b1a5ea30d6fec6e89b75396dcd2/src/Symfony/Bundle/WebProfilerBundle/Resources/views/Icon/twig.svg?short_path=bfb4e6d#L1).

screenshot:
screenshot


stof:
thanks

PR #663, opened 2017-05-25
Improvement to IDE-friendliness when 'wrapper_class' option is used

Xymanek:
Sets the `class` property in service definition to assist IDEs in knowing what class the service really is. Since connection services use a factory, the `class` option doesn't affect anything. I've tested the logic with following "hack" and it works without any problems:

```php
class DataConnectionCompilerPass implements CompilerPassInterface
{
    public function process (ContainerBuilder $container)
    {
        $service = $container->findDefinition('doctrine.dbal.data_connection');
        $service->setClass(DataConnectionWrapper::class);
    }
}
```

P.S. Why so many `private`s in `\Doctrine\DBAL\Connection`? Makes extending it so hard....

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -210,6 +210,11 @@ class DoctrineExtension extends AbstractDoctrineExtension
             ))
         ;
 
+        // Set class in case "wrapper_class" option was used to assist IDEs
+        if (isset($options['wrapperClass'])) {
+            $def->setClass($options['wrapperClass']);
+        }
+
         if (!empty($connection['use_savepoints'])) {
             $def->addMethodCall('setNestTransactionsWithSavepoints', array($connection['use_savepoints']));
         }
@@ -220,11 +225,6 @@ class DoctrineExtension extends AbstractDoctrineExtension
                 new Reference(sprintf('doctrine.dbal.%s_connection', $name))
             ));
             $container->setDefinition(sprintf('doctrine.dbal.%s_shard_manager', $name), $shardManagerDefinition);
-        }
-
-        // Set class in case "wrapper_class" option was used to assist IDEs
-        if (isset($options['wrapperClass'])) {
-            $def->setClass($options['wrapperClass']);
         }
     }
 
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -221,6 +221,11 @@ class DoctrineExtension extends AbstractDoctrineExtension
             ));
             $container->setDefinition(sprintf('doctrine.dbal.%s_shard_manager', $name), $shardManagerDefinition);
         }
+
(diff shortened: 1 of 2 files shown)


Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -196,6 +197,28 @@ public function testDbalLoad()
...
+                ))),
+                array(),
+                array('dbal' => array('default_connection' => 'foo')),
+                array()
+            ),
+            $container
+        );
+
+        $this->assertEquals(TestWrapperClass::class, $container->getDefinition('doctrine.dbal.default_connection')->getClass());
+        $this->assertEquals(null, $container->getDefinition('doctrine.dbal.second_connection')->getClass());
```
stloyd: `assertNull()`

Xymanek: Ooops

Xymanek: Anything else? Don't want to pollute with commits

Ocramius:
> P.S. Why so many privates in `\Doctrine\DBAL\Connection`? Makes extending it so hard....

It's not meant to be extended.

Xymanek:
Anything else needs to be done?

DependencyInjection/DoctrineExtension.php
```diff
@@ -221,6 +221,11 @@ protected function loadDbalConnection($name, array $connection, ContainerBuilder
             ));
             $container->setDefinition(sprintf('doctrine.dbal.%s_shard_manager', $name), $shardManagerDefinition);
         }
+
+        // Set class in case "wrapper_class" option was used to assist IDEs
+        if (isset($options['wrapperClass'])) {
+            $def->setClass($options['wrapperClass']);
+        }
```
weaverryan: I would move this up to after line 211 - so right after `$def = `

weaverryan:
+1 The `wrapperClass` option is ultimately used in the DBAL to create a class of that specific instance. So, yea, technically, the definition's should match the actual class that will be created.

Xymanek:
@weaverryan Fixed and fixed

Ocramius:
@Xymanek can you please rebase instead of merging `master`? Commit Xymanek/DoctrineBundle@124573478feb51e4908e72077f1c06982d80ff15 should go away

Xymanek:
@Ocramius can you please give me a hint on how to do that? I kinda know what you mean but not sure how to do it cross-repository...

Ocramius:
@Xymanek I'll do that now while merging :-)

If you never did it before, consider reading https://www.atlassian.com/git/tutorials/rewriting-history

Ocramius:
Manually rebased and merged via doctrine/DoctrineBundle@1818d3d60bff83bdeea79cbb852eed104217998f

Thanks @Xymanek!

PR #668, opened 2017-05-31
Require PHP 7

alcaeus:
As with ORM itself, the bundle will require PHP 7 in the future. Older Symfony versions may be removed in a separate pull request. I haven't made the effort to use the `::class` constant and short array syntax throughout, I'll leave that for a separate pull request.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,15 +7,14 @@ cache:
     - $HOME/.composer/cache
 
 php:
-  - 7.0
   - 7.1
   - nightly
 
 matrix:
   include:
-    - php: 7.0
+    - php: 7.1
       env: deps=low SYMFONY_DEPRECATIONS_HELPER=weak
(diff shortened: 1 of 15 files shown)


lcobucci:
We're actually going to require php 7.1, so I'd suggest to already do it here

.travis.yml
```diff
@@ -7,20 +7,15 @@ cache:
     - $HOME/.composer/cache
 
 php:
-  - 5.5
-  - 5.6
   - 7.0
```
Ocramius: We should drop this too, as per @lcobucci's review

alcaeus:
@Ocramius @lcobucci Done.

Ocramius:
>  let's forget this fake "mario"

f677d-tumblr_mrbqtkq9cw1r0ralmo1_r2_500

Issue #673, opened 2017-06-27
BC break between 1.6.7 -> 1.6.8 regarding driver selection


Majkl578:
After upgrading from 1.6.7 to 1.6.8, all our tests started failing. While investigating, found that DBAL driver is somehow not picked up correctly.
All functional tests failing with:
```
Doctrine\DBAL\Exception\DriverException: An exception occured in driver: could not find driver
```

It suddenly tries to load pdo_mysql instead of pdo_pgsql.

Bisected this breakage to commit 99a62a7c4139c0ccb6969de5eff1447b1a6cbaf4.

This is what our DBAL config looks like:
```
doctrine:
    dbal:
        default_connection: xxx
        connections:
            xxx:
                server_version: 9.4
                driver: "%database_driver%"
                host: "%database_host%"
                port: "%database_port%"
                dbname: "%database_name%"
                user: "%database_user%"
                password: "%database_password%"
                charset: UTF8
                mapping_types:
                    bit: boolean
                    tsvector: string
        types:
            foo:
                class: X\Y\Types\FooType # <-- commented type
```

Ocramius:
Indeed, `$platform = $connection->getDatabasePlatform();` is the culprit here.

@mikeSimonson how much of an issue would a reversal of the bugfix be? I think this one is more critical...

mikeSimonson:
Does that mean that somehow the $connection->getDatabasePlatform(); call override the server version config value ? Reading the code it seems like it shouldn't be the case (it bails out if it's already set).

@Majkl578 Can you dig further with xdebug to see how that value gets changed ?

stof:
@mikeSimonson  `$connection->getDatabasePlatform()` needs to know the server version, to get the right platform. If you have not configured it explicitly in the config, DBAL tries to guess it, and that requires connecting to the server.

However, I don't see any reason for using MySQL (unless your project does weird thing changing the database config at runtime to move it from mysql to postgresql, and the code now connects before your magic.

@Ocramius I'm in favor of reverting the bugfix and deprecating the feature entirely. Marking a type as commented should be done in the Type class itself rather than in the bundle config (DBAL supports marking a type as commented in the Type itself since 2.4)

mikeSimonson:
Does that mean that somehow the $connection->getDatabasePlatform(); call override the server version config value ? Reading the code it seems like it shouldn't be the case (it bails out if it's already set).

@Majkl578 Can you dig further with xdebug to see how that value gets changed ?

[...]

PR #676, opened 2017-06-29
Revert "Fix an issue were some commented type (#623)"

sebastianfeldmann:
This reverts commit 99a62a7.

There was an issue with the database driver configuration.
You can see the whole discussion here:
https://github.com/doctrine/DoctrineBundle/issues/673

### DIFF
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -61,9 +61,6 @@ class ConnectionFactory
             foreach ($mappingTypes as $dbType => $doctrineType) {
                 $platform->registerDoctrineTypeMapping($dbType, $doctrineType);
             }
-        }
-        if (!empty($this->commentedTypes)) {
-            $platform = $connection->getDatabasePlatform();
             foreach ($this->commentedTypes as $type) {
                 $platform->markDoctrineTypeCommented(Type::getType($type));
             }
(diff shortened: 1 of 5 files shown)


Ocramius:
As discussed with @sebastianfeldmann, we need:

 * some sort of explanation of what to do with types that need to be registered with the platform, and how to avoid them
 * some sort of regression prevention test that avoids #673 from happening again. Mocking the connection object and verifying that `getDatabasePlatform` is NOT called would be best

sebastianfeldmann:
@mikeSimonson patch is "correct" but revealed another problem

stof:
why closing the PR ? The discussion on #673 seems to agree than the new situation is worse than the old one, so reverting to the old one is better (and there is a better way to mark types as commented in DBAL anyway)

Ocramius:
> seems to agree

Yes, but the correct solution is to still register types with the platform. The previous solution was only registering types if they were also db-types, not just commented types.

The patch does not need to be reverted, it is correct. What needs to happen is better documentation around the issue, and removing `->getDatabasePlatform()` calls overall (although it can't happen without a major release)

stof:
@Ocramius no, the correct solution is to mark types as commented in the type itself, so that the platform itself registers the type as commented *the first time it needs to load the list of commented types*. This way, there is no need to register them as commented early anymore (forcing thr usage of the platform early too)

stof:
And registering early without calling `getDatabasePlatform()` is impossible, as registering happens on the platform object. This is why using the DBAL feature of lettign the type itself know it needs to be commented is better.

Ocramius:
The fact that the feature shouldn't be used is a thing, but the fact that
it didn't work at all when type platform mappings were missing completely
was a legit (and fixed) bug that exposed a deeper problem.

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

PR #677, opened 2017-06-29
Enhance exception message

sebastianfeldmann:
The error occures if doctrine needs to establish a connection to
determine the platform version to register the types.
The idea is to provide some more information so the user can
understand what and why this is happening.

This is discussed at:
https://github.com/doctrine/DoctrineBundle/issues/673

### DIFF
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -16,7 +16,10 @@ namespace Doctrine\Bundle\DoctrineBundle;
 
 use Doctrine\Common\EventManager;
 use Doctrine\DBAL\Configuration;
+use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DriverManager;
+use Doctrine\DBAL\DBALException;
+use Doctrine\DBAL\Exception\DriverException;
 use Doctrine\DBAL\Types\Type;
 
 /**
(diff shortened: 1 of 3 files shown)


mikeSimonson:
@sebastianfeldmann Nice
Thanks

mikeSimonson:
Can you add a test for this that trigger the exception ?

sebastianfeldmann:
I can definitely try ;)

sebastianfeldmann:
I added a unit test, found and fixed an error, thats good, but now I face a different problem.

The exception we are trying to catch is of type `DriverException`.
The minimum composer requirement for Doctrine DBAL is `2.3`. The `DriverException` which I'm trying to catch and enhance was added in Doctrine DBAL `2.4`.

### List of options
1. Bump up the version requirement to am minimum of 2.4, I don't know if that's a good Idee. I think changes like that require a major version update :/
2. Catch a more generic exception type, but then we would change exceptions we definitely don't want to change, so not a good idea either.
3. Search for some specific exception messages and only change those, but that's definitely not a good idea.

So I'm kind of stuck. Any suggestions?

Ocramius:
> Bump up the version requirement to am minimum of 2.4, I don't know if that's a good Idee. I think changes like that require a major version update :/

Feel free to do that - bumping dependencies is *NOT* a BC break

 > Catch a more generic exception type, but then we would change exceptions we definitely don't want to change, so not a good idea either.

Would avoid that - too easy to cause other bugs

 > Search for some specific exception messages and only change those, but that's definitely not a good idea.

This makes the BC boundary too complex to tackle.

I would suggest bumping to `"doctrine/dbal": "^2.5.12"`

sebastianfeldmann:
I followed your advice and bumped the dependency to `"doctrine/dbal": "^2.5.12"`.

To "mock" the `DriverManager::getConnection` call I had to create a "FakeDriver" class.
I put the class at the bottom of the `ConnectionFactoryTest` file. Hope that's ok.

Additionally I had to change the exception type.
The exception we catch is a `DriverException`. To create a `DriverException` the previous exception has to implement the `DriverExceptionInterface`. Sadly the `DriverException` we catch doesn't implement the `DriverExceptionInterface` itself. So it is not possible to put a `DriverException` into a `DriverException`.
So I endet up throwing a `Doctrine\DBALException` to pass the `DriverException` as the previous  exception.

Hope that wasn't to confusing ;)

kimhemsoe:
@sebastianfeldmann Thanks! very nice

PR #681, opened 2017-07-28
[Webprofiler] Improve SQL explain table display

mimol91:
Table which is shown in Weprofiler is not clean while values in sql explain table are long.
This PR adds horizontal scroll for long tables. 

**Related PR in Symfony Webprofiler:** https://github.com/symfony/symfony/pull/23706
_(It does not cause any backward compatibility problems)_

**Before:**
before

**After:**
after

### DIFF
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -263,7 +263,7 @@
                                 </div>
 
                                 {% if query.explainable %}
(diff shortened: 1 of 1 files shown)


mikeSimonson:
Thanks for the the PR, I am going to merge this right away but I will wait for @stof advice on the second before to merge it.

PR #684, opened 2017-08-06
Register commands using PSR-4 based Service Discovery

GuilhemN:
Commands auto-registration will be deprecated soon (see https://github.com/symfony/symfony/pull/23805) so we need to stop using it.

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -72,6 +72,20 @@ class DoctrineExtension extends AbstractDoctrineExtension
 
         $this->adapter->loadServicesConfiguration($container);
 
+        // Use a prototype when the ORM is available.
+        // This avoids listing all ORM commands by hand.
+        if (class_exists(Version::class)) {
+            $loader = new XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
+
+            $prototype = new Definition();
+            $prototype
+                ->setPublic(false)
+                ->addTag('console.command')
+            ;
+
+            $loader->registerClasses($prototype, 'Doctrine\Bundle\DoctrineBundle\Command\\', __DIR__.'/../Command/*');
+        }
+
         if (!empty($config['dbal'])) {
             $this->dbalLoad($config['dbal'], $container);
         }
--- a/DoctrineBundle.php
+++ b/DoctrineBundle.php
@@ -21,6 +21,7 @@ use Doctrine\Bundle\DoctrineBundle\Command\Proxy\ImportDoctrineCommand;
 use Doctrine\Bundle\DoctrineBundle\Command\Proxy\RunSqlDoctrineCommand;
 use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\EntityListenerPass;
 use Doctrine\ORM\Proxy\Autoloader;
+use Doctrine\ORM\Version;
 use Symfony\Component\Console\Application;
 use Symfony\Component\DependencyInjection\Compiler\PassConfig;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
@@ -140,11 +141,8 @@ class DoctrineBundle extends Bundle
      */
     public function registerCommands(Application $application)
     {
-        // Use the default logic when the ORM is available.
-        // This avoids listing all ORM commands by hand.
-        if (class_exists('Doctrine\\ORM\\Version')) {
-            parent::registerCommands($application);
-
+        if (class_exists(Version::class)) {
+            // Commands are registered in the extension
             return;
         }
 
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "php": "^7.1",
         "symfony/framework-bundle": "~2.7|~3.0|~4.0",
         "symfony/console": "~2.7|~3.0|~4.0",
-        "symfony/dependency-injection": "~2.7|~3.0|~4.0",
+        "symfony/dependency-injection": "~3.3|~4.0",
         "doctrine/dbal": "^2.5.12",
         "jdorn/sql-formatter": "~1.1",
         "symfony/doctrine-bridge": "~2.7|~3.0|~4.0",
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
     ],
     "require": {
         "php": "^7.1",
-        "symfony/framework-bundle": "~2.7|~3.0|~4.0",
+        "symfony/framework-bundle": "~3.0|~4.0",
         "symfony/console": "~2.7|~3.0|~4.0",
         "symfony/dependency-injection": "~3.3|~4.0",
         "doctrine/dbal": "^2.5.12",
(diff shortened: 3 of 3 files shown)


DependencyInjection/DoctrineExtension.php
```diff
@@ -72,6 +72,20 @@ public function load(array $configs, ContainerBuilder $container)
...
         $this->adapter->loadServicesConfiguration($container);
 
+        // Use the default logic when the ORM is available.
+        // This avoids listing all ORM commands by hand.
+        if (class_exists('Doctrine\ORM\Version') && method_exists(XmlFileLoader::class, 'registerClasses')) {
+            $loader = new XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
+
+            $prototype = new Definition();
+            $prototype
+                ->setPublic(False)
```
Koc: wrong case

GuilhemN: Indeed, fixed

DependencyInjection/DoctrineExtension.php
```diff
@@ -72,6 +72,20 @@ public function load(array $configs, ContainerBuilder $container)
 
         $this->adapter->loadServicesConfiguration($container);
 
+        // Use the default logic when the ORM is available.
+        // This avoids listing all ORM commands by hand.
+        if (class_exists('Doctrine\ORM\Version') && method_exists(XmlFileLoader::class, 'registerClasses')) {
```
Ocramius: `Version::class`

DependencyInjection/DoctrineExtension.php
```diff
@@ -72,6 +72,20 @@ public function load(array $configs, ContainerBuilder $container)
...
+            $loader = new XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
+
+            $prototype = new Definition();
+            $prototype
+                ->setPublic(false)
+                ->addTag('console.command')
+            ;
+
+            $loader->registerClasses($prototype, 'Doctrine\Bundle\DoctrineBundle\Command\\', '../../Command/*');
```
Ocramius: Please add `__DIR__`

DoctrineBundle.php
```diff
@@ -143,7 +144,10 @@ public function registerCommands(Application $application)
         // Use the default logic when the ORM is available.
         // This avoids listing all ORM commands by hand.
         if (class_exists('Doctrine\\ORM\\Version')) {
-            parent::registerCommands($application);
+            if (!method_exists(FileLoader::class, 'registerClasses')) {
+                // BC sf < 3.3
```
Ocramius: Could bump the dependency instead

GuilhemN: Is it ok? It will forbid using sf 2.x.

DependencyInjection/DoctrineExtension.php
GuilhemN: done

DependencyInjection/DoctrineExtension.php
GuilhemN: done

DoctrineBundle.php
Ocramius: Yes, that is OK. This is not a bugfix, so it will land in the next minor anyway

Ocramius:
@GuilhemN the commit messages: make them descriptive, please

GuilhemN:
@Ocramius I rebased, these commits were just here so you could know the changes I've done, they didn't bring much value otherwise.

DoctrineBundle.php
GuilhemN: ok, done

DependencyInjection/DoctrineExtension.php
```diff
@@ -72,6 +72,20 @@ public function load(array $configs, ContainerBuilder $container)
 
         $this->adapter->loadServicesConfiguration($container);
 
+        // Use a prototype when the ORM is available.
+        // This avoids listing all ORM commands by hand.
+        if (class_exists(Version::class) && method_exists(XmlFileLoader::class, 'registerClasses')) {
```
ogizanagi: ` && method_exists(XmlFileLoader::class, 'registerClasses')` part of the condition could be removed now that the DI dependency has been bumped to 3.3.

GuilhemN: indeed, updated

DoctrineBundle.php
```diff
@@ -140,11 +140,8 @@ public function shutdown()
      */
     public function registerCommands(Application $application)
     {
-        // Use the default logic when the ORM is available.
-        // This avoids listing all ORM commands by hand.
         if (class_exists('Doctrine\\ORM\\Version')) {
```
Ocramius: `Version::class`

GuilhemN: fixed

kimhemsoe:
Are we OK with dropping support for sf2.7?
Seems to me that it would be possible to support 3.4 without dropping support for 2.7 without much trouble.

kimhemsoe:
Sorry, we pull in symfony/dependency-injection 3.3, which require finder, config and yaml to be 3.3 if installed, if i am correct?

Is that any nasty surprises here? Like is that any issues with console which yaml require to be 3.3?

GuilhemN:
> Are we OK with dropping support for sf2.7?
Seems to me that it would be possible to support 3.4 without dropping support for 2.7 without much trouble.

@Ocramius told me it was ok (https://github.com/doctrine/DoctrineBundle/pull/684#discussion_r131547007).

> Sorry, we pull in symfony/dependency-injection 3.3, which require finder, config and yaml to be 3.3 if installed, if i am correct?
>
>Is that any nasty surprises here? Like is that any issues with console which yaml require to be 3.3?

Indeed I had to bump symfony/framework-bundle and symfony/yaml.

Ocramius:
@kimhemsoe yeah, dropping symfony 2.x and <3.3 is perfectly OK for new improvement releases.

kimhemsoe:
OK i will try to get the last bug fix in for a 1.7 release then. And then we set this for 1.8

GuilhemN:
https://github.com/symfony/symfony/pull/23805 has been merged :)

chalasr:
DI <3.3 layers should be removed once this merged.

Ocramius:
@kimhemsoe please rebase instead

kimhemsoe:
Will do, also need to fix that travis failure we have. 

Hmm.. did not see that the edit ended up with a merge.

DependencyInjection/DoctrineExtension.php
```diff
@@ -73,6 +73,20 @@ public function load(array $configs, ContainerBuilder $container)
 
         $this->adapter->loadServicesConfiguration($container);
 
+        // Use a prototype when the ORM is available.
+        // This avoids listing all ORM commands by hand.
+        if (class_exists(Version::class)) {
```
Majkl578: This type of check should be avoided since Version class is being removed in 3.0, see https://github.com/doctrine/doctrine2/pull/6653.
Probably good enough for now (already used in registerCommands()), but should be handled differently in future.

nicolas-grekas:
can be closed

PR #685, opened 2017-08-09
Add autowiring for Doctrine\DBAL\Connection

stof:
Although this is a class rather than an interface, it is the main API of Doctrine DBAL, which has more helper methods than the driver connection interface. So using it in typehints make sense when using these helpers.

### DIFF
--- a/Tests/DependencyInjection/DoctrineExtensionTest.php
+++ b/Tests/DependencyInjection/DoctrineExtensionTest.php
@@ -16,8 +16,12 @@ namespace Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection;
 
 use Doctrine\Bundle\DoctrineBundle\DependencyInjection\DoctrineExtension;
 use Doctrine\Bundle\DoctrineBundle\Tests\Builder\BundleConfigurationBuilder;
+use Doctrine\Common\Persistence\ManagerRegistry;
+use Doctrine\Common\Persistence\ObjectManager;
 use Doctrine\Common\Proxy\AbstractProxyFactory;
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver\Connection as DriverConnection;
+use Doctrine\ORM\EntityManagerInterface;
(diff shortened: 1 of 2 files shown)


stof:
Well, we don't have tests asserting classes available for autowiring currently AFAIK.

Ocramius:
Yeah, but since the autowiring is squishy by nature, a regression prevention mechanism would be quite useful...

stof:
@Ocramius I added a test ensuring the presence of our autowiring aliases, to ensure they don't get removed by mistake.

I haven't tried writing a test actually performing autowiring, as this is much more complex to implement (and btw, in 3.3, the autowiring would work without this PR when you have only 1 connection defined, but would trigger a deprecation warning because making it work without the alias would involve getting through the BC layer for the old autowiring rules)

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -28,6 +28,30 @@
...
+    public function testAutowiringAlias()
+    {
+        $container = $this->getContainer();
+        $extension = new DoctrineExtension();
+        $config = BundleConfigurationBuilder::createBuilderWithBaseValues()->build();
+
+        $extension->load(array($config), $container);
+
+        $expectedAliases = array(
+            'Doctrine\DBAL\Driver\Connection' => 'database_connection',
```
Ocramius: These can all use `::class`

stof: yeah, I forgot that this bundles allows it 😄

stof: done

kimhemsoe:
@mikeSimonson Dont worry i was on it, was just holding it back to see if 1.7 had issues before i merged this. 

Think i will do a new 1.7.x later with these changes, before 1.8 which will drop support for old symfony versions.

mikeSimonson:
@kimhemsoe Sorry, I didn't wanted to let you all the work.

Issue #639, opened 2017-03-21
Exclude tests from classmap


garak:
I'm watching autoload_classmap.php on a project of mine, production env. The classmap is optimized via `--optimize-autoloader --classmap-authoritative` flags.
I see a lot of definitions about tests, see below.
I was wondering if we could exclude that, using `exclude-from-classmap` directive.

Relevant rows:
```
    'Doctrine\\Bundle\\DoctrineBundle\\Tests\\BundleTest' => $vendorDir . '/doctrine/doctrine-bundle/Tests/BundleTest.php',
    'Doctrine\\Bundle\\DoctrineBundle\\Tests\\Command\\CreateDatabaseDoctrineTest' => $vendorDir . '/doctrine/doctrine-bundle/Tests/Command/CreateDatabaseDoctrineTest.php',
    'Doctrine\\Bundle\\DoctrineBundle\\Tests\\ContainerTest' => $vendorDir . '/doctrine/doctrine-bundle/Tests/ContainerTest.php',
    'Doctrine\\Bundle\\DoctrineBundle\\Tests\\DataCollector\\DoctrineDataCollectorTest' => $vendorDir . '/doctrine/doctrine-bundle/Tests/DataCollector/DoctrineDataCollectorTest.php',
    'Doctrine\\Bundle\\DoctrineBundle\\Tests\\DependencyInjection\\AbstractDoctrineExtensionTest' => $vendorDir . '/doctrine/doctrine-bundle/Tests/DependencyInjection/AbstractDoctrineExtensionTest.php',
    'Doctrine\\Bundle\\DoctrineBundle\\Tests\\DependencyInjection\\DoctrineExtensionTest' => $vendorDir . '/doctrine/doctrine-bundle/Tests/DependencyInjection/DoctrineExtensionTest.php',
    'Doctrine\\Bundle\\DoctrineBundle\\Tests\\DependencyInjection\\TestDatetimeFunction' => $vendorDir . '/doctrine/doctrine-bundle/Tests/DependencyInjection/TestDatetimeFunction.php',
    'Doctrine\\Bundle\\DoctrineBundle\\Tests\\DependencyInjection\\TestFilter' => $vendorDir . '/doctrine/doctrine-bundle/Tests/DependencyInjection/TestFilter.php',
[...]

PR #641, opened 2017-03-22
PSR-4 folders repository reorganization

e-moe:
Fix #639

PR #688, opened 2017-08-12
Add .gitattributes

garak:
This is also fixing #639

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+.gitattributes   export-ignore
+.gitignore       export-ignore
+.travis.yml      export-ignore
+phpunit.xml.dist export-ignore
+/Resources/doc   export-ignore
+/Tests           export-ignore


mikeSimonson:
@garak Thanks

alcaeus:
I'm not sure this is a good idea. This will cause a severe difference between running `composer install --prefer-source` and `composer install --prefer-dist`. IMO it would be better to split up into a `lib` and `tests` folder as we do on many other repositories (and as #641 suggested). What do you think @mikeSimonson?

garak:
What would be the problem with such difference?

alcaeus:
For starters, the issue you linked above is only fixed for some people. People with source installs still have the test sources available in their autoloader.

garak:
Ah OK, I didn't realize that. Anyway, the issue is still open

Ocramius:
@alcaeus we do it on all other repos and it's fine. The tests are NOT public API anyway, so whoever is installing the package as a dependency shouldn't care about whether they are there or not.

mikeSimonson:
Well the next steps are

- to reorganize the repo with the tests outside the rest of the src folder
- and the autoload-dev feature of composer.json

Issue #689, opened 2017-08-18
Phantom MySql Connection


JustBlackBird:
I'm using `DoctrineBundle@1.6.7` and have the following config:

```
doctrine:
    dbal:
        default_connection: xxx
        connections:
            xxx:
                server_version: 9.4
                driver: "%database_driver%"
                host: "%database_host%"
                port: "%database_port%"
                dbname: "%database_name%"
                user: "%database_user%"
                password: "%database_password%"
                charset: UTF8
        types:
            foo:
                class: X\Y\Types\FooType # This is a commented type
```

Then I compile a container I have not only one connection but two: one `xxx` and another one is `default_connection`. The latter one has the following credentials:

```
driver: "pdo_mysql"
host: "localhost"
port: ~
user: root
password: ~
```

The main question is _where does the phantom connection come from?_
___
When I upgrade DoctrineBundle from `1.6.7` to `1.6.8` I get the same result as described in #673.

JustBlackBird:
I've traced the problem a little and found a strange place in [`DependencyInjection\Configuration`](https://github.com/doctrine/DoctrineBundle/blob/master/DependencyInjection/Configuration.php#L69..L84)

This block is executed for `types` array which just cannot have `default_connection`
```
                        $v['default_connection'] = isset($v['default_connection']) ? (string) $v['default_connection'] : 'default';
                        $v['connections'] = array($v['default_connection'] => $connection);
```

stof:
Can you run `bin/console debug:config doctrine` and post the output ?

JustBlackBird:
Sure. Here it is:
```
doctrine:
    dbal:
        types:
            uuid:
                class: Ramsey\Uuid\Doctrine\UuidType
                commented: true
            binary_flag:
                class: SharedKernelBundle\Doctrine\BinaryFlagType
                commented: true
            numeric_id:
                class: SharedKernelBundle\Doctrine\NumericIdType
                commented: true
            year:
                class: SharedKernelBundle\Doctrine\YearType
                commented: true
            date_immutable:
                class: VasekPurchart\Doctrine\Type\DateTimeImmutable\DateImmutableType
                commented: true
            datetime_immutable:
                class: VasekPurchart\Doctrine\Type\DateTimeImmutable\DateTimeImmutableType
                commented: true
            datetimetz_immutable:
                class: VasekPurchart\Doctrine\Type\DateTimeImmutable\DateTimeTzImmutableType
                commented: true
            time_immutable:
[...]

PR #690, opened 2017-08-18
Add tests to illustrate phantom connection

JustBlackBird:
This is a test case to illustrate #689


JustBlackBird:
This example was incorrect

PR #692, opened 2017-08-30
Add services tagged with 'doctrine.dbal.logger' to the connection's chain logger

renan:
I've run into a use case where I need more than one `SQLLogger` on my application. I could've simply redefined the whole service that is provided by Doctrine Bundle but thought supporting this use case in the Bundle itself would be more beneficial as it opens other bundles to also hook into Doctrine SQL Logging.

Services that are tagged with `doctrine.dbal.logger` are added to all connections.

### DIFF
--- a/DependencyInjection/Compiler/DoctrineDBALLoggerPass.php
+++ b/DependencyInjection/Compiler/DoctrineDBALLoggerPass.php
@@ -53,7 +53,7 @@ final class DoctrineDBALLoggerPass implements CompilerPassInterface
 
         if ($reflectionClass->isAbstract()) {
             throw new InvalidArgumentException(sprintf(
-                'The service "%s" tagged "%s" must not be abstract.',
+                'The service "%s" tagged "%s" cannot be abstract.',
                 $serviceId,
                 self::TAG_NAME
             ));
--- a/DependencyInjection/Compiler/DoctrineDBALLoggerPass.php
+++ b/DependencyInjection/Compiler/DoctrineDBALLoggerPass.php
@@ -1,0 +1,91 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
+
+use Doctrine\DBAL\Logging\SQLLogger;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
+use Symfony\Component\DependencyInjection\Reference;
+
+final class DoctrineDBALLoggerPass implements CompilerPassInterface
+{
+    private const TAG_NAME = 'doctrine.dbal.logger';
+    private const LOGGER_INTERFACE = SQLLogger::class;
+    private const BASE_CHAIN_NAME = 'doctrine.dbal.logger.chain';
+
+    public function process(ContainerBuilder $container)
+    {
+        $serviceList = $container->findTaggedServiceIds(self::TAG_NAME, true);
+        $serviceList = array_keys($serviceList);
+
+        foreach ($serviceList as $serviceId) {
+            $this->ensureIsValidServiceDefinition($container, $serviceId);
+            $this->attachServiceToAllConnections($container, $serviceId);
+        }
+    }
+
+    private function ensureIsValidServiceDefinition(ContainerBuilder $container, string $serviceId): void
+    {
+        $serviceDefinition = $container->getDefinition($serviceId);
+
+        $serviceClass = $container->getParameterBag()->resolveValue(
+            $serviceDefinition->getClass()
+        );
+        $reflectionClass = $container->getReflectionClass($serviceClass);
+
+        if (!$reflectionClass) {
+            throw new InvalidArgumentException(sprintf(
+                'Class "%s" used for service "%s" cannot be found.',
+                $serviceClass,
+                $serviceId
+            ));
+        }
+
+        if (!$reflectionClass->implementsInterface(self::LOGGER_INTERFACE)) {
+            throw new InvalidArgumentException(sprintf(
+                'The service "%s" tagged "%s" must implement "%s".',
+                $serviceId,
+                self::TAG_NAME,
+                self::LOGGER_INTERFACE
+            ));
+        }
+
+        if ($reflectionClass->isAbstract()) {
+            throw new InvalidArgumentException(sprintf(
+                'The service "%s" tagged "%s" must not be abstract.',
+                $serviceId,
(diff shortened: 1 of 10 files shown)


DependencyInjection/Compiler/DoctrineDBALLoggerPass.php
```diff
@@ -0,0 +1,91 @@
...
+                'The service "%s" tagged "%s" must implement "%s".',
+                $serviceId,
+                self::TAG_NAME,
+                self::LOGGER_INTERFACE
+            ));
+        }
+
+        if ($reflectionClass->isAbstract()) {
+            throw new InvalidArgumentException(sprintf(
+                'The service "%s" tagged "%s" must not be abstract.',
```
mikeSimonson: can't be

DependencyInjection/Compiler/DoctrineDBALLoggerPass.php
```diff
@@ -0,0 +1,91 @@
...
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
+use Symfony\Component\DependencyInjection\Reference;
+
+final class DoctrineDBALLoggerPass implements CompilerPassInterface
+{
+    private const TAG_NAME = 'doctrine.dbal.logger';
+    private const LOGGER_INTERFACE = SQLLogger::class;
```
stof: I would remove this constant, and use `SQLLogger::class` directly in the `implementsInterface` call

DependencyInjection/Compiler/DoctrineDBALLoggerPass.php
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    private function createEmptyChain(ContainerBuilder $container, string $chainId, string $connectionId): void
+    {
+        $chainReference = new Reference($chainId);
+
+        $container->setDefinition($chainId, new Definition(self::BASE_CHAIN_NAME));
+
+        $configurationDefinition = $container->getDefinition($connectionId . '.configuration');
+        $configurationDefinition->addMethodCall('setSQLLogger', array($chainReference));
```
stof: what if the connection had a logger injected in it already, but not a chain one ? you need to inject the existing logger into your chain

DependencyInjection/DoctrineExtension.php
```diff
@@ -143,29 +143,35 @@ protected function loadDbalConnection($name, array $connection, ContainerBuilder
...
+        unset($connection['profiling']);
 
-                $loggerId = 'doctrine.dbal.logger.chain.'.$name;
-                $container->setDefinition($loggerId, $chainLogger);
-                $logger = new Reference($loggerId);
-            } else {
-                $logger = $profilingLogger;
+        if (!empty($loggers)) {
+            $chainLogger = new $defitionClassname('doctrine.dbal.logger.chain');
+            foreach ($loggers as $logger) {
```
stof: if we have only 1 logger, we should not wrap it in a chain. No need to add overhead for nothing.

renan: Wrapping was done to make it easier to add more loggers, overhead should be non-existent, but I will revert this so that I don't change `DoctrineExtension` at all

stof: Overhead is going through the chain logger layer for each logging call, which may be a lot if you have lots of DB queries (I agree that the cost of the DB query outweights the logger cost anyway though).
Making the compiler pass deal with an existing logger set in the configuration already would be better

renan:
The feedback has been applied.

I undid all changes to `DoctrineExtension` so the previous added overhead is gone.

If the configuration is using a chain logger it will simply append the new loggers into it. Otherwise it will wrap the existing logger plus the new loggers into a new chain.

DependencyInjection/Compiler/DoctrineDBALLoggerPass.php
```diff
@@ -0,0 +1,146 @@
...
+            $chainDefinition = $this->createChainDefinitionAttachedToConnection($container, $connectionName, $configurationDefinition);
+
+            foreach ($serviceList as $serviceId) {
+                $serviceReference = new Reference($serviceId);
+                $chainDefinition->addMethodCall('addLogger', [$serviceReference]);
+            }
+        }
+    }
+
+    private function ensureIsValidServiceDefinition(ContainerBuilder $container, string $serviceId): void
```
kimhemsoe: Is this needed?

I have never noticed any other doing this much for detecting miss configurations of services and symfony it self is happy to fail when you try to retrieve the service from the container.

kimhemsoe: Anyone have a comment on this?

We don't have this anywhere else neither have i seen others do it.

kimhemsoe:
@renan Is it possible for you to have a go at resolving the conflict?

DependencyInjection/Compiler/DoctrineDBALLoggerPass.php
```diff
@@ -0,0 +1,146 @@
...
+
+    private function createChainAndAttachToConnection(
+        ContainerBuilder $container,
+        string $connectionName,
+        Definition $configurationDefinition
+    ): Definition {
+        $chainId = self::BASE_CHAIN_NAME . '.' . $connectionName;
+
+        $chainReference = new Reference($chainId);
+        $chainDefinition = new Definition(self::BASE_CHAIN_NAME);
```
soukicz: Shouldn't there be `LoggerChain::class`?

I am getting error `Symfony\Component\DependencyInjection\Exception\InvalidArgumentException: "'doctrine.dbal.logger.chain'" is not a valid class name for the "doctrine.dbal.logger.chain.default" service.` with self::BASE_CHAIN_NAME

DependencyInjection/Compiler/DoctrineDBALLoggerPass.php
ostrolucky: Yep it's not needed

ostrolucky:
Hey! Thank you for your contribution. I've reviewed this PR in detail as part of #EUFOSSA and I'm afraid this is not mergeable in current state because of following reasons:
- there are 3 feedback comments left unaddressed
- it disregards current `dbal.connections.*.logging` option. This is pretty big one and can create lot of confusion.
- it's undocumented and too implicit, while custom service definition is pretty explicit and easy to do via decorator
- it creates ChainLogger even if there is only single logger
- it adds too much complexity while less complex solution would be to contribute option to inject iterable of loggers into LoggerChain and utilize here [!tagged option available since Symfony 3.4](https://symfony.com/blog/new-in-symfony-3-4-simpler-injection-of-tagged-services). In such case we could just make our own definitions tag and always injected services tagged as such.

PR #697, opened 2017-09-18
Allow PHP 5.5 again

nicolas-grekas:
Not allowing php 5.5 for no real reason doesn't play well with Symfony 3.4 LTS, which still supports 5.5.
All Symfony 3.4 users should be allowed to benefit from Symfony 3.4 features that the doctrine bundle could leverage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,16 @@ cache:
     - $HOME/.composer/cache
 
 php:
+  - 5.5
+  - 5.6
+  - 7.0
   - 7.1
   - nightly
 
 matrix:
   include:
+    - php: 5.6
+      env: SYMFONY_VERSION="2.8.*@dev"
     - php: 7.1
       env: deps=low SYMFONY_DEPRECATIONS_HELPER=weak
     - php: 7.1
@@ -33,7 +38,7 @@ install:
   - if [ "$deps" = "low" ]; then composer --prefer-lowest --prefer-stable update; else composer update; fi;
 
 script:
-  - ./vendor/bin/phpunit -v --coverage-clover ./build/logs/clover.xml
+  - ./vendor/bin/simple-phpunit -v --coverage-clover ./build/logs/clover.xml
 
 after_script:
     - php ./vendor/bin/coveralls -v
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         }
     ],
     "require": {
-        "php": "^7.1",
+        "php": "^5.5.9|^7.0",
         "symfony/framework-bundle": "~2.7|~3.0|~4.0",
         "symfony/console": "~2.7|~3.0|~4.0",
         "symfony/dependency-injection": "~2.7|~3.0|~4.0",
@@ -40,8 +40,7 @@
         "symfony/property-info": "~2.8|~3.0|~4.0",
         "symfony/phpunit-bridge": "~2.7|~3.0|~4.0",
         "twig/twig": "~1.12|~2.0",
-        "satooshi/php-coveralls": "^1.0",
-        "phpunit/phpunit": "^6.1"
+        "satooshi/php-coveralls": "^1.0"
     },
     "suggest": {
         "symfony/web-profiler-bundle": "To use the data collector.",


Ocramius:
`1.5.x` is stable and is perfectly OK to work with symfony `3.4.x`

nicolas-grekas:
Is it maintained? Can we submit 3.4 leveraging features to it? I don't see any 1.5 branch that I could submit against.
Please reopen.

nicolas-grekas:
Calling to other maintainers of the repo, please reopen so that the discussion could at least happen.
PHP 5.5 has been ditched without any community discussion in #668
This move will prevent Symfony 3.4 from being useful to its full user base.

Ocramius:
> Can we submit 3.4 leveraging features to it? I don't see any 1.5 branch that I could submit against.

Only bugfixes.

 > This move will prevent Symfony 3.4 from being useful to its full user base.

Users would just rely on 1.5.x: don't see a problem here.

nicolas-grekas:
> Users would just rely on 1.5.x: don't see a problem here.

this means: users who don't use 7.1, I see the problem.

Ocramius:
There is no problem: want new libraries, get new environments.

If there is a critical bug or a security issue, it will be backported.

If you want shiny new things to run on old or even [unmaintained](https://secure.php.net/supported-versions.php) PHP versions, then pay a dedicated maintainer to manually cherry-pick stuff in your own fork.

Ocramius:
Another addition here: while Symfony offers an LTS, that does not apply to Doctrine. Just use the oldstable version.

nicolas-grekas:
We're just talking about the bundle here.
Here is the problem:
- ppl will start with Symfony 3.4 and PHP 7.0, because that's just allowed and common with Ubuntu Xenial.
- they will want to use Doctrine, so they will install the bundle v1.5, no choice.
- they will get deprecations because the 1.5 bundle is out of sync.
- they will not be able to fix the deprecations without upgrading both the bundle AND their PHP version.

There are two basic problems here:
- many ppl prefer to stick to the PHP version of their distrib. so many will not be able to install 7.1
- forcing TWO changes at once on an upgrade path is calling for troubles.

So, the current policy *for the bundle* is just brutal. It's nothing smooth nor user friendly.
That's a serious issue to me.

Ocramius:
> is just brutal.

Yes it is, and it is intentional.

We will require PHP 7.1 or newer for any new minor and major releases.

All what you listed above has been discussed ad-nauseam hundreds of times, and it is a common problem with distros, not this project.

As I already said, users can simply continue to rely on existing released versions, and will receive security patches anyway.

nicolas-grekas:
There is not a single line of code in the bundle that require 7.1.

Note that I'm not arguing at all about Doctrine - Doctrine is your(+co-maintainers) project and I don't want to interfere with the policy you want to have there.

But here, applying this doctrinal policy on the bundle just break the Symfony experience - and keeps any community contributions away (mines included, I don't like loosing my OSS time like this, even if the bundle would need and benefit from some upgrades.)

Ocramius:
> There is not a single line of code in the bundle that require 7.1.

Well aware of that: it is in first place a maintenance concern, not a compatibility concern. We simply reduced the surface of what we deal with on our side, and allow for patches which include newer syntax and newer tooling.

 > just break the Symfony experience

What "Symfony experience" is broken when using 1.5.x?

If there is something specifically broken, there should be a bug report.

If 1.5.x works, then nothing to be changed here.

nicolas-grekas:
Eg this #695 will never be allowed to <7.1 users. For no technical reason but a composer.json entry.

Ocramius:
#695 is a new feature, so it will only be in newer minor/major releases

nicolas-grekas:
It's a new feature from a libs pov maybe, but it's a bug fix from the php-pm pov (see https://github.com/symfony/symfony/issues/18244 and https://github.com/symfony/symfony/issues/23984).
Here, this is just forcing *your* pov to everyone.
Again on Doctrine no pb with that, an thanks for doing the hard work there.
On this bundle that is real core to a basic Symfony stack, I think this policy is not appropriate because it is not aligned to Symfony's policy.

Ocramius:
> It's a new feature from a libs pov maybe, but it's a bug fix from the php-pm pov (see symfony/symfony#18244 and symfony/symfony#23984).

Thise are both feature requests, since the framework was not designed against that stack in first place. 

 > Here, this is just forcing *your* pov to everyone.

Solution:

selection_086

According to your logic, any feature in symfony would need adjustments and maintenance to be aligned in any bundle that symfony-standard relies upon. That is not realistic and we also do not want that on our side.

 > I think this policy is not appropriate because it is not aligned to Symfony's policy.

Ok, this is where I think we cut the line. If you want the project to be in Symfony, then maintain it on your own terms and under the Symfony organisation. If symfony wants to be pushy, it can be pushy on its own.

Locking here: this project stays on 7.1+.

PR #698, opened 2017-09-18
Fix CI

nicolas-grekas:


.travis.yml
```diff
@@ -33,7 +35,7 @@ install:
   - if [ "$deps" = "low" ]; then composer --prefer-lowest --prefer-stable update; else composer update; fi;
 
 script:
-  - ./vendor/bin/phpunit -v --coverage-clover ./build/logs/clover.xml
+  - ./vendor/bin/simple-phpunit -v --coverage-clover ./build/logs/clover.xml
```
stof: If you use simple-phpunit, I suggest you to enable caching for it:

- define a `$SYMFONY_PHPUNIT_DIR` env variable to put phpunit outside the `vendor` folder
- put this folder in the Travis cache
- (optional) add `./vendor/bin/simple-phpunit install` in the install section so that the install instructions are separate from the test output.

stof:
CI is still broken

nicolas-grekas:
Hum, dunno where this error comes from, no time to look at it, letting for someone else.

PR #708, opened 2017-10-02
Fallback to "twig" service, when "templating" service not available

### LINKED ISSUE #707: Exception, when attempting to EXPLAIN a query from Web Profiler
Using `Product` entity from https://symfony.com/doc/current/doctrine.html documentation and modified `DefaultController::editAction` to:

1. accept `Product $product` parameter
2. have `@Route("/edit-product/{product}")` annotation

Then I'm accessing `/edit-product/2` url and viewing performed DB queries in Web Profiler.

When I click on `Explain query` link I'm getting `An error occurred while loading the query explanation.` message instead of actual EXPLAIN and Developer Console of Browser shows that underlying AJAX request failed with HTTP 500 response.

Looking deeper shows that error was:

```
An exception has been thrown during the rendering of a template ("You have requested a non-existent service "templating". Did you mean one of these: "templating.helper.logout_url", "templating.helper.security"?"). (500 Internal Server Error)
```

The https://github.com/doctrine/DoctrineBundle/blob/master/Controller/ProfilerController.php#L82 code is using `templating` service without checking if it exists upfront. It should be doing check like in  `ControllerTrailt::render` method (see https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/Controller/ControllerTrait.php#L231-L247) instead.

P.S.
I've used Symfony Flex to create a project and templates are rendered correctly with Twig.

aik099:
Fixes exception about non-existing `templating` service usage, when:

1. viewing SELECT query EXPLAIN
2. `symfony/templating` component isn't installed
3. `symfony/twig-bundle` component is installed

Closes #707

### DIFF
--- a/Controller/ProfilerController.php
+++ b/Controller/ProfilerController.php
@@ -79,10 +79,21 @@ class ProfilerController implements ContainerAwareInterface
             return new Response('This query cannot be explained.');
         }
 
-        return $this->container->get('templating')->renderResponse('@Doctrine/Collector/explain.html.twig', array(
+        $template = '@Doctrine/Collector/explain.html.twig';
+        $templateParams = array(
             'data' => $results,
             'query' => $query,
-        ));
+        );
+
+        if ($this->container->has('templating')) {
+            return $this->container->get('templating')->renderResponse($template, $templateParams);
+        }
+
+        if ($this->container->has('twig')) {
+            return new Response($this->container->get('twig')->render($template, $templateParams));
+        }
+
+        throw new \LogicException('You can not use the "explainAction" method if the Templating Component or the Twig Bundle are not available.');
     }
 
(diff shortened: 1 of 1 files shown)


stof:
This is overkill. We can use the twig service directly, as we only need to render a Twig template. See #6810

stof:
I don't need to account for the `twig` service being unavailable, as this controller is used inside the web profiler, and you cannot use the web profiler without twig.

aik099:
@stof , I see that you've sent PR with alternative solution at same time with me. Should I close this PR?

stof:
@aik099 yeah, as I commented before, this approach is overkill. I suggest merging my PR instead which is simpler.

aik099:
OK. I guess we both like to send PRs in parallel. Next time please make note in task if you plan to also send a PR.

PR #709, opened 2017-10-02
Fix profiler query explaining without the templating component

### LINKED ISSUE #707: Exception, when attempting to EXPLAIN a query from Web Profiler

stof:
Fixes #707

Should be backported to 1.6.x too.

### DIFF
--- a/Controller/ProfilerController.php
+++ b/Controller/ProfilerController.php
@@ -79,10 +79,10 @@ class ProfilerController implements ContainerAwareInterface
             return new Response('This query cannot be explained.');
         }
 
-        return $this->container->get('templating')->renderResponse('@Doctrine/Collector/explain.html.twig', array(
+        return new Response($this->container->get('twig')->render('@Doctrine/Collector/explain.html.twig', array(
             'data' => $results,
             'query' => $query,
-        ));
+        )));
     }
 
     private function explainSQLServerPlatform(Connection $connection, $query)


ArrowStrike:
We have the same problem in "doctrine/doctrine-bundle": "^1.6".
And this guy fix it perfectly
https://github.com/ssaki/DoctrineBundle/commit/8e91477358a3814c7b3c979e645adc88014c6ed2
Please, can you add this to Doctrine?

goetas:
Same issue with `doctrine/doctrine-bundle` 1.8.x, the solution by ssaki/DoctrineBundle@8e91477 worked pretty well.

kimhemsoe:
@goetas Can you open a new issue and how to reproduce this. I want to spam this issue more and this sounds like a different issue.

kimhemsoe:
Reported here https://github.com/doctrine/DoctrineBundle/issues/757

PR #710, opened 2017-10-03
Fix deprecation with Symfony 3.4

nicolas-grekas:
Sibling to https://github.com/symfony/symfony/pull/24409
I chose to made as less changes as possible and not create a PSR-11 service locator in a patch release.

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -47,11 +47,6 @@ class DoctrineExtension extends AbstractDoctrineExtension
     private $defaultConnection;
 
     /**
-     * @var array
-     */
-    private $entityManagers;
-
-    /**
      * @var SymfonyBridgeAdapter
      */
     private $adapter;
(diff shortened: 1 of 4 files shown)


nicolas-grekas:
There are 3 references to `service_container` in the bundle, for these services:
- `doctrine.orm.%s_entity_listener_resolver` (`Doctrine\Bundle\DoctrineBundle\Mapping\ContainerAwareEntityListenerResolver`)
- `doctrine.dbal.connection.event_manager` (`Symfony\Bridge\Doctrine\ContainerAwareEventManager`)
- `doctrine` (`Doctrine\Bundle\DoctrineBundle\Registry`)

We could remove them all, but that will need more work, and a bigger patch.
Unless you think otherwise, I'd propose to make that happen on a next minor release.

nicolas-grekas:
(tests fixed, HHVM failure is unrelated)

DependencyInjection/DoctrineExtension.php
```diff
@@ -47,11 +47,6 @@ class DoctrineExtension extends AbstractDoctrineExtension
     private $defaultConnection;
 
     /**
-     * @var array
-     */
-    private $entityManagers;
```
nicolas-grekas: this property is just not used anywhere but locally, thus is dead code

PR #714, opened 2017-10-10
Clear grouped queries on each collect

ossinkine:
Right now the collector keep grouped queries even you create new collector instance and it's wrong.

### DIFF
--- a/DataCollector/DoctrineDataCollector.php
+++ b/DataCollector/DoctrineDataCollector.php
@@ -23,6 +23,7 @@ class DoctrineDataCollector extends BaseCollector
 
     /** @var int|null */
     private $invalidEntityCount;
+    private $groupedQueries;
 
     public function __construct(ManagerRegistry $registry)
     {
@@ -147,6 +148,7 @@ class DoctrineDataCollector extends BaseCollector
         $this->data['entities'] = $entities;
         $this->data['errors']   = $errors;
         $this->data['caches']   = $caches;
+        $this->groupedQueries = null;
     }
 
     public function getEntities()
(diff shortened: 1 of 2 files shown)


mikeSimonson:
@ossinkine The goal of this approach was to avoid duplicating the data that needs to be stored in the sqlite database of the collector.

How did you ended up with 2 instance of the datacollector on one request ?

ossinkine:
@mikeSimonson When running tests. First running test creates a client and dispatches a request. Then second test creates a new client and so on. All clients in all tests will contain the same data in profiler.
Obviously `groupedQueries` should be cleared when `data` is cleared, because `groupedQueries` are created from `data`.
It's not clear for me why data is duplicated in sqlite collector?

mikeSimonson:
From memory, at the end of the request the dataCollector object gets serialized in an sqlite db and for some weird reason the static properties are not serialized but the data it contains can be reconstructed from the rest of the object.

We need to verify that in the code and look at what exactly is going on when the dataCollector gets stored.

ossinkine:
Sounds like a dirty hack 😄

kimhemsoe:
@mikeSimonson Is intended or should behavior get revised?

mikeSimonson:
@kimhemsoe 

The intended behavior is that is should recalculate the groupedQueries every time collect is called. I didn't knew that collect could be called multiple times on the same instance.
The reason for the static variable is because it's static it won't get serialized in sqlite.

I think that this can stay as is.
The data will be duplicated in the sqlite database but it's dev only so I not sure that it's a big deal. 

It still lacks a test.

And it also looks like it would be simpler to find a way to unset that static variable when collect is called.
Or even simply to rebuild the entire array every time the getGroupedQueries is called (perf is not a concern I think).

ossinkine:
There should not be a static variable here, bacause it keeps current object state but not any global state.

mikeSimonson:
It was there because it cached the result of the getGroupedQueries method and while caching that result it still didn't duplicated it in the serialized version of the data. And for me can be removed entirely unless caching it really did provided a real speed benefit.

ossinkine:
If you have issue with serialization maybe would be better to use `__sleep`/`__wakeup` magic methods or implement `Serializable` interface.

mikeSimonson:
It does implements Serializable, and it only deals with ``$this->data``, so the PR as is, is good enough.
@ossinkine Can you add a test that show that wen collect is called a second time the groupedQueries gets updated ?

ossinkine:
@mikeSimonson Please review

mikeSimonson:
@ossinkine Thanks.

Manually merged in https://github.com/doctrine/DoctrineBundle/commit/69700f4ffcfa4fc21087a1adffe944e536420a14 .

PR #715, opened 2017-10-11
Fix Symfony 3.4 auto-registration deprecation

Jean-Beru:
Commands auto-registration is deprecated in Symfony 3.4. This PR will avoid `Auto-registration of the command "Doctrine\Bundle\DoctrineCacheBundle\Command\xxx" is deprecated since Symfony 3.4 and won't be supported in 4.0. Use PSR-4 based service discovery instead.` exceptions.

### DIFF
--- a/DoctrineBundle.php
+++ b/DoctrineBundle.php
@@ -133,4 +133,11 @@ class DoctrineBundle extends Bundle
             }
         }
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function registerCommands(Application $application)
+    {
+    }
 }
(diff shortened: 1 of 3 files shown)


DoctrineBundle.php
```diff
@@ -133,24 +133,4 @@ public function shutdown()
             }
         }
     }
-
-    /**
-     * {@inheritDoc}
-     */
-    public function registerCommands(Application $application)
```
nicolas-grekas: I think this needs to still be implemented, in order to prevent calling the parent.
Otherwise, when the ORM is not loaded, this will still register the related commands, using the legacy way.

stof: hmm, indeed

Jean-Beru: Indeed. Done.

Jean-Beru:
Services have been updated to add a `class` to each command.

Issue #678, opened 2017-07-07
doctrine:database:drop --if-exists --force throws error when database does not exist and using URL parameter


jmsche:
Hi,

Using the `url` parameter, if I launch the following command:
`bin/console d:d:d --if-exists --force`

I'll get the following error:
`An exception occured in driver: SQLSTATE[HY000] [1049] Unknown database 'tasks_test'`

It's due to the call to `parseDatabaseUrl` done here: https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/DriverManager.php#L144

As the `url` param is defined, it will refill the `dbname` param, which is not what is wanted here.

Nijusan:
PR for this: https://github.com/doctrine/DoctrineBundle/pull/767

weaverryan:
This should be fixed, and the fix is included in the latest release. If you're still having issues, please let us know!

Cheers!

Thomas-Negrault:
With `1.10.2` I still have an issue executing database drop with `--if-exists`
config:
```
#config/packages/test/doctrine.yaml
doctrine:
    dbal:
        driver: pdo_sqlite
        url: "sqlite:///%kernel.cache_dir%/behat.db"
```

```
bin/console d:d:d --if-exists --env=test --force -vvv

In DBALException.php line 32:
                                                                                                           
  [Doctrine\DBAL\DBALException]                                                                            
  Operation 'Doctrine\DBAL\Platforms\AbstractPlatform::getListDatabasesSQL' is not supported by platform.  
                                                                                                           

Exception trace:
 () at /var/www/vendor/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php:32
 Doctrine\DBAL\DBALException::notSupported() at /var/www/vendor/doctrine/dbal/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php:2727
 Doctrine\DBAL\Platforms\AbstractPlatform->getListDatabasesSQL() at /var/www/vendor/doctrine/dbal/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php:97
 Doctrine\DBAL\Schema\AbstractSchemaManager->listDatabases() at /var/www/vendor/doctrine/doctrine-bundle/Command/DropDatabaseDoctrineCommand.php:105
 Doctrine\Bundle\DoctrineBundle\Command\DropDatabaseDoctrineCommand->execute() at /var/www/vendor/symfony/console/Command/Command.php:255
[...]

PR #717, opened 2017-10-17
Throw error in drop command if dbname is refilled

enumag:
Fix for #678.

Should I add a separate return code?

Also this could be a BC break for databases that do not require the database to not be opened. Not sure what to do about that. Maybe not throw the error here, try to run it but add a message to the later fail?

The current behavior is a real WTF on postgres - it throws `cannot drop the currently open database` because the code that should prevent it is not working correctly as described in #678.

### DIFF
--- a/Command/DropDatabaseDoctrineCommand.php
+++ b/Command/DropDatabaseDoctrineCommand.php
@@ -99,6 +99,13 @@ EOT
             // as some vendors do not allow dropping the database connected to.
             $connection->close();
             $connection = DriverManager::getConnection($params);
+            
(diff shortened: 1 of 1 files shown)


mikeSimonson:
Can you add a test for this ?

enumag:
No since there are no tests for this command at the moment.

kimhemsoe:
@enumag We need a test covering the changes even if there is missing tests :) Only way we can ensure that your change wont break later.

enumag:
@kimhemsoe In that case you'll have to find someone to write them because it won't be me.

kimhemsoe:
Closing then.

PR #719, opened 2017-10-20
Adding a new option to pull your repositories from the service container

weaverryan:
Hi guys!

tl;dr; This PR will allow users to create *true* repository services  so they can be autowired! Replaces #658.

I would like to stop recommending that people use traditional Doctrine repositories, and instead create normal services for their repository classes: https://blog.fervo.se/blog/2017/07/06/doctrine-repositories-autowiring/

This pull requests makes this much easier by adding:

A) A trait & abstract class to make custom repository classes easy
B) A custom entity repository factory that returns your repository services, instead of creating new `EntityRepository` like normal.

## Example

First, the feature needs to be enabled:

```yml
doctrine:
    orm:
        use_service_repositories: true
```

Suppose we have a `BlogPost` entity and we want a custom repository.

```php
/**
 * @ORM\Entity(repositoryClass="App\Repository\BlogPostRepository")
 */
class BlogPost
{

}
```

Because the feature is enabled, the `repositoryClass` is now actually your repository's service id (which is usually your class name anyways). It should look like this:

```php
class BlogRepository extends AbstractServiceRepository
{
    private $registry;

    {
        $this->registry = $registry;
    }

    protected function getEntityRepository()
    {
        return $this->registry->getRepository(BlogPost::class);
    }
}
```

The `AbstractServiceRepository` has just one abstract method (`getEntityRepository`) and gives you *all* of the same public functions from `EntityRepository`.

To use the repository, you can now auto-wire the service, or fetch it from the entity manager like normal:

```php
public function indexAction()
{
    $posts = $this->getDoctrine()->getRepository(BlogPost::class)->findAll();
}
```

If you don't want a custom repository, that's fine - we will give you a `DefaultServiceRepository` object, which implements `ObjectRepository` and has all the same methods as `EntityRepository`.

Cheers!

### DIFF
--- a/DependencyInjection/Compiler/ServiceRepositoryCompilerPass.php
+++ b/DependencyInjection/Compiler/ServiceRepositoryCompilerPass.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of the Doctrine Bundle
+ *
+ * The code was originally distributed inside the Symfony framework.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ * (c) Doctrine Project, Benjamin Eberlei <kontakt@beberlei.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * @author Ryan Weaver <ryan@knpuniversity.com>
+ */
+class ServiceRepositoryCompilerPass implements CompilerPassInterface
+{
+    const REPOSITORY_SERVICE_TAG = 'doctrine.repository_service';
+
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->hasDefinition('doctrine.orm.container_repository_factory')) {
+            return;
+        }
+
+        $locatorDef = $container->findDefinition($container->getDefinition('doctrine.orm.container_repository_factory')->getArgument(0));
+        $repoServiceIds = array_keys($container->findTaggedServiceIds('doctrine.repository_service'));
+        $repoReferences = array_map(function($id) {
+            return new Reference($id);
+        }, $repoServiceIds);
+        $locatorDef->replaceArgument(0, array_combine($repoServiceIds, $repoReferences));
+    }
+}
--- a/DependencyInjection/Compiler/ServiceRepositoryCompilerPass.php
+++ b/DependencyInjection/Compiler/ServiceRepositoryCompilerPass.php
@@ -15,6 +15,7 @@
 namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
 
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\Compiler\ServiceLocatorTagPass;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Reference;
 
@@ -31,11 +32,13 @@ class ServiceRepositoryCompilerPass implements CompilerPassInterface
             return;
         }
 
-        $locatorDef = $container->findDefinition($container->getDefinition('doctrine.orm.container_repository_factory')->getArgument(0));
+        $locatorDef = $container->getDefinition('doctrine.orm.container_repository_factory');
         $repoServiceIds = array_keys($container->findTaggedServiceIds(self::REPOSITORY_SERVICE_TAG));
         $repoReferences = array_map(function($id) {
             return new Reference($id);
         }, $repoServiceIds);
-        $locatorDef->replaceArgument(0, array_combine($repoServiceIds, $repoReferences));
+
+        $ref = ServiceLocatorTagPass::register($container, array_combine($repoServiceIds, $repoReferences));
+        $locatorDef->replaceArgument(0, $ref);
     }
 }
--- a/DependencyInjection/Compiler/ServiceRepositoryCompilerPass.php
+++ b/DependencyInjection/Compiler/ServiceRepositoryCompilerPass.php
@@ -32,7 +32,7 @@ class ServiceRepositoryCompilerPass implements CompilerPassInterface
         }
 
         $locatorDef = $container->findDefinition($container->getDefinition('doctrine.orm.container_repository_factory')->getArgument(0));
-        $repoServiceIds = array_keys($container->findTaggedServiceIds('doctrine.repository_service'));
+        $repoServiceIds = array_keys($container->findTaggedServiceIds(self::REPOSITORY_SERVICE_TAG));
         $repoReferences = array_map(function($id) {
             return new Reference($id);
         }, $repoServiceIds);
--- a/DependencyInjection/Compiler/ServiceRepositoryCompilerPass.php
+++ b/DependencyInjection/Compiler/ServiceRepositoryCompilerPass.php
@@ -33,6 +33,12 @@ class ServiceRepositoryCompilerPass implements CompilerPassInterface
         }
 
         $locatorDef = $container->getDefinition('doctrine.orm.container_repository_factory');
+
+        if (!class_exists(ServiceLocatorTagPass::class)) {
+            // 3.2 and lower compatibility
+            $locatorDef->replaceArgument(0, new Reference('service_container'));
+        }
+
         $repoServiceIds = array_keys($container->findTaggedServiceIds(self::REPOSITORY_SERVICE_TAG));
         $repoReferences = array_map(function($id) {
             return new Reference($id);
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -487,7 +487,7 @@ class Configuration implements ConfigurationInterface
                     ->scalarNode('quote_strategy')->defaultValue('doctrine.orm.quote_strategy.default')->end()
                     ->scalarNode('entity_listener_resolver')->defaultNull()->end()
                     ->scalarNode('repository_factory')->defaultNull()->end()
-                    ->booleanNode('use_service_repositories')->defaultValue(false)->end()
+                    ->booleanNode('use_service_repositories')->defaultFalse()->end()
                 ->end()
                 ->children()
                     ->arrayNode('second_level_cache')
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -487,6 +487,7 @@ class Configuration implements ConfigurationInterface
                     ->scalarNode('quote_strategy')->defaultValue('doctrine.orm.quote_strategy.default')->end()
                     ->scalarNode('entity_listener_resolver')->defaultNull()->end()
                     ->scalarNode('repository_factory')->defaultNull()->end()
+                    ->booleanNode('use_service_repositories')->defaultValue(false)->end()
                 ->end()
                 ->children()
                     ->arrayNode('second_level_cache')
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -17,6 +17,7 @@ namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection;
 use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\ServiceRepositoryCompilerPass;
 use Doctrine\Bundle\DoctrineBundle\Repository\EntityRepositoryInterface;
 use Doctrine\ORM\Version;
+use Symfony\Component\DependencyInjection\Compiler\ServiceLocatorTagPass;
 use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
 use Symfony\Component\DependencyInjection\Alias;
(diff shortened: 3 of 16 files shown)


Majkl578:
I'm personally strongly against this addition, it extends service locator antipattern by yet another indirect lock-in dependency. Service locator should be phased out, not bloated even further.

weaverryan:
@Majkl578 When you say `service locator antipattern`, are you referring to the repository system in the EntityManager, or Symfony's container? 

The goal of this PR is to encourage people to create real services and use proper DI. So, that's a big win! :) The purpose of *also* allowing these services to be fetched via the norml `$em->getRepository()` is because (A) it's familiar to people and (B) users will still be able to fetch a repository and use the basic methods (e.g. `find()`) without needing to create a custom service for every entity. But in practice, as soon as they have a real service, we would recommend they just DI it (which will be even easier than using `$em->getRepository()` anyways, thanks autowiring).

Cheers!

Ocramius:
s/`AbstractServiceRepository`/`RepositoryLocator`

weaverryan:
@Ocramius Do you want me to rename `AbstractServiceRepository` to `RepositoryLocator`? Can you explain that a bit more? I'm happy to rename it to the best choice. In my mind, I was thinking that the user is creating a "repository that is a service"... hence having them extend this optional `AbstractServiceRepository` class.

DependencyInjection/Configuration.php
```diff
@@ -487,6 +487,7 @@ private function getOrmEntityManagersNode()
                     ->scalarNode('quote_strategy')->defaultValue('doctrine.orm.quote_strategy.default')->end()
                     ->scalarNode('entity_listener_resolver')->defaultNull()->end()
                     ->scalarNode('repository_factory')->defaultNull()->end()
+                    ->booleanNode('use_service_repositories')->defaultValue(false)->end()
```
stof: use `defaultFalse()` for consistency

Ocramius:
Oh, I misunderstood the feature then! The abstract looked completely different to me.

I don't think we should add this inheritance level on top of what we already have though...

DependencyInjection/Compiler/ServiceRepositoryCompilerPass.php
```diff
@@ -0,0 +1,41 @@
...
+    const REPOSITORY_SERVICE_TAG = 'doctrine.repository_service';
+
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->hasDefinition('doctrine.orm.container_repository_factory')) {
+            return;
+        }
+
+        $locatorDef = $container->findDefinition($container->getDefinition('doctrine.orm.container_repository_factory')->getArgument(0));
+        $repoServiceIds = array_keys($container->findTaggedServiceIds('doctrine.repository_service'));
```
stof: why not using the constant ?

Resources/config/orm.xml
```diff
@@ -102,6 +102,17 @@
             <factory class="%doctrine.orm.entity_manager.class%" method="create" />
         </service>
 
+        <service id="doctrine.orm.container_repository_factory" class="Doctrine\Bundle\DoctrineBundle\Repository\ContainerRepositoryFactory" public="false">
+            <argument type="service">
+                <service class="Symfony\Component\DependencyInjection\ServiceLocator">
```
stof: won't work before Symfony DI 3.3. Min version is 2.7

Resources/config/orm.xml
```diff
@@ -102,6 +102,17 @@
...
+                    <tag name="container.service_locator" />
+                    <!-- services specified in compiler pass -->
```
stof: A better idea when using 3.3+ and creating a collector in a compiler pass is to use `\Symfony\Component\DependencyInjection\Compiler\ServiceLocatorTagPass::register`, which deduplicates service locators

Repository/ContainerRepositoryFactory.php
```diff
@@ -0,0 +1,64 @@
...
+    public $container;
+
+    private $registry;
+
+    private $genericRepositories = array();
+
+    public function __construct(ContainerInterface $container, RegistryInterface $registry)
+    {
+        $this->container = $container;
+        $this->registry = $registry;
```
stof: the registry looks unused

Repository/DefaultServiceRepository.php
```diff
@@ -0,0 +1,46 @@
...
+        $this->entityManager = $entityManager;
+        $this->className = $className;
+    }
+
+    /**
+     * @return EntityRepository
+     */
+    private function getEntityRepository()
+    {
+        return $this->entityManager->getRepository($this->getClassName());
```
lyrixx: are you sure it's `$this->getClassName()` and not `$this->className`. 
If you are sure, it mean the className property is useless (private + never used)

Resources/config/orm.xml
lyrixx: And may add an id on this service to avoid:
```php
        $locatorDef = $container->findDefinition($container->getDefinition('doctrine.orm.container_repository_factory')->getArgument(0));
```
It will be easier to understand

DependencyInjection/DoctrineExtension.php
```diff
@@ -438,8 +444,27 @@ protected function loadOrmEntityManager(array $entityManager, ContainerBuilder $
...
-            $methods['setRepositoryFactory'] = new Reference($entityManager['repository_factory']);
+        if ($entityManager['use_service_repositories']) {
+            if ($entityManager['repository_factory']) {
+                throw new InvalidArgumentException('The "repository_factory" option cannot be set when "use_service_repositories" is set to true.');
+            }
+
+            $serviceRepoFactoryDef = $container->getDefinition('doctrine.orm.container_repository_factory');
+            if (class_exists(ServiceLocatorTagPass::class)) {
+
+                $serviceRepoFactoryDef->replaceArgument(0, new Reference('service_container'));
```
lyrixx: This looks wrong. it should be a service locator.

Anyway I think it will be easier to deal with the compat in the compiler pass

weaverryan:
@Ocramius 

> I don't think we should add this inheritance level on top of what we already have though...

I'm trying to make it so that users can stop needing to extend the `EntityRepository` class and instead create a pure service where they control the `__construct()` method.  The `AbstractServiceRepository` is actually optional: users can just create a pure service & not extend anything. But it's needed for users that want to get the nice `find**` methods without doing any work. It's a way to get the "niceness" of `EntityRepository`, but give users back their `__construct()` method. I can't think of another way to accomplish this. There *is* also a trait that users can use, but the abstract class makes it clear the method they need to implement.

Tobion:
As this currently targets master which requires php 7.1, it should IMO add type declarations everywhere.

Tests/ServiceRepositoryTest.php
```diff
@@ -0,0 +1,83 @@
...
+                )),
+                'use_service_repositories' => true,
+            ),
+        )), $container);
+
+        $container->register(TestCustomRepoRepository::class)
+            ->addTag('doctrine.repository_service');
+
+        $container->getDefinition('doctrine.orm.container_repository_factory')
+            ->setPublic(true);
```
Tobion: why does it need to be made public?

weaverryan: Good catch! I'll remove that

Repository/EntityRepositoryInterface.php
```diff
@@ -0,0 +1,87 @@
...
+use Doctrine\ORM\QueryBuilder;
+
+/**
+ * An interface that all entity repositories should implement.
+ *
+ * This contains all of the public methods found in Doctrine\ORM\EntityRepository
+ *
+ * @author Ryan Weaver <ryan@knpuniversity.com>
+ */
+interface EntityRepositoryInterface extends ObjectRepository
```
Tobion: I think this should be moved to Doctrine\ORM in the end. And let \Doctrine\ORM\EntityRepository implement it.

Repository/DefaultServiceRepository.php
weaverryan: Great catch. Yes, there is a problem here - adding a test and fixing now!

Repository/EntityRepositoryInterface.php
weaverryan: I agree in principle. But I'm not sure when Doctrine 2.6 will be released, or if this could be put into that in time. I'd like to make this feature the default recommended way of doing things in Symfony 4 (and even update the generator to create these new service repositories to bump people in this direction).

There is a similar issue with the `RepositoryTrait` - it has copy-and-paste from `EntityRepository`... which in a more perfect world, could be replaced by a trait that lives in `Doctrine\ORM`. Of course, this could easily be done later without BC breaks.

So... in my perfect world, we would merge this PR, but also make these changes (add interface + the trait) in `Doctrine\ORM` and then use them later.

DependencyInjection/DoctrineExtension.php
weaverryan: Good call - I updated to be in the pass. to keep 2.7 support, the class itself accepts a locator or Symfony's `ContainerInterface` for older 3.2 and lower.

Tobion:
I really like this approach. Well done @weaverryan !

If the em is closed I only need to `$this->registry->resetManager()` and the service repo works again.

I wonder if we should try to automatically enable this behavior without the `use_service_repositories` config. We could just check if the configured repo exists as a service and if it implements the new `EntityRepositoryInterface`. Then  use it. If not, fall back to the standard `\Doctrine\ORM\Repository\DefaultRepositoryFactory`.
The only problem would be that we can't automatically use `DefaultServiceRepository` when no repo has been configured. But speaking of DefaultServiceRepository, it does not seem to support resetting the entity manger because it injects the entity manager directly instead of the registry. So it does not have any advantage over the default EntityManager from doctrine, does it?

Tobion:
While this approach is a good solution to using service repositories with inheritance. In theory I must agree with https://www.tomasvotruba.cz/blog/2017/10/16/how-to-use-repository-with-doctrine-as-service-in-symfony/ that using repositories with composition is cleaner and solves even more problems like type declarations.

lyrixx:
Just a side question: How did you build the list of method declared in `EntityRepositoryInterface` ? (and why)

Repository/AbstractServiceRepository.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+namespace Doctrine\Bundle\DoctrineBundle\Repository;
+
+/**
+ * Abstract helper class for creating custom service repository classes.
+ *
+ * @author Ryan Weaver <ryan@knpuniversity.com>
+ */
+abstract class AbstractServiceRepository implements EntityRepositoryInterface
```
Majkl578: This class is useless since there's both interface + trait and no content, it only abuses inheritance.

Repository/ContainerRepositoryFactory.php
```diff
@@ -0,0 +1,77 @@
...
+        if (null !== $repositoryServiceId) {
+            if (!$this->container->has($repositoryServiceId)) {
+                throw new \RuntimeException(sprintf('Could not find the repository service for the "%s" entity. Make sure the "%s" service exists and is tagged with "%s"', $entityName, $repositoryServiceId, ServiceRepositoryCompilerPass::REPOSITORY_SERVICE_TAG));
+            }
+
+            return $this->container->get($repositoryServiceId);
+        }
+
+        $repositoryHash = $entityClass . spl_object_hash($entityManager);
+        if (!isset($this->genericRepositories[$repositoryHash])) {
```
Majkl578: If my assumption is correct, this will leak (old now-invalid repositories being kept in memory) when EM is resetted.

Repository/EntityRepositoryInterface.php
Majkl578: This is pretty serious decision to introduce such interface in an integration bundle, which is IMHO not desired. It will become a huge pain due to BC when Doctrine 3 is released and/or when EntityRepositoryInterface is added to Doctrine itself.
I don't think it should exist here at all.

Repository/RepositoryTrait.php
```diff
@@ -0,0 +1,258 @@
...
+use Doctrine\ORM\Query;
+use Doctrine\ORM\Query\ResultSetMappingBuilder;
+use Doctrine\ORM\QueryBuilder;
+
+/**
+ * A helpful trait when creating your own repository.
+ *
+ * @author Ryan Weaver <ryan@knpuniversity.com>
+ */
+trait RepositoryTrait
```
Majkl578: By basically copying whole EntityRepository into this trait, you're now moving responsibility of bugfixes and backporting features to this bundle and unDRY the whole concept.

Repository/DefaultServiceRepository.php
```diff
@@ -0,0 +1,54 @@
...
+
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\EntityRepository;
+
+/**
+ * Default repository that's used for service repositories.
+ *
+ * @author Ryan Weaver <ryan@knpuniversity.com>
+ */
+class DefaultServiceRepository implements EntityRepositoryInterface
```
Majkl578: Should be final since it's not to be extended any further.

Repository/EntityRepositoryInterface.php
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * Select all elements from a selectable that match the expression and
+     * return a new collection containing these elements.
+     *
+     * @param \Doctrine\Common\Collections\Criteria $criteria
+     *
+     * @return \Doctrine\Common\Collections\Collection
+     */
+    public function matching(Criteria $criteria);
```
Majkl578: This is not correct, it's part of separate interface, not responsibility of Repository.

weaverryan:
See #725 for an alternative. Making this work with multiple EM's, etc is doable, but #725 is much easier, and I think a great solution.

PR #725, opened 2017-10-24
New Trait to make service repositories very easy

weaverryan:
Hi guys!

This is a much simpler approach that replaces #719. The big difference between this and #719 is that the behavior of `$em->getRepository()` is unchanged: it still returns the normal `EntityRepository`. But with the new`RepositoryTrait`, creating a new service that decorates your repository is very easy. And the PR (as you can see) is about 50 times simpler :).

Usage:

* You are still allowed to map a `repositoryClass` on your entity, but generally, you will not (create a service instead).

* Create a repository service:

```php
<?php

namespace App\Repository;

use Doctrine\Bundle\DoctrineBundle\Repository\RepositoryTrait;
use Symfony\Bridge\Doctrine\RegistryInterface;
use App\Entity\CoolStuff;

class CoolStuffRepository
{
    use RepositoryTrait;

    private $registry;

    {
        $this->registry = $registry;
    }

    // 2 abstract methods to fuel the trait
    protected function getEntityManager()
    {
        return $this->registry->getManager();
    }
    protected function getClassName()
    {
        return CoolStuff::class;
    }

    // add custom methods here
}
```

The `RepositoryTrait` gives you all the existing methods on `EntityRepository`, like `createQueryBuilder()` and `find**` (but no magic methods). If the user wants to make their class implement the `ObjectRepository` interface, that's totally up to them. You can also ignore our trait and create your services all on your own.

To use it, just DI your class like normal. 

For entities with NO custom repository, users can still use `$em->getRepository()`. But when they need custom methods, they would create a custom service and start using it.

Thanks!

### DIFF
--- a/Repository/RepositoryTrait.php
+++ b/Repository/RepositoryTrait.php
@@ -1,0 +1,174 @@
+<?php
+
+/*
+ * This file is part of the Doctrine Bundle
+ *
+ * The code was originally distributed inside the Symfony framework.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ * (c) Doctrine Project, Benjamin Eberlei <kontakt@beberlei.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Doctrine\Bundle\DoctrineBundle\Repository;
+
+use Doctrine\Common\Collections\Criteria;
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\EntityRepository;
+use Doctrine\ORM\NativeQuery;
+use Doctrine\ORM\Query;
+use Doctrine\ORM\Query\ResultSetMappingBuilder;
+use Doctrine\ORM\QueryBuilder;
+
+/**
+ * A helpful trait when creating your own repository.
+ *
+ * @author Ryan Weaver <ryan@knpuniversity.com>
+ */
+trait RepositoryTrait
+{
+    /**
+     * @return EntityManagerInterface
+     */
+    abstract protected function getEntityManager();
+
+    /**
+     * @return string The class name for your entity.
+     */
+    abstract protected function getClassName();
+
+    /**
+     * @see EntityRepository::createQueryBuilder()
+     *
+     * @param string $alias
+     * @param string $indexBy The index for the from.
+     *
+     * @return QueryBuilder
+     */
+    public function createQueryBuilder($alias, $indexBy = null)
+    {
+        return $this->getRepository()->createQueryBuilder($alias, $indexBy);
+    }
+
+    /**
+     * @see EntityRepository::createResultSetMappingBuilder()
+     *
+     * @param string $alias
+     *
+     * @return ResultSetMappingBuilder
+     */
+    public function createResultSetMappingBuilder($alias)
+    {
+        return $this->getRepository()->createResultSetMappingBuilder($alias);
+    }
+
+    /**
+     * @see EntityRepository::createNamedQuery()
+     *
+     * @param string $queryName
+     *
+     * @return Query
+     */
+    public function createNamedQuery($queryName)
+    {
+        return $this->getRepository()->createNamedQuery($queryName);
+    }
+
+    /**
+     * @see EntityRepository::createNativeNamedQuery()
+     *
+     * @param string $queryName
+     *
+     * @return NativeQuery
+     */
+    public function createNativeNamedQuery($queryName)
+    {
+        return $this->getRepository()->createNativeNamedQuery($queryName);
+    }
+
+    /**
+     * @see EntityRepository::clear()
+     *
+     * @return void
+     */
+    public function clear()
+    {
+        $this->getRepository()->clear();
+    }
+
+    /**
+     * @see EntityRepository::find()
+     *
+     * @param mixed    $id          The identifier.
+     * @param int|null $lockMode    One of the \Doctrine\DBAL\LockMode::* constants
+     *                              or NULL if no specific lock mode should be used
+     *                              during the search.
+     * @param int|null $lockVersion The lock version.
+     *
+     * @return object|null The entity instance or NULL if the entity can not be found.
+     */
+    public function find($id, $lockMode = null, $lockVersion = null)
+    {
+        return $this->getRepository()->find($id, $lockMode, $lockVersion);
+    }
+
+    /**
+     * @see EntityRepository::findAll()
+     *
+     * @return array The entities.
+     */
+    public function findAll()
+    {
+        return $this->getRepository()->findAll();
+    }
+
+    /**
+     * @see EntityRepository::findBy()
+     *
+     * @param array      $criteria
+     * @param array|null $orderBy
+     * @param int|null   $limit
+     * @param int|null   $offset
+     *
+     * @return array The objects.
+     */
+    public function findBy(array $criteria, array $orderBy = null, $limit = null, $offset = null)
+    {
+        return $this->getRepository()->findBy($criteria, $orderBy, $limit, $offset);
+    }
+
+    /**
+     * @see EntityRepository::findOneBy()
+     *
+     * @param array $criteria
+     * @param array|null $orderBy
+     *
+     * @return object|null The entity instance or NULL if the entity can not be found.
+     */
+    public function findOneBy(array $criteria, array $orderBy = null)
+    {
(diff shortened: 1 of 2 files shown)


Tests/Repository/RepositoryTraitTest.php
```diff
@@ -0,0 +1,65 @@
...
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\EntityRepository;
+use Doctrine\ORM\QueryBuilder;
+use PHPUnit\Framework\TestCase;
+
+class RepositoryTraitTest extends TestCase
+{
+    public function testBasicTraitFunctionality()
+    {
+        $em = $this->getMockBuilder(EntityManagerInterface::class)->getMock();
```
Tobion: `createMock` everywhere

Repository/RepositoryTrait.php
```diff
@@ -0,0 +1,174 @@
...
+ * A helpful trait when creating your own repository.
+ *
+ * @author Ryan Weaver <ryan@knpuniversity.com>
+ */
+trait RepositoryTrait
+{
+    /**
+     * @return EntityManagerInterface
+     */
+    abstract protected function getEntityManager();
```
Tobion: type declarations pls

nicolas-grekas:
Shouldn't we do this in the bridge, in order to have in work on PHP 5.5, thus all SF 3.4 apps?

Tobion:
@nicolas-grekas it would only be part of sf 4.1 which does not support php 5.5

fabpot:
I agree with @Tobion in principle. But it looks like we need to workaround artificial limitations introduced in this bundle (having a different PHP requirement than Symfony itself). So, I'm +1 to move this to the bridge for 3.4. Or, we revert all changes related to the bump from PHP 5.5 to 7.1 in the bundle (it looks like relatively easy to do). Whatever makes this change available for our users soon.

beberlei:
What about a solution based on Doctrine ORMs own `RepositoryFactory`? Framework integration is what it was made for: https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Repository/RepositoryFactory.php

This solution here only makes the creatoin of your own repository simple, but it does not facilitate creating services for it automatically. Also it requires to "not use" the `EntityManager#getRepository` which can be easy to forget.

```
class SymfonyRepositoryFactory implements RepositoryFactory
{
    private $managedRepositories = [];

    public function getRepository(EntityManagerInterface $entityManager, $entityName)
    {
        $metadata = $entityManager->getClassMetadata($entityName);
        $repositoryHash = $metadata->getName() . spl_object_hash($entityManager);

        if ($this->container->has($metadata->customRepositoryClassName)) {
            $repository = $this->container->get($metadata->customRepositoryClassName);

            return $repository;
        } else if (isset($this->managedRepositories[$repositoryHash])) {
            return $this->managedRepositories[$repositoryHash];
        }

        $repositoryClassName = $metadata->customRepositoryClassName
            ?: $entityManager->getConfiguration()->getDefaultRepositoryClassName();

        return $this->managedRepositories[$repositoryHash] = new $repositoryClassName($entityManager, $metadata);
    }
}
```

beberlei:
I am not up to date on autowiring and how this all works in Symfony 3 (still doing the "old ways"), so correct me if i am wrong on basics of the new approach.

Is the goal to automatically register repositories that are defined in `@Entity(repositoryClass)` as a service in the container? Or is the goal for custom repositories to be created by the Symfony container?

For the second goal, my factory in the comment above would allow developers to call `EntityManager#getRepository()` and still get a symfony controlled repository back, or fetch it by same service name and get the same instance, so it avoids the problem of having 2 repositories (Symfony and Doctrine layer) and remembering which to use. It might be a good complement to @weaverryan PR here.

For the first goal, which was also part of the previous pull request, there still would need to be DIC related code.

Also why use a trait? The EntityRepository is perfectly fine as a base class, even when using `doctrine` as a service:

```
class CoolStuffRepository extends EntityRepository
{
    private $registry;

    {
        $this->registry = $registry;
        $manager = $registry->getManagerForClass(CoolStuff::class);
        parent::__construct($manager, $manager->getClassMetadata(CoolStuff::class));
    }
}
```
is less code than the trait based solution, but requires a bit of "magic foo" in the constructor which is not necessarily easy for developers. So we can simplify by shipping a layer for that:

```
class CoolStuffRepository extends EntityRegistryRepository
{
    public function __construct(RegistryInterface $registry, /** and other services I need here */)
    {
        parent::__construct($registry);
    }
}

class EntityRegistryRepository extends EntityRepository
{
    protected $registry;

    {
        $this->registry = $registry;
        $manager = $registry->getManagerForClass(CoolStuff::class);
        parent::__construct($manager, $manager->getClassMetadata(CoolStuff::class));
    }
}
```

beberlei:
@weaverryan your repository in the PR description is too simple, and you would need to document and explain getManagerForClass() somewhere even if you add this simple approach to the docs and discuss context for using one or the other. 

Hiding the relevant registry code away like in my last class `EntityRegistryRepository` will also allow entities to change the EntityManager without leading to bugs here using the default one, thus automatically making the right decision instead of the easy one.

Tobion:
> What about a solution based on Doctrine ORMs own RepositoryFactory? 

That was what was first tried in https://github.com/doctrine/DoctrineBundle/pull/719
But using repos using composition instead of inheritance is cleaner and more flexible. See https://www.tomasvotruba.cz/blog/2017/10/16/how-to-use-repository-with-doctrine-as-service-in-symfony/

This means `@Entity(repositoryClass)` will not use the service and is not being used when you want repos as service. 
The trait is just a helper to expose the common EntityRepository methods in your own repos if you want that. And it's most likely getting merged in the doctrine bridge isntead: https://github.com/symfony/symfony/pull/24693

Tobion:
@beberlei Your EntityRegistryRepository solution seems interesting as well and works with multiple EMs.

weaverryan:
@beberlei I have a quick question: can an entity only be managed by exactly *one* EntityManager? I was assuming that there was some edge case where it could be in multiple, and was designing #719 (with great complexity) around this idea.

I also think your `EntityRegistryRepository` idea is interesting. But we also have another problem: this bundle was bumped to PHP 7.1 - it's way ahead of Symfony's PHP requirements (3.4 will still allow 5.5.9). I don't know if this can be reversed (I know the tests fail on 5.5.9, but due to `createMock()` - not sure what (if anything) was changed in the code for 7.1).

weaverryan:
@beberlei Also, are there any issues with `getManagerForClass()`? I mean, it loops over all of the EM's to find the correct one. Is there a possible issue with instantiating (thanks to this loop) other EM's to find the correct one?

weaverryan:
Sorry, one more minor issue with this approach (which I'm mentioning to get your thoughts): the `EntityManager::getRepository()` method advertises that it returns an `EntityRepository`. In order for people to fulfill this in their code, their services MUST extend `EntityRepository` (so, your example above of extending it is mandatory). There is actually no interface (e.g. `EntityRepositoryInterface`) that can be used here - `ObjectRepository` only has *some* of the methods. If users fail to extend `EntityRepository`, anything that type-hints `EntityRepository` could fail.  I suppose we could make the new repository factory throw an exception if your service does not extend this class. But, it is a bit unfortunate that it's being forced. The approach in symfony/symfony#24693 has the advantage that the services can be pure - you can choose whatever methods you want (it has the disadvantage that you sort of have 2 different repositories ideas now)

weaverryan:
Btw, it may not matter, but I've just proposed re-adding PHP 5.5 support to this bundle, which was trivial: https://github.com/doctrine/DoctrineBundle/pull/726

weaverryan:
The more I think about the Benjamin’s approach (which is similar to what I tried, but has some simplicities I missed), the more I like it.

But I just thought of one other issue. A long time ago, I noticed injecting repositories (that were services) caused some performance issues, because it takes a bit of time to instantiate them (probably the class metadata stuff - a penalty you pay, even if you do not use the repository during the request). So, in a perfect world, the constructor would be lazy, and we would only load the class metadata when we need to. But EntityRepository stores that on a protected property... which means we cannot do something clever and load it only on demand. That seems like an old flaw, and I’m not sure what to do about it. Thoughts?

kimhemsoe:
I think top that the solution for this problem belongs in the bridge. 

I will let the PR stay open for now so the discussion wont get lost.

weaverryan:
Closing now in favor of #727. Basically, we should merge #727 or symfony/symfony#24693. These are the 2 competing approaches.

Thanks!

PR #705, opened 2017-09-29
Backport some critical issues/changes to the 1.6 branch

fabpot:
I've just created a 1.6 branch where I can backport critical changes. That would partially solve the issue mentioned in #704.

PR #726, opened 2017-10-26
Reverts the bump from PHP 5 to PHP 7.1: re-adds php 5.5 support

weaverryan:
Hi guys!

This bundle was bumped to PHP 7.1 only, but the next version of Symfony (3.4) will still support 5.5. This inconsistency makes the latest version of this bundle not usable by some users (including Ubuntu LTS). This re-adds PHP 5 support without breaking BC.

I believe Fabien (https://github.com/doctrine/DoctrineBundle/pull/725#issuecomment-339421794) is also in favor of this - I don't think it makes sense to bump the version yet, especially when this bundle really isn't using any PHP 7 features.

Thanks!

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -808,7 +808,7 @@ class DoctrineExtension extends AbstractDoctrineExtension
     /**
      * @return string
      */
-    private function getDefinitionClassname(): string
+    private function getDefinitionClassname()
     {
         return class_exists(ChildDefinition::class) ? ChildDefinition::class : DefinitionDecorator::class;
     }
--- a/Tests/DataCollector/DoctrineDataCollectorTest.php
+++ b/Tests/DataCollector/DoctrineDataCollectorTest.php
@@ -28,7 +28,7 @@ class DoctrineDataCollectorTest extends TestCase
     public function testCollectEntities()
     {
         $manager = $this->getMockBuilder('Doctrine\ORM\EntityManager')->disableOriginalConstructor()->getMock();
-        $config = $this->createMock('Doctrine\ORM\Configuration');
+        $config = $this->getMockBuilder('Doctrine\ORM\Configuration')->getMock();
         $factory = $this->getMockBuilder('Doctrine\Common\Persistence\Mapping\AbstractClassMetadataFactory')
             ->setMethods(array('getLoadedMetadata'))->getMockForAbstractClass();
         $collector = $this->createCollector(array('default' => $manager));
@@ -83,7 +83,7 @@ class DoctrineDataCollectorTest extends TestCase
      */
     private function createCollector(array $managers)
     {
-        $registry = $this->createMock('Doctrine\Common\Persistence\ManagerRegistry');
+        $registry = $this->getMockBuilder('Doctrine\Common\Persistence\ManagerRegistry')->getMock();
         $registry
             ->expects($this->any())
             ->method('getConnectionNames')
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         }
     ],
     "require": {
-        "php": "^7.1",
+        "php": "^5.5.9|^7.0",
         "symfony/framework-bundle": "~2.7|~3.0|~4.0",
         "symfony/console": "~2.7|~3.0|~4.0",
         "symfony/dependency-injection": "~2.7|~3.0|~4.0",
@@ -41,7 +41,7 @@
         "symfony/phpunit-bridge": "~2.7|~3.0|~4.0",
         "twig/twig": "~1.12|~2.0",
         "satooshi/php-coveralls": "^1.0",
-        "phpunit/phpunit": "^6.1"
+        "phpunit/phpunit": "^4.8.36|^5.7|^6.4"
     },
     "suggest": {
         "symfony/web-profiler-bundle": "To use the data collector.",
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,14 +7,19 @@ cache:
     - $HOME/.composer/cache
 
 php:
+  - 5.5
+  - 5.6
+  - 7.0
   - 7.1
   - nightly
 
 matrix:
   include:
-    - php: 7.1
+    - php: 5.5
       env: deps=low SYMFONY_DEPRECATIONS_HELPER=weak
-    - php: 7.1
+    - php: 5.6
+      env: SYMFONY_VERSION="2.8.*@dev"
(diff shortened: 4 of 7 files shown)


Tests/BundleTest.php
```diff
@@ -14,13 +14,12 @@
...
 namespace Doctrine\Bundle\DoctrineBundle\Tests;
 
-use PHPUnit\Framework\TestCase;
 use Symfony\Bridge\Doctrine\DependencyInjection\CompilerPass\DoctrineValidationPass;
 use Symfony\Bridge\Doctrine\DependencyInjection\CompilerPass\RegisterEventListenersAndSubscribersPass;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Doctrine\Bundle\DoctrineBundle\DoctrineBundle;
 
-class BundleTest extends TestCase
+class BundleTest extends \PHPUnit_Framework_TestCase
```
alcaeus: Not necessary due to the forward compatibility layer in 4.8 and 5.7.

composer.json
```diff
@@ -24,7 +24,7 @@
         }
     ],
     "require": {
-        "php": "^7.1",
+        "php": ">=5.5.9",
```
alcaeus: `^5.5.9 || ^7.0` - no need to allow PHP 8.

composer.json
```diff
@@ -41,7 +41,7 @@
         "symfony/phpunit-bridge": "~2.7|~3.0|~4.0",
         "twig/twig": "~1.12|~2.0",
         "satooshi/php-coveralls": "^1.0",
-        "phpunit/phpunit": "^6.1"
+        "phpunit/phpunit": "~4"
```
alcaeus: `^4.8.36 || ^5.7 || ^6.4` would do the trick and ensure you've got the `PHPUnit\Framework\TestCase` class

DependencyInjection/DoctrineExtension.php
```diff
@@ -84,6 +84,20 @@ public function load(array $configs, ContainerBuilder $container)
 
             $this->ormLoad($config['orm'], $container);
         }
+
+        if (PHP_VERSION_ID < 70000) {
```
stof: I suggest to keep this out to simplify the code

Tests/DataCollector/DoctrineDataCollectorTest.php
```diff
@@ -28,7 +28,7 @@ class DoctrineDataCollectorTest extends TestCase
     public function testCollectEntities()
     {
         $manager = $this->getMockBuilder('Doctrine\ORM\EntityManager')->disableOriginalConstructor()->getMock();
-        $config = $this->createMock('Doctrine\ORM\Configuration');
+        $config = $this->getMock('Doctrine\ORM\Configuration');
```
stof: you should use `getMockBuilder()->getMock()` to stay compatible with PHPUnit 6

Majkl578:
I'm against this change.

PHP 5.5 had EOL over a year ago, 5.6 is in security only mode without active support and 7.0 reaches that phase next month.
Also latest versions od Doctrine packages require 7.1 (in this case most notably DBAL 2.6 and upcoming ORM 2.6) and this change will only cause confusion among users, there is already a lot of noise regarding PHP 7.1 now, this would only make it worse.

Anyone who needs legacy PHP version gets DBAL 2.5, ORM 2.5 and DoctrineBundle 1.6, but there is no reason to keep support to relics in latest versions even further.

_This is really concerning, I fear what will be next, maybe PR to degrade ORM to 5.5 too?_

composer.json
```diff
@@ -24,7 +24,7 @@
         }
     ],
     "require": {
-        "php": "^7.1",
+        "php": "^5.5.9 || ^7.0",
```
Majkl578: Rest of the constraints use `^5.5.9|^7.0` format, please don't introduce different one.

fabpot:
@Majkl578 This bundle is all about the integration with Symfony, so having the same PHP requirements make sense IMHO. It makes sense also because we need to make changes in DoctrineBundle to support new Symfony 3.4 features.

Majkl578:
> This bundle is all about the integration with Symfony, so having the same PHP requirements make sense IMHO.

Same could be said about it in terms of Doctrine.

As far as I know fixes related to Symfony 3.4 were already backported to 1.6.x: #705, #710, #715, so users of unmaintained PHP versions are _just fine_ with those.

Why do you think anyone would desperately need Symfony 3.4 and at the same time wouldn't need Doctrine 2.6 (which requires PHP 7.1, but has important features like immutable datetime types)?

alcaeus:
@Majkl578 I think the idea is to allow users of Symfony 3.4 to also get support for the new features (autowiring etc.) in the core bundles. Considering DoctrineBundle is one of the default bundles, it makes sense to loosen the PHP requirements to match the Symfony version it targets. That way, people get new Symfony features but not the new Doctrine features. As for ORM itself, it's a different story. People will have to use 2.5 if they want to use anything older than 7.1, but that shouldn't affect the bundle (in this case).

composer.json
alcaeus: Ah, I keep forgetting which format is the one we use. Sorry.

Ocramius:
FWIW, I discussed this with @nicolas-grekas while at Polycon.io: change is fine as long as maintenance is up to the symfony team again.

This is a decision @kimhemsoe has to take: he's the person in charge here.

I personally won't touch and won't care for pre-PHP-7.1 support unless it's security issues.

fabpot:
Indeed, the Symfony core team is taking over the bundles as demonstrated by the activity during the last few weeks on this bundle and DoctrineCacheBundle. It's not to say that the Doctrine team cannot help, and @kimhemsoe and others are more than welcome to continue working on it, but we think that the Doctrine bundles should be more aligned with Symfony choices.

weaverryan:
I've fixed the bad version constraint format :)

composer.json
```diff
@@ -41,7 +41,7 @@
         "symfony/phpunit-bridge": "~2.7|~3.0|~4.0",
         "twig/twig": "~1.12|~2.0",
         "satooshi/php-coveralls": "^1.0",
-        "phpunit/phpunit": "^6.1"
+        "phpunit/phpunit": "^4.8.36 || ^5.7 || ^6.4"
```
xabbuh: `^4.8.36|^5.7|^6.4`?

weaverryan:
Ping! If we're going to do it, I'd like to merge this and #727 soon. It will affect the Symfony 4 docs and some other things.

Thanks!

PR #727, opened 2017-10-29
Adding option to allow repositories to be fetched from the service container

weaverryan:
Hi guys!

Sorry again for the noise :). The goal is to allow repositories to be added to the container easily (i.e. in an autowired way). There are 2 possible approaches:

~~1) In symfony/symfony#24693, we add a new trait to make creating services that are similar to repositories easily.~~

2) In this PR, we allow for the `$em->getRepository()` to return a service, instead of it being created by Doctrine.

I'm currently leaning towards the approach in this PR, as it keeps only *one* idea of a "repository" and migrating to the new system would be super easy. This idea was inspired by @beberlei in #725. ~~I have a few pending questions:~~

Answered: https://github.com/doctrine/DoctrineBundle/pull/727#issuecomment-340747136
~~A) Can an entity only be managed by exactly one EntityManager (as you seem to suggest)? This PR (which uses your code) suggests this.~~

~~B) Are there any issues with `Registry::getManagerForClass()`? It loops over all of the EM's to find the correct one. Is there a possible issue with instantiating (thanks to this loop) other EM's?~~

## Example

The feature is automatically enabled. But existing repositories will still be created the "default" way.

The users will configure their `repositoryClass` like normal. But now, `repositoryClass` is *allowed* to be a service id (which usually == the class name these days, but that's not important).

```php
/**
 * @ORM\Entity(repositoryClass="App\Repository\CoolStuffRepository")
 */
class CoolStuff
{

}
```

Your repository class will (with default config) auto-registered as a service. So, without touching any config, you create the class and it just works:

```php
class CoolStuffRepository extends ServiceEntityRepository /* this class extends EntityRepository, implements ServiceEntityRepositoryInterface */
{
    /*
     * If you want more args, you will need to call the parent __construct
    {
        parent::__construct($registry, CoolStuff::class);
    }
    */

    // your custom methods
}
```

That's it! The new `ContainerRepositoryFactory` is capable of fetching repositories from the container *or* creating them like normal. How does it know which to do?

A) If your class still extends `EntityRepository`, then it's created like normal. This is a great BC/upgrade path.
B) If your class implements a new `ServiceEntityRepositoryInterface`, this is a signal to fetch from the container.

The entire system is functionally-tested, including strict errors if you mess up (e.g. implement the new interface, but forget to register your service).

Your new services *are* required to extend `EntityRepository`, because outside systems could be type-hinting against this. If you don't like that, you could ignore this feature and create your own service classes outside of the repository system.

Cheers!


Resources/config/dbal.xml
```diff
@@ -64,6 +64,7 @@
             <argument>%doctrine.default_entity_manager%</argument>
         </service>
         <service id="Doctrine\Common\Persistence\ManagerRegistry" alias="doctrine" public="false" />
+        <service id="Symfony\Bridge\Doctrine\RegistryInterface" alias="doctrine" public="false" />
```
alcaeus: The alias looks wrong here.

DependencyInjection/DoctrineExtension.php
```diff
@@ -375,6 +378,9 @@ protected function ormLoad(array $config, ContainerBuilder $container)
                 $def->addTag('doctrine.event_subscriber');
             }
         }
+
+        $container->registerForAutoconfiguration(EntityRepository::class)
```
alcaeus: Shouldn’t the tag only be applied to the new ServiceEntityRepository?

Tests/DependencyInjection/Fixtures/Bundles/RepositoryServiceBundle/Repository/TestCustomRepoRepository.php
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Fixtures\Bundles\RepositoryServiceBundle\Repository;
+
+use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
+use Fixtures\Bundles\RepositoryServiceBundle\Entity\TestCustomRepoEntity;
+use Symfony\Bridge\Doctrine\RegistryInterface;
+
+class TestCustomRepoRepository extends ServiceEntityRepository
+{
+    public function __construct(RegistryInterface $registry)
```
alcaeus: Having to override the constructor is going to feel weird. People are likely to forget to override it or to change the signature. One option to avoid this is to loop through the metadata array from the entity manager and look for an entity that has our class as custom repository. It would be nice if we could even move that to the service itself to avoid looking up that information on every instantiation, but i believe changing this will improve DX.

DependencyInjection/DoctrineExtension.php
nicolas-grekas: doing so wuld allow removing the config option altogether, which would be nice, isn't it?

Resources/config/dbal.xml
weaverryan: Yea...  this is a bit unrelated. The issue is that the existing alias is a bit difficult, because there is an abstract class in the Bridge by the same name. It’s one of those situations where the user needs to guess the correct use statement, and only one will work. I’d like to add this ORM-specific alias and tell people to target it (but also mention the other interface, for those who care).

Tests/DependencyInjection/Fixtures/Bundles/RepositoryServiceBundle/Repository/TestCustomRepoRepository.php
weaverryan: The need to override the constructor is indeed the weirdest part :/ (but you only need to do it if you need to add more constructor args, otherwise the default constructor is fine). So, I’m interested in your idea, but I don’t understand yet. Where would we loop over the class metadata? And how would that help? Doesn’t the user need to (somehow) do some work/chores in their service to set the protected properties that EntityRepository needs internally?

DependencyInjection/DoctrineExtension.php
weaverryan: Interesting... to remove the option entirely, it would mean that, when an entity has a custom repository, we would sometimes instantiate that class (like current behavior) and sometimes fetch it from the container. Effectively, if their repo extends the new ServiceEntityRepository, then we fetch from container (or error if not exists). Otherwise, we instantiate.

I like this because it makes migrating to the new system even easier and removes this option. But is it too magic? The choice of your base class will determine how your repo is fetched/created.

nicolas-grekas: That's what autoconfig is already, adding behavior based on the base classes :)

beberlei:
@weaverryan to answer your questions:

a.) Theoretically, an entity can be in multiple managers, but I haven't seen someone using it this way and would consider it a 1% special case where its ok that repository as service + autowiring doesn't work. Other code in the DoctrineBundle will not work (Forms, Validators) if you are using one entity with multiple managers.

b.) You are right, this might cause additional overhead with multiple EMs, and you need to probably look out for issues with automatically connecting to databases, although afaik this problem has been fixed recently? Again, this method is used in form and validation component already, so any potential problems should have surfaced there already.

In general I think this approach works for the 80-90% use case and as a workaround users can just register the service manually to provide the correct entity manager.

Repository/ContainerRepositoryFactory.php
```diff
@@ -0,0 +1,73 @@
...
+     * {@inheritDoc}
+     */
+    public function getRepository(EntityManagerInterface $entityManager, $entityName)
+    {
+        $metadata = $entityManager->getClassMetadata($entityName);
+        $repositoryServiceId = $metadata->customRepositoryClassName;
+
+        if (null !== $repositoryServiceId) {
+            if (!$this->container->has($metadata->customRepositoryClassName)) {
+                throw new \RuntimeException(sprintf('Could not find the repository service for the "%s" entity. Make sure the "%s" service exists and is tagged with "%s".', $entityName, $repositoryServiceId, ServiceRepositoryCompilerPass::REPOSITORY_SERVICE_TAG));
```
beberlei: This is a bit harsh, wouldn't it be better to default to the original use by checking `$customRepositoryClassName` does not implement the `ServiceEntityRepository` and instantiate it with the regular approach?

beberlei: The way to do this would be remove the `if (null !== $repositoryServiceId)` check to get back into the default code starting in line 64+.

weaverryan:
Thank you for the answers!

> Am I correct in assuming that this requires users to create a service for each repository themselves?
> Is this desired or would an even better result be that repository services are automagically registered?

Yes, you're correct! But the registration of the services can already be automated with the "service auto-registration" feature in 3.3. And for repositories that don't have a custom class, auto-registering a service with an `EntityRepository` class doesn't help the user much anyways, as you wouldn't be able to autowire those services as arguments to other services. So, I think it's not a problem :).

I'm working on a few changes related to your comments and others right now...

beberlei:
@weaverryan can you point me to details of the "service auto-registration" feature of 3.3? I am still living in my small symfony 2 world ;-)

weaverryan:
@beberlei LOL, of course :). 

New Symfony projects ship with this optional config: https://github.com/symfony/symfony-standard/blob/3.3/app/config/services.yml#L19-L23

This automatically registers every service in the container. The service id === the class name. Those services are *private* (important because it means that unused services are removed before dumping the compiled container... so no extra overhead if some classes in `src/` are *not* actually services). Thanks to `_defaults` (https://github.com/symfony/symfony-standard/blob/3.3/app/config/services.yml#L8-L15), these services are also autowired (+ private and "autoconfigured" - fancy word that means we will try to add tags automatically based on the interface of services).

The `Repository` directory is excluded from the auto-registration (https://github.com/symfony/symfony-standard/blob/3.3/app/config/services.yml#L23). That's actually not needed, just a minor performance boost in dev. If this PR is merged, users would uncomment that and their repositories would instantly become services (well, autowiring will fail until they change their base class to the new, autowiring friendly `ServiceEntityRepository` from this PR... and stolen from you).

So, the user won't care much if their default "EntityRepository" repositories are in the container or not... using `$em->getRepository()` is even easier than manually fetching/wiring up some service id.

But as soon as they *do* have a custom repo class, they'll want that to be auto-registered so that it is available for autowiring. That's actually pretty simple - only the `ServiceEntityRepository` is necessary for that. Most of this PR is *really* centered around making `$em->getRepository()` also smart enough to now get things from the container so that there's consistency.

weaverryan:
Update! The config option is gone. The feature is now *always* on, but you need to activate it by making your repository implement a new interface. So, all existing repositories will work exactly like before.

The new functionality is still Symfony 3.3+ only, and clear exceptions are thrown if you're on a lower version. Actually, most of the ugliness in this PR is keeping 3.2 and lower compat. If we bumped the next version of this bundle to 3.3, we could remove all of that.

There are no current blockers. This is ready again for review.

DependencyInjection/Compiler/ServiceRepositoryCompilerPass.php
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        $repoReferences = array_map(function($id) {
+            return new Reference($id);
+        }, $repoServiceIds);
+
+        $ref = ServiceLocatorTagPass::register($container, array_combine($repoServiceIds, $repoReferences));
+        $locatorDef->replaceArgument(0, $ref);
+    }
+}
```
fabpot: missing newline

DependencyInjection/Compiler/ServiceRepositoryCompilerPass.php
```diff
@@ -0,0 +1,52 @@
...
+    public function process(ContainerBuilder $container)
+    {
+        $locatorDef = $container->getDefinition('doctrine.orm.container_repository_factory');
+
+        $repoServiceIds = array_keys($container->findTaggedServiceIds(self::REPOSITORY_SERVICE_TAG));
+
+        // Symfony 3.2 and lower sanity check
+        if (!class_exists(ServiceLocatorTagPass::class)) {
+            if (!empty($repoServiceIds)) {
+                throw new RuntimeException(sprintf('The "%s" tag can only be used with Symfony 3.3 or higher. Remove the tag from the following services (%s) or upgrade to Symfony 3.3.', self::REPOSITORY_SERVICE_TAG, implode(', ', $repoServiceIds)));
```
fabpot: I would say "upgrade to Symfony 3.3 or higher".

DependencyInjection/DoctrineExtension.php
```diff
@@ -375,6 +378,25 @@ protected function ormLoad(array $config, ContainerBuilder $container)
...
+         * Compatibility for Symfony 3.2 and lower: gives the service a default argument.
+         * When DoctrineBundle requires 3.3 or higher, this can be moved to an anonymous
+         * service in orm.xml.
+         *
+         * This is replaced with a true locator by ServiceRepositoryCompilerPass.
+         * This makes that pass technically optional (good for tests).
+         */
+        if (class_exists(ServiceLocator::class)) {
+            $container->getDefinition('doctrine.orm.container_repository_factory')
+                ->replaceArgument(0, (new Definition(ServiceLocator::class))->setArgument(0, array()));
```
alcaeus: Nitpick: use short array syntax.

Tests/DependencyInjection/Fixtures/Bundles/RepositoryServiceBundle/Repository/TestCustomRepoRepository.php
alcaeus: I thought that the constructor override was necessary to get the new repositories to play nice with service auto-registration. Since there will most likely be a repository for a single entity (unlike the default `EntityRepository`), it would make sense trying to deduct the entity class name from the entity manager in `ServiceEntityRepository` if it isn't given:
- If entity class name is missing, fetch a list of all metadata from the document managers (yes, plural unfortunately).
- Loop over metadata, looking for an entity with a `customRepositoryClass` of `get_class($this)`.
- If a single entity is found, use it. If more/less than one entity is found, throw an exception.

weaverryan:
Nit-pick handled :)

fabpot:
FYI, I've bumped the version of the bundle to 1.8 before merging this PR, which will be part of 1.8.0 stable.

DependencyInjection/DoctrineExtension.php
```diff
@@ -375,6 +378,25 @@ protected function ormLoad(array $config, ContainerBuilder $container)
                 $def->addTag('doctrine.event_subscriber');
             }
         }
+
+        // if is for Symfony 3.2 and lower compat
```
fabpot: typo: `it` instead of `if`

Repository/ContainerRepositoryFactory.php
```diff
@@ -0,0 +1,102 @@
...
+        $this->container = $container;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getRepository(EntityManagerInterface $entityManager, $entityName)
+    {
+        $metadata = $entityManager->getClassMetadata($entityName);
+        $repositoryServiceId = $metadata->customRepositoryClassName;
```
Taluu: unless I read something wrong, this variable doesn't seem to be used anywhere within the body of this method

yceruto: https://github.com/doctrine/DoctrineBundle/pull/727/files/4efcb135344e55403a0fd193c23a8c7249cfd7d0#diff-f36aa436e9aae3ea54bbaca0caf377e3R65 ;)

yceruto:
@weaverryan I tried this feature with Symfony 4 and it look great! Thanks.

Just a minor DX issue:
```bash
Cannot autowire service "App\Repository\PostRepository": argument "$entityClass" of method "Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository::__construct()" must have a type-  
  hint or be given a value explicitly.
```
So I had to configure each repository manually to complete the service definition. Is there a way to skip this step?

weaverryan:
Hmm. Did you change your entity’s base class to the new ServiceEntityRepository? That should fix autowiring :)

yceruto:
yep, I did.

ogizanagi:
@yceruto : Overwritting the `ServiceEntityRepository::__construct` to pass the `$entityClass` argument yourself in own `PostRepository` is a required step. I don't think we can/want to skip it, as that'll mean accessing the doctrine metadata during compile time to get the entity class associated to the repository service and also because the service constructor does belong to you, so you're free to not expose the `$entityClass` argument the same way or at all.

yceruto:
@ogizanagi I thought extending from `ServiceEntityRepository` was enough (https://gist.github.com/weaverryan/eec33e4a6d3f98efd3b56939bc08697d), that mean I need to do this also for each repository, no?
```php
{
    parent::__construct($registry, CoolStuff::class);
}
```

ostrolucky:
Yes. Not a fan myself of this constructor overriding requirement. Not sure how is this

```diff
+use App\Entity\Company;
+use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
+use Doctrine\Common\Persistence\ManagerRegistry;
 
-class CompanyRepository extends Repository
+class CompanyRepository extends ServiceEntityRepository
 {
+    public function __construct(ManagerRegistry $registry, $entityClass)
+    {
+        parent::__construct($registry, Company::class);
+    }
+
 ////////////////
+use App\Entity\Project\Project;
+use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
+use Doctrine\Common\Persistence\ManagerRegistry;
 
-class ProjectRepository extends Repository
+class ProjectRepository extends ServiceEntityRepository
 {
+    public function __construct(ManagerRegistry $registry, $entityClass)
+    {
+        parent::__construct($registry, Project::class);
+    }
+
 ////////////////
+use App\Entity\User\UserAdmin;
+use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
+use Doctrine\Common\Persistence\ManagerRegistry;
 
-class UserAdminRepository extends Repository
+class UserAdminRepository extends ServiceEntityRepository
 {
+    public function __construct(ManagerRegistry $registry, $entityClass)
+    {
+        parent::__construct($registry, UserAdmin::class);
+    }
+
 ////////////////
+use App\Entity\User\UserClient;
+use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
+use Doctrine\Common\Persistence\ManagerRegistry;
 
-class UserClientRepository extends Repository
+class UserClientRepository extends ServiceEntityRepository
 {
+    public function __construct(ManagerRegistry $registry, $entityClass)
+    {
+        parent::__construct($registry, UserClient::class);
+    }
+
 ////////////////
+use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
+use Doctrine\Common\Persistence\ManagerRegistry;
 
-class UserPermissionRepository extends Repository
+class UserPermissionRepository extends ServiceEntityRepository
 {
+    public function __construct(ManagerRegistry $registry, $entityClass)
+    {
+        parent::__construct($registry, UserPermission::class);
+    }
+
```


Better than this:

```yml
services:
    repository_factory:
        public: false
        abstract: true
        factory: ['@doctrine', getRepository]

    App\Repository\CompanyRepository:
        parent: repository_factory
        arguments: [App\Entity\Company]

    App\Repository\UserAdminRepository:
        parent: repository_factory
        arguments: [App\Entity\User\UserAdmin]

    App\Repository\UserClientRepository:
        parent: repository_factory
        arguments: [App\Entity\User\UserClient]

    App\Repository\UserPermissionRepository:
        parent: repository_factory
        arguments: [App\Entity\Project\UserPermission]

    App\Repository\ProjectRepository:
        parent: repository_factory
        arguments: [App\Entity\Project\Project]
```

Can I have a go with different approach, where I access metadata during compilation? I will do the benchmark as part of that.

Koc:
@ostrolucky accessing metadata during compilation is not good idea.

Koc:
@weaverryan why not create `ServiceEntityRepository` abstract with method `getEntityClass`?

alcaeus:
> accessing metadata during compilation is not good idea.

Please elaborate as to why it's a bad idea.

> why not create `ServiceEntityRepository` abstract with method `getEntityClass`

IIRC, @weaverryan didn't want users to implement additional methods - it should "just work". On top of that, entities and repositories are tied together using the mapping (namely the `repositoryClass` option in the `Entity` mapping). There shouldn't be the need for the repository class to be tied to an entity in a second place. This is also a great antipattern, as from a Doctrine perspective an entity has one repository class, but a repository class may be used for multiple entities.

Koc:
@stof explained why https://github.com/symfony/symfony/issues/18673#issuecomment-216183042 and some time ago was some attempts to deprecate this behavior https://github.com/symfony/symfony/pull/18728

yceruto:
> @weaverryan didn't want users to implement additional methods - it should "just work".

Users need override the `__constructor` method though to make it work.

> There shouldn't be the need for the repository class to be tied to an entity in a second place.

It mean we've already a antipattern now `parent::__construct($registry, CoolStuff::class);`? We can reuse this repository for another entity?

alcaeus:
@yceruto 

> It mean we've already a antipattern now `parent::__construct($registry, CoolStuff::class);`? We can reuse this repository for another entity?

Correct - I've also mentioned that this approach feels wrong in one of the pull requests involved with this feature. I still think that this shouldn't have to be handled by the user.

alcaeus:
@koc the issues you linked deal with fetching a service from an unbuilt container builder. Since this bundle is the place where the services in questions are generated, there are ways to work around this if you wanted to.

bobvandevijver:
I wonder why this is merged without any documentation, as it is actually I feature which is pretty awesome!

xabbuh:
@bobvandevijver see http://symfony.com/doc/current/doctrine.html#querying-for-objects-the-repository

Repository/ContainerRepositoryFactory.php
michaelperrin: 10 years later 😆I agree with @Taluu , the variable doesn't seem to be used, as it used only used for the message of the `RuntimeException` that is thrown a few lines below.

But the `$customRepositoryName` class contains exactly the same value, and could be used instead. Maybe @weaverryan can tell more about it :)

PR #735, opened 2017-11-19
Fix #730 compatibility with web profiler 4

### LINKED ISSUE #730: WebProfilerExtension dumpData() array given Type error.
Hey,

I have problem with Doctrine profiling on MySQL database.
`Type error: Argument 2 passed to Symfony\Bundle\WebProfilerBundle\Twig\WebProfilerExtension::dumpData() must be an instance of Symfony\Component\VarDumper\Cloner\Data, array given`

This is coming from:
`vendor/doctrine/doctrine-bundle/Resources/views/Collector/db.html.twig:241`

I changed the profiler_dump to dump and it contains array with my query params:
Example:
`array(1) { [0]=> string(36) "bc54bece-c234-11e7-85e6-0242ac140003" }`

I am using: symfony/orm-pack ^1.0 with symfony 4.0 beta-2.

ostrolucky:
Doctrine profiler does not work in Symfony 4. This issue is completely ignored, so I had to give it a try. I have extended travis matrix to test Symfony 4 and fixed all Symfony 4 test failures.


composer.json
```diff
@@ -39,9 +39,14 @@
...
         "symfony/phpunit-bridge": "~2.7|~3.0|~4.0",
-        "twig/twig": "~1.12|~2.0",
+        "twig/twig": "~1.26|~2.0",
         "satooshi/php-coveralls": "^1.0",
-        "phpunit/phpunit": "^4.8.36|^5.7|^6.4"
+        "phpunit/phpunit": "^4.8.36|^5.7|^6.4",
+        "symfony/web-profiler-bundle": "~2.7|~3.0|~4.0"
+    },
+    "conflict": {
+        "symfony/http-foundation": "<2.6",
```
ostrolucky: db.html.twig uses `request.query.getBoolean`. `ParameterBag#getBoolean` has been introduced with 2.6

composer.json
```diff
@@ -39,9 +39,13 @@
...
         "symfony/phpunit-bridge": "~2.7|~3.0|~4.0",
-        "twig/twig": "~1.12|~2.0",
+        "twig/twig": "~1.26|~2.0",
         "satooshi/php-coveralls": "^1.0",
-        "phpunit/phpunit": "^4.8.36|^5.7|^6.4"
+        "phpunit/phpunit": "^4.8.36|^5.7|^6.4",
+        "symfony/web-profiler-bundle": "~2.7|~3.0|~4.0"
+    },
+    "conflict": {
+        "symfony/http-foundation": "<2.6"
```
ostrolucky: `db.html.twig` uses `request.query.getBoolean()`. `ParameterBag#getBoolean` has been introduced with 2.6

composer.json
```diff
@@ -29,7 +29,7 @@
         "symfony/console": "~2.7|~3.0|~4.0",
         "symfony/dependency-injection": "~2.7|~3.0|~4.0",
         "doctrine/dbal": "^2.5.12",
-        "jdorn/sql-formatter": "~1.1",
+        "jdorn/sql-formatter": "^1.2.16",
```
ostrolucky: We use `SqlFormatter::$variable_attributes`, introduced with 1.2.16

composer.json
```diff
@@ -39,9 +39,13 @@
         "symfony/validator": "~2.7|~3.0|~4.0",
         "symfony/property-info": "~2.8|~3.0|~4.0",
         "symfony/phpunit-bridge": "~2.7|~3.0|~4.0",
-        "twig/twig": "~1.12|~2.0",
+        "twig/twig": "~1.26|~2.0",
```
ostrolucky: Increased to be able to call `Twig_Environment::addRuntimeLoader` in tests. Notice it's require-dev section

.travis.yml
```diff
@@ -18,20 +18,19 @@ matrix:
...
-    - env: deps=dev
 
 env:
   global:
     - deps=no
 
 before_install:
   - composer self-update
-  - if [ "$SYMFONY_VERSION" != "" ]; then composer require --no-update symfony/symfony:${SYMFONY_VERSION}; fi;
+  - if [ "$SYMFONY_VERSION" != "" ]; then jq "(.require, .\"require-dev\")|=(with_entries(if .key|test(\"^symfony/\") then .value|=\"${SYMFONY_VERSION}\" else . end))" composer.json|ex -sc 'wq!composer.json' /dev/stdin; fi;
```
ostrolucky: Old approach replaced to be able to keep paths used in tests in `../vendor/symfony/{component}` format, instead of introducing switches in tests to determine if `../vendor/symfony/symfony/{component}` exists. Simple regex not used to keep `conflicts` section without change.

DataCollector/DoctrineDataCollector.php
```diff
@@ -136,6 +137,19 @@ public function collect(Request $request, Response $response, \Exception $except
             }
         }
 
+        // Might be good idea to replicate this block in doctrine bridge so we can drop this from here after some time.
+        // This code is compatible with such change, because cloneVar is supposed to check if input is already cloned.
+        foreach ($this->data['queries'] as &$queries) {
+            foreach ($queries as &$query) {
+                // HttpKernel < 3.2 compatibility layer
+                if (!method_exists($this, 'cloneVar')) {
```
mikeSimonson: Shouldn't both foreach be inside that if ?

ostrolucky: indeed

.travis.yml
stof: much simpler: `composer require symfony/lts` (be careful, versions there are just 2 and 3, not 2.8 and 3.4)

.travis.yml
```diff
@@ -11,27 +11,33 @@ php:
...
 matrix:
   include:
     - php: 5.5
       env: deps=low SYMFONY_DEPRECATIONS_HELPER=weak
     - php: 5.6
-      env: SYMFONY_VERSION="2.8.*@dev"
-    - php: 5.6
+      env: SYMFONY_VERSION="2.8.*"
+    - php: 7.0
+      env: SYMFONY_VERSION="3.2.*"
```
stof: don't had it. 3.2 is unmaintained

.travis.yml
```diff
@@ -11,27 +11,33 @@ php:
...
+    - php: 7.1
+      env: SYMFONY_VERSION="3.4.*@beta"
+    - php: 7.2
+      env: SYMFONY_VERSION="4.0.*@beta"
```
stof: -1 for that. This is already covered by the `deps=dev` job. I'm only adding version-specific jobs for LTS versions (I don't want to update this config file every 6 months to change the versions, too painful)

DependencyInjection/Compiler/EntityListenerPass.php
```diff
@@ -47,6 +47,7 @@ public function process(ContainerBuilder $container)
                 }
 
                 $resolver = $container->findDefinition($resolverId);
+                $resolver->setPublic(true);
```
stof: would be better to use a ServiceLocator (by changing the lazy-loading listener resolver) rather than forcing to use public services

Resources/views/Collector/db.html.twig
```diff
@@ -463,3 +465,22 @@
...
+        <tbody>
+        {% for key, value in data %}
+            <tr>
+                <th scope="row">{{ key }}</th>
+                <td>{{ value }}</td>
+            </tr>
+        {% endfor %}
+        </tbody>
+    </table>
+{% endmacro %}
```
stof: please add the final EOL

Tests/ProfilerTest.php
```diff
@@ -0,0 +1,85 @@
...
+            'page' => 'foo',
+            'profile' => $profile,
+            'collector' => $this->collector,
+            'queries' => $this->logger->queries,
+        ]);
+
+        $output = str_replace(["\e[37m", "\e[0m", "\e[32;1m", "\e[34;1m"], "", $output);
+        $this->assertContains("SELECT * FROM foo WHERE bar IN ('foo', 'bar');", $output);
+    }
+}
```
stof: missing final EOL

Tests/ProfilerTest.php
```diff
@@ -0,0 +1,85 @@
...
+use Symfony\Bridge\Twig\Extension\RoutingExtension;
+use Symfony\Bundle\WebProfilerBundle\Twig\WebProfilerExtension;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\DataCollector\RequestDataCollector;
+use Symfony\Component\HttpKernel\Fragment\FragmentHandler;
+use Symfony\Component\HttpKernel\Profiler\Profile;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+
+class ProfilerTest extends \PHPUnit\Framework\TestCase
```
stof: please add a use statement

.travis.yml
ostrolucky: Good idea, though you already said disadvantages yourself. I would like to keep ability to pick exact symfony version

.travis.yml
ostrolucky: Not exactly true, end of support is in January. Until then we need to make sure it works. If you notice, there are BC layers introduced in my patch and I would like them to get tested.

.travis.yml
ostrolucky: Which deps=dev? The one which was allowed to fail? We can't afford for this bundle to stay incompatible with next Symfony version and figure it out only at RC phase again. It must have been a while since tests started to fail for Symfony 4 and it was ignored. These constraints should be set in the way it's clear that something is wrong when tests fail, but on the other side aren't supposed to be expected to fail randomly, which would be the case with dev dependencies. I suggest to have matrix for latest dev=alpha then. That's when failures are starting to be unwanted.

DependencyInjection/Compiler/EntityListenerPass.php
ostrolucky: Not sure what do you mean. I don't want to change behaviour. Tests expect this service to be public, meaning having it directly accessible from container. Making it otherwise would be BC break. I have done this change only to fix tests.

weaverryan:
Hey guys! Do we need any more changes on this? I'm tracking to make sure we get a release and tag by next week.

ostrolucky:
It's finished from my POV

ostrolucky:
Let's please try to do the merge during this Symfony RC phase, so I can make analogous patch for Doctrine bridge 3.4 which will allow us to drop this workaround https://github.com/ostrolucky/DoctrineBundle/blob/b1e881aacb1af71293ceca13da852c01d9fc1afe/DataCollector/DoctrineDataCollector.php#L140-L149 when Symfony 2.* is EOL. If we won't make it, that workaround might have to stay there during whole life cycle of Symfony 3.4.

weaverryan:
@ostrolucky Unless it's a pretty serious/obvious bug fix, nothing at this point will make it into 3.4. So, I think that workaround will need to stay. But, no big deal.

But yes, since this is broken in Symfony 4, I'd love to see a quick merge + tag so that (if there are any more issues), we can find them.

weaverryan:
Thanks @mikeSimonson! I already see a new 1.8.0 tag... but I think this isn't included in it? Is that correct? Could we create a 1.8.1 with this fix?

mikeSimonson:
@weaverryan done

PR #740, opened 2017-11-27
Remove aliases for common interfaces

alcaeus:
#640 added a bunch of aliases for autowiring, but some of them don't make sense because the interfaces might also be implemented by other bundles. Thus, interfaces in `Doctrine\Common` shouldn't be used as service aliases for ORM-specific implementations.

For example, this conflicts with https://github.com/doctrine/DoctrineMongoDBBundle/pull/441 - in order to not confuse people I believe it's best if none of the mappers registers service aliases for the common interfaces.

### DIFF
--- a/Resources/config/dbal.xml
+++ b/Resources/config/dbal.xml
@@ -63,7 +63,6 @@
             <argument>%doctrine.default_connection%</argument>
             <argument>%doctrine.default_entity_manager%</argument>
         </service>
-        <service id="Doctrine\Common\Persistence\ManagerRegistry" alias="doctrine" public="false" />
         <service id="Symfony\Bridge\Doctrine\RegistryInterface" alias="doctrine" public="false" />
 
         <service id="doctrine.twig.doctrine_extension" class="Doctrine\Bundle\DoctrineBundle\Twig\DoctrineExtension" public="false">
--- a/Resources/config/orm.xml
+++ b/Resources/config/orm.xml
@@ -75,7 +75,6 @@
     </parameters>
 
     <services>
-        <service id="Doctrine\Common\Persistence\ObjectManager" alias="doctrine.orm.entity_manager" public="false" />
         <service id="Doctrine\ORM\EntityManagerInterface" alias="doctrine.orm.entity_manager" public="false" />
 
         <!--- Deprecated Annotation Metadata Reader Service alias, use annotation_reader service -->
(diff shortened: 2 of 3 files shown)


stof:
Well, for projects using only one of these bundles, autowiring would do exactly what they need for these common interfaces currently. Only projects using both the ORM and the ODM would face issues with these aliases (but then, it is because there is simply no way to decide automatically).

I think it is better to make autowiring work for all people using only one of the Doctrine mappers rather than breaking it for them just because people using multiple mappers cannot benefit from this feature.

alcaeus:
> I think it is better to make autowiring work for all people using only one of the Doctrine mappers rather than breaking it for them just because people using multiple mappers cannot benefit from this feature.

I disagree. By that logic, we'd be adding the same aliases in the pull request mentioned above. If I'm not mistaken, the last alias to be registered wins. Since this is determined by the order in which the bundles are registered, adding a second mapper later and adding it to the list (or having it loaded automatically using flex) will actually break the current logic. I believe that it's better to avoid this situation entirely by not teaching people to typehint the `Doctrine\Common` interfaces. If you want to keep your code framework agnostic, you'll most likely have some glue in place to decide whether to use ORM or ODM.

weaverryan:
I'm very conflicted: we're trying to solve a 1% use-case. I don't want to make life more difficult for the 99%!

But, @alcaeus brought up another good example: the Doctrine Event Subscribers (`EventSubscriber` from Doctrine\Common). Since this is a shared interface, we can't add autoconfigure to `DoctrineBundle` without making the ODM go crazy. It seems that in the world of autowiring/autoconfigure, each library really needs its *own* interfaces, otherwise things get crossed.

So, I'm a soft +1 for this, though I don't like the BC break. In this *one* situation, perhaps we should unset the aliases if the ODM is present. And in future situations (like event subscribers), we need to be more careful not to do too much with shared interfaces.

stof:
well, auto-configuration of subscribers also break if you define multiple ORM entity managers though

alcaeus:
@stof yes, in the sense that only the default entity manager can be typehinted as it's the only one being aliased to the entity manager interface. However, I can be sure I'm getting an entity manager. Autowiring any object manager should not be possible: you don't know what you're getting as it depends on the order the bundles are loaded.

kimhemsoe:
I accept this BC break now, better now then much later.

PR #748, opened 2017-12-07
Always register the Entity listener

eightyfive:
Regardless of doctrine/orm version, you _must_ register the Entity listener in the Entity file. I'm using `symfony/orm-pack: 1.0.4` which points to `doctrine/orm: ^2.4.5` and Listener does not work unless Entity listener is properly declared/registered in Entity class:
```
* @ORM\EntityListeners({"App\Entity\UserListener"})
```

### DIFF
--- a/Resources/doc/entity-listeners.rst
+++ b/Resources/doc/entity-listeners.rst
@@ -32,7 +32,7 @@ entity manager it should be registered with. Example:
             </services>
         </container>
 
-If you use a version of doctrine/orm < 2.5 you have to register the entity listener in your entity as well:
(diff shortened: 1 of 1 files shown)


mikeSimonson:
@eightyfive Can you check in your composer.lock the version of doctrine orm that you have installed.
Because symfony/orm-pack: 1.0.4 require the orm ^2.5.11.

There is likely another dependency that you have and that maintain you to <2.5

Thanks

eightyfive:
Hey,

That must have been the case, I deleted my project meanwhile.

Thanks anyway.

PR #749, opened 2017-12-13
Increase dbal version

silverbackdan:
There are many fixes and improvements in dbal 2.6 - naive PR to see if travis builds pass with this simple update.
https://github.com/doctrine/dbal/releases


silverbackdan:
It seems with the version updates the 7.1 build that is passing in the master branch still. I haven't checked in enough detail to be that confident but thought it might be better to raise this as a PR than just an issue. I came across this as I was using the latest MariaDB which wasn't synchronising the database schema.

alcaeus:
The changes are unnecessary: the original constraints also cover the constraints you’ve added.

silverbackdan:
Oh yes of course, sorry.

PR #751, opened 2017-12-17
Removing more headers

mikeSimonson:


sroze:
What's the reasoning behind it?

mikeSimonson:
https://github.com/doctrine/DoctrineModule/pull/617#issuecomment-356407504

Majkl578:
@sroze Also to quote the license itself:
> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

This requirement is fulfilled by the LICENSE file in the root which is part of Composer/ZIP distribution or GitHub forks. Should anyone copy any part of the library (i.e. single class) to their code, it's their responsibility to retain copyright, not the vendor's.
From the maintainer's point of view, it's unnecessary extra work to keep it in sync, plus it's 17 lines of boilerplate in each file ([~7700 lines in ORM](https://github.com/doctrine/doctrine2/pull/6600) for example).

sroze:
Fair enough, thanks for your answer 👍

PR #756, opened 2018-01-12
merge the shard config instead of replacing

oshanrube:
this will avoid causing the missing db driver exception


kimhemsoe:
Hi and thanks. Before we can do anything a test is needed covering the changes.

kimhemsoe:
Looks like a BC break for me. Marking as wont fix. Please write if you believe different :)

PR #764, opened 2018-01-22
Added env variables resolving to DoctrineExtension

snetilov:
Currently I have no way to work with doctrine config that depends on environmental variables as mergeExtensionConfigurationPass is registered before ResolveEnvPlaceholdersPass. This commit allows to work with doctrine config like this:

doctrine:
orm:
metadata_cache_driver:
type: "%env(DOCTRINE_METADATA_CACHE_TYPE)%"
host: "%env(DOCTRINE_METADATA_CACHE_HOST)%"
port: "%env(DOCTRINE_METADATA_CACHE_PORT)%"
query_cache_driver:
type: "%env(DOCTRINE_QUERY_CACHE_TYPE)%"
host: "%env(DOCTRINE_QUERY_CACHE_HOST)%"
port: "%env(DOCTRINE_QUERY_CACHE_PORT)%"
result_cache_driver:
type: "%env(DOCTRINE_RESULT_CACHE_TYPE)%"
host: "%env(DOCTRINE_RESULT_CACHE_HOST)%"
port: "%env(DOCTRINE_RESULT_CACHE_PORT)%"

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -59,6 +59,10 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $configuration = $this->getConfiguration($configs, $container);
         $config = $this->processConfiguration($configuration, $configs);
 
+        if (method_exists($container, 'resolveEnvPlaceholders')) {
+            $config = $container->resolveEnvPlaceholders($config, true);
+        }
+        
         $this->adapter->loadServicesConfiguration($container);
 
         if (!empty($config['dbal'])) {


kimhemsoe:
Thanks @snetilov 

We need a check for if the method "resolveEnvPlaceholders" exists and some tests.

Alternative you could use a parameters file to archive the same result.

snetilov:
There's a method_exists check in commit. I've tried to use parameters.yml. No success.

kimhemsoe:
Need a test for this and figure out why travis is failing.

kimhemsoe:
@snetilov Sorry that i missed the method_exists() think i was a bit to fast there :)

stof:
Well, using env variables for the host and port should work fine. But the type cannot be resolved using the environment variable at runtime, as it controls which class is used by the service (and so resolving it at runtime would require moving *all* the config processing at runtime). what is your use case for changing the type based on the environment ?

DependencyInjection/DoctrineExtension.php
```diff
@@ -59,6 +59,10 @@ public function load(array $configs, ContainerBuilder $container)
         $configuration = $this->getConfiguration($configs, $container);
         $config = $this->processConfiguration($configuration, $configs);
 
+        if (method_exists($container, 'resolveEnvPlaceholders')) {
+            $config = $container->resolveEnvPlaceholders($config, true);
```
stof: resolving env placeholders in the whole config is a no-go, as this would mean that DB credentials are not read at runtime anymore for instance (requiring a cache clearing after changing the env variable, while using `%env()%` is precisely meant to avoid that)

snetilov:
In dev environment I'm using 'array' as all three caches. But in production I would like to have an option to redefine cache type to memcached. Since all my application is running in Docker env variables seemed a good choise.
Now I'm using SYMFONY__ constants but they are deprecated now. And I have no other way to define cache type in runtime.

stof:
@snetilov Symfony has a prod and a dev environment, which can have different config files. Why not just relying on that ? your containers should already select the right Symfony environment for prod or dev anyway for things to work fine.

snetilov:
I'll try to use env variables only in 'type'. Maybe this will solve my issue.

stof:
@snetilov no, that's precisely the place where you should **not** use them.

snetilov:
Why not use them?

stof:
Have you read my previous comment ? The type must be known at compile-time. The main feature of `%env()%` is precisely that they are **not** resolved at compile-time, but at runtime, allowing to change the environment variable without forcing to clear the cache.

The memcached credentials are placed where using it is legitimate btw.

snetilov:
I understand my case when depending on environment ( I mean production servers, staging servers, testing servers and development servers. Not symfony environments!) I set different environmental variable to different values, in one case pointing cache to redis, in another to memcached, or even to consul. Clearing the cache and exiting maintenance period of application.
But I can't understand the ideology behind "The main feature of %env()% is precisely that they are not resolved at compile-time, but at runtime, allowing to change the environment variable without forcing to clear the cache.". Isn't this just a constraint that makes switching cache difficult in runtime? What stops me from clearing the cache after changing cache type?

snetilov:
By the way, this probles arises from the fact that doctrine takes all heavy decisions at container compile time and not in runtime by utilizing DI as an example.

stof:
> Isn't this just a constraint that makes switching cache difficult in runtime?

Switching cache type requires instantiating a different class, and so changing the definition of the service (unless we move all the config processing at runtime, but this requires making the bundle more complex, and might even have a performance impact).
And your proposal here is to force clearing the Symfony cache for changes in environment variables used **anywhere** in the DoctrineBundle configuration. This is not good, as it means you cannot swap the DB credentials easily anymore for instance.

> By the way, this probles arises from the fact that doctrine takes all heavy decisions at container compile time and not in runtime by utilizing DI as an example.

We do use DI. But DI does not mean that you do all processing of the config at runtime (the concept of DI does not even have a relation to runtime or compile-time, and some DIC don't have a compile-time at all).

> What stops me from clearing the cache after changing cache type?

When building optimized setups which should be able to start really fast, you would often do the Symfony cache-warmup at build time. And then, starting your new container already has the Symfony cache ready, meaning it starts in a few ms rather than in several seconds (the warmup of the Symfony cache takes time, precisely because we consider that doing things during warmup is better than doing them later when possible as `later` means someone is waiting for the HTTP response)

snetilov:
> And your proposal here is to force clearing the Symfony cache for changes in environment variables used anywhere in the DoctrineBundle configuration. This is not good, as it means you cannot swap the DB credentials easily anymore for instance.

This makes sens.

> And then, starting your new container already has the Symfony cache ready

My setup is so complex that it's easier for me to implement rolling update in docker swarm. But now I understand you. My proposal will make life harder for a lot of people and simplifies it only for some.

By the way, this setup works for me for now:
config_prod.yml:
`doctrine:
    orm:
        metadata_cache_driver:
            type: "memcached"
            host: "%env(MEMCACHED_HOST)%"
            port: "%env(MEMCACHED_PORT)%"
        query_cache_driver:
            type: "memcached"
            host: "%env(MEMCACHED_HOST)%"
            port: "%env(MEMCACHED_PORT)%"
        result_cache_driver:
            type: "memcached"
            host: "%env(MEMCACHED_HOST)%"
            port: "%env(MEMCACHED_PORT)%"`

kimhemsoe:
super, guessing we can close this then?

snetilov:
Yes. Thanks.

PR #767, opened 2018-01-26
Fixed command doctrine:database:drop --if-exists

Nijusan:
command still failed when the database does not exist altough the --if-exists option was called. problem with symfony4 seems to be the url key in the $params array.

adds unsetting url key of params array because otherwise dropping database with if-exists option still fails when getConnection($params) is called


Nijusan:
@kimhemsoe hi. i am not that familiar with the further process, can i somehow support more? or is everything fine up to this point?

kimhemsoe:
Thanks for asking.

We need a test covering the changes, to make sure it will not break again in the future. Unfortunately the commands are lacking testing. I would recommend you take a look at CreateDatabaseDoctrineTest to how it can be done. 

The tests are currently failing with the newest changes from symfony, but you can safely ignore that.

Nijusan:
@kimhemsoe hi kim, i've added a test class for the command. but i am not able to write a concrete test for the above problem. i've written one and outcommented it, because it needs a mysql connection.
the problem is, that this failure only happens with a mysql connection, or better, it does not fail with sqlite. it seems no problem with sqlite to connect to a database which does not exists, where else mysql gives the error that the database xy does not exist.

i am stuck at this point, i tried do mock the DriverManager to inspect the params array with a callback to check if the url key exists, because it shoudnt, but DriverManager is a final class.

any idea? maybe the commited class is a helping start for further testing...
maybe ur travis has a local mysql connection, then maybe this is still helpful?

Nijusan:
any update on this issue?

kimhemsoe:
@Nijusan I am a bit busy atm the moment at work. Doctrine is something i do in my sparetime, but i hope to have time soon to take a look at this and all the other test which is failing on master.

jwage:
@Nijusan What is the error that causes this to fail?

Nijusan:
@jwage the problem occurs if the database connection is configured via url like it is done in symfony 4.   then https://github.com/Nijusan/DoctrineBundle/blob/cde9878791d8b1076f08e36dec0afb8ac72ade83/Command/DropDatabaseDoctrineCommand.php#L90 fails with the same reason u already unset the dbname parameter.  mysql tries to connect and use that given database, but if it does not exist, it cannot connect. so it fails in this line and never reaches the possibility for --if-exists

jwage:
@Nijusan Forgive me, I don't understand. If all the connection information is in the `url` option and you unset it how is it able to connect without a database name when it gets here https://github.com/doctrine/DoctrineBundle/pull/767/files#diff-b46232fa79d53e95a9b5efea5b606a8dR90

Nijusan:
@jwage no problem. dont be sorry for asking: its a simple answer, it is not the only information. doctrine already resolved all information into the parameters array:

```
» bin/console d:d:d
array:11 [
  "driver" => "pdo_mysql"
  "charset" => "utf8mb4"
  "url" => "mysql://foo:bar@127.0.0.1:3306/test_db"
  "host" => "127.0.0.1"
  "port" => "3306"
  "user" => "foo"
  "password" => "bar"
  "driverOptions" => []
  "serverVersion" => "5.7"
  "defaultTableOptions" => []
  "dbname" => "test_db"
]
```

Tests/Command/DropDatabaseDoctrineTest.php
```diff
@@ -0,0 +1,130 @@
...
+            ->getMock();
+
+        $mockContainer->expects($this->any())
+            ->method('get')
+            ->with('doctrine')
+            ->willReturn($mockDoctrine);
+
+        return $mockContainer;
+    }
+}
```
jwage: Can you add a new line at the end of this file?

Tests/Command/DropDatabaseDoctrineTest.php
```diff
@@ -0,0 +1,130 @@
...
+
+        $commandTester = new CommandTester($command);
+        $commandTester->execute(
+            array_merge(array('command' => $command->getName(), '--force' => true))
+        );
+
+        $this->assertContains("Dropped database for connection named " . sys_get_temp_dir() . "/" . $dbName . "" , $commandTester->getDisplay());
+    }
+
+    /*
```
jwage: If this test is not needed, can you remove it instead of leaving it commented out?

Tests/Command/DropDatabaseDoctrineTest.php
```diff
@@ -0,0 +1,130 @@
...
+namespace Doctrine\Bundle\DoctrineBundle\Tests\Command;
+
+use Doctrine\Bundle\DoctrineBundle\Command\DropDatabaseDoctrineCommand;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Console\Application;
+use Symfony\Component\Console\Tester\CommandTester;
+
+class DropDatabaseDoctrineTest extends TestCase
+{
+
```
jwage: Remove this errant extra new line.

Tests/Command/DropDatabaseDoctrineTest.php
```diff
@@ -0,0 +1,130 @@
+<?php
+
```
jwage: Put only 1 new line between the opening `<?php` and the namespace.

jwage:
@Nijusan Ok, I understand now. I requested a few changes.

Nijusan:
@jwage thank u. i did the changes as mentioned

kimhemsoe:
@Nijusan Is it possible for you to fix the remaining issues and do a rebase? Thanks :)

Nijusan:
@kimhemsoe done :)
sry was kinda busy lately

kimhemsoe:
@Nijusan Perfect, can you fix the cs issues and then we are good to go :D

Nijusan:
@kimhemsoe it looks good now :)

mikeSimonson:
@Nijusan Thanks for the PR and the tests

PR #768, opened 2018-01-27
Added missing quoting of scalar values for yaml files

kimhemsoe:
FIxed a few deprecation errors regarding yaml and scalar values


kimhemsoe:
@alcaeus seems i was sleeping alot... this already in master. I messed this one up badly :D

PR #771, opened 2018-01-29
Build Stages

Majkl578:
Migrating Travis builds to Build Stages & sync with other Doctrine projects.

Same amount of green/red stages as before, but it'd be indeed better to fix CI in master beforehand.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: php
-
 sudo: false
+dist: trusty
 
 cache:
   directories:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -14,35 +14,55 @@ php:
...
     - php: 5.5
       env: deps=low SYMFONY_DEPRECATIONS_HELPER=weak
+      install:
+        - travis_retry composer update -n --prefer-lowest --prefer-stable --prefer-dist
+
     - php: 5.6
       env: SYMFONY_VERSION="2.8.*"
+
     - php: 7.0
       env: SYMFONY_VERSION="3.2.*"
```
stof: This should be `3.4` not `3.2`. 3.2 is not a LTS (and is EOLed)

Majkl578: I think there is 3.4 already in separate stage. I didn't touch these, feel free to send PR.

PR #772, opened 2018-01-29
Align CS with Doctrine CS 2.1

Majkl578:
- disable sniffs incompatible with PHP 5.5,
- disable `array` type specification where not needed,
- `doctrine/coding-standard` is not a direct dev-dependency in composer.json because of the PHP 5.5 nonsense - thus only installed on Travis.

Since there is no CS rule/stage currently, I was wondering whether it should follow Doctrine CS or Symfony CS. Since the bundle resides under `doctrine`, I thought Doctrine CS is more appropriate, but feel free to discuss or reject in favor of Symfony CS. 😎 

CS stage separately: https://travis-ci.org/doctrine/DoctrineBundle/jobs/334826059

### DIFF
--- a/Controller/ProfilerController.php
+++ b/Controller/ProfilerController.php
@@ -1,6 +1,5 @@
 <?php
 
-
 namespace Doctrine\Bundle\DoctrineBundle\Controller;
 
 use Doctrine\DBAL\Connection;
@@ -11,8 +10,6 @@ use Symfony\Component\HttpFoundation\Response;
 
 /**
  * ProfilerController.
- *
- * @author Christophe Coevoet <stof@notk.org>
  */
 class ProfilerController implements ContainerAwareInterface
 {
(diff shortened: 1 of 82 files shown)


Controller/ProfilerController.php
```diff
@@ -32,54 +29,54 @@ public function setContainer(ContainerInterface $container = null)
...
+        /** @var \Doctrine\DBAL\Connection $connection */
         $connection = $this->container->get('doctrine')->getConnection($connectionName);
         try {
             if ($connection->getDatabasePlatform() instanceof SQLServerPlatform) {
                 $results = $this->explainSQLServerPlatform($connection, $query);
             } else {
                 $results = $this->explainOtherPlatform($connection, $query);
             }
-        } catch (\Exception $e) {
+        } catch (\Throwable $e) {
```
Majkl578: Oops, this one needs to be reverted, because 5.x. :/ Sniff is already excluded though.

PR #777, opened 2018-02-14
Fix getting service repository from factory with different em

eshishko:
There is the problem, when is multiple entity managers, and try to get repository with diferent from default:

`$this->getDoctrine()->getRepository(Customer::class, 'master')`

it will use the default one inside repository.

### DIFF
--- a/Repository/ContainerRepositoryFactory.php
+++ b/Repository/ContainerRepositoryFactory.php
@@ -53,6 +53,9 @@ final class ContainerRepositoryFactory implements RepositoryFactory
                 if (! $repository instanceof EntityRepository) {
                     throw new \RuntimeException(sprintf('The service "%s" must extend EntityRepository (or a base class, like ServiceEntityRepository).', $repositoryServiceId));
                 }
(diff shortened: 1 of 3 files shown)


stof:
Making the entity manager mutable in there will cause all sort of issues IMO.

If you have 2 entity managers, you must have 2 repository instances instead of mutating a single one based on which code requested it last (because some code may have requested it earlier and keep it for later usage).
So this does not make your code work with multiple entity manager (it will create lots of WTF bugs).

Thus, this code recreates the issue related to entity manager resetting: to support resetting, DoctrineBundle wraps the entity manager in a proxy. When defining the repository service, it injects the proxy itself. But the entity manager received as argument here is the inner one (as the entity manager itself is passing `$this` to the factory), meaning you get a direct reference to the inner manager. Defining a repository as service worked fine with this implementation **only** because it was not injecting the entity manager inside the repository on its own.

stof:
Note that having the same entity class managed by multiple entity managers is not even something which is meant to be supported in DoctrineBundle. It can create lots of weird issues.

PR #779, opened 2018-02-15
Import mapping for bundle less

barell:
This pull request is to support bundle-less Symfony application.
There are two new options added:
- --bundle-less - if enabled, the command will use bundle argument as a path to the application source code
- --root-namespace - if provided it will use namespace when creating new classes (default for bundle-less will be App so don't have to provide it by default)

Example usage if we want to import annotation style entities:
`php bin/console doctrine:mapping:import --bundle-less src annotation`
This will create entity classes in src/Entity/ directory using App\Entity namespace.

`php bin/console doctrine:mapping:import --bundle-less --root-namespace=Foo src annotation`
This will create entity classes in src/Entity/ directory using Foo\Entity namespace.

The default behaviour of this command is not affected when not using --bundle-less option.

### DIFF
--- a/Command/ImportMappingDoctrineCommand.php
+++ b/Command/ImportMappingDoctrineCommand.php
@@ -1,33 +1,18 @@
 <?php
 
-/*
- * This file is part of the Doctrine Bundle
- *
- * The code was originally distributed inside the Symfony framework.
- *
- * (c) Fabien Potencier <fabien@symfony.com>
- * (c) Doctrine Project, Benjamin Eberlei <kontakt@beberlei.de>
- *
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
- */
-
 namespace Doctrine\Bundle\DoctrineBundle\Command;
 
-use Symfony\Component\Console\Input\InputArgument;
-use Symfony\Component\Console\Input\InputOption;
-use Symfony\Component\Console\Input\InputInterface;
-use Symfony\Component\Console\Output\OutputInterface;
 use Doctrine\ORM\Mapping\Driver\DatabaseDriver;
+use Doctrine\ORM\Tools\Console\MetadataFilter;
 use Doctrine\ORM\Tools\DisconnectedClassMetadataFactory;
 use Doctrine\ORM\Tools\Export\ClassMetadataExporter;
-use Doctrine\ORM\Tools\Console\MetadataFilter;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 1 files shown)


Command/ImportMappingDoctrineCommand.php
```diff
@@ -59,9 +61,19 @@ protected function configure()
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $bundle = $this->getApplication()->getKernel()->getBundle($input->getArgument('bundle'));
+        if ($input->getOption('bundle-less')) {
```
Ocramius: Much like with previous patches attempting to understand where files are located, I don't really understand why the user can't simply  perform the import and then copy the files over.

barell: Can you import if there is no single bundle in your app?

Ocramius: There's surely at least DoctrineBundle, since you installed it? :P

barell: Obviously you are right, I didn't think about this kind of workaround - but still it's not perfect :)

Command/ImportMappingDoctrineCommand.php
```diff
@@ -29,6 +29,8 @@ protected function configure()
             ->addOption('shard', null, InputOption::VALUE_REQUIRED, 'The shard connection to use for this command')
             ->addOption('filter', null, InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'A string pattern used to match entities that should be mapped.')
             ->addOption('force', null, InputOption::VALUE_NONE, 'Force to overwrite existing mapping files.')
+            ->addOption('bundle-less', null, InputOption::VALUE_NONE, 'Bundle argument will be used as destination path')
+            ->addOption('root-namespace', null, InputOption::VALUE_REQUIRED, 'Specify root namespace')
```
weaverryan: Could we instead not have the bundle-less option? What I mean is: allow the bundle option to be a bundle *or* a namespace. First, see if it’s a bundle. If it’s not, assume it’s the namespace. Then, why not just add an optional “path” argument, but throw an exception if it’s not passed if a namespace is used: require it in that case.

The final command would look like:

bin/console doctrine:mapping:import App annotation --path=src/

I think that’s simple enough. It’s a rare command: so just let them choose the namespace and path. For XML, the path we would recommend in the docs is config/doctrine

jwage: @weaverryan I like your suggestion. @barell what do you think?

Majkl578:
Closing in favor of  #791 which is nearly complete and built on top of this.

PR #790, opened 2018-03-23
Deprecating doctrine:generate:entities and pointing to MakerBundle

weaverryan:
Since the Doctrine devs aren't interested in maintaining the generation commands and since we implemented them in MakerBundle, I hope this will be a nice win for everyone: let's point them elsewhere and avoid issues being created related to Symfony 4 support :).

### DIFF
--- a/Command/GenerateEntitiesDoctrineCommand.php
+++ b/Command/GenerateEntitiesDoctrineCommand.php
@@ -73,7 +73,11 @@ EOT
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         trigger_error('The doctrine:generate:entity command has been deprecated.', E_USER_DEPRECATED);
-        $output->writeln(' <comment>NOTE:</comment> The <info>doctrine:generate:entities</info> command has been deprecated. Use <info>make:entity --regenerate</info> from MakerBundle instead.');
+        $output->writeln([
+            ' <comment>NOTE:</comment> The <info>doctrine:generate:entities</info> command has been deprecated.',
+            '       To read more about the differences between anemic and rich models go here <info>http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/tutorials/getting-started.html#adding-behavior-to-entities</info>.',
+            '       If you wish to generate your entities, use <info>make:entity --regenerate</info> from MakerBundle instead.',
+        ]);
 
         $manager = new DisconnectedMetadataFactory($this->getContainer()->get('doctrine'));
 
--- a/Command/GenerateEntitiesDoctrineCommand.php
+++ b/Command/GenerateEntitiesDoctrineCommand.php
@@ -72,6 +72,9 @@ EOT
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        trigger_error('The doctrine:generate:entity command has been deprecated.', E_USER_DEPRECATED);
+        $output->writeln(' <comment>NOTE:</comment> The <info>doctrine:generate:entities</info> command has been deprecated. Use <info>make:entity --regenerate</info> from MakerBundle instead.');
+
         $manager = new DisconnectedMetadataFactory($this->getContainer()->get('doctrine'));
 
         try {


Command/GenerateEntitiesDoctrineCommand.php
```diff
@@ -72,6 +72,9 @@ class is supposed to extend which. You have to adjust the entity
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        trigger_error('The doctrine:generate:entity command has been deprecated.', E_USER_DEPRECATED);
+        $output->writeln(' <comment>NOTE:</comment> The <info>doctrine:generate:entities</info> command has been deprecated. Use <info>make:entity --regenerate</info> from MakerBundle instead.');
```
Ocramius: I'd likely suggest pointing out the bad practice and point it at the docs instead. We (doctrine project) do not endorse MakerBundle in any way, as it just perpetuates an already too widespread misuse of mappings.

Nyholm:
I like the idea of deprecating this command. How about writing a message like: 

>  <comment>NOTE:</comment> The <info>doctrine:generate:entities</info> command has been deprecated. Read more at https://symfony.com/doc/current/bundles/DoctrineBundle/index.html

weaverryan:
I like @Nyholm's idea. DoctrineBundle is the intersection of Symfony & Doctrine. This would allow us to give your recommendation on the matter, but also point people to the Symfony resource that *is* available, should they choose it.

Majkl578:
There is currently no recommendation regarding this and [Symfony explicitly rejected to do so](https://github.com/symfony/symfony-docs/issues/8893). So at current state it's a no from me.
In meanwhile, [Doctrine documentation has been updated](https://github.com/doctrine/doctrine2/pull/6902).

Nyholm:
Thank you for your input. I have not seen these PRs before. 

So if I understand you correctly you are fine with this PR **if** there is a section in the docs that explains why this command was deprecated. The symfony docs will also say something like: "If you still really want this generate feature, then look at the MakerBundle".

Ocramius:
> Read more at https://symfony.com/doc/current/bundles/DoctrineBundle/index.html

Great idea if the docs are coming from this repo, but since we disagree on practices (as @Majkl578 pointed out): no.

Link http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/tutorials/getting-started.html#adding-behavior-to-entities instead.

Command/GenerateEntitiesDoctrineCommand.php
jwage: I'd suggest changing the language to something like this:

    $output->writeln(' <comment>NOTE:</comment> The <info>doctrine:generate:entities</info> command has been deprecated. To read more about the differences between anemic and rich models go here <link to our docs>. If you wish to generate your entities, use <info>make:entity --regenerate</info> from MakerBundle instead.');

weaverryan:
Hi guys!

I've just updated the message to talk about anemic models and link to the new section on the Doctrine documentation. THEN, at the end, I mention the new command if the user still wants it. It looks like this:

screen shot 2018-03-24 at 7 55 22 am

If you guys want to educate on anemic versus rich models, this is much better than just having a command that doesn't work :).

kimhemsoe:
+1 from here.

Will merge if no objections.

Is there any symfony documentation about using dto's + doctrine with forms vs letting using form + doctrine? If not i believe it could be a nice addition.

weaverryan:
@kimhemsoe 

> Is there any symfony documentation about using dto's + doctrine with forms vs letting using form + doctrine? If not i believe it could be a nice addition.

Not really. And we talked internally, and I think this is something we need to improve - both showing forms with model objects (or at least mentioning that more specifically) and also showing data mappers, which already has a PR (https://github.com/symfony/symfony-docs/pull/6900) we just need to "un-stuck" it :).

kimhemsoe:
@weaverryan  Thanks

jwage:
Thanks @weaverryan and @Ocramius!

PR #791, opened 2018-03-23
Fix doctrine:mapping:import to work in Symfony 4 (no bundle)

### LINKED ISSUE #736: Symfony 4 doctrine:mapping:import missing bundle
Reported originally by @pedroresende in https://github.com/symfony/symfony/issues/25013

-----

| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Symfony version  | 4.0.0-beta4

If I ty to run the command

```
php bin/console doctrine:mapping:import
```

I receive the following message 

```
Not enough arguments (missing: "bundle").
```

However in Symfony 4 there is no bundle.

weaverryan:
Hi guys!

This replaces and builds off of #779, which fixes #736.

It's VERY simple: the first argument can now be a bundle (like before) or a namespace:

A) If it's a bundle, everything works 100% like before.
B) If it's *not* a bundle, we assume it's the namespace. Then, there's no magic: we require them to explicitly pass us the path where the files should be generated. No fancy guess work.

Unfortunately, this is very difficult to test. So, I've purposefully made the diff very light and consistent. I've tested this in a REAL app, by copying the first 3 examples in the `setHelp()` to test everything. It all works perfectly.

Thanks!

### DIFF
--- a/Command/ImportMappingDoctrineCommand.php
+++ b/Command/ImportMappingDoctrineCommand.php
@@ -1,33 +1,18 @@
 <?php
 
-/*
- * This file is part of the Doctrine Bundle
- *
- * The code was originally distributed inside the Symfony framework.
- *
- * (c) Fabien Potencier <fabien@symfony.com>
- * (c) Doctrine Project, Benjamin Eberlei <kontakt@beberlei.de>
- *
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
- */
-
 namespace Doctrine\Bundle\DoctrineBundle\Command;
 
-use Symfony\Component\Console\Input\InputArgument;
-use Symfony\Component\Console\Input\InputOption;
-use Symfony\Component\Console\Input\InputInterface;
-use Symfony\Component\Console\Output\OutputInterface;
 use Doctrine\ORM\Mapping\Driver\DatabaseDriver;
+use Doctrine\ORM\Tools\Console\MetadataFilter;
 use Doctrine\ORM\Tools\DisconnectedClassMetadataFactory;
 use Doctrine\ORM\Tools\Export\ClassMetadataExporter;
-use Doctrine\ORM\Tools\Console\MetadataFilter;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
 
 /**
  * Import Doctrine ORM metadata mapping information from an existing database.
- *
- * @author Fabien Potencier <fabien@symfony.com>
- * @author Jonathan H. Wage <jonwage@gmail.com>
  */
 class ImportMappingDoctrineCommand extends DoctrineCommand
 {
@@ -68,7 +53,7 @@ Use the <info>--force</info> option, if you want to override existing mapping fi
 
 <info>php %command.full_name% "MyCustomBundle" xml --force</info>
 EOT
-            );
+        );
     }
 
     /**
@@ -89,21 +74,21 @@ EOT
             $namespace = $input->getOption('root-namespace');
         }
 
-        $type = $input->getArgument('mapping-type') ? $input->getArgument('mapping-type') : 'xml';
-        if ('annotation' === $type) {
+        $type     = $input->getArgument('mapping-type') ? $input->getArgument('mapping-type') : 'xml';
+        if ($type === 'annotation') {
             $destPath .= '/Entity';
         } else {
             $destPath .= '/Resources/config/doctrine';
         }
-        if ('yaml' === $type) {
+        if ($type === 'yaml') {
             $type = 'yml';
         }
 
-        $cme = new ClassMetadataExporter();
+        $cme      = new ClassMetadataExporter();
         $exporter = $cme->getExporter($type);
         $exporter->setOverwriteExistingFiles($input->getOption('force'));
 
-        if ('annotation' === $type) {
+        if ($type === 'annotation') {
             $entityGenerator = $this->getEntityGenerator();
             $exporter->setEntityGenerator($entityGenerator);
         }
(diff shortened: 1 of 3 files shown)


weaverryan:
@kimhemsoe I see you added the "needs test" label. I'm not sure this is feasible - and only one of the commands in this bundle has a test at *all*. What do you think?

Ocramius:
Should still be tested though: the fact that other classes do not have any
tests is not a precondition to keep doing that

weaverryan:
Fair enough :). I'm looking into it, and will hopefully have something soonish.

kimhemsoe:
@weaverryan You are ok with me merging https://github.com/doctrine/DoctrineBundle/pull/793 now before this? It will most likely mess your PR a bit up

weaverryan:
@kimhemsoe Yea, no problem - I can clean it up on my end :)

Command/ImportMappingDoctrineCommand.php
```diff
@@ -23,17 +23,25 @@ protected function configure()
...
             ->setDescription('Imports mapping information from an existing database')
             ->setHelp(<<<EOT
 The <info>%command.name%</info> command imports mapping information
 from an existing database:
 
+Generate annotation mappings into the src/ directory using App as the namespace:
+<info>php %command.full_name% App\\\Entity annotation --path=src/Entity</info>
+
+Generate annotation mappings into the config/doctrine/ directory using App as the namespace:
+<info>php %command.full_name% App\\\Entity xml --path=config/doctrine</info>
```
weaverryan: The non-bundle version is totally explicit: pass the exact namespace and the exact path - no guessing or assuming (like we do with bundles). It may make the command slightly less user-friendly, but is more flexible, and it's a rare command to use anyways.

composer.json
```diff
@@ -25,7 +25,7 @@
     ],
     "require": {
         "php": "^5.5.9|^7.0",
-        "symfony/framework-bundle": "~2.7|~3.0|~4.0",
+        "symfony/framework-bundle": "^2.7.22|~3.0|~4.0",
```
weaverryan: Needed so we get symfony/symfony#20442, else the commands in the test aren't registered correctly with deps=low

weaverryan:
@kimhemsoe I've rebased for #793 and added a new test for `ImportMappingDoctrineCommandTest`, which tests the old use-cases and the new one. 

And, tests are green 🎆 and my last commit should fix all the CS stuff with my changes for the new 4.0 :).

Tests/Command/ImportMappingDoctrineCommandTest.php
```diff
@@ -0,0 +1,146 @@
...
+
+class ImportMappingDoctrineCommandTest extends TestCase
+{
+    /** @var Kernel|null */
+    private $kernel;
+
+    /** @var CommandTester|null */
+    private $commandTester;
+
+    public function setup()
```
Majkl578: protected

Tests/Command/ImportMappingDoctrineCommandTest.php
```diff
@@ -0,0 +1,146 @@
...
+            ->get('doctrine')
+            ->getConnection();
+        $connection->executeQuery('CREATE TABLE product (id integer primary key, name varchar(20), hint text)');
+
+        $application         = new Application($this->kernel);
+        $command             = $application->find('doctrine:mapping:import');
+        $this->commandTester = new CommandTester($command);
+    }
+
+    public function tearDown()
```
Majkl578: protected

Command/ImportMappingDoctrineCommand.php
```diff
@@ -59,19 +67,31 @@ protected function configure()
...
-        $bundle = $this->getApplication()->getKernel()->getBundle($input->getArgument('bundle'));
-
-        $destPath = $bundle->getPath();
-        $type     = $input->getArgument('mapping-type') ? $input->getArgument('mapping-type') : 'xml';
-        if ($type === 'annotation') {
-            $destPath .= '/Entity';
-        } else {
-            $destPath .= '/Resources/config/doctrine';
-        }
+        $type = $input->getArgument('mapping-type') ? $input->getArgument('mapping-type') : 'xml';
```
Majkl578: This is odd. It should either use hasArgument() or be simply `?:`.

Command/ImportMappingDoctrineCommand.php
```diff
@@ -59,19 +67,31 @@ protected function configure()
...
         if ($type === 'yaml') {
             $type = 'yml';
         }
 
+        $bundles = $this->getContainer()->getParameter('kernel.bundles');
+        if (isset($bundles[$input->getArgument('name')])) {
```
Majkl578: Better have multiple `$input->getArgument('name')` calls extracted into a variable.

Command/ImportMappingDoctrineCommand.php
```diff
@@ -59,19 +67,31 @@ protected function configure()
...
+            $bundle    = $this->getApplication()->getKernel()->getBundle($input->getArgument('name'));
+            $namespace = $bundle->getNamespace() . '\\Entity';
```
Majkl578: No need to escape.

Tests/Command/ImportMappingDoctrineCommandTest.php
```diff
@@ -0,0 +1,146 @@
...
+        ]);
+
+        $expectedMetadataPath = sys_get_temp_dir() . '/import_mapping_bundle/Entity/Product.php';
+        $this->assertFileExists($expectedMetadataPath);
+        $this->assertContains('namespace Doctrine\Bundle\DoctrineBundle\Tests\Command\Entity;', file_get_contents($expectedMetadataPath), 'File contains correct namespace');
+    }
+
+    /**
+     * @expectedException \InvalidArgumentException
+     * @expectedExceptionMessageRegExp /The \-\-path option is required/
```
Majkl578: Escaping - is superfluous.

weaverryan:
Comments addressed! Tests green! Woo!

jwage:
:+1: @weaverryan can you squash now and then I think we can merge.

kimhemsoe:
@weaverryan Thanks

weaverryan:
Thanks guys! Sorry I didn't get the squashing done :)

kimhemsoe:
@weaverryan My fault, i was a bit fast on the trigger.

PR #792, opened 2018-03-23
Fixing ContainerTest where we try to fetch private services

weaverryan:
Some of the tests on master are failing. They're not a big: in `ContainerTest`, we're fetching every service we created and checking its type. That's a very pragmatic test, but doesn't work when services become private. For this one test, I've made all services public. This is a fairly common tactic in tests for Symfony bundles. And, it should make the tests happy again :).

Cheers!

### DIFF
--- a/Tests/TestCase.php
+++ b/Tests/TestCase.php
@@ -6,6 +6,7 @@ use Doctrine\Bundle\DoctrineBundle\DependencyInjection\DoctrineExtension;
 use Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection\TestType;
 use Doctrine\Common\Annotations\AnnotationReader;
 use PHPUnit\Framework\TestCase as BaseTestCase;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\Compiler\ResolveChildDefinitionsPass;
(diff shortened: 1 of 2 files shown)


Tests/TestCase.php
```diff
@@ -65,12 +66,32 @@ public function createYamlBundleTestContainer()
...
+            }
+        }
+
+        foreach ($container->getAliases() as $id => $alias) {
+            if (false !== strpos($id, 'doctrine')) {
+                $alias->setPublic(true);
+            }
+        }
+    }
+}
```
jwage: Can you add a new line at the end of this file?

jwage:
Are the tests still failing for other reasons now?

kimhemsoe:
My plan for fixing them was removing the failing test and add new tests which were no longer tested. 

I don't mind a quick fix for now.

weaverryan:
Tests and CS should be happy now. I think there is one invalid CS failure still, but all good otherwise.

Majkl578:
Yeah the catch error is a BC break caused by CodeSniffer upstream, it's not yet been handled here.

weaverryan:
@Majkl578 Ah, thanks - I was wondering. That's now the *only* violation, so this should be ready.

kimhemsoe:
@weaverryan Thanks

PR #793, opened 2018-03-25
Apply Doctrine CS v4

carusogabriel:
I just couldn't add `composer.lock` :(

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ jobs:
       env: CODING_STANDARDS
       php: 7.1
       install:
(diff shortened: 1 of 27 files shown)


kimhemsoe:
@carusogabriel Thanks

Majkl578:
> I just couldn't add composer.lock :(

Maybe, in this case, we can explicitly put those into require with specific versions. Like:
```yml
 travis_retry composer require -n --prefer-dist --dev doctrine/coding-standard:4.0.0 squizlabs/php_codesniffer:3.2.4 slevomat/coding-standard:4.5.2
```

PR #796, opened 2018-03-29
If the database config comes from a URL, ensure that the database nam…

rtfjr86:
…e is not set.

### DIFF
--- a/Command/DropDatabaseDoctrineCommand.php
+++ b/Command/DropDatabaseDoctrineCommand.php
@@ -96,7 +96,7 @@ EOT
         // If the database config comes from a URL,
         // ensure that the database name is not set.
         if (isset($params['url'])) {
-            $params['url'] = str_replace('/'.basename($params['url']), '', $params['url']);
+            $params['url'] = str_replace('/' . basename($params['url']), '', $params['url']);
         }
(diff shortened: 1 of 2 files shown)


Ocramius:
@rtfjr86 please add a test scenario checking what happens if the given values collide

Command/DropDatabaseDoctrineCommand.php
```diff
@@ -92,6 +92,13 @@ protected function execute(InputInterface $input, OutputInterface $output)
         // Reopen connection without database name set
         // as some vendors do not allow dropping the database connected to.
         $connection->close();
+
+        // If the database config comes from a URL,
+        // ensure that the database name is not set.
+        if (isset($params['url'])) {
+            $params['url'] = str_replace('/'.basename($params['url']), '', $params['url']);
```
Ocramius: Should probably throw an exception instead

Ocramius:
Wouldn't mock it at all, but I see that the Travis setup doesn't have a running server. @kimhemsoe any suggestions here?

kimhemsoe:
Can you do the test with sqlite? Noticed that Ryan have been doing that in some of the other tests.

PR #797, opened 2018-04-02
Fix the coding style

mikeSimonson:


DataCollector/DoctrineDataCollector.php
```diff
@@ -24,7 +24,7 @@ class DoctrineDataCollector extends BaseCollector
     /** @var int|null */
     private $invalidEntityCount;
 
-    /** @var [] */
+    /** @var []string */
```
Majkl578: `string[]`

mikeSimonson: @Majkl578 Thanks

PR #798, opened 2018-04-05
Update CreateDatabaseDoctrineCommand.php

themizzi:
Setting the path dynamically, which is useful to switch between in memory and path based SQLite dbs, with an env var requires setting the key `path` on the params. Isset returns true in this case even if the path is empty. `!empty` is a better check with this in mind.

### DIFF
--- a/Command/CreateDatabaseDoctrineCommand.php
+++ b/Command/CreateDatabaseDoctrineCommand.php
@@ -72,7 +72,7 @@ EOT
             }
         }
 
-        $hasPath = isset($params['path']);
+        $hasPath = !empty($params['path']);
         $name    = $hasPath ? $params['path'] : (isset($params['dbname']) ? $params['dbname'] : false);
         if (! $name) {
             throw new \InvalidArgumentException("Connection does not contain a 'path' or 'dbname' parameter and cannot be dropped.");


Command/CreateDatabaseDoctrineCommand.php
```diff
@@ -72,7 +72,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
             }
         }
 
-        $hasPath = isset($params['path']);
+        $hasPath = !empty($params['path']);
```
kimhemsoe: Could the same not be archived with "isset($params['path']) && $params['path'] !== ''"  ?

That way we can avoid empty() :)

alcaeus:
After taking another look, we've decided not to merge this. An empty path is very much valid in this context, which is why we use `isset` to check this. The correct solution would be to not pass an empty string, which can be an issue when dealing with environment variables.

Luckily, Symfony 4.3 will ship with a `default:` env var processor which allows you to fall back to a default value from a parameter, or receive `null` instead of an empty string. If you don't want to wait for Symfony 4.3 or can't upgrade when it is released, you can create an env var processor that does what Symfony 4.3 will ship.

PR #801, opened 2018-04-16
Use EntityManagerInterface

Basster:
Let ManagerConfigurator use EntityManagerInterface instead of EntityManager, so you can use custom EntityManager implementations by extending EntityManagerDecorator.   

BC-Break: that raises the `doctrine/orm` requirement from `~2.3` to `~2.4`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "doctrine/doctrine-cache-bundle": "~1.2"
     },
     "require-dev": {
-        "doctrine/orm": "~2.3",
+        "doctrine/orm": "~2.4",
         "symfony/yaml": "~2.7|~3.0|~4.0",
         "symfony/validator": "~2.7|~3.0|~4.0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -34,7 +34,7 @@
         "doctrine/doctrine-cache-bundle": "~1.2"
     },
     "require-dev": {
-        "doctrine/orm": "~2.3",
+        "doctrine/orm": "~2.4",
```
Ocramius: @kimhemsoe we don't support `2.4` since ages: can we bump this to `^2.6.2` ASAP?

Basster: 2.4 was the first release I've found having `EntityManagerInterface` and `EntityManagerDecorator`. The rest is on you...

kimhemsoe: @Ocramius It was on my todo list for 1.9, maybe bump it to 2.5.x first?

kimhemsoe:
Thanks @Basster

PR #802, opened 2018-04-19
Fix explaining queries

### LINKED ISSUE #757: [Profiler] Explain is broken

stof:
Recent Symfony versions are wrapping the params in a VarDumper Data when collecting the profiler info. This ensures that they get unwrapped properly when trying to explain the query.

Fixes #757

### DIFF
--- a/Controller/ProfilerController.php
+++ b/Controller/ProfilerController.php
@@ -8,6 +8,7 @@ use Symfony\Component\DependencyInjection\ContainerAwareInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Profiler\Profiler;
+use Symfony\Component\VarDumper\Cloner\Data;
 
 /**
  * ProfilerController.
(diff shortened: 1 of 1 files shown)


stof:
Recent means at least 3.4+. I don't remember about older 3.x versions (and I don't have any project using these older unmaintained 3.x versions to test)

Controller/ProfilerController.php
```diff
@@ -77,14 +78,27 @@ private function explainSQLServerPlatform(Connection $connection, $query)
         } else {
             $sql = 'SET SHOWPLAN_TEXT ON; GO; SET NOEXEC ON; ' . $query['sql'] . '; SET NOEXEC OFF; GO; SET SHOWPLAN_TEXT OFF;';
         }
-        $stmt = $connection->executeQuery($sql, $query['params'], $query['types']);
+
+        $params = $query['params'];
+
+        if ($params instanceof Data) {
```
mikeSimonson: Can you create a function with an explicit name and call it in both place ?

kimhemsoe:
@stof Is it possible to add a test for it?

stof:
The proper way to test this in a way making the test strong a regarding changes in the bridge collector would involve building a fake app running some DB query, and having a functional test running a request on this fake app, then opening the profiler, and then trying to explain the query in it (by loading the URL of the partial). This would involve a bunch of work, and I won't have time to work on this during the next few days (and I cannot guarantee it for the next few weeks, even though the French holidays coming in May may give me time for it, but in conflict with other OSS stuff). Is it required to build such test covering the profiler explaining feature for the bug fix to be merged ?

kimhemsoe:
I will merge in now then. I very much would a test for this if you or someone else find the time to todo it :)

kimhemsoe:
it is a long running bug, so lets fix it for now and then we can clean up the mess later...

@stof Thanks

hvt:
Would be great if a release could be made that includes this PR.

PR #806, opened 2018-04-19
Add 'doctrine.event_subscriber' tag for subscribers automatically

voronkovich:


kimhemsoe:
I would say no for the same reasons that we want to remove ObjectManager and etc in the next Major release.

@alcaeus Do you agree?

alcaeus:
Correct: other mappers use the very same interface for their event subscribers, so for anyone using two mappers (e.g. ORM and MongoDB ODM), this would cause problems because an ODM event subscriber would be automatically registered as event subscriber in the ORM. If we want to provide autoconfiguration for event subscribers, we need mapper-specific interfaces.

voronkovich:
@alcaeus, Thanks for the explanation!

PR #807, opened 2018-04-24
Bump branch alias

XWB:


composer.json
```diff
@@ -59,7 +59,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.8.x-dev"
+            "dev-master": "1.9.x-dev"
```
mikeSimonson: Shouldn't that be 1.10 now that we have a 1.9.1 ?

alcaeus: When we have a support branch for 1.9, `dev-master` should be 1.10.

stof: Do we actually need to maintain multiple branches in parallel ? Can't we just have master, and then deciding whether the next release is 1.9.2 or 1.10 based on what it contains (according to semver) ? Maintaining multiple branches creates more work.

alcaeus: Absolutely - would you still want to bump branch-alias in that case? I'm not sure if it should be done preemptively or only after changes requiring a new minor release have been pushed to the branch.

stof: I would keep it to 1.9.x then, until we have minor-level changes

alcaeus:
Thanks @XWB!

PR #817, opened 2018-05-11
[Messenger] Wire the transaction middleware factory when component is enabled

ogizanagi:
Relates to https://github.com/symfony/symfony/pull/27128

How should we process for the CI?

### DIFF
--- a/DependencyInjection/Compiler/MessengerPass.php
+++ b/DependencyInjection/Compiler/MessengerPass.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\Messenger\MessageBusInterface;
+
+/**
+ * Class for Symfony Messenger component integrations
+ */
+class MessengerPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        // Remove wired services if the Messenger component actually isn't enabled:
+        if (!$container->hasAlias('message_bus') || !is_subclass_of($container->findDefinition('message_bus')->getClass(), MessageBusInterface::class)) {
+            $container->removeDefinition('doctrine.orm.messenger.middleware_factory.transaction');
+            $container->removeDefinition('messenger.middleware.doctrine_transaction_middleware');
+        }
+    }
+}
--- a/DependencyInjection/Compiler/MessengerPass.php
+++ b/DependencyInjection/Compiler/MessengerPass.php
@@ -16,10 +16,12 @@ class MessengerPass implements CompilerPassInterface
      */
     public function process(ContainerBuilder $container)
     {
-        // Remove wired services if the Messenger component actually isn't enabled:
-        if (!$container->hasAlias('message_bus') || !is_subclass_of($container->findDefinition('message_bus')->getClass(), MessageBusInterface::class)) {
-            $container->removeDefinition('doctrine.orm.messenger.middleware_factory.transaction');
-            $container->removeDefinition('messenger.middleware.doctrine_transaction_middleware');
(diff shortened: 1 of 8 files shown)


composer.json
```diff
@@ -61,5 +62,6 @@
         "branch-alias": {
             "dev-master": "1.9.x-dev"
         }
-    }
+    },
+    "minimum-stability": "dev"
```
sroze: Should that really be added?

ogizanagi: That would be required to get the beta2 until released, but I don't think it'll stay here. That was just the way to install it locally for me.
Anyway, I'm waiting some instructions/plans/some help to deal with the CI as we're even unlikely to keep the messenger component in `require-dev` due to the sf 2.7 support in this bundle.

DependencyInjection/Compiler/MessengerPass.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+class MessengerPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        // Remove wired services if the Messenger component actually isn't enabled:
+        if (!$container->hasAlias('message_bus') || !is_subclass_of($container->findDefinition('message_bus')->getClass(), MessageBusInterface::class, true)) {
```
yceruto: The 3rd argument of `is_subclass_of()` is already `true` by default, right?

ogizanagi: `true` :)

PHP API consistency (see [`is_a`](http://php.net/manual/fr/function.is-a.php)) tricked me!

ogizanagi:
Should be ready. [New tests on the new travis build](https://travis-ci.org/doctrine/DoctrineBundle/jobs/378660126#L649-L667) should not be skipped anymore once https://github.com/symfony/symfony/pull/27128 is merged.

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -676,6 +677,31 @@ public function testAnnotationsBundleMappingDetectionWithVendorNamespace()
         $this->assertEquals('Fixtures\Bundles\Vendor\AnnotationsBundle\Entity', $calls[0][1][1]);
     }
 
+    public function testMessengerIntegration()
+    {
+        if (! interface_exists(MessageBusInterface::class)) {
```
Majkl578: I'd argue that it should be a dev-dependency, otherwise it's untested.

ogizanagi: Oops, I though `symfony/symfony` was used on CI for specific versions. So I expected it to be available in the new build.
The issue with adding it into the `dev-dependencies` (see first commit) is that it'll conflict with lowest dependencies tested for this packages due to the component requirements. So it must be explicitly added on a specific build. WDYT?

Majkl578: You can override it for now using `^4.1@beta`, the `@beta` part would be dropped once 4.1 is stable. But yes, it wouldn't allow SF pre-3.4. Since there is only one (or two) CI job to test SF 2.x, I think it'd be better to drop this dependency only in such job.
(Another argument for dropping 2.x soon.)

ogizanagi:
And it's now green. Also tested on a real app, works perfectly fine :)

weaverryan:
Can you open a docs PR on Symfony too?

ogizanagi:
@weaverryan : Done in https://github.com/symfony/symfony-docs/pull/9774

ogizanagi:
Symfony 4.1 is released. Could we move forward with this one now? :)

(Note: Code styles issues are not related to these changes)

ogizanagi:
Hi there 👋 Any update?
Symfony 4.1 is released since a month now, it would be awesome for end-users to be able to use the transaction middleware out-of-the-box thanks to this :)

sroze:
Anybody? :)

Nyholm:
Friendly ping

ogizanagi:
@Nyholm : As mentioned previously, these CS issues aren't actually related to this PR :)

kimhemsoe:
@ogizanagi Thanks

PR #825, opened 2018-06-08
Better error message when missing entity mapping

Tobion:
Otherwise you just get an error like

> php.CRITICAL: Call to a member function getClassMetadata() on null

@weaverryan


Repository/ServiceEntityRepository.php
```diff
@@ -28,6 +28,13 @@ public function __construct(ManagerRegistry $registry, $entityClass)
     {
         $manager = $registry->getManagerForClass($entityClass);
 
+        if (null === $manager) {
+            throw new \LogicException(sprintf(
```
jwage: Can you add a use statement for `LogicException`?

Also, is it possible to add a unit test which covers this scenario?

Repository/ServiceEntityRepository.php
```diff
@@ -28,6 +28,13 @@ public function __construct(ManagerRegistry $registry, $entityClass)
...
+                'Could not find the entity manager for class "%s". Did you forget to configure doctrine where to find the entity mapping?',
```
weaverryan: How about:

... Did you forget -> Check your Doctrine configuration to make sure it is configured to load this entity’s metadata.

Tobion: done

Repository/ServiceEntityRepository.php
Tobion: both done

Repository/ServiceEntityRepository.php
```diff
@@ -4,6 +4,7 @@
 
 use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\ORM\EntityRepository;
+use \LogicException;
```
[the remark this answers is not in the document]
Ocramius: This is possibly something we could fix in CS /cc @carusogabriel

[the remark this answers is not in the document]
carusogabriel: @ocramius We have a rule to forbid that. We just need to merge https://github.com/doctrine/DoctrineBundle/pull/855 :)

PR #832, opened 2018-07-03
Inject the EntityManager instead of the factory in DoctrineExtractor

dunglas:
Follows symfony/symfony#27829

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "symfony/dependency-injection": "~2.7|~3.0|~4.0",
         "doctrine/dbal": "^2.5.12",
         "jdorn/sql-formatter": "^1.2.16",
-        "symfony/doctrine-bridge": "^4.2",
+        "symfony/doctrine-bridge": "~2.7|~3.0|~4.0",
         "doctrine/doctrine-cache-bundle": "~1.2"
     },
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -30,7 +30,7 @@
         "symfony/dependency-injection": "~2.7|~3.0|~4.0",
         "doctrine/dbal": "^2.5.12",
         "jdorn/sql-formatter": "^1.2.16",
-        "symfony/doctrine-bridge": "~2.7|~3.0|~4.0",
+        "symfony/doctrine-bridge": "^4.2",
```
dunglas: We'll be able to use feature detection before merging.

stof: aren't we already able to do it ?

dunglas: Not sure if there is a new class yet.

kimhemsoe:
Dont know the details, but i am guessing this should be merged?

composer.json
dunglas: done

dunglas:
Ready for a last review. It should be merged before the release of 4.2, because without this patch DoctrineBundle triggers a deprecation.

kimhemsoe:
@dunglas Thanks

PR #835, opened 2018-07-17
Add missing default_dbname configuration option

Deamon:
Hi all,

In doctrine/dbal v2.6, was introduced a parameter default_dbname for postgre driver.
The bundle do not have the configuration option availaible so here is a PR.

I'm new to contributing on doctrine so I'm open to any suggestions.


jwage:
@kimhemsoe thoughts?

kimhemsoe:
@Deamon Thanks :)

PR #839, opened 2018-08-16
Complete example

ThomasLandauer:
Question:  Is `src/EventListener` the recommended location for an entity listener?

This integrates the findings of @piotrekkr at https://github.com/symfony/symfony-docs/issues/5013  
Since all basics are now covered, I'd say you can close the issue.


kimhemsoe:
@ThomasLandauer Thanks

PR #849, opened 2018-09-19
Fix handling of lazy entity listeners

nicolas-grekas:
Discussed in https://github.com/symfony/symfony/issues/28304#issuecomment-422710505

When an entity listener is declared as lazy at the definition level (and not at the tag level), then it is registered using the `ContainerAwareEntityListenerResolver::register()`. This class in turn uses `get_class()` to create its internal map.

But `get_class()` on a lazy-proxy returns the generated class of the proxy - not the class of the listener implementation.

Makes sense?

### DIFF
--- a/DependencyInjection/Compiler/EntityListenerPass.php
+++ b/DependencyInjection/Compiler/EntityListenerPass.php
@@ -52,9 +52,9 @@ class EntityListenerPass implements CompilerPassInterface
                     $this->attachToListener($container, $name, $id, $attributes);
                 }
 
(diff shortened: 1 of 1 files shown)


kimhemsoe:
@nicolas-grekas Thanks

kimhemsoe:
@nicolas-grekas Emh.. whops.. was that supposed to be merged with master?

nicolas-grekas:
that's a bug fix, so should go on the lowest maintained branch I suppose?

PR #853, opened 2018-09-25
Fix deprecation for symfony/config 4.2+

### LINKED ISSUE #851: A tree builder without a root node is deprecated since Symfony 4.2 and will not be supported anymore in 5.0.
Do you want a modification like :
        if (method_exists(TreeBuilder::class, 'getRootNode')) {
            $treeBuilder = new TreeBuilder($rootName);
            $rootNode = $treeBuilder->getRootNode();
        } else {
            $treeBuilder = new TreeBuilder();
            $rootNode = $treeBuilder->Root($rootName);
        }
I can do the PR.

stof:
Closes #851


DependencyInjection/Configuration.php
```diff
@@ -33,8 +33,14 @@ public function __construct($debug)
      */
     public function getConfigTreeBuilder()
     {
-        $treeBuilder = new TreeBuilder();
-        $rootNode    = $treeBuilder->root('doctrine');
+        $treeBuilder = new TreeBuilder('doctrine');
```
Majkl578: Please use `use function` instead of `\`.

stof: not possible: ` "php": "^5.5.9|^7.0",` in composer.json

mazarini: Please, put the "new" statement in the "if" (with/without parms).
There is  other "new TreeBuilder()" in the file.

Majkl578: @stof Using `\fn` directly violates our coding standard so it's invalid. If, due to PHP 5.5 compatibily, `use function` cannot be used, the function should be kept in _fallback_ mode.
This breaks build: https://travis-ci.org/doctrine/DoctrineBundle/jobs/443601176#L649

PR #855, opened 2018-09-26
Apply Doctrine CS v5

carusogabriel:


.travis.yml
```diff
@@ -63,7 +63,7 @@ jobs:
       env: CODING_STANDARDS
       php: 7.1
       install:
-        - travis_retry composer require -n --prefer-dist --dev doctrine/coding-standard:^4.0
+        - travis_retry composer require -n --prefer-dist --dev doctrine/coding-standard:^5.0
```
carusogabriel: Is there some reason why we don't require it in the `composer.json` itself?

jwage: I think it is probably just old and can be updated to match other repos.

Majkl578: The reason is that DoctrineBundle sadly still supports dead PHP versions.

Ocramius:
@carusogabriel rebase, and then I'll just merge this :+1:

carusogabriel:
@Ocramius Should be RTM :+1:

Ocramius:
Hmm, still got PHP 5 on this thing, and it is failing. Can you please remove it, and bump the minimum version to 7.1?

Yes, it really is time.

carusogabriel:
@Ocramius I got a running build when I opened this PR: https://travis-ci.org/doctrine/DoctrineBundle/builds/433228829.

Maybe we should retry?

> Can you please remove it, and bump the minimum version to 7.1?

See @Majkl578's https://github.com/doctrine/DoctrineBundle/pull/855#issuecomment-434117003, as well, I'd go with keeping this PR only related to Coding Standard changes :)

Ocramius:
Right, then :ship:

DependencyInjection/Configuration.php
```diff
@@ -2,10 +2,12 @@
 
 namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection;
 
+use ReflectionClass;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\NodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
+use function method_exists;
```
nicolas-grekas: This line breaks the Symfony CI, fixed in #869

PR #858, opened 2018-10-01
Custom doctrine repositories as services per default

Deltachaos:
Injecting doctrine repositories as service instead of the entity manager is best practice in Symfony. On the other hand the service definitions to do this look the same for 99.9% percent of the cases:

```
    AppBundle\Entity\MyCustomRepository:
        class: AppBundle\Entity\MyCustomRepository
        factory: [ "@doctrine.orm.default_entity_manager", "getRepository" ]
        arguments:
          - AppBundle\Entity\MyCustom
```

This was ok for the time where we had no Symfony autowiring as the services had to be defined explicitly all the time. Now there is autowiring and the developers are used to not define they're services on they're own anymore if the want to stick to the default. So we should also support this in the DoctrineBundle.

This PR adds a CompilerPass that checks if custom repositories have already been defined explicitly and if not creates the basic definition automatically. With this you can autowire the repositories without any future action needed.

### DIFF
--- a/DependencyInjection/Compiler/FallbackServiceRepositoryCompilePass.php
+++ b/DependencyInjection/Compiler/FallbackServiceRepositoryCompilePass.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
+
+use Doctrine\Common\Persistence\Mapping\ClassMetadataFactory;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+
+class FallbackServiceRepositoryCompilePass implements CompilerPassInterface
+{
+    /**
+     * @param ContainerBuilder $container
+     *
+     * @throws \Exception
+     */
+    public function process(ContainerBuilder $container)
+    {
+        foreach (array_keys($container->getDefinitions()) as $id) {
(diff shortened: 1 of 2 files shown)


DependencyInjection/Compiler/FallbackServiceRepositoryCompilePass.php
```diff
@@ -0,0 +1,72 @@
...
+    {
+        foreach (array_keys($container->getDefinitions()) as $id) {
+            if (!preg_match('/^doctrine\.orm\.[^.]+_entity_manager$/', $id)) {
+                continue;
+            }
+
+            $this->processMetadata(
+                $container,
+                $id,
+                $container->get($id)->getMetadataFactory()
```
stof: Instantiating a service in a compiler pass is a total no-go. There is exactly **no** guarantee that this will work, as the container is not fully configured yet. And even if it works at some point in time in your project, any change to any service definition could break it. So doing that in an open-source bundle which has to work in any project is impossible.

stof:
The example you gave does not work when using `Registry->reset()` to reset your entity manager (for instance in a long-running process), as the repository created by the default Doctrine factory would be referencing the old instance of the entity manager, not the new one, and so your repository service instance would be broken.

When building the new system of repositories defined as a service, making it work fine for the resetting feature was a requirement I gave @weaverryan. The automatic registration of old-style repository services has already been rejected multiple times in the past in this bundle, precisely for this reason (we were even actively discouraging to use it, as we knew we would get bug reports later on when devs would face the weird bugs related to it)

kimhemsoe:
@Deltachaos Closing for now, if you to see this in the bundle, the requirements is in the comments :)

PR #868, opened 2018-10-29
[Messenger] Wire the Doctrine transport factory when component is enabled

vincenttouzet:
This is related to this PR : https://github.com/symfony/symfony/pull/29007

### DIFF
--- a/Resources/config/messenger.xml
+++ b/Resources/config/messenger.xml
@@ -11,5 +11,11 @@
         <service id="messenger.middleware.doctrine_transaction" class="Symfony\Bridge\Doctrine\Messenger\DoctrineTransactionMiddleware" abstract="true" public="false">
             <argument type="service" id="doctrine" />
         </service>
+
+        <service id="messenger.transport.doctrine.factory" class="Symfony\Component\Messenger\Transport\Doctrine\DoctrineTransportFactory">
+            <argument type="service" id="doctrine" />
+
+            <tag name="messenger.transport_factory" />
+        </service>
     </services>
 </container>


Resources/config/messenger.xml
```diff
@@ -11,5 +11,12 @@
         <service id="messenger.middleware.doctrine_transaction" class="Symfony\Bridge\Doctrine\Messenger\DoctrineTransactionMiddleware" abstract="true" public="false">
             <argument type="service" id="doctrine" />
         </service>
+
+        <service id="messenger.transport.doctrine.factory" class="Symfony\Component\Messenger\Transport\Doctrine\DoctrineTransportFactory">
+            <argument type="service" id="doctrine" />
+            <argument type="service" id="messenger.transport.serializer" on-invalid="ignore" />
```
sroze: This needs to be removed now.

greg0ire:
Why target master? Is there a BC-break?

Resources/config/messenger.xml
greg0ire: Indeed, there is no second argument: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Messenger/Transport/Doctrine/DoctrineTransportFactory.php#L29

sroze:
The class is only available in Symfony 4.3.

vincenttouzet:
@greg0ire No just a new feature for the next version in Symfony 4.3. But maybe I target the wrong branch ? Which one must be the target ?

greg0ire:
Ah maybe the doctrine bundle is special and new features should go on master. Note that branch 1.10.x does not seem to have an incompatibility with symfony 4.3, but I'm not really sure what the workflow is for this bundle, so disregard my comment for now.

alcaeus:
`master` is good here, but I'll take a closer look over the next few days 👍

weaverryan:
Here is the class it's wiring: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Messenger/Transport/Doctrine/DoctrineTransportFactory.php

If the user doesn't have symfony/messenger 4.3, then they won't have this class. Normally, I think that's not a problem - because as long as they don't try to use the feature, it won't case an error. But because this is tagged as a transport, wouldn't a Symfony 4.2 user get an error about a missing class if they tried to use any messenger transport? @vincenttouzet can you check that?

sroze:
Indeed, because it's tagged as a transport factory, it will be used... therefore this would fail for a Symfony 4.2 user. We need to add, in a compiler pass, that if the class does not exist, we remove this definition.

alcaeus:
Superseded by #944. Thanks @vincenttouzet!

PR #869, opened 2018-10-30
Fix PHP 5.6+only syntax that breaks the Symfony CI

nicolas-grekas:
Introduced in #855

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -7,7 +7,6 @@ use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\NodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
-use function method_exists;
 
 /**
  * This class contains the configuration information for the bundle


stof:
This change was indeed wrong due to the min PHP version supported by the bundle.

Majkl578:
The commit message is wrong - `use function` is not 7.1-only syntax, it exists as of 5.6.

@kimhemsoe Can we just drop <7.1 support in master? It's been source of mutliple issues and confusion (last time [here](https://github.com/doctrine/DoctrineBundle/pull/853#discussion_r220539226), [here](https://github.com/doctrine/DoctrineBundle/pull/855#discussion_r220397344) and [here](https://github.com/doctrine/DoctrineBundle/pull/855#issuecomment-434115106)), it's really time.

nicolas-grekas:
Commit message updated.

kimhemsoe:
@nicolas-grekas Thanks

nicolas-grekas:
@kimhemsoe thanks to you. We're back to green :)

Ocramius:
Folks, this change is invalid: the fact that symfony mocks `method_exists` within the DoctrineBundle namespaces is already a nightmare.

This needs to be fixed downstream, not here.

Ocramius:
Hmm, no, my comment was incorrect: this is because of PHP version support.

Sending a patch to finally put a tombstone on it, once and for all.

PR #873, opened 2018-11-04
Drop oldstable dependency version support: bump *all* dependencies to current stable release

Ocramius:
Ref: https://github.com/doctrine/DoctrineBundle/pull/869#issuecomment-435697515

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,37 +28,15 @@ script:
 
 jobs:
   include:
-    - php: 5.5
-      env: deps=low SYMFONY_DEPRECATIONS_HELPER=weak
-      install:
-        - travis_retry composer update -n --prefer-lowest --prefer-stable --prefer-dist
-
-    - php: 5.6
-      env: SYMFONY_VERSION="2.8.*"
-
-    - php: 7.0
-      env: SYMFONY_VERSION="3.2.*"
-
-    - php: 7.1
-      env: stability=RC SYMFONY_DEPRECATIONS_HELPER=weak SYMFONY_VERSION="3.4.*"
-      install:
-        - travis_retry composer update -n --prefer-dist
-
-    - php: 7.1
+    - php: 7.2
       env: stability=dev SYMFONY_DEPRECATIONS_HELPER=weak SYMFONY_VERSION="4.1.*"
       install:
         - composer require --dev "symfony/messenger" --no-update
         - travis_retry composer update -n --prefer-dist
 
-    - php: 7.2
-      env: stability=alpha
-      install:
-        - composer config minimum-stability alpha
-        - travis_retry composer update -n --prefer-dist
-
     - stage: Code Quality
       env: CODING_STANDARDS
-      php: 7.1
+      php: 7.2
       install:
         - travis_retry composer require -n --prefer-dist --dev doctrine/coding-standard:^5.0
       script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,8 @@ cache:
     - $HOME/.composer/cache
 
 php:
-  - 5.5
-  - 5.6
-  - 7.0
-  - 7.1
   - 7.2
+  - 7.3
   - nightly
 
 env:
--- a/composer.json
+++ b/composer.json
@@ -24,28 +24,25 @@
         }
     ],
     "require": {
-        "php": "^5.5.9|^7.0",
-        "symfony/framework-bundle": "^2.7.22|~3.0|~4.0",
-        "symfony/console": "~2.7|~3.0|~4.0",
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -24,28 +24,25 @@
...
     "require": {
-        "php": "^5.5.9|^7.0",
-        "symfony/framework-bundle": "^2.7.22|~3.0|~4.0",
-        "symfony/console": "~2.7|~3.0|~4.0",
-        "symfony/dependency-injection": "~2.7|~3.0|~4.0",
-        "doctrine/dbal": "^2.5.12",
-        "jdorn/sql-formatter": "^1.2.16",
-        "symfony/doctrine-bridge": "~2.7|~3.0|~4.0",
-        "doctrine/doctrine-cache-bundle": "~1.2"
+        "php": "^7.2",
```
Majkl578: Since SF 4 requires PHP 7.1, this bundle should IMO do the same, at least until 5.0 GA.

.travis.yml
```diff
@@ -31,37 +28,15 @@ script:
...
-      install:
-        - composer config minimum-stability alpha
-        - travis_retry composer update -n --prefer-dist
-
     - stage: Code Quality
       env: CODING_STANDARDS
-      php: 7.1
+      php: 7.2
       install:
         - travis_retry composer require -n --prefer-dist --dev doctrine/coding-standard:^5.0
```
Majkl578: We can now move this to composer.json directly. 👍

composer.json
jwage: Yes, this needs to stay 7.1 for now.

Ocramius: No need to follow symfony's dependency versioning: the bundle can define its own dependency versions (and it should)

.travis.yml
Ocramius: :+1: but would need the usual `composer.lock` ballet too

composer.json
jwage: This is a bundle for the Symfony eco-system and it is built for Symfony and should follow the Symfony release cycle they follow and document publicly.

.travis.yml
Majkl578: Let's keep it to separate PR, we can enable some 7.1+ sniffs that were excluded and also use `use function`. 😉 cc @carusogabriel if you're interested

composer.json
Ocramius: This is something we discussed, discussed and discussed again ad-nauseam.

If the package still has to live under `doctrine/*`, then keep it under our maintenance with our constraints and our ability to maintain it.

If it is part of the Symfony LTS cycle yaddayaddawhatever, then let's push it under `symfony/*` and call it a day.

Ocramius: Also (and extremely important), this targets `master`, not the current release. Dependency bumps are indeed for the next minor version.

jwage: I think it is best if we follow the release roadmap that Symfony publicly documents.

If you merge this to master and release it as 1.10, then when Symfony releases a 4.x version with new features and we add them in 1.11, DoctrineBundle will be incompatible with Symfony and the 7.1 users won't be able to upgrade Symfony.

DoctrineBundle is a part of the Symfony eco-system and we have people from the Symfony team (@weaverryan @nicolas-grekas @javiereguiluz) that actively help maintain this bundle so I don't think it is a problem for us to play nice with the Symfony eco-system.

Ocramius: > and the 7.1 users won't be able to upgrade Symfony.

This is not a problem

jwage: Symfony eco-system still supports 7.1 and this bundle is built for that eco-system.

Ocramius: Again: not a problem. You get the new features when you get to upgrade.

As mentioned above, this is a bundle, and if it should be maintained outside the doctrine organisation then I don't understand why it is in here in first place. If the release goes hand in hand with symfony, make it part of the symfony subtree split..

Ocramius:
>  a bug fix version

That isn't `master`

composer.json
jwage: Because that is just where it ended up several years ago. Where it is in github is irrelevant imo. Github has features to handle this, we have outside collaborators who have the ability to maintain this bundle (and they do). You can control notifications per repository, etc.

I can't think of any reason why it can't be moved but I also don't think it is a very big deal for the repository to live here and follow the documented release lifecycle of Symfony.

Ocramius: >  I also don't think it is a very big deal for the repository to live here and follow the documented release lifecycle of Symfony.

It is a big deal for my inbox, from what I can see.

Alright then, closing as `won't fix` here, will simply mute the repo.

Ocramius:
Closing as "won't fix" due to discussions above. Dependency constraints will stay as they are.

PR #882, opened 2018-11-30
Make travis green

chalasr:
Fixing the build on Symfony 4.1 and adding a build for 4.2.


chalasr:
Given the maintenance period of the Symfony 2.8 branch ended recently, I suggest to drop support for Symfony 2.x in the next minor.

alcaeus:
@chalasr I agree, I would also suggest dropping support for PHP < 7.1 as we've started doing in the other Doctrine bundles.

kimhemsoe:
@chalasr Thanks

PR #884, opened 2018-12-03
Better feature detection for DoctrineExtractor

### LINKED ISSUE #883: Symfony 4.2 compatibility
Error:
```
TypeError: Argument 1 passed to Symfony\Bridge\Doctrine\PropertyInfo\DoctrineExtractor::__construct() must implement interface Doctrine\Common\Persistence\Mapping\ClassMetadataFactory, instance of Doctrine\ORM\EntityManager given
```

Possible cause:
https://github.com/doctrine/DoctrineBundle/blob/82d2c63cd09acbde2332f55d9aa7b28aefe4983d/DependencyInjection/DoctrineExtension.php#L801-L803

dunglas:
Fixes #883

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -25,7 +25,6 @@ use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\DependencyInjection\ServiceLocator;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Messenger\MessageBusInterface;
-use Symfony\Component\PropertyInfo\PropertyInitializableExtractorInterface;
 
 /**
  * DoctrineExtension is an extension for the Doctrine DBAL and ORM library.
@@ -798,9 +797,8 @@ class DoctrineExtension extends AbstractDoctrineExtension
     private function loadPropertyInfoExtractor($entityManagerName, ContainerBuilder $container)
     {
         $propertyExtractorDefinition = $container->register(sprintf('doctrine.orm.%s_entity_manager.property_info_extractor', $entityManagerName), DoctrineExtractor::class);
-        if (interface_exists(PropertyInitializableExtractorInterface::class)) {
-            $argumentId = sprintf('doctrine.orm.%s_entity_manager', $entityManagerName);
-        } else {
+        $constructorParameterType = (new \ReflectionClass(DoctrineExtractor::class))->getConstructor()->getParameters()[0]->getType();
+        if (null !== $constructorParameterType && $constructorParameterType->getName() === ClassMetadataFactory::class) {
             $argumentId = sprintf('doctrine.orm.%s_entity_manager.metadata_factory', $entityManagerName);
 
             $metadataFactoryDefinition = $container->register($argumentId, ClassMetadataFactory::class);
@@ -809,6 +807,8 @@ class DoctrineExtension extends AbstractDoctrineExtension
                 'getMetadataFactory',
             ]);
             $metadataFactoryDefinition->setPublic(false);
+        } else {
+            $argumentId = sprintf('doctrine.orm.%s_entity_manager', $entityManagerName);
         }
 
         $propertyExtractorDefinition->addArgument(new Reference($argumentId));


kimhemsoe:
@dunglas Alot of tests failing, you have an what is going on?

alcaeus:
@kimhemsoe I believe #882 should fix them.

dunglas:
@alcaeus not totally, I use PHP features that aren't supported by PHP5. I need to take a deeper look

alcaeus:
@dunglas Could you give this another look please?

greg0ire:
@dunglas php 5 seems to have been dropped since you started working on this, let me restart the build :)

https://github.com/doctrine/DoctrineBundle/blob/ad278efe707c57e24aa54cd9d500f7f211c2c816/composer.json#L27

greg0ire:
Still failing, I'm afraid you will have to rebase.

greg0ire:
That would mean red builds on that branch, or dropping php 5 there

xabbuh:
I changed the target branch which makes the build slightly look better. But it seems that the fix still needs some work (see the failing tests).

👍 for me to release 1.10.1 without this PR (and release 1.10.2 once this one is ready).

alcaeus:
> for me to release 1.10.1 without this PR (and release 1.10.2 once this one is ready).

Yeah, thought of that as well. Will do 👍

alcaeus:
@dunglas can you take another look at this please?

dunglas:
I'll try to take a look asap, unfortunately it's not on top of my todo list right now tbh

alcaeus:
That's fine - I'm just not aware of what the issue is and how to fix it. Maybe @doctrine/team-symfony-integration can advise how to proceed here?

silverbackdan:
Does this not just need a rebase now as the travis config has been updated to reflect the new php minimum requirements? I might be wrong but I think I'll try creating a PR with the new travis config and this update and see if it works to help get this resolved.

silverbackdan:
> Does this not just need a rebase now as the travis config has been updated to reflect the new php minimum requirements? I might be wrong but I think I'll try creating a PR with the new travis config and this update and see if it works to help get this resolved.

Sorry, missed the conversation regarding 1.10.x as well.

alcaeus:
After taking a look, I can see why it's not at the top of your todo list. We can't reliably detect whether the Doctrine Bridge is installed in version 4.2 (or newer) by checking features. The check introduced here won't work as there is no parameter type in the `DoctrineExtractor` constructor, so the check will always fail.

The only check I came up with was checking the type hint of the `DoctrineChoiceLoader` argument (to distinguish Symfony 3 from Symfony 4) and checking for the existence of the `DoctrineTransactionMiddlewareFactory` class, but that matches both Symfony 4.0 and 4.2. There are no other obvious changes in the bridge that would help during version detection.

We could use a package like `ocramius/package-versions` to go version-based on this check, but all other checks check the code so I didn't just want to go that route.

greg0ire:
How about using this: https://github.com/symfony/doctrine-bridge/compare/4.1...v4.2.0#diff-52766cedbcf59f925f7316471f14d51fR29 ?

Something like `is_a(DoctrineType::class, ResetInterface::class, true)` could work, apparently: https://3v4l.org/s0Rp9

alcaeus:
Good catch, that would work. I'll test that later today unless somebody else wants to try it earlier.

greg0ire:
I'm working on a PR, but will not have time to test it :P

greg0ire:
See https://github.com/doctrine/DoctrineBundle/pull/917

xabbuh:
see #918

greg0ire:
How about an approach where we detect this property: https://github.com/symfony/doctrine-bridge/compare/4.1...v4.2.0#diff-0ddb11f82d9ee9635a016dbb98a03de9R31 ? It's supported in php 5: https://secure.php.net/manual/en/reflectionclass.hasproperty.php

xabbuh:
@greg0ire At least for me the linked diff is empty.

greg0ire:
BTW, I don't really understand your point, @alcaeus :

>The check introduced here won't work as there is no parameter type in the DoctrineExtractor constructor, so the check will always fail.

There is a parameter type… in v 4.1 , and the code in this PR checks that.

greg0ire:
I reused #917 for this

alcaeus:
Closing in favour of #917.

PR #896, opened 2018-12-11
Unset parameters for sharded databases as well

### LINKED ISSUE #895: Create command fails for sharded databases
Creating a new database for the global database of a sharded configuration fails, because the command tries to connect to the database that doesn't exist yet.

The cause of this is that the dbname, url and path parameters are not unset for sharded databases, while they are unset for 'normal' connections.

cinamo:
Creating a new database for the global database of a sharded configuration fails, because the command tries to connect to the database that doesn't exist yet.

The cause of this is that the dbname, url and path parameters are not unset for sharded databases, while they are unset for 'normal' connections.

### DIFF
--- a/Command/CreateDatabaseDoctrineCommand.php
+++ b/Command/CreateDatabaseDoctrineCommand.php
@@ -63,13 +63,13 @@ EOT
             $shards = $params['shards'];
             // Default select global
             $params = array_merge($params, $params['global']);
-            unset($params['global']['dbname']);
+            unset($params['global']['dbname'], $params['global']['path'], $params['global']['url']);
             if ($input->getOption('shard')) {
                 foreach ($shards as $i => $shard) {
                     if ($shard['id'] === (int) $input->getOption('shard')) {
                         // Select sharded database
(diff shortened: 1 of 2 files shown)


cinamo:
@kimhemsoe The issue is, in the provided test there's no `url` or `path` parameter set, which is why that test succeeds. The default Symfony 4 .env configuration uses a `url` parameter for the global connection, though...

Would you recommend I add a `url` parameter in the existing test? Or do you have another suggestion maybe? Thanks!

kimhemsoe:
Yes, a test showing what you are trying to fix would be nice.

cinamo:
@kimhemsoe Updated a test that fails on the non-fixed version

stof:
Do we need the same kind of fix for MasterSlaveConnection too ?

xabbuh:
@alcaeus Should this be backported to the `1.10.x` branch as it is part of that milestone?

alcaeus:
@xabbuh Yes - we've had a minor hiccup with the history. I'll fix it today before releasing.

PR #897, opened 2018-12-13
Format code samples properly

greg0ire:


greg0ire:
I suspect these issues come from suggestions I made through the Github UI. The diff of db73431 looks sane though :confused:

PR #898, opened 2018-12-13
[Documentation][Entity Listeners] Mention lazy tag attribute

dmaicher:
The `lazy` tag attribute was not mentioned in the docs.

This might be superfluous in case https://github.com/doctrine/DoctrineBundle/issues/899 can/will be done

### DIFF
--- a/Resources/doc/entity-listeners.rst
+++ b/Resources/doc/entity-listeners.rst
@@ -20,6 +20,7 @@ entity manager it should be registered with. Example:
                         event: preUpdate
                         entity: App\Entity\User
                         entity_manager: custom
+
     .. code-block:: xml
 
         <?xml version="1.0" ?>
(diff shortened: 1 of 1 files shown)


Resources/doc/entity-listeners.rst
```diff
@@ -15,11 +15,7 @@ entity manager it should be registered with. Example:
                 class: \UserListener
                 tags:
                     - { name: doctrine.orm.entity_listener }
--
-                        name: doctrine.orm.entity_listener
-                        event: preUpdate
-                        entity: App\Entity\User
-                        entity_manager: custom
+
```
greg0ire: This looks wrong, and seems related to #897

Resources/doc/entity-listeners.rst
```diff
@@ -60,3 +57,34 @@ If you use a version of doctrine/orm < 2.5 you have to register the entity liste
 See also
 https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/events.html#entity-listeners
 for more info on entity listeners and the resolver required by Symfony.
+
+
+Lazy Entity Listeners
+---------------------
+
+You can use the `lazy` attribute on the tag to make sure the listener services are only instantiated when they are actually used.
```
greg0ire: You need to double quote keywords, and I think linkes should be wrapped to 72 chars.

```suggestion
You can use the ``lazy`` attribute on the tag to make sure the listener
services are only instantiated when they are actually used.
```

Resources/doc/entity-listeners.rst
dmaicher: Indeed :+1:

SenseException:
There's currently only a master branch documentation of DoctrineBundle on the Doctrine website, but should we put this addition to earlier supported versions of the docs?

dmaicher:
Also related is this issue I created: https://github.com/doctrine/DoctrineBundle/issues/899

If we decide to implement that change then we might not even need to merge this PR.

Resources/doc/entity-listeners.rst
alcaeus: #897 was merged. Please rebase to remove the change here.

alcaeus:
> This might be superfluous in case #899 can/will be done.

Since 1.10 will be around a bit longer, it would still make sense to merge this.

Resources/doc/entity-listeners.rst
dmaicher: Rebased. Keeping the new empty line is ok? Improves readability when editing it imho

dmaicher:
This could be merged? :blush:

alcaeus:
Ah, this should go into 1.10.x. Could you please change the base of the PR to 1.10.x and make sure no commits from master are in the PR? Thanks!

greg0ire:
so `git rebase --onto 1.10.x master patch-1` (never tried it, but I saw that in the manual the other day)

dmaicher:
Done :wink:

alcaeus:
Okay. Build failure seems to be because travis-ci didn't choose the right branch when running the file. I've restarted the failing step to see if it will work this time.

alcaeus:
Thanks @dmaicher!

Issue #900, opened 2018-12-20
MasterSlaveConnection and configuration of default_table_options 


Spawnrad:
Hi,

When i use MasterSlave configuration on doctrine the parameter default_table_options is not taken.
But once i remove the slave configuration it work and my tables is create with my charset configuration.

Here my doctrine configuration : 

```
doctrine:
    dbal:
        driver:   pdo_mysql
        host:     "%env(DATABASE_MASTER_HOST)%"
        dbname:   "%env(DATABASE_DBNAME)%"
        user:     "%env(DATABASE_USER)%"
        password: "%env(DATABASE_PASSWORD)%"
        slaves:
            slave1:
                host:     "%env(DATABASE_SLAVE_HOST)%"
                dbname:   "%env(DATABASE_DBNAME)%"
                user:     "%env(DATABASE_USER)%"
                password: "%env(DATABASE_PASSWORD)%"
        charset: utf8mb4
        default_table_options:
            charset: utf8mb4
            collate: utf8mb4_unicode_ci  
```
Its looks like that the defaultTableOptions have no data when i use MasterSlave configuration : 

```
array:6 [
  "driver" => "pdo_mysql"
  "slaves" => array:1 [
    "slave1" => array:6 [
      "host" => "db:3306"
      "dbname" => "databasename"
      "user" => "root"
      "password" => "root"
      "port" => null
      "driver" => "pdo_mysql"
    ]
  ]
  "driverOptions" => []
  "master" => array:8 [
    "host" => "db:3306"
    "dbname" => "databasename"
    "user" => "root"
    "password" => "root"
    "charset" => "utf8mb4"
    "port" => null
    "defaultTableOptions" => array:2 [
      "charset" => "utf8mb4"
      "collate" => "utf8mb4_unicode_ci"
    ]
    "driver" => "pdo_mysql"
  ]
  "wrapperClass" => "Doctrine\DBAL\Connections\MasterSlaveConnection"
  "defaultTableOptions" => []
]
```

Once i remove the slave for example i have that result and it work very well : 

```
array:9 [
  "driver" => "pdo_mysql"
  "host" => "db:3306"
  "dbname" => "databasename"
  "user" => "root"
  "password" => "root"
  "charset" => "utf8mb4"
  "port" => null
  "driverOptions" => []
  "defaultTableOptions" => array:2 [
    "charset" => "utf8mb4"
    "collate" => "utf8mb4_unicode_ci"
  ]
]
```

Thank you

[...]

PR #901, opened 2018-12-21
Keep defaultTableOptions when MasterSlaveConnection

Spawnrad:
Hi, i have open an issue to explain : https://github.com/doctrine/DoctrineBundle/issues/900

I figure out that the defaultTableOptions is not keep when we use MasterSlaveConnection.


alcaeus:
Closing in favour of #915 since we need to fix this in 1.10. I also added tests. Thanks for the fix @Spawnrad!

PR #903, opened 2019-01-07
Remove superfluous if statement in catch clause

davidmaes:


alcaeus:
@davidmaes can you fix this in 1.10.x please? We'll down merge it to master afterwards. Thanks!

PR #906, opened 2019-01-22
Update examples in entity listeners documentation

YetiCGN:
The code example is weirdly misleading because the same class is attached to the `preUpdate` event twice. Furthermore, the YAML example is wrong and inconsistent with the XML example. This PR fixes that.


YetiCGN:
@alcaeus Sure thing!

alcaeus:
Thanks @YetiCGN!

PR #907, opened 2019-01-23
Fix configuration to prevent an error

bnd170:
Delete first tag that will override inmediatly after


alcaeus:
This should be fixed in the 1.10 documentation as well. @bnd170 could you change the base and rebase your branch? Thanks!

alcaeus:
Closing as both #906 and #910 improve this documentation.

PR #909, opened 2019-01-24
Autowire event subscriber

Baachi:
Hi!

We just discovered, that the `EventSubscriber` Interface is not enabled for autowiring.


alcaeus:
This PR cannot be merged: the `Doctrine\Common\EventSubscriber` interface is shared among multiple object mappers; registering every event subscriber with an ORM EntityManager causes problems. (see also https://github.com/doctrine/DoctrineBundle/pull/806).

If you want to change this, you'll have to introduce an ORM specific marker interface for this purpose.

Baachi:
Ah got it! Thanks @alcaeus for the explanation.

stof:
note that you can still use that code in your own project (put these 2 lines in your `Kernel::configureContainer` method) in case your project uses only the ORM but not other Doctrine O*M projects.

nicolas-grekas:
Or better: use "instance of" under "_defaults" in services.yaml

PR #912, opened 2019-01-30
Updated the value of path option to match the official recipe

javiereguiluz:
This problem was spotted by @nicolas-grekas. He suggested to use instead the same value as in the official recipe (https://github.com/symfony/recipes/blob/master/doctrine/doctrine-bundle/1.6/manifest.json). I fully agree with that and I hope you do too. Thanks.

### DIFF
--- a/Resources/doc/configuration.rst
+++ b/Resources/doc/configuration.rst
@@ -957,8 +957,8 @@ can configure. The following block shows all possible configuration keys:
                 driver_class:             MyNamespace\MyDriverImpl
                 options:
                     foo: bar
-                path:                     "%kernel.data_dir%/data.sqlite" # SQLite specific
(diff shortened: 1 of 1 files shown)


greg0ire:
Also `git log  -S data_dir` in symfony/symfony only yields https://github.com/symfony/symfony/commit/9e4aebf3c95ac3f0d3f8dbbe14f002e5edf97cc4 and https://github.com/symfony/symfony/commit/1a45bb6d6358387f1a0d3d00792f63fe1069f5a3 , as if that directory never was a thing… weird

javiereguiluz:
@alcaeus I may have rebased this successfully ... but please double check it. Thanks.

alcaeus:
Thanks @javiereguiluz!

PR #913, opened 2019-01-30
Added some missing contents to an XML config sample

javiereguiluz:
In https://symfony.com/doc/current/bundles/DoctrineBundle/configuration.html#doctrine-dbal-configuration you can see that the XML config doesn't show highlighted contents. This PR may fix it.

### DIFF
--- a/Resources/doc/configuration.rst
+++ b/Resources/doc/configuration.rst
@@ -1003,9 +1003,42 @@ can configure. The following block shows all possible configuration keys:
                 http://symfony.com/schema/dic/doctrine/doctrine-1.0.xsd">
 
             <doctrine:config>
+                <!--
+                    SQLite specific options:
+                    - path
+                    - memory
+                -->
+                <!--
+                    Oracle specific options:
+                    - pooled (SERVER=POOLED)
+                    - service (SERVICE_NAME instead of SID)
+                    - servicename (SERVICE_NAME)
+                -->
+                <!--
+                    PostgreSQL specific options:
+                    - default_dbname (default_dbname)
+                    - sslmode (LIBPQ-CONNECT-SSLMODE)
+                    - sslrootcert (LIBPQ-CONNECT-SSLROOTCERT)
+                    - sslcert (LIBPQ-CONNECT-SSLCERT)
+                    - sslkey (LIBPQ-CONNECT-SSLKEY)
+                    - sslcrl (LIBPQ-CONNECT-SSLCRL)
+                -->
+                <!--
+                    IBM DB2 specific options:
+                    - protocol (PROTOCOL)
+                -->
+                <!--
+                    SQL Anywhere specific options:
+                    - server (ServerName)
+                -->
+                <!--
+                    oci8 specific options:
+                    - sessionMode (session_mode)
+                -->
+
                 <doctrine:dbal
                     name="default"
-                    url="mysql://user:secret@localhost:1234/otherdatabase" <!-- this would override the values below -->
+                    url="mysql://user:secret@localhost:1234/otherdatabase"
                     dbname="database"
                     host="localhost"
                     port="1234"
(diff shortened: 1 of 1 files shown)


Resources/doc/configuration.rst
```diff
@@ -993,52 +993,59 @@ can configure. The following block shows all possible configuration keys:
...
+            xmlns:doctrine="http://symfony.com/schema/dic/doctrine"
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
+                http://symfony.com/schema/dic/services/services-1.0.xsd
+                http://symfony.com/schema/dic/doctrine
+                http://symfony.com/schema/dic/doctrine/doctrine-1.0.xsd">
+
+            <doctrine:config>
+                <doctrine:dbal
+                    name="default"
+                    url="mysql://user:secret@localhost:1234/otherdatabase" <!-- this would override the values below -->
```
xabbuh: these inline comments are the actual issue, we cannot use them in XML inside tags

javiereguiluz: I see. It makes sense now. Any idea to solve this? Thanks.

SenseException: Good catch. I thought this was about inconsistent XML, but the Doctrine docs are having issues too with them: https://www.doctrine-project.org/projects/doctrine-bundle/en/latest/configuration.html#doctrine-dbal-configuration

xabbuh: In the Symfony docs we usually add comments above the tag where we want to comment some attributes. That might be a bit too much here But maybe not all of the comments are really necessary.

javiereguiluz: So, what should we do? Keep the comments and lose highlighting (not a big deal) or move the comments at the beginning (it may look crowded, as Christian said).

greg0ire: Maybe provide several examples with comments at the beginning? That would help grouping things that go together (url vs non-url, RDMBS-specific settings, etc…)

alcaeus: I extracted the comments. @xabbuh can you please take another look?

Resources/doc/configuration.rst
```diff
@@ -1003,33 +1003,48 @@ can configure. The following block shows all possible configuration keys:
...
+                <!-- server: SQL Anywhere specific (ServerName) -->
+                <!-- service: Oracle specific (SERVICE_NAME instead of SID) -->
+                <!-- servicename: Oracle specific (SERVICE_NAME) -->
+                <!-- sessionMode: oci8 driver specific (session_mode) -->
+                <!-- default_dbname: PostgreSQL specific (default_dbname) -->
+                <!-- sslmode: PostgreSQL specific (LIBPQ-CONNECT-SSLMODE) -->
+                <!-- sslrootcert: PostgreSQL specific (LIBPQ-CONNECT-SSLROOTCERT) -->
+                <!-- sslcert: PostgreSQL specific (LIBPQ-CONNECT-SSLCERT) -->
+                <!-- sslkey: PostgreSQL specific (LIBPQ-CONNECT-SSLKEY) -->
+                <!-- sslcrl: PostgreSQL specific (LIBPQ-CONNECT-SSLCRL) -->
```
greg0ire: How about grouping things that go together in one comment?
```
                <!-- 
                      sslmode: PostgreSQL specific (LIBPQ-CONNECT-SSLMODE)
                      sslrootcert: PostgreSQL specific (LIBPQ-CONNECT-SSLROOTCERT)
                      sslcert: PostgreSQL specific (LIBPQ-CONNECT-SSLCERT)
                      sslkey: PostgreSQL specific (LIBPQ-CONNECT-SSLKEY)
                      sslcrl: PostgreSQL specific (LIBPQ-CONNECT-SSLCRL)
                -->
```

Resources/doc/configuration.rst
```diff
@@ -1003,33 +1003,48 @@ can configure. The following block shows all possible configuration keys:
                 http://symfony.com/schema/dic/doctrine/doctrine-1.0.xsd">
 
             <doctrine:config>
+                <!-- path: SQLite specific -->
+                <!-- memory: SQLite specific -->
+                <!-- multiple-active-result-sets: driver specific -->
+                <!-- pooled: Oracle specific (SERVER=POOLED) -->
```
greg0ire: To be moved next to other Oracle specific parameters below?

Resources/doc/configuration.rst
alcaeus: Good idea!

PR #917, opened 2019-02-04
Detect doctrine bridge version

### LINKED ISSUE #883: Symfony 4.2 compatibility

greg0ire:
Although it is in the PropertyInfo sub namespace, the DoctrineExtractor
class is in the doctrine bridge, which means we should detect the
doctrine bridge version and not the PropertyInfo component version.

Fixes #883, Closes #884 

I did this in a hurry and had no time to test it.


DependencyInjection/DoctrineExtension.php
```diff
@@ -800,7 +800,7 @@ protected function loadOrmCacheDrivers(array $entityManager, ContainerBuilder $c
     private function loadPropertyInfoExtractor($entityManagerName, ContainerBuilder $container)
     {
         $propertyExtractorDefinition = $container->register(sprintf('doctrine.orm.%s_entity_manager.property_info_extractor', $entityManagerName), DoctrineExtractor::class);
-        if (interface_exists(PropertyInitializableExtractorInterface::class)) {
+        if (is_a(DoctrineType::class, ResetInterface::class, true)) {
```
xabbuh: This will fail when the Form component is not installed and thus the parent class of `DoctrineType` does not exist.

xabbuh: Looks like this then:

```
$ bin/console cache:clear

 // Clearing the cache for the dev environment with debug true                                                          

PHP Fatal error:  Class 'Symfony\Component\Form\AbstractType' not found in /Users/christian/projects/doctrine-bundle-917/vendor/symfony/doctrine-bridge/Form/Type/DoctrineType.php on line 29
```

greg0ire: ah hence the build failure… and I'm afraid this means my solution can't work. Closing.

greg0ire:
The only concern about this is the name of the property not being covered by the BC-promise. If this gets merged, I shall make a PR on `DoctrineExtractor` to explain the situation and forbid any changes on that property.

DependencyInjection/DoctrineExtension.php
```diff
@@ -800,7 +800,7 @@ protected function loadOrmCacheDrivers(array $entityManager, ContainerBuilder $c
     private function loadPropertyInfoExtractor($entityManagerName, ContainerBuilder $container)
     {
         $propertyExtractorDefinition = $container->register(sprintf('doctrine.orm.%s_entity_manager.property_info_extractor', $entityManagerName), DoctrineExtractor::class);
-        if (interface_exists(PropertyInitializableExtractorInterface::class)) {
+        if ((new ReflectionClass(DoctrineExtractor::class))->hasProperty('entityManager')) {
```
xabbuh: What about using `property_exists()` here?

greg0ire: Even better! No need to bother with reflection! 👍

PR #918, opened 2019-02-04
fix Doctrine bridge version detection

xabbuh:
Since we cannot use feature changes to detect which version of the
Doctrine bridge is installed I suggest that we parse the changelog and
check if entries for 4.2 are present.


alcaeus:
This certainly gets you points for creativeness, but I wonder if it's more sensible to just install [`ocramius/package-versions`](https://packagist.org/packages/ocramius/package-versions)?

xabbuh:
@alcaeus We still support PHP 5 in this branch, but ocramius/package-versions requires PHP 7.

alcaeus:
@xabbuh ah, that's correct - we can't use that then 😞 

Closing this one in favour of #917 as it uses code to detect the version instead of checking the changelog. Again, points for creativity here 😉

PR #919, opened 2019-02-05
provide better error when doctrine/orm is missing

xabbuh:


greg0ire:
Ah the cs-fix failed… @doctrine/doctrinecore I had the same issue today when contributing, checking cs locally is hard, I had to run `composer require -n --prefer-dist --dev doctrine/coding-standard:^5.0` (copied that form .travis.yml) to be able to run `phpcbf`. Is that the right way to do this?

Majkl578:
@greg0ire That is solved in master, can't be done in 1.10.x due to PHP 5.x compat.

alcaeus:
Thanks @xabbuh!

PR #929, opened 2019-03-13
Use namespaced Twig

rosier:
Switch to namespaced classes

### DIFF
--- a/Tests/ProfilerTest.php
+++ b/Tests/ProfilerTest.php
@@ -17,16 +17,16 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Fragment\FragmentHandler;
 use Symfony\Component\HttpKernel\Profiler\Profile;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
-use Twig_Environment;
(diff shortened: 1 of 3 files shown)


xabbuh:
@alcaeus No, namespaced classes have been introduced much longer ago.

alcaeus:
Oh, I must have missed that then. After taking another look, aliases have been added in Twig 1.34.0 for 1.x and 2.4.0 for 2.x, so those are the versions to be added in composer.json.

rosier:
Just noticed that the twig deprecations make the tests also fail on php v7.x in the 1.10 branch.

Should I change the target to 1.10.x also?

alcaeus:
Technically it would be possible, but I'm not sure about effectively changing the twig version requirements in a patch release, which would be necessary for this change. Instead, I'd like to change the builds to not fail on deprecations as we've done in other packages where this has become an issue in version branches.

stof:
Well, Symfony itself already requires Twig versions shipping namespaced classes. So it is already impossible to install an older Twig version if you use a maintained version of Symfony.

rosier:
Status: Needs Review

rosier:
No problem. Thanks for the heads up.

PR #930, opened 2019-03-13
CI: Test against PHP 7.4snapshot instead of nightly (8.0)

rosier:
composer install fails on php v8.0 / nightly


alcaeus:
Small correction here: this is already in 1.10.x, but I haven't merged it down yet. Will do so over the next few days; closing here in the meantime.

PR #935, opened 2019-03-20
Removing deprecated call to Configuration::setSchemaAssetsFilter() + extendable

weaverryan:
Hi!

If you're using doctrine/dbal 2.9 or higher, this removes the deprecated call to `Configuration:: setFilterSchemaAssetsExpression()`. It also makes the "schema filter" extendable via a tag. The use-case is when a 3rd-party bundle manages a table automatically for you through DBAL and you don't want, for example, Doctrine migrations (https://github.com/doctrine/migrations/blob/cf7f94b0bcfbd7d9c240aee759dc6619bc9ad796/lib/Doctrine/Migrations/Generator/DiffGenerator.php#L117) / `doctrine:schema:update` to suggest removing that table because it's not found in the mapping metadata.

One real-world use-case will be for #868: when the Doctrine transport is added to Messenger, a table will be created for storing messages (either automatically by Messenger or manually by the user - their choice). DoctrineBundle itself could use this hook to avoid migrations trying to remove that table once it's created.

Thanks!

### DIFF
--- a/Dbal/RegexSchemaAssetFilter.php
+++ b/Dbal/RegexSchemaAssetFilter.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\Dbal;
+
+use Doctrine\DBAL\Schema\AbstractAsset;
+
+class RegexSchemaAssetFilter
+{
+    /** @var string */
+    private $filterExpression;
+
+    public function __construct(string $filterExpression)
+    {
+        $this->filterExpression = $filterExpression;
+    }
+
+    public function __invoke($assetName) : bool
+    {
+        if ($assetName instanceof AbstractAsset) {
+            $assetName = $assetName->getName();
+        }
+
+        return preg_match($this->filterExpression, $assetName);
+    }
+}
--- a/DependencyInjection/Compiler/DbalSchemaFilterPass.php
+++ b/DependencyInjection/Compiler/DbalSchemaFilterPass.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
+
+use Doctrine\DBAL\Configuration;
+use LogicException;
+use Symfony\Component\DependencyInjection\ChildDefinition;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Processes the doctrine.dbal.schema_filter
+ */
+class DbalSchemaFilterPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        $filters = $container->findTaggedServiceIds('doctrine.dbal.schema_filter');
+
+        if (count($filters) > 0 && ! method_exists(Configuration::class, 'setSchemaAssetsFilter')) {
+            throw new LogicException('The doctrine.dbal.schema_filter tag is only supported when using doctrine/dbal 2.9 or higher.');
+        }
+
+        $connectionFilters = [];
+        foreach ($filters as $id => $tagAttributes) {
+            foreach ($tagAttributes as $attributes) {
+                $name = isset($attributes['connection']) ? $attributes['connection'] : $container->getParameter('doctrine.default_connection');
+
+                if (! isset($connectionFilters[$name])) {
+                    $connectionFilters[$name] = [];
+                }
+
+                $connectionFilters[$name][] = new Reference($id);
+            }
+        }
+
+        foreach ($connectionFilters as $name => $references) {
+            $configurationId = sprintf('doctrine.dbal.%s_connection.configuration', $name);
+
+            if (! $container->hasDefinition($configurationId)) {
+                continue;
+            }
+
+            $definition = new ChildDefinition('doctrine.dbal.schema_asset_filter_manager');
+            $definition->setArgument(0, $references);
+
+            $id = sprintf('doctrine.dbal.%s_schema_asset_filter_manager', $name);
+            $container->setDefinition($id, $definition);
+            $container->findDefinition($configurationId)
+                ->addMethodCall('setSchemaAssetsFilter', [new Reference($id)]);
+        }
+    }
+}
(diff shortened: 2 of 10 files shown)


Dbal/RegexSchemaAssetFilter.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\Dbal;
+
+use Doctrine\DBAL\Schema\AbstractAsset;
+
+class RegexSchemaAssetFilter implements SchemaAssetFilterInterface
```
stof: Shouldn't this be part of DBAL instead ?

DependencyInjection/Compiler/DbalSchemaFilterPass.php
```diff
@@ -0,0 +1,56 @@
...
+ */
+class DbalSchemaFilterPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        // feature require doctrine/dbal 2.9 or higher
+        if (!method_exists(\Doctrine\DBAL\Configuration::class, 'setSchemaAssetsFilter')) {
```
stof: Shouldn't we fail if the tag is used for older DBAL versions instead of silently ignoring the config ?

DependencyInjection/Compiler/DbalSchemaFilterPass.php
```diff
@@ -0,0 +1,56 @@
...
+            return;
+        }
+
+        $filters = $container->findTaggedServiceIds('doctrine.dbal.schema_filter');
+
+        $connectionFilters = [];
+        foreach ($filters as $id => $tagAttributes) {
+            foreach ($tagAttributes as $attributes) {
+                $name = isset($attributes['connection']) ? $attributes['connection'] : $container->getParameter('doctrine.default_connection');
```
stof: should ignoring the name apply it to the default connection, or to all connections ?

Dbal/SchemaAssetFilterInterface.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\Dbal;
+
+/**
+ * Classes consumed by SchemaAssetsFilterManager to filter DBAL assets.
+ */
+interface SchemaAssetFilterInterface
```
stof: This creates a new API while DBAL works with a callable here.

Why not relying on the DBAL signature directly (and `SchemaAssetsFilterManager` becomes a callable implementing a `All` feature over a list of other callables returning a boolean).

DependencyInjection/Compiler/DbalSchemaFilterPass.php
weaverryan: I was wondering about this too. Making it apply to all connections is actually more useful - as, for example, Symfony's session system could (in theory) say: "Hey, ignore a `session` table that exits in any connection... I'm not sure which connection is used for the PDO instance I'm using".

But, that also seems like a possible, legitimate edge-case: I want to ignore `session` on connection 1, but I really DO have a `session` table on connection2 that I'm managing.

Dbal/RegexSchemaAssetFilter.php
weaverryan: Possibly. But, would they want this there? They don't actually need this class, and so they would have this class just to support other people composition class-based callbacks.

weaverryan:
Feedback handled!

Dbal/RegexSchemaAssetFilter.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\Dbal;
+
+use Doctrine\DBAL\Schema\AbstractAsset;
+
+class RegexSchemaAssetFilter
+{
+    private $filterExpression;
```
alcaeus:
```suggestion
    /** @var string */
    private $filterExpression;
```

Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -761,10 +764,60 @@ public function testDbalOracleInstancename()
 
     public function testDbalSchemaFilter()
     {
+        if (method_exists(\Doctrine\DBAL\Configuration::class, 'setSchemaAssetsFilter')) {
+            $this->markTestSkipped('Test only applies to doctrine/dbal 2.9 or lower');
```
alcaeus: Shouldn't this be 2.8 or lower since 2.9 contains the method?

```suggestion
            $this->markTestSkipped('Test only applies to doctrine/dbal 2.8 or lower');
```

DependencyInjection/Compiler/DbalSchemaFilterPass.php
alcaeus: I agree with Ryan here: users should know which connection handles which tables and can build their ignore lists accordingly.

Dbal/RegexSchemaAssetFilter.php
alcaeus: Kinda agree that this should be provided by DBAL, but then they only made it `callable` so there's nothing really we can do. For now we can have the classes here, maybe we can revisit this in a later DBAL release.

weaverryan:
Updated and rebased! Thanks for checking it!

alcaeus:
Some tests still seem to fail: https://travis-ci.org/doctrine/DoctrineBundle/jobs/514116227#L347

> 1) Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection\XmlDoctrineExtensionTest::testDbalSchemaFilter
Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: You have requested a non-existent service "doctrine.dbal.default_connection.configuration".
/home/travis/build/doctrine/DoctrineBundle/vendor/symfony/dependency-injection/ContainerBuilder.php:1019
/home/travis/build/doctrine/DoctrineBundle/Tests/DependencyInjection/AbstractDoctrineExtensionTest.php:773
> 2) Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection\YamlDoctrineExtensionTest::testDbalSchemaFilter
Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: You have requested a non-existent service "doctrine.dbal.default_connection.configuration".
/home/travis/build/doctrine/DoctrineBundle/vendor/symfony/dependency-injection/ContainerBuilder.php:1019
/home/travis/build/doctrine/DoctrineBundle/Tests/DependencyInjection/AbstractDoctrineExtensionTest.php:773

weaverryan:
Sorry about that! It's better now - only php nightly failing for unrelated reasons.

alcaeus:
I've rebased the branch on top of master to fix build the nightly build issue and fixed CS. Thanks for building this @weaverryan!

alcaeus:
Thanks @weaverryan!

Issue #870, opened 2018-10-30
Remove aliases for common interfaces (again)


kimhemsoe:
We have tried this before but had to be reverted in #805

alcaeus:
Fixed in #1048.

PR #936, opened 2019-03-24
Allow to typehint `ConnectionRegistry`

sroze:
This would allow to typehint the `ConnectionRegistry` interface to be able to get a connection.

### DIFF
--- a/Resources/config/dbal.xml
+++ b/Resources/config/dbal.xml
@@ -64,6 +64,7 @@
             <argument>%doctrine.default_entity_manager%</argument>
         </service>
         <service id="Doctrine\Common\Persistence\ManagerRegistry" alias="doctrine" public="false" />
+        <service id="Doctrine\Common\Persistence\ConnectionRegistry" alias="doctrine" public="false" />
         <service id="Symfony\Bridge\Doctrine\RegistryInterface" alias="doctrine" public="false" />
 
         <service id="doctrine.twig.doctrine_extension" class="Doctrine\Bundle\DoctrineBundle\Twig\DoctrineExtension" public="false">


Resources/config/dbal.xml
```diff
@@ -64,6 +64,7 @@
             <argument>%doctrine.default_entity_manager%</argument>
         </service>
         <service id="Doctrine\Common\Persistence\ManagerRegistry" alias="doctrine" public="false" />
+        <service id="Doctrine\Common\Persistence\ConnectionRegistry" alias="doctrine" public="false" />
```
alcaeus: I'm very much against adding more interfaces from `doctrine/persistence` and tying them to ORM. Existing aliases will be dropped in 2.0, see #870. Do we have other options here?

sroze: Another bundle? :/ What's the reasoning behind not having interfaces from `doctrine/persistence` in this bundle? 🤔 If ORM changes and `doctrine` is not anymore the service holding the `ConnectionRegistry` anymore, we "just" have to release a new version of this bundle isn't it?

alcaeus: ORM isn’t the only mapper. The surrounding interfaces will be removed in 2.0 as the assumption that any interface from persistence should be owned by ORM goes out the door once another mapper (e.g. MongoDB ODM) is used.

weaverryan: We solved this in general up until now by having an sub-interface that’s ORM-only iirc.

But if that doesn’t exist, it is tricky: not using these interfaces for autowiring optimizes for the 1% that use ORM and ODM simultaneously. 

We could use the named autowiring thing that was added in 4.2 to make it ConnectionRegistry $dbalRegistry (where you need the type + arg name to match)

sroze: > We could use the named autowiring thing that was added in 4.2 to make it ConnectionRegistry $dbalRegistry (where you need the type + arg name to match)

But it wouldn't be a good DX for the 99% :/

Majkl578:
Besides what @alcaeus [said](https://github.com/doctrine/DoctrineBundle/pull/936#discussion_r268437090), I'd rather prefer deprecating ConnectionRegistry than giving it more meaning.

weaverryan:
Could the user instead already use the RegistryInterface from the Symfony Bridge to get the connection?

alcaeus:
Yes, the ConnectionRegistry interface is extended by the ManagerRegistry interface, so this shouldn’t be too much of a problem.

@sroze did you have a particular use-case for adding this?

sroze:
Just wanted to get a connection by its name so thought I had to inject `ConnectionRegistry`. I'm happy if `ManagerRegistry` has the methods I need... but it wasn't straight-forward to me.

alcaeus:
Good to see it worked. I'll take a look at ConnectionRegistry and how to proceed with it. Closing here in the meantime.

PR #910, opened 2019-01-28
Made entity listener documentation less ambiguous

romaricdrigon:
Hello,

Entity listener documentation was giving different syntaxes, without explaining was was required exactly and for which versions to use those. It was hard to understand the entity annotation is required if you use the "short" service definition, or that you can use the "long" comprehensive service definition in place.
It was misleading for new users, and even for senior ones, I got it wrong today.

This change proposal hopes to make it clearer, but explaining the 2 different syntaxes one after the other.

PR #937, opened 2019-03-29
Improvement in the listener description

nietzscheson:
The description is confused. Improvement in the listener description.

### DIFF
--- a/Resources/doc/entity-listeners.rst
+++ b/Resources/doc/entity-listeners.rst
@@ -14,12 +14,8 @@ entity manager it should be registered with. Example:
             user_listener:
                 class: \UserListener
                 tags:
-                    - { name: doctrine.orm.entity_listener }
-                    -
-                        name: doctrine.orm.entity_listener
-                        event: preUpdate
-                        entity: App\Entity\User
-                        entity_manager: custom
+                    - { name: doctrine.orm.entity_listener, entity: App\Entity\User, event: preUpdate }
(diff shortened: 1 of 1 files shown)


Resources/doc/entity-listeners.rst
```diff
@@ -14,12 +14,8 @@ entity manager it should be registered with. Example:
             user_listener:
                 class: \UserListener
                 tags:
-                    - { name: doctrine.orm.entity_listener }
-                    -
-                        name: doctrine.orm.entity_listener
-                        event: preUpdate
-                        entity: App\Entity\User
-                        entity_manager: custom
+                    - { name: doctrine.orm.entity_listener, entity: App\Entity\User, event: preUpdate }
```
SenseException: The example shows the minimum attributes for a tag and what other attributes are available. Your changes drop this visualization. If your goal was to reduce confusion by removing one tag, you should create a new configuration block with the minimum and maximum amount of tag attributes.

You also need to adapt the XML part for the configuration to be equal to the YAML ones.

Can you please also add the `lazy` attribute to the configuration block with the maximum amount of tag attributes?

alcaeus:
My apologies - I had an open merge from `1.10.x` to `master` sitting on my hard drive which is why this appeared confusing. In fact, #906 and #910 updated these blocks for clarity. I'm closing this PR as I don't think we should be updating that once again.

PR #939, opened 2019-03-31
Add support for validator auto-mapping

dunglas:
Adds support for symfony/symfony#27735 in the bundle.

Related config:

```yaml
framework:
    validation:
        auto_mapping:
            App\Entity\: []
```

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -12,6 +12,7 @@ use LogicException;
 use Symfony\Bridge\Doctrine\DependencyInjection\AbstractDoctrineExtension;
 use Symfony\Bridge\Doctrine\Messenger\DoctrineTransactionMiddleware;
 use Symfony\Bridge\Doctrine\PropertyInfo\DoctrineExtractor;
+use Symfony\Bridge\Doctrine\Validator\DoctrineLoader;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Alias;
 use Symfony\Component\DependencyInjection\ChildDefinition;
@@ -22,6 +23,7 @@ use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\Messenger\MessageBusInterface;
 use Symfony\Component\Messenger\Transport\Doctrine\DoctrineTransportFactory;
+use Symfony\Component\Validator\Mapping\Loader\LoaderInterface;
 
 /**
  * DoctrineExtension is an extension for the Doctrine DBAL and ORM library.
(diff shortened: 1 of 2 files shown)


DependencyInjection/DoctrineExtension.php
```diff
@@ -760,6 +762,23 @@ private function loadPropertyInfoExtractor($entityManagerName, ContainerBuilder
         $propertyExtractorDefinition->addTag('property_info.type_extractor', ['priority' => -999]);
     }
 
+    /**
+     * Loads a validator loader for each defined entity manager.
+     *
+     * @param string $entityManagerName
+     */
+    private function loadValidatorLoader($entityManagerName, ContainerBuilder $container)
```
greg0ire: This package requires php 7.1, I think you can use type hints here:

```suggestion
    private function loadValidatorLoader(string: $entityManagerName, ContainerBuilder $container): void
```

dunglas:
I tested this PR on a real project, and it's working properly.

Majkl578:
> and it's working properly

This will need some sort of test otherwise it can break in future. :|

dunglas:
Yes I totally agree. At least a functional test is mandatory.

greg0ire (approved):

SenseException (approved):

SenseException (changes requested):
A phpcs issue is still there. Otherwise this is :+1:

DependencyInjection/DoctrineExtension.php
```diff
@@ -761,6 +763,21 @@ private function loadPropertyInfoExtractor($entityManagerName, ContainerBuilder
         $propertyExtractorDefinition->addTag('property_info.type_extractor', ['priority' => -999]);
     }
 
+    /**
+     * Loads a validator loader for each defined entity manager.
+     */
+    private function loadValidatorLoader(string $entityManagerName, ContainerBuilder $container) : void
+    {
+        if (!class_exists(DoctrineLoader::class)) {
```
greg0ire: This should fix the build:
```suggestion
        if (! class_exists(DoctrineLoader::class)) {
```

DependencyInjection/DoctrineExtension.php
```diff
@@ -761,6 +763,21 @@ private function loadPropertyInfoExtractor($entityManagerName, ContainerBuilder
         $propertyExtractorDefinition->addTag('property_info.type_extractor', ['priority' => -999]);
     }
 
+    /**
+     * Loads a validator loader for each defined entity manager.
+     */
+    private function loadValidatorLoader(string $entityManagerName, ContainerBuilder $container) : void
+    {
+        if (! class_exists(DoctrineLoader::class)) {
```
stof: This requires testing for LoaderInterface first as the bridge does not prevent installing its latest version without the Validator component (which is expected) and then that would be a fatal error due to the missing interface.

stof (changes requested):

SenseException (approved):

stof (approved):

dunglas:
Thanks a lot for finishing this one Andreas!

tarjei:
It seems that the blogpost for this feature has not been updated with the info about auto enabling it:
https://symfony.com/blog/new-in-symfony-4-3-automatic-validation

PR #942, opened 2019-04-02
Update build pipeline

alcaeus:
This cleans up the build pipeline and removes obsolete configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,6 @@ php:
   - 7.3
   - 7.4snapshot
 
-env:
-  global:
-    - deps=no
-
 before_install:
   - mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{,.disabled} || echo "xdebug not available"
   - composer self-update
@@ -31,7 +27,7 @@ jobs:
   include:
     # Tests the lowest set of dependencies
     - php: 7.1
-      env: deps=low SYMFONY_DEPRECATIONS_HELPER=999999
+      env: stability=lowest SYMFONY_DEPRECATIONS_HELPER=999999
       install:
         - COMPOSER_MEMORY_LIMIT=-1 travis_retry composer update -n --prefer-lowest --prefer-stable --prefer-dist
 
@@ -41,20 +37,20 @@ jobs:
 
     # Test against latest Symfony 3.4 dev version
     - php: 7.1
-      env: stability=dev SYMFONY_DEPRECATIONS_HELPER=999999 SYMFONY_VERSION="3.4.*"
+      env: SYMFONY_VERSION="3.4.*" SYMFONY_DEPRECATIONS_HELPER=999999
       install:
         - travis_retry composer update -n --prefer-dist
 
     # Test against latest Symfony 4.1 dev version
     - php: 7.1
-      env: stability=dev SYMFONY_DEPRECATIONS_HELPER=999999 SYMFONY_VERSION="4.1.*"
+      env: SYMFONY_VERSION="4.1.*" SYMFONY_DEPRECATIONS_HELPER=999999
       install:
         - composer require --dev "symfony/messenger:4.1.*" --no-update
         - travis_retry composer update -n --prefer-dist
 
     # Test against latest Symfony 4.2 dev version
     - php: 7.1
-      env: stability=dev SYMFONY_VERSION="4.2.*"
+      env: SYMFONY_VERSION="4.2.*"
       install:
         - composer require --dev "symfony/messenger:4.2.*" --no-update
         - travis_retry composer update -n --prefer-dist
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,6 +75,9 @@ jobs:
 
     - stage: Coverage
       php: 7.2
+      install:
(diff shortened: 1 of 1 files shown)


stof:
>     * why are we not testing against php nighly? And is that different version than 7.4?

nightly is 8.0-dev

>     * why are we not testing with alpha packages? Fixing bundle after symfony is fully released is too late

most packages don't release alpha releases. So it would run with stable releases. The job running with stability=dev does a better job at that.

>     * "deps" env was unused, that's why it was removed? And can you point me to where is "stability" flag used pls? Maybe it's travis ci specific thing but I'm not familiar enough with that.

not used anywhere. the actual implementation is that the setup instructions have an additional command to configure it. But having it as an env variable in the matrix allows to distinguish the job from other ones (it could be done by naming the job if that's supported)

ostrolucky:
> most packages don't release alpha releases. So it would run with stable releases. The job running with stability=dev does a better job at that.

Are you sure about that? minimum-stability alpha should cover beta and RC packages too. That's why I added it there when there was Symfony beta available. IIRC DoctrineBundle didn't work with it and wasn't fixed until RC versions and test suite was green. Doesn't matter alpha releases are rare, what matters is this covers all - alpha, beta, RC.

> not used anywhere

so we should give it meaning to make it less confusing (not saying that it has to be in this PR). composer command should be same everywhere and we should do --minimum-stability according this flag

.travis.yml
```diff
@@ -31,7 +27,7 @@ jobs:
   include:
     # Tests the lowest set of dependencies
     - php: 7.1
-      env: deps=low SYMFONY_DEPRECATIONS_HELPER=999999
+      env: stability=lowest SYMFONY_DEPRECATIONS_HELPER=999999
```
stof: `lowest` is not actually a stability value. This is confusing.

.travis.yml
```diff
@@ -81,6 +70,9 @@ jobs:
 
     - stage: Coverage
       php: 7.2
+      install:
+        - composer require --dev "symfony/messenger:4.2.*" --no-update
```
stof: why do we need messenger ?

.travis.yml
```diff
@@ -41,37 +37,30 @@ jobs:
 
     # Test against latest Symfony 3.4 dev version
```
stof: this comment looks wrong, as it seems to test against latest stable 3.4

stof:
@ostrolucky testing against dev stability allows to catch that much earlier.
the issue is that stability=dev is currently used only in the 7.4snapshot job, which is allowed to fail due to PHP nightly being unstable.
If we want to catch failures from future Symfony versions, we need to make that run in a mandatory job instead (or at least in a less volatile job than PHP snapshot)

ostrolucky:
Dev stability is too unstable. We need to fail build when actual public releases start to happen. But yes I agree with you with avoiding mixing this with nightly

.travis.yml
Majkl578: Using lowercase env variable is confusing as well.

Majkl578:
Testing against dev dependencies leads to erratic build pipeline, producing suddenly failing builds.
In ORM we only test against dev depenedencies during CRON task, that way it won't break PRs and releases: https://github.com/doctrine/orm/blob/ebb53acda96e304a6f91d4a5a2d76378a4e1d210/.travis.yml#L54-L61
So 👎 for using minimum-stability=dev for regular builds but 👍 for introducing it as part of CRON build only.

.travis.yml
alcaeus: The messenger component is needed to test the integration there, which is conditional on whether the extension is present.

.travis.yml
alcaeus: These are only used to label builds in the build view. The `stability` env variable is not used anywhere else.

alcaeus:
Having dev builds as part of the cronjob is a good idea, helps avoid bothering contributors with build issues. I'll work in your suggestion then re-request reviews.

alcaeus:
Thanks for your feedback. We now test:
* Most up-to-date stable dependencies on PHP 7.1, 7.2, 7.3 and 7.4snapshot (failures allowed)
* The lowest set of dependencies on PHP 7.1
* Latest Symfony 3.4, 4.1, 4.2 releases on PHP 7.3
* phpcs and Coverage is checked on PHP 7.3
* During cronjob runs, a build with `minimum-stability=dev` is run on PHP 7.3

This should cover most scenarios. As discussed above, I've dropped alpha builds since there is no advantage compared to testing against dev versions. Once the first release candidates for a new Symfony release come along (e.g. for Symfony 4.3), we'll add those to the build pipeline as well. Until then this is tested implicitly when running with a dev stability.

I've also removed the `stability` label from builds and instead used uppercase labels as we've used them before. This is only done to mark builds that can't easily be distinguished otherwise.

.travis.yml
```diff
@@ -31,56 +31,47 @@ jobs:
...
-      install:
-        - composer config minimum-stability alpha
-        - travis_retry composer update -n --prefer-dist
-
-    # Test dev level software
-    - php: nightly
-      env: stability=dev
+    # Test dev versions
+    - php: 7.3
+      if: type = cron
```
ostrolucky: noice!

PR #944, opened 2019-04-04
Wire messenger transport factory

alcaeus:
Supersedes #868.

This adds a service for the DoctrineTransportFactory class to the messenger integration config, but doesn't tag it as transport factory. This way, there shouldn't be any errors due to the class not being present in older versions of symfony/messenger.
The tag is then added in DoctrineExtension if the class in question actually exists.

I'm not sure how to best test this since all we're checking is whether the service has a tag; this information is no longer available in the compiled container. Any suggestions?

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -367,13 +367,7 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $container->registerForAutoconfiguration(ServiceEntityRepositoryInterface::class)
             ->addTag(ServiceRepositoryCompilerPass::REPOSITORY_SERVICE_TAG);
 
-        // If the Messenger component is installed and the doctrine transaction middleware is available, wire it:
-        if (! interface_exists(MessageBusInterface::class) || ! class_exists(DoctrineTransactionMiddleware::class)) {
-            return;
-        }
-
-        $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
-        $loader->load('messenger.xml');
+        $this->loadMessengerServices($container);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


DependencyInjection/DoctrineExtension.php
```diff
@@ -793,4 +788,22 @@ public function getConfiguration(array $config, ContainerBuilder $container)
...
+    {
+        // If the Messenger component is installed and the doctrine transaction middleware is available, wire it:
+        if (! interface_exists(MessageBusInterface::class) || ! class_exists(DoctrineTransactionMiddleware::class)) {
+            return;
+        }
+
+        $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
+        $loader->load('messenger.xml');
+
+        if (! class_exists(DoctrineTransportFactory::class)) {
```
greg0ire: I was going to ask why load a file with a class that does not exist, but I guess it's in case other services with class that do exist are added later? Should the definition maybe be removed from the container if the class does not exist?

alcaeus: The file was added previously. I thought it was best to add the service there as well but to leave it untagged. Is it necessary to delete the service? We can also make the service private and have it removed automatically during container compilation.

greg0ire: > The file was added previously.

Oh right, didn't see that other service, my bad.

> Is it necessary to delete the service?

I don't think it could be harmful, my suggestion was maybe a bit paranoid :P

fabpot: There is no need to remove the service, it will be removed automatically by the container when not used.

Resources/config/messenger.xml
```diff
@@ -11,5 +11,13 @@
         <service id="messenger.middleware.doctrine_transaction" class="Symfony\Bridge\Doctrine\Messenger\DoctrineTransactionMiddleware" abstract="true" public="false">
             <argument type="service" id="doctrine" />
         </service>
+
+        <!--
+        The following service isn't tagged as transport factory because the class may not exist.
+        The tag is added conditionally in DoctrineExtension.
+        -->
+        <service id="messenger.transport.doctrine.factory" class="Symfony\Component\Messenger\Transport\Doctrine\DoctrineTransportFactory">
```
fabpot: You should add `public="false"` to allow the container to remove the service of not tagged.

alcaeus: Added, thanks for the clarification!

PR #892, opened 2018-12-10
Drop support for PHP < 7.1 and Symfony < 3.4

alcaeus:
As per discussion in Slack and #885, this PR removes support for PHP < 7.1 (as they are no longer supported) as well as Symfony < 3.4 (since they are in security support only). This will target the 1.11 minor release.

The remaining build failure is due to a deprecation which isn't ignored in the Symfony 3.4 build. I have no idea which dependency combination causes that, so if anyone would have a few pointers for me, they'd be very much appreciated.

PR #950, opened 2019-04-07
abstain from registering incomplete services

xabbuh:
fixes symfony/symfony#30948

#FOSSHackathons #EUFOSSA

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -24,6 +24,7 @@ use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\Messenger\MessageBusInterface;
 use Symfony\Component\Messenger\Transport\Doctrine\DoctrineTransportFactory;
+use Symfony\Component\PropertyInfo\PropertyInfoExtractorInterface;
 use Symfony\Component\Validator\Mapping\Loader\LoaderInterface;
 
 /**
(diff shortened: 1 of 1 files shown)


xabbuh:
ping @dmaicher @dunglas

dmaicher:
Can confirm that this fixes the issue.

But it also means (without having `symfony/property-info`) I cannot enable the new doctrine validator mapping anymore so it actually works? Or am I missing something?

xabbuh:
something else seems to be wrong, @dmaicher and me are investigating

xabbuh:
In fact, this was not an issue on `1.10.x` at all. I picked the code that already prevent this from happen into `master`.

This is ready to be reviewed.

dmaicher:
@alcaeus somehow this 

https://github.com/doctrine/DoctrineBundle/blob/1.10.x/DependencyInjection/DoctrineExtension.php#L358

was lost during some merge up into master?

DependencyInjection/DoctrineExtension.php
```diff
@@ -355,11 +355,17 @@ protected function ormLoad(array $config, ContainerBuilder $container)
 
         $config['entity_managers'] = $this->fixManagersAutoMappings($config['entity_managers'], $container->getParameter('kernel.bundles'));
 
+        $loadPropertyInfoExtractor = interface_exists('Symfony\Component\PropertyInfo\PropertyInfoExtractorInterface')
```
alcaeus: Can you refactor this to use the `::class` constants?

alcaeus:
This was inadvertently removed in #892. Thanks @xabbuh and @dmaicher for digging into this. I'll see what I can do to prevent this in the future.

alcaeus:
Thanks @xabbuh!

xabbuh:
Thank you for fixing the class.

Issue #899, opened 2018-12-13
[Entity Listeners] always lazy load via ServiceLocator


dmaicher:
Since soon support for Symfony < 3.4 will be dropped we could use a `ServiceLocator` here 

https://github.com/doctrine/DoctrineBundle/blob/master/Mapping/ContainerAwareEntityListenerResolver.php#L20

to and get rid of the `lazy` attribute completely?

https://github.com/doctrine/DoctrineBundle/blob/master/DependencyInjection/Compiler/EntityListenerPass.php#L44

So we don't have to mark the listeners `public` anymore and we can always lazily fetch them.

Or is there any reason **to not always lazy load** them?

alcaeus:
Just a heads up, I've decided to leave this out from #892 but will be changing this in time for a 1.11 release so we can deprecate the original behaviour. I just need some more time to wrap my head around the logic.

dmaicher:
Looking into this now at EUFOSSA hackathon :+1:

alcaeus:
Closing as #951 was merged 🎉

PR #951, opened 2019-04-07
Add ContainerEntityListenerResolver + load listeners lazy by default

dmaicher:
EUFOSSA

This implements https://github.com/doctrine/DoctrineBundle/issues/899

- rename `ContainerAwareEntityListenerResolver` into `ContainerEntityListenerResolver` + deprecate old class
- make entity listener services lazy by default (if no `lazy` tag attribute specified)
- use service locator as a smaller container for resolving services

I tested this on my huge work project which uses Symfony 3.4 and we have 3 entity listeners:

```
Information for Service "doctrine.orm.default_entity_listener_resolver"
=======================================================================

 ---------------- ------------------------------------------------------------------------ 
  Option           Value                                                                   
 ---------------- ------------------------------------------------------------------------ 
  Service ID       doctrine.orm.default_entity_listener_resolver                           
  Class            Doctrine\Bundle\DoctrineBundle\Mapping\ContainerEntityListenerResolver  
  Tags             -                                                                       
  Calls            registerService, registerService, registerService                       
  Public           yes                                                                     
  Synthetic        no                                                                      
  Lazy             no                                                                      
  Shared           yes                                                                     
  Abstract         no                                                                      
  Autowired        no                                                                      
  Autoconfigured   no                                                                      
  Arguments        Service(service_locator.hpxy8ox)                                        
 ---------------- ------------------------------------------------------------------------ 
```

### DIFF
--- a/DependencyInjection/Compiler/EntityListenerPass.php
+++ b/DependencyInjection/Compiler/EntityListenerPass.php
@@ -2,10 +2,12 @@
 
 namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
 
+use Doctrine\Bundle\DoctrineBundle\Mapping\ContainerEntityListenerResolver;
 use Doctrine\Bundle\DoctrineBundle\Mapping\EntityListenerServiceResolver;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\Compiler\ServiceLocatorTagPass;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
 use Symfony\Component\DependencyInjection\Reference;
 
@@ -45,13 +47,8 @@ class EntityListenerPass implements CompilerPassInterface
                     $this->attachToListener($container, $name, $id, $attributes);
                 }
 
-                $class = $resolver->getClass();
-
-                if (substr($class, 0, 1) === '%') {
-                    // resolve container parameter first
-                    $class = $container->getParameterBag()->resolveValue($resolver->getClass());
-                }
-                $resolverSupportsLazyListeners = is_a($class, EntityListenerServiceResolver::class, true);
+                $resolverClass                 = $this->getResolverClass($resolver, $container);
+                $resolverSupportsLazyListeners = is_a($resolverClass, EntityListenerServiceResolver::class, true);
 
                 $lazyByAttribute = isset($attributes['lazy']) && $attributes['lazy'];
                 if ($lazyByAttribute && ! $resolverSupportsLazyListeners) {
@@ -70,10 +67,15 @@ class EntityListenerPass implements CompilerPassInterface
 
                     $resolver->addMethodCall('registerService', [$listener->getClass(), $id]);
 
-                    if (! isset($lazyServiceReferencesByResolver[$resolverId])) {
-                        $lazyServiceReferencesByResolver[$resolverId] = [];
+                    // if the resolver uses the default class we will use a service locator for all listeners
+                    if ($resolverClass === ContainerEntityListenerResolver::class) {
+                        if (! isset($lazyServiceReferencesByResolver[$resolverId])) {
+                            $lazyServiceReferencesByResolver[$resolverId] = [];
+                        }
+                        $lazyServiceReferencesByResolver[$resolverId][$id] = new Reference($id);
+                    } else {
+                        $listener->setPublic(true);
                     }
-                    $lazyServiceReferencesByResolver[$resolverId][$id] = new Reference($id);
                 } else {
                     $resolver->addMethodCall('register', [new Reference($id)]);
                 }
@@ -81,7 +83,7 @@ class EntityListenerPass implements CompilerPassInterface
         }
 
         foreach ($lazyServiceReferencesByResolver as $resolverId => $listenerReferences) {
-            $container->findDefinition($resolverId)->replaceArgument(0, ServiceLocatorTagPass::register($container, $listenerReferences));
+            $container->findDefinition($resolverId)->setArgument(0, ServiceLocatorTagPass::register($container, $listenerReferences));
         }
     }
 
@@ -106,5 +108,17 @@ class EntityListenerPass implements CompilerPassInterface
         }
 
         $container->findDefinition($listenerId)->addMethodCall('addEntityListener', $args);
+    }
+
+    private function getResolverClass(Definition $resolver, ContainerBuilder $container) : string
+    {
+        $resolverClass = $resolver->getClass();
+
+        if (substr($resolverClass, 0, 1) === '%') {
+            // resolve container parameter first
+            $resolverClass = $container->getParameterBag()->resolveValue($resolver->getClass());
+        }
+
+        return $resolverClass;
     }
 }
--- a/DependencyInjection/Compiler/EntityListenerPass.php
+++ b/DependencyInjection/Compiler/EntityListenerPass.php
@@ -3,6 +3,7 @@
 namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
 
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\Compiler\ServiceLocatorTagPass;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
 use Symfony\Component\DependencyInjection\Reference;
@@ -18,6 +19,8 @@ class EntityListenerPass implements CompilerPassInterface
     public function process(ContainerBuilder $container)
     {
         $resolvers = $container->findTaggedServiceIds('doctrine.orm.entity_listener');
+
+        $lazyServiceReferencesByResolver = [];
 
         foreach ($resolvers as $id => $tagAttributes) {
             foreach ($tagAttributes as $attributes) {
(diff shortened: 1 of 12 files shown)


DependencyInjection/Compiler/EntityListenerPass.php
```diff
@@ -41,35 +44,44 @@ public function process(ContainerBuilder $container)
                     $this->attachToListener($container, $name, $id, $attributes);
                 }
 
-                if (isset($attributes['lazy']) && $attributes['lazy']) {
+                $interface = 'Doctrine\\Bundle\\DoctrineBundle\\Mapping\\EntityListenerServiceResolver';
```
maidmaid: `EntityListenerServiceResolver::class` ?

Mapping/ContainerEntityListenerResolver.php
```diff
@@ -0,0 +1,111 @@
...
+    {
+        if ($className === null) {
+            $this->instances = [];
+
+            return;
+        }
+
+        $className = $this->normalizeClassName($className);
+
+        if (! isset($this->instances[$className])) {
```
drupol: This if could be removed: https://3v4l.org/2Zj6D

Mapping/ContainerAwareEntityListenerResolver.php
```diff
@@ -2,107 +2,16 @@
...
-
-    /**
-     * @param string $className
-     *
-     * @return string
-     */
-    private function normalizeClassName($className)
-    {
-        return trim($className, '\\');
+        @trigger_error(sprintf('The class "%s" is deprecated since 1.11 and will be removed in 2.0 Use "%s" instead.', self::class, 'Doctrine\Bundle\DoctrineBundle\Mapping\ContainerEntityListenerResolver'), E_USER_DEPRECATED);
```
OskarStark: Dot to finish the sentence before “Use”

OskarStark:
Maybe add a test that the deprecation is raised?

dmaicher:
> Maybe add a test that the deprecation is raised?

@alcaeus should we test this? I think you said its not needed?

Mapping/ContainerAwareEntityListenerResolver.php
```diff
@@ -2,107 +2,16 @@
 
 namespace Doctrine\Bundle\DoctrineBundle\Mapping;
 
-use InvalidArgumentException;
-use RuntimeException;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
-class ContainerAwareEntityListenerResolver implements EntityListenerServiceResolver
+/**
+ * @deprecated since 1.11 and will be removed in 2.0. Use ContainerEntityListenerResolver instead.
```
nicolas-grekas: since *Doctrine* [...]
so that ppl can figure out their deprecation logs more easily (should be a policy really :) )

DependencyInjection/Compiler/EntityListenerPass.php
```diff
@@ -41,35 +45,43 @@ public function process(ContainerBuilder $container)
...
+                if (substr($class, 0, 1) === '%') {
+                    // resolve container parameter first
+                    $class = $container->getParameterBag()->resolveValue($resolver->getClass());
+                }
+                $resolverSupportsLazyListeners = is_a($class, EntityListenerServiceResolver::class, true);
+
+                $lazyByAttribute = isset($attributes['lazy']) && $attributes['lazy'];
+                if ($lazyByAttribute && ! $resolverSupportsLazyListeners) {
+                    throw new InvalidArgumentException(
+                        sprintf('Lazy-loaded entity listeners can only be resolved by a resolver implementing %s.', EntityListenerServiceResolver::class)
```
greg0ire: IMO You should break the line in the `sprintf` rather than in the exception, it gives a more even distribution of text:

```php
<?php
throw new InvalidArgumentException(sprintf(
    'Lazy-loaded entity listeners can only be resolved by a resolver implementing %s.',
    EntityListenerServiceResolver::class
));
```

Mapping/ContainerEntityListenerResolver.php
```diff
@@ -0,0 +1,107 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\Mapping;
+
+use InvalidArgumentException;
+use Psr\Container\ContainerInterface;
+use RuntimeException;
+
+class ContainerEntityListenerResolver implements EntityListenerServiceResolver
```
greg0ire: Should this class be `final`? Since that packaged moved from the sf namespace to Doctrine it would make sense to apply that kind of policy to new types / methods, right?

Mapping/ContainerEntityListenerResolver.php
```diff
@@ -0,0 +1,107 @@
...
+
+        return $this->instances[$className];
+    }
+
+    /**
+     * @param string $serviceId
+     *
+     * @return object
+     */
+    private function resolveService($serviceId)
```
greg0ire:
```suggestion
    private function resolveService(string $serviceId)
```

Mapping/ContainerEntityListenerResolver.php
```diff
@@ -0,0 +1,107 @@
...
+            } else {
+                $this->instances[$className] = new $className();
+            }
+        }
+
+        return $this->instances[$className];
+    }
+
+    /**
+     * @param string $serviceId
```
greg0ire: This should be removed in favor of a type hint
```suggestion
```

Mapping/ContainerEntityListenerResolver.php
```diff
@@ -0,0 +1,107 @@
...
+     *
```
greg0ire:
```suggestion
```

Mapping/ContainerEntityListenerResolver.php
```diff
@@ -0,0 +1,107 @@
...
+
+        return $this->container->get($serviceId);
+    }
+
+    /**
+     * @param string $className
+     *
+     * @return string
+     */
+    private function normalizeClassName($className)
```
greg0ire:
```suggestion
    private function normalizeClassName(string $className): string
```

Mapping/ContainerEntityListenerResolver.php
```diff
@@ -0,0 +1,107 @@
...
+
+namespace Doctrine\Bundle\DoctrineBundle\Mapping;
+
+use InvalidArgumentException;
+use Psr\Container\ContainerInterface;
+use RuntimeException;
+
+class ContainerEntityListenerResolver implements EntityListenerServiceResolver
+{
+    /** @var ContainerInterface */
```
OskarStark: This php doc could be removed

Mapping/ContainerEntityListenerResolver.php
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        return $this->container->get($serviceId);
+    }
+
+    /**
+     * @param string $className
+     *
+     * @return string
+     */
```
greg0ire: The whole docblock can be removed

Mapping/ContainerAwareEntityListenerResolver.php
```diff
@@ -2,107 +2,16 @@
...
-
-    /**
-     * @param string $className
-     *
-     * @return string
-     */
-    private function normalizeClassName($className)
-    {
-        return trim($className, '\\');
+        @trigger_error(sprintf('The class "%s" is deprecated since 1.11 and will be removed in 2.0. Use "%s" instead.', self::class, 'Doctrine\Bundle\DoctrineBundle\Mapping\ContainerEntityListenerResolver'), E_USER_DEPRECATED);
```
alcaeus: I already suspected this yesterday, but the `DebugClassLoader` indeed throws a deprecation once you have added `@deprecated` to the class above. We can drop the deprecation in the constructor 🎉

Mapping/ContainerEntityListenerResolver.php
alcaeus: No: while the property is assigned a value in the constructor, thus implicitly setting a type, static analysis can't assume that this is the only allowed type. To have this information for static analysis, we still keep `@var` around until we've got typed properties.

alcaeus:
> > Maybe add a test that the deprecation is raised?
> 
> @alcaeus should we test this? I think you said its not needed?

I don't think it's necessary: the deprecation is dropped anyways and left for the `DebugClassLoader` to take care of, so I'd argue we don't need to test anything here.

Mapping/ContainerEntityListenerResolver.php
OskarStark: Thank you 👍🏻

Mapping/ContainerEntityListenerResolver.php
dmaicher: Currently it cannot be final as the old/legacy class `ContainerAwareEntityListenerResolver` extends this new class.

DependencyInjection/Compiler/EntityListenerPass.php
```diff
@@ -41,35 +45,44 @@ public function process(ContainerBuilder $container)
...
-                    $resolver->addMethodCall('registerService', [$listener->getClass(), $id]);
+                    $lazyServiceReferencesByResolver[$resolverId][$id] = new Reference($id);
                 } else {
                     $resolver->addMethodCall('register', [new Reference($id)]);
                 }
             }
         }
+
+        foreach ($lazyServiceReferencesByResolver as $resolverId => $listenerReferences) {
+            $container->findDefinition($resolverId)->replaceArgument(0, ServiceLocatorTagPass::register($container, $listenerReferences));
```
dmaicher: Actually this is currently not covered by any tests. Will add some tests for this today :blush:

Mapping/ContainerEntityListenerResolver.php
dmaicher: maybe we can add `@final` for now and add the keyword in version2? @alcaeus ?

Mapping/ContainerEntityListenerResolver.php
dmaicher: Actually need to revert this :smile: 

```
 11 | ERROR | [ ] Property
    |       |     \Doctrine\Bundle\DoctrineBundle\Mapping\ContainerEntityListenerResolver::$container
    |       |     does not have @var annotation.
    |       |     (SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingPropertyTypeHint)
```

DependencyInjection/Compiler/EntityListenerPass.php
dmaicher: Added a test now.

dmaicher: @alcaeus there is one possible BC break with this that I just thought about...

What if someone uses a custom entity resolver that just `EntityListenerServiceResolver` but does not take any constructor argument for the container/serviceLocator?

Maybe we should only do this optimization if the resolver is actually an `ContainerEntityListenerResolver`? And not some custom implementation?

dmaicher: @alcaeus there is one possible BC break with this I just thought about.

What if someone uses a custom entity listener resolver that implements `EntityListenerServiceResolver` but is not an instance of `ContainerEntityListenerResolver`?

Then this might fail as we cannot guarantee that the first constructor argument is actually a `ContainerInterface`...

So should we maybe only do this optimization if its our class `ContainerEntityListenerResolver`?

Mapping/ContainerEntityListenerResolver.php
alcaeus: Yes, `@final` works for now. It would mean that people using the old class would probably two warnings from the debug class loader (deprecation of the old class as well as warning because the new class would be final), but I'm hoping it will just get them to migrate away faster ;)

DependencyInjection/Compiler/EntityListenerPass.php
dmaicher: I fixed this case. So now we are only doing this in case the class is `ContainerEntityListenerResolver`

dmaicher:
Ready for another round of reviews. Also tested it again with my large work project and it works fine :blush:

ostrolucky:
I guess you guys talked about that, but can you involve others in your decision to deprecate whole class, instead of relaxing __construct typehint in existing class?

alcaeus:
> I guess you guys talked about that, but can you involve others in your decision to deprecate whole class, instead of relaxing __construct typehint in existing class?

Relaxing the type hint from the Symfony `ContainerInterface` to the PSR `ContainerInterface` would have worked without other changes as the original type implements the new type and constructor signature changes don't generate method mismatches in extending classes. However, we decided to deprecate the class to get rid of the `ContainerAware` naming and for consistency with other classes like `ContainerRepositoryFactory`. At the time, I wouldn't have known who else to include in the discussion.

DependencyInjection/Compiler/EntityListenerPass.php
```diff
@@ -41,35 +47,44 @@ public function process(ContainerBuilder $container)
...
-                        $class = $container->getParameterBag()->resolveValue($resolver->getClass());
-                    }
+                    $resolver->addMethodCall('registerService', [$listener->getClass(), $id]);
 
-                    if (! is_a($class, $interface, true)) {
-                        throw new InvalidArgumentException(
-                            sprintf('Lazy-loaded entity listeners can only be resolved by a resolver implementing %s.', $interface)
-                        );
+                    // if the resolver uses the default class we will use a service locator for all listeners
+                    if ($resolverClass === ContainerEntityListenerResolver::class) {
```
alcaeus: I'd like to get rid of this logic for 2.0, but for now this is perfectly fine 👍

ostrolucky:
I would say old name would fit this change too, since it's still aware of container, just not the main one. But yeah I don't mind

alcaeus:
Thanks @dmaicher! 🎉

PR #954, opened 2019-04-13
[Profiling] Add query backtrace data

ottaviano:
image

[dbal pull request](https://github.com/doctrine/dbal/pull/3513) for sql logger


Resources/views/Collector/db.html.twig
```diff
@@ -254,6 +254,11 @@
                                         &nbsp;&nbsp;
                                         <a class="link-inverse" href="{{ path('_profiler', { panel: 'db', token: token, page: 'explain', connection: connection, query: i }) }}" onclick="return explain(this);" data-target-id="explain-{{ i }}-{{ loop.parent.loop.index }}">Explain query</a>
                                     {% endif %}
+
+                                    {% if query.backtrace is defined %}
+                                        &nbsp;&nbsp;
+                                        <a href="#" class="sf-toggle link-inverse" data-toggle-selector="#backtrace-query-{{ i }}-{{ loop.parent.loop.index }}" data-toggle-alt-content="Hide query backtrace">View query backtrace</a>
```
greg0ire: Maybe make things easier for the reader:
```suggestion
                                        <a 
                                            href="#" 
                                            class="sf-toggle link-inverse" 
                                            data-toggle-selector="#backtrace-query-{{ i }}-{{ loop.parent.loop.index }}" 
                                            data-toggle-alt-content="Hide query backtrace"
                                        >View query backtrace</a>
```

Resources/views/Collector/db.html.twig
```diff
@@ -267,6 +272,36 @@
...
+                                                {% for trace in query.backtrace %}
+                                                    <tr>
+                                                        <td>{{ loop.index }}</td>
+                                                        <td>
+                                                            <span class="text-small">
+                                                                {% set line_number = trace.line|default(1) %}
+                                                                {% set file_path = trace.file|format_file(line_number)|striptags|replace({ (' at line ' ~ line_number): '' }) %}
+                                                                {% set file_path_parts = file_path|split(constant('DIRECTORY_SEPARATOR')) %}
+
+                                                                {{ file_path_parts[:-1]|join(constant('DIRECTORY_SEPARATOR')) }}{{ constant('DIRECTORY_SEPARATOR') }}<strong>{{ file_path_parts|last }}</strong>
```
greg0ire:
```suggestion
                                                                {{ file_path_parts[:-1]|join(constant('DIRECTORY_SEPARATOR')) -}}
                                                                {{ constant('DIRECTORY_SEPARATOR') -}}
                                                                <strong>{{ file_path_parts|last }}</strong>
                                                                
```

Resources/views/Collector/db.html.twig
```diff
@@ -267,6 +272,36 @@
...
+                                                                {{- trace.type }}<span class="status-warning">{{ trace.function }}</span> (line {{ line_number }})
```
greg0ire:
```suggestion
                                                                {{- trace.type -}}
                                                                <span class="status-warning">{{ trace.function }}</span>
                                                                (line {{ line_number }})
```

alcaeus:
Note: keeping this on hold until https://github.com/doctrine/dbal/pull/3513 is merged.

Resources/views/Collector/db.html.twig
```diff
@@ -267,6 +272,36 @@
...
+                                                </tr>
+                                            </thead>
+                                            <tbody>
+                                                {% for trace in query.backtrace %}
+                                                    <tr>
+                                                        <td>{{ loop.index }}</td>
+                                                        <td>
+                                                            <span class="text-small">
+                                                                {% set line_number = trace.line|default(1) %}
+                                                                {% set file_path = trace.file|format_file(line_number)|striptags|replace({ (' at line ' ~ line_number): '' }) %}
```
stof: why using `|format_file(line_number)` if you remove both the tags (making the file clickable) and the line number ?

javiereguiluz:
@ottaviano thanks a lot for proposing this feature. I like it a lot! However, as a potential user of this feature, I'd like to see the following trace instead of what you showed in your screenshot:

image

In my opinion, the trace should only display my own code, to show which of my lines of code was responsible for the database query. All the other lines are internal Doctrine stuff which I wouldn't care.

But all this is just my opinion, so let's wait and see if others agree. Thanks!

Resources/views/Collector/db.html.twig
ottaviano: it's for making relative paths

alcaeus:
@javiereguiluz I'm on the fence about this. On one hand, it would definitely make the trace easier to understand, but it may leave holes in the trace regarding "how do we end up here". I believe if we were to do this, we should allow users to toggle this and switch from the shortened trace to the full trace. If we keep the numbering intact, it would show the user that we're omitting some information to make it understandable.

Another problem is, what do we remove? For example, what if the query is run because I use a ParamConverter annotation? Hiding `vendor` entirely would completely hide the trace as the query is entirely triggered by vendor code. We need a good system to decide what entries to hide by default.

Resources/views/Collector/db.html.twig
ottaviano: applied, thanks

Resources/views/Collector/db.html.twig
ottaviano: done, thanks

Resources/views/Collector/db.html.twig
ottaviano: I prefer to keep inline version, like other links in this template

Resources/config/dbal.xml
```diff
@@ -7,6 +7,7 @@
     <parameters>
         <parameter key="doctrine.dbal.logger.chain.class">Doctrine\DBAL\Logging\LoggerChain</parameter>
         <parameter key="doctrine.dbal.logger.profiling.class">Doctrine\DBAL\Logging\DebugStack</parameter>
+        <parameter key="doctrine.dbal.logger.backtrace.class">Doctrine\DBAL\Logging\BacktraceLogger</parameter>
```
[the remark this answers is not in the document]
ottaviano: done 👍

Resources/views/Collector/db.html.twig
```diff
@@ -267,6 +272,36 @@
                                 {% if query.explainable %}
                                     <div id="explain-{{ i }}-{{ loop.parent.loop.index }}" class="sql-explain"></div>
                                 {% endif %}
+
+                                {% if query.backtrace is defined %}
+                                    <div id="backtrace-query-{{ i }}-{{ loop.parent.loop.index }}" class="hidden">
```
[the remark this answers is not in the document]
ottaviano: agree, fixed

Resources/views/Collector/db.html.twig
```diff
@@ -267,6 +272,36 @@
...
+                                        <table>
+                                            <thead>
+                                                <tr >
```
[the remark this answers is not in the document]
ottaviano: done 👍

greg0ire:
I think maybe this could be done later if done at all… plus, a lot of similar PRs have been rejected in the past, so IMO we should focus on the PR first before potentially wasting brainpower here.

ostrolucky:
So what are the downsides of automatically enabling this when profiling is enabled? I'm not sure about exposing new option just for this.

alcaeus:
Not a fan of having many configuration options either, but in this case I agree that it should be opt-in due to the amount of data that's potentially collected.

ostrolucky:
Same is true when `profiling` though. My idea was to enable both in such case

ottaviano:
@ostrolucky I add a new option for disabling this feature by default as the logger saves in memory the trace of **each** query. When a page runs 10 queries it will contain a lot of data.

ottaviano:
status: need review

alcaeus:
> @ostrolucky I add a new option for disabling this feature by default as the logger saves in memory the trace of **each** query. When a page runs 10 queries it will contain a lot of data.

Agree here - we're collecting a lot of data on top of usual profiling data, so it makes sense to add another option for this.

As for the rest of the review, I'll hold off until there's something to review: with  being closed, the functionality needs to be moved here. I would suggesting putting it in the `Doctrine\Bundle\DoctrineBundle\Dbal\Logging` namespace. For now, we can extend the `DebugStack` class as the data collector implementation depends on the `DebugStack` as well and uses its query structure. We'll have to talk about how to move forward with this internally.

ottaviano:
I moved `BacktraceLogger` from Dbal to this Bundle as suggested in https://github.com/doctrine/DoctrineBundle/pull/954#issuecomment-487410710

status: need review

Dbal/Logging/BacktraceLogger.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\Dbal\Logging;
+
+use Doctrine\DBAL\Logging\DebugStack;
+
+class BacktraceLogger extends DebugStack
```
greg0ire:
```suggestion
final class BacktraceLogger extends DebugStack
```

Dbal/Logging/BacktraceLogger.php
```diff
@@ -0,0 +1,23 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function startQuery($sql, ?array $params = null, ?array $types = null)
```
greg0ire:
```suggestion
    public function startQuery($sql, ?array $params = null, ?array $types = null) : void
```

DependencyInjection/Configuration.php
```diff
@@ -143,6 +143,7 @@ private function getDbalConnectionsNode()
                 ->scalarNode('schema_filter')->end()
                 ->booleanNode('logging')->defaultValue($this->debug)->end()
                 ->booleanNode('profiling')->defaultValue($this->debug)->end()
+                ->booleanNode('profiling_backtrace')->defaultValue(false)->end()
```
greg0ire: Please add a call to `info` here.

DependencyInjection/DoctrineExtension.php
```diff
@@ -132,9 +132,16 @@ protected function loadDbalConnection($name, array $connection, ContainerBuilder
...
-        if ($connection['profiling']) {
-            $profilingLoggerId = 'doctrine.dbal.logger.profiling.' . $name;
-            $container->setDefinition($profilingLoggerId, new ChildDefinition('doctrine.dbal.logger.profiling'));
+
+        if ($connection['profiling'] || $connection['profiling_backtrace']) {
+            if ($connection['profiling_backtrace']) {
+                $profilingAbstractId = 'doctrine.dbal.logger.backtrace';
+            } else {
+                $profilingAbstractId = 'doctrine.dbal.logger.profiling';
+            }
```
greg0ire: Maybe us a ternary here
```php
<?php
$profilingAbstractId = $connection['profiling_backtrace'] ?
    'doctrine.dbal.logger.backtrace' : 
    'doctrine.dbal.logger.profiling';
```

I did not mutualize the prefix to keep things greppable

Tests/Dbal/Logging/BacktraceLoggerTest.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\Tests\Dbal\Logging;
+
+use Doctrine\Bundle\DoctrineBundle\Dbal\Logging\BacktraceLogger;
+use PHPUnit\Framework\TestCase;
+
+class BacktraceLoggerTest extends TestCase
+{
+    public function testBacktraceLogged()
```
greg0ire:
```suggestion
    public function testBacktraceLogged() : void
```

Dbal/Logging/BacktraceLogger.php
ottaviano: done

Dbal/Logging/BacktraceLogger.php
ottaviano: done

DependencyInjection/Configuration.php
ottaviano: done

DependencyInjection/DoctrineExtension.php
ottaviano: done

Tests/Dbal/Logging/BacktraceLoggerTest.php
ottaviano: done

alcaeus:
I've updated the PR a little bit:
* I changed the behaviour to always require `profiling` to be enabled instead of treating the new option separate. I think collecting backtraces should be seen as an addition to the original profiling, so it makes sense to make this a two-step thing.
* While doing that, I renamed `profiling_backtrace` to `profiling_collect_backtrace` to better explain what the option does. The help text also specifies that profiling needs to be enabled.
* I added two connections to the extension test to ensure the new functionality is properly tested and behaves as we expect.

Thanks @ottaviano for adding this great feature!

ottaviano:
very nice, thank you @alcaeus 👍

DependencyInjection/Configuration.php
```diff
@@ -143,6 +143,10 @@ private function getDbalConnectionsNode()
                 ->scalarNode('schema_filter')->end()
                 ->booleanNode('logging')->defaultValue($this->debug)->end()
                 ->booleanNode('profiling')->defaultValue($this->debug)->end()
+                ->booleanNode('profiling_collect_backtrace')
```
ostrolucky: Shouldn't this validate that profiling is enabled? Currently if it's not, this will have no effect and such spent be behavior tend to problematic

alcaeus: I was considering that but decided against it. The idea I had when changing behaviour was to avoid people having to explicitly set `profiling_collect_backtrace: '%kernel.debug%'` to collect traces. So, `profiling_collect_backtrace: true` should enable trace collection when profiling itself is enabled, but not do anything otherwise. That's why there's no warning: having `profiling: false` and `profiling_collect_backtrace: true` is a perfectly valid scenario.

In the end, we may also opt to have `profiling` have multiple levels (e.g. `none`, `queries`, `backtrace`) to avoid having multiple switches, but I didn't want to add that kind of behaviour just yet - we can always do this in a future version.

Resources/views/Collector/db.html.twig
stof: but why not keeping the full feature of `|format_file` which makes them clickable ?

greg0ire:
So glad to see this so long awaited feature finally be delivered! Great job @ottaviano!

PR #956, opened 2019-04-22
ping/close connection middleware services

insidestyles:
https://github.com/symfony/symfony/pull/31061

### DIFF
--- a/Tests/DependencyInjection/DoctrineExtensionTest.php
+++ b/Tests/DependencyInjection/DoctrineExtensionTest.php
@@ -663,6 +663,10 @@ class DoctrineExtensionTest extends TestCase
 
         $this->assertNotNull($middlewarePrototype = $container->getDefinition('messenger.middleware.doctrine_transaction'));
         $this->assertCount(1, $middlewarePrototype->getArguments());
+        $this->assertNotNull($middlewarePrototype = $container->getDefinition('messenger.middleware.doctrine_ping_connection'));
+        $this->assertCount(1, $middlewarePrototype->getArguments());
+        $this->assertNotNull($middlewarePrototype = $container->getDefinition('messenger.middleware.doctrine_close_connection'));
+        $this->assertCount(1, $middlewarePrototype->getArguments());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


DependencyInjection/DoctrineExtension.php
```diff
@@ -849,6 +851,14 @@ private function loadMessengerServices(ContainerBuilder $container) : void
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('messenger.xml');
 
+        if (class_exists(DoctrineCloseConnectionMiddleware::class)) {
```
alcaeus: Since the services are private and abstract, they'd be removed from the container if no-one uses them, right? If so, it would be better to just move both services into `messenger.xml` instead of conditionally loading extra service configs.

insidestyles: Is it ok if some updates the bundle, he has to update messenger component (for example 4.2 to 4.3)?

insidestyles: @alcaeus: updated but it depends on symfony PR

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -662,6 +662,8 @@ public function testMessengerIntegration()
         $extension->load([$config], $container);
 
         $this->assertNotNull($middlewarePrototype = $container->getDefinition('messenger.middleware.doctrine_transaction'));
+        $this->assertNotNull($middlewarePrototype = $container->getDefinition('messenger.middleware.doctrine_ping_connection'));
+        $this->assertNotNull($middlewarePrototype = $container->getDefinition('messenger.middleware.doctrine_close_connection'));
```
alcaeus: This changes the test - the assertCount call below needs to be duplicated as well.

insidestyles: added

sroze:
Merged :)

alcaeus:
Ok, let me figure out how to make sure the build is stable and then merge this.

alcaeus:
Build failure is unrelated and needs to be addressed separately. For now, I'm merging this. Thanks @insidestyles!

sroze:
Thank you both @insidestyles & @alcaeus.

PR #961, opened 2019-05-09
Test against Symfony 4.3 beta

alcaeus:


SenseException:
Should a failure be allowed for the beta?

alcaeus:
> Should a failure be allowed for the beta?

I don't think that's necessary.

PR #964, opened 2019-05-12
Fix non existent service when using symfony cache pools

deguif:
Thanks for publishing `1.11.0` version.

When updating to `1.11.0` and simplifying my previous hacky configuration by using the new method for declaring cache configuration based on the symfony cache pools, I got one error.

```
The service "doctrine.orm.default_entity_manager" has a dependency on a non-existent service "doctrine.orm.cache.pool.my_cache_pool".
```

By looking at the code, the `createPoolCacheDefinition` returns a service id, which will be aliased later, but this service id was never registered on the container (instead it was its alias).

I updated the code so that the service id is registered and could later be aliased.

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -739,7 +739,7 @@ class DoctrineExtension extends AbstractDoctrineExtension
                 break;
 
             case 'pool':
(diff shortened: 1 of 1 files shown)


alcaeus:
Eh, that’s not good. I’ll take a look at the fix tomorrow morning and will publish a bug fix release as well. 👍

deguif:
You're welcome.
Thanks for maintaining doctrine bundle.

PR #965, opened 2019-05-12
Wiring of the new PropertyAccessExtractor

dunglas:
Wiring for symfony/symfony#31486.

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -25,6 +25,7 @@ use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\Messenger\MessageBusInterface;
 use Symfony\Component\Messenger\Transport\Doctrine\DoctrineTransportFactory;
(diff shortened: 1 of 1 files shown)


alcaeus:
LGTM. @dunglas can you fix the coding style error?

```
FILE: DependencyInjection/DoctrineExtension.php
--------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
--------------------------------------------------------------------------------
 799 | ERROR | [x] Use early exit to reduce code nesting.
     |       |     (SlevomatCodingStandard.ControlStructures.EarlyExit.EarlyExitNotUsed)
--------------------------------------------------------------------------------
```

dunglas:
Actually, for this specific case, it just increases the cyclomatic complexity. But done in another commit.

alcaeus:
I agree there - this is an edge case in the sniff that I don’t like either. I’ll see if we can update the coding standard or fix the sniff.

dunglas:
It has been discussed here: https://github.com/slevomat/coding-standard/issues/371#issuecomment-401648617

alcaeus:
I've changed the base to 1.11, but this is good to go. Thanks for the fix @dunglas!

PR #970, opened 2019-05-14
Fix BC break with DoctrineType::reset() for all Symfony versions

javer:
Fixes BC break with mistakenly removed `kernel.reset` tag from `form.type.entity` service, because `DoctrineType` implements `ResetInterface` only in Symfony 4.2+, therefore all previous versions starting from 3.4 are affected.

This BC break was introduced in version 1.11.0 in commit 6b25ea4

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -23,6 +23,7 @@ use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Messenger\MessageBusInterface;
 use Symfony\Component\Messenger\Transport\Doctrine\DoctrineTransportFactory;
 use Symfony\Component\PropertyInfo\PropertyAccessExtractorInterface;
@@ -342,6 +343,10 @@ class DoctrineExtension extends AbstractDoctrineExtension
     {
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('orm.xml');
+
+        if (class_exists(AbstractType::class)) {
+            $container->getDefinition('form.type.entity')->addTag('kernel.reset', ['method' => 'reset']);
+        }
 
         $entityManagers = [];
         foreach (array_keys($config['entity_managers']) as $name) {


DependencyInjection/DoctrineExtension.php
```diff
@@ -343,6 +345,10 @@ protected function ormLoad(array $config, ContainerBuilder $container)
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('orm.xml');
 
+        if (class_exists(AbstractType::class) && method_exists(DoctrineType::class, 'reset')) {
```
greg0ire: I think the both parts of the condition will always be true, see https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/Form/AbstractType.php and https://github.com/symfony/symfony/blob/3.4/src/Symfony/Bridge/Doctrine/Form/Type/DoctrineType.php#L280

Shouldn't the condition use `interface_exists(ResetInterface::class)`?

alcaeus: I'm not sure that's sufficient: `ResetInterface` will exist when `symfony/contracts` is provided in a new-enough version. However, auto registration of the `kernel.reset` tag is dependent on `symfony/framework-bundle` 4.2 or newer. According to the changeling, we could check for the `addLink` method in `AbstractController`:
```suggestion
        // This check can be removed when support for Symfony < 4.2 is dropped
        if (method_exists(AbstractController::class, 'addLink')) {
```

javer:
@alcaeus Fixed, please check.

DependencyInjection/DoctrineExtension.php
```diff
@@ -343,6 +344,11 @@ protected function ormLoad(array $config, ContainerBuilder $container)
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('orm.xml');
 
+        // This check can be removed when support for Symfony < 4.2 is dropped
+        if (!method_exists(AbstractController::class, 'addLink')) {
```
xabbuh: This checks the version of the FrameworkBundle which doesn't have to be the same as the one of the DoctrineBridge (which provides the type).

alcaeus: So we also need to check whether the type implements the ResetInterface?

DependencyInjection/DoctrineExtension.php
```diff
@@ -343,6 +345,11 @@ protected function ormLoad(array $config, ContainerBuilder $container)
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('orm.xml');
 
+        // This check can be removed when support for Symfony < 4.2 is dropped
+        if (! interface_exists(ResetInterface::class) || ! is_a(DoctrineType::class, ResetInterface::class)) {
```
xabbuh: Shouldn't it be the other way around (i.e. `interface_exists(ResetInterface::class) || is_a(DoctrineType::class, ResetInterface::class)`)?

javer: We should add this service tag only in case when `DoctrineType` does not implement `ResetInterface`. But before this we need to check whether `ResetInterface` exists, and if not - we have old Symfony and also need to add this tag.

Or do you propose change condition from `(! a || ! b)` to `(! (a && b))` ?

alcaeus: We need to add the tag ourselves if any of the following conditions are true:
* there is no ResetInterface
* the type doesn’t implement ResetInterface
* FrameworkBundle does not have autoconfiguration for ResetInterface (checked by method_exists on the controller)

javer:
@alcaeus @xabbuh What do you think about this solution?

DependencyInjection/DoctrineExtension.php
```diff
@@ -343,6 +345,13 @@ protected function ormLoad(array $config, ContainerBuilder $container)
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('orm.xml');
 
+        // This check can be removed when support for Symfony < 4.2 is dropped
+        if (! interface_exists(ResetInterface::class)
+            || ! is_a(DoctrineType::class, ResetInterface::class)
+            || ! $container->getDefinition('form.type.entity')->hasTag('kernel.reset')) {
```
alcaeus: If this works reliably (I’m not sure because I don’t know when autoconfig is applied, ping @nicolas-grekas) I don’t believe we need the other checks at all 🤔

DependencyInjection/DoctrineExtension.php
stof: we should not rely on autoconfiguration IMO, and always add the tag explicitly when we need to have it.

stof: if we only support versions defining the interface in the DoctrineType, we should put the tag directly in XML.

javer:
@alcaeus It seems to be the most reliable solution according to above comments, because method `reset` could be removed from `DoctrineType` in the future versions of DoctrineBridge.

alcaeus:
> @alcaeus It seems to be the most reliable solution according to above comments, because method `reset` could be removed from `DoctrineType` in the future versions of DoctrineBridge.

That would be a BC break which would only be possible in a major release: this bundle would not install such a version without modifications.

javer:
@alcaeus So we return to the first solution, which was correct, because it:
* checks whether `symfony/form` component is installed
* checks whether `DoctrineType` class has method `reset`
* adds `kernel.reset` tag to the service regardless it's already added or not

alcaeus:
The safe choice:

* check whether `symfony/form` is installed. If not, skip the rest and don't do anything.
* check whether `ResetInterface` exists. If not, skip the rest of the checks and add the tag.
* check whether `DoctrineType` implements `ResetInterface`. If not, skip the rest of the checks and add the tag.
* check whether `AbstractController` has an `addLink` method. If not, add the tag.

Note that checking for the `reset` method is unnecessary - the minimum version we require has that. It's the `ResetInterface` that's only available in newer versions and is also tied to autoconfiguration.

The "yeah, this won't break choice":
* Just add the tag.

As I said - I couldn't find any side effects from having two `kernel.reset` tags in there, so we may as well just add the tag. I'm also not sure if the tag would even be added a second time by the autoconfiguration mechanism.

javer:
@alcaeus Done, please check.

javer:
@alcaeus @xabbuh Hi guys, I have tested this fix with Symfony 4.2 and 4.3 and have just realized that even in these recent versions of Symfony `kernel.reset` tag isn't added automatically, because automatic tagging works only for autoconfigured services: https://github.com/symfony/symfony/blob/4.3/src/Symfony/Component/DependencyInjection/Compiler/ResolveInstanceofConditionalsPass.php#L50
But this service is not marked as autoconfigured, even when you make all your services in the app autoconfigurable, because `autoconfigure` option works only for services defined in the same yml file: https://symfony.com/doc/current/service_container.html#the-autoconfigure-option

Actually this `ResetInterface` haven't ever worked since adding it by @nicolas-grekas in commit https://github.com/symfony/symfony/commit/898203649f0301a38d11095a9b787d2349549004#diff-0b86d71e1a20e98ac050dd33c4e79373R29

So I modified code to explicitly add `kernel.reset` tag always if we have `symfony/form` component installed. All other checks aren't necessary and even wrong.

xabbuh:
The check for the existence of the method is still necessary, isn't it? Otherwise the kernel might call try to call the method because of the tag while it actually doesn't exist.

javer:
@xabbuh Well, actually this check is not so critical, because removing `reset` method from DoctrineType will be a BC break in Symfony Doctrine Bridge, but I added this check for safety.

By the way, we returned to the implementation I already did 16 days ago: https://github.com/doctrine/DoctrineBundle/pull/970#issuecomment-493423484

xabbuh:
Ah sorry, that's my mistake. I thought the method wasn't present in Symfony 3.4.

xabbuh:
So if we don't need the check, can't we just add the tag in the XML config file where we define the service config?

javer:
No, unfortunately we cannot hardcode this tag, because this bundle doesn't have `symfony/form` as a dependency, so we should add this tag only when application uses this component.

I have already tried to do this unconditionally, and unit tests were failed.

DependencyInjection/DoctrineExtension.php
alcaeus: The second check is not needed as we require at least Symfony 3.4 which will always have the `reset` method. Since I'm not aware of Symfony 5 dropping it, we should not use the check.

```suggestion
        if (class_exists(AbstractType::class)) {
```

javer: I agree, removed extra check.

PR #975, opened 2019-05-28
Allow Symfony 5.0

nicolas-grekas:
Same as two years ago: this is required to unlock our CI and will help ensure Doctrine is compatible from day 1.


alcaeus:
Note: due to https://github.com/doctrine/DoctrineCacheBundle/issues/156 we still need to remove the dependency on cache-bundle to make the whole stack work on Symfony 5. Please let me know what we can do to make sure the Symfony Test Suite runs in the meantime.

PR #981, opened 2019-06-11
Deprecate configuring caches through DoctrineCacheBundle

alcaeus:
With this PR, we are deprecating configuring cache drivers for ORM through DoctrineCacheBundle, as the bundle is deprecated (see https://github.com/doctrine/DoctrineCacheBundle/issues/156). Instead of assuming `array` as default, the bundle now uses an appropriate cache pool (`app` or `system` depending on the cache in question) if no explicit configuration was given. This avoids deprecation notices for people that did not add any cache configuration.

Caches should either be configured by passing a symfony/cache pool or service ID. The latter allows using `doctrine/cache` classes, but without the comfort of configuring the caches through bundle config. The implicit requirement on doctrine/cache-bundle will be removed in DoctrineBundle 2.0.


alcaeus:
Coverage tests are broken because of a deprecation:
```
1x: The "Symfony\Component\HttpKernel\EventListener\TranslatorListener" class is deprecated since Symfony 4.3 and will be removed in 5.0, use LocaleAwareListener instead.
    1x in ImportMappingDoctrineCommandTest::setup from Doctrine\Bundle\DoctrineBundle\Tests\Command
```

Can't figure out what's using it though, especially since this doesn't happen in any of the other builds. Could this be a problem because of the code coverage whitelist?

SenseException:
The service of the event-subscriber is called depending on some configuration.

https://github.com/symfony/framework-bundle/blob/v4.3.1/DependencyInjection/FrameworkExtension.php#L1085

It seems to depend on `!$this->isConfigEnabled($container, $config)` above. Can you confirm that? Is there an enabled or disabled translator?

alcaeus:
Turns out the reason for the failure was that the coverage tests installed symfony/messenger in version 4.2. This also locks the FrameworkBundle to 4.2, which still uses the now deprecated class. Locally, I had symfony/messenger 4.3 installed, which doesn't cause the issue. I've fixed .travis.yml in #983 which will be merged up soon.

composer.json
```diff
@@ -37,6 +37,7 @@
     "require-dev": {
         "doctrine/orm": "^2.6",
         "symfony/cache": "^3.4|^4.1",
+        "symfony/translation": "^3.4|^4.1",
```
xabbuh: Can be reverted now that the bugfix has been merged, right?

DependencyInjection/DoctrineExtension.php
```diff
@@ -739,20 +741,42 @@ protected function loadCacheDriver($driverName, $entityManagerName, array $drive
...
+                $usesDeprecatedCacheType = false;
+                $serviceId               = $this->createPoolCacheDefinition($container, $driverMap['pool']);
                 break;
 
             case 'provider':
                 $serviceId = sprintf('doctrine_cache.providers.%s', $driverMap['cache_provider']);
                 break;
         }
 
+        if ($usesDeprecatedCacheType) {
```
xabbuh: Couldn't this be done in the `Configuration` class?

alcaeus: How so? I’d have to check the actual value given to the type configuration option and can’t deprecate the entire configuration option.

composer.json
alcaeus: I could do that if I bump the minimum Symfony versions to the latest patch release after fixes have been released. Otherwise, the prefer-lowest build would fail. Since I’m not in a hurry, I’ll just wait for the next patch releases and properly fix it then 👍

DependencyInjection/DoctrineExtension.php
xabbuh: Using a normalizer and calling `trigger_error()` should work, right?

Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -294,6 +304,9 @@ public function testLoadSingleConnection()
         $this->assertDICDefinitionMethodCallOnce($configDef, 'setDefaultRepositoryClassName', ['Acme\Doctrine\Repository']);
     }
 
+    /**
+     * @group legacy
```
stof: we need to split this test in 2. We need a non-legacy test covering the loading of multiple connections (using non-deprecated settings for the cache).

Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -258,6 +265,9 @@ public function testLoadSimpleSingleConnectionWithoutDbName()
         ]);
     }
 
+    /**
+     * @group legacy
```
stof: should not be a legacy test. We need a non-legacy test covering this case.

Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -417,6 +445,9 @@ public function testEntityManagerRedisMetadataCacheDriverConfigurationWithDataba
         $this->assertDICDefinitionMethodCallOnce($definition, 'select', [1]);
     }
 
+    /**
+     * @group legacy
```
stof: should not be legacy

Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -570,6 +601,9 @@ public function testSetQuoteStrategy()
         $this->assertDICDefinitionMethodCallOnce($def2, 'setQuoteStrategy', [0 => new Reference('doctrine.orm.quote_strategy.ansi')]);
     }
 
+    /**
+     * @group legacy
```
stof: we need a non-legacy test covering the SLC config.

Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -222,6 +227,8 @@ public function testLoadSimpleSingleConnection()
 
     /**
      * The PDO driver doesn't require a database name to be to set when connecting to a database server
+     *
+     * @group legacy
```
stof: should not be a legacy test.

Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -188,6 +190,9 @@ public function testDbalLoadSavepointsForNestedTransactions()
         $this->assertCount(0, $calls);
     }
 
+    /**
+     * @group legacy
```
stof: should not be a legacy test.

stof:
the rule for marked a test as `@group legacy` (and so allowing it to use the BC layer to test it) is that this test gets removed when removing the corresponding deprecated feature (or converted to a test expected an exception, if the deprecation is meant to be replaced by an exception forbidding a given case). Many tests here should be adapted to keep covering different parts of the bundle without using deprecated APIs.

DependencyInjection/DoctrineExtension.php
alcaeus: I'll give it a shot 👍

alcaeus:
@stof you're right - I didn't think of those. I've updated the test and included some legacy tests to test existing functionality. There may be a few more changes necessary once we drop old stuff, but that's alright.

PR #986, opened 2019-06-24
fix in cache configuration in doctrine/packages/doctrine.yml

DominikStyp:
Since following configuration doesn't work (and it should, since [php_file cache is implemented](https://www.doctrine-project.org/projects/doctrine-cache/en/latest/index.html#phpfilecache). I've introduced simple fix in Configuration file.
```
result_cache_driver:
            type: php_file
            directory: '%kernel.cache_dir%/doctrine/php_file'
            extension: 'php'
            umask: 0002
```


alcaeus:
As per #981, we’re deprecating configuring caches through DoctrineCacheBundle in 1.12 for removal in 2.0. We’re not adding any new features for those caches, sorry.

PR #988, opened 2019-06-28
Fix empty auto mapping

alanpoulain:
Even if the auto-mapping was disabled or not present in the framework configuration, it was still enabled since the Doctrine bundle would register a tagged validator loader.
This PR fixes it.

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -813,7 +813,9 @@ class DoctrineExtension extends AbstractDoctrineExtension
      */
     private function loadValidatorLoader(string $entityManagerName, ContainerBuilder $container) : void
     {
-        if (! interface_exists(LoaderInterface::class) || ! class_exists(DoctrineLoader::class)) {
+        $autoMapping = $container->hasParameter('validator.auto_mapping') ? $container->getParameter('validator.auto_mapping') : false;
+
+        if (! $autoMapping || ! interface_exists(LoaderInterface::class) || ! class_exists(DoctrineLoader::class)) {
             return;
         }
 
--- a/Tests/ContainerTest.php
+++ b/Tests/ContainerTest.php
@@ -70,7 +70,7 @@ class ContainerTest extends TestCase
         }
         $this->assertInstanceOf(DoctrineExtractor::class, $container->get('doctrine.orm.default_entity_manager.property_info_extractor'));
 
-        if (class_exists(DoctrineLoader::class)) {
+        if (class_exists(DoctrineLoader::class) && $container->getParameter('validator.auto_mapping')) {
             $this->assertInstanceOf(DoctrineLoader::class, $container->get('doctrine.orm.default_entity_manager.validator_loader'));
         } else {
             $this->assertFalse($container->has('doctrine.orm.default_entity_manager.validator_loader'));
--- a/Tests/DataCollector/DoctrineDataCollectorTest.php
+++ b/Tests/DataCollector/DoctrineDataCollectorTest.php
@@ -57,11 +57,13 @@ class DoctrineDataCollectorTest extends TestCase
             'sql' => 'SELECT * FROM foo WHERE bar = :bar',
             'params' => [':bar' => 1],
             'executionMS' => 32,
+            'types' => [],
         ];
         $logger->queries[] = [
             'sql' => 'SELECT * FROM foo WHERE bar = :bar',
             'params' => [':bar' => 2],
             'executionMS' => 25,
+            'types' => [],
         ];
         $collector         = $this->createCollector([]);
         $collector->addLogger('default', $logger);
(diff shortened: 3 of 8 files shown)


weaverryan:
I see dunglas approved this, but I’m missing a detail. Normally, if this bundle adds a service with a tag, but then another bundle (FrameworkBundle) never uses that tag (because the feature is disable), then adding the service with the tag would have no effect. Why is it different here? I suspect I’m missing a detail :)

alanpoulain:
It's because the tagged services are always used within the compiler pass:
https://github.com/symfony/symfony/blob/835f6b0373b89929e008860985d76317a0040435/src/Symfony/Component/Validator/DependencyInjection/AddAutoMappingConfigurationPass.php#L60-L65

The removal of the property info loader has been done but not the Doctrine one:
https://github.com/symfony/symfony/blob/835f6b0373b89929e008860985d76317a0040435/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L1263

It seems there was a recent fix because of this:
https://github.com/symfony/symfony/pull/31749

But the Doctrine loader is still present in the container.

@dunglas am I right?

alanpoulain:
I've done a PR in Symfony too: https://github.com/symfony/symfony/pull/32278

weaverryan:
But the tags are only used in that compiler if the auto mapping service exists (otherwise it exits at the beginning of the function). You’re right that the doctrine loader would still exist in the container, but if nothing is using it, it won’t actually do anything and will ultimately be removed from the container, right? I mean, please tell me if you’re seeing different behavior - I’m just lightly reading the code and could be missing the point :)

alanpoulain:
I think you misunderstand this line:
https://github.com/symfony/symfony/blob/835f6b0373b89929e008860985d76317a0040435/src/Symfony/Component/Validator/DependencyInjection/AddAutoMappingConfigurationPass.php#L39-L41

`'validator.auto_mapping'` corresponds to the configuration, not the service and is always here:
https://github.com/symfony/symfony/blob/835f6b0373b89929e008860985d76317a0040435/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L1262

There is no auto-mapping service.

weaverryan:
Ah, yep, yep - thank you :). It’s a little weird for one bundle to rely on a parameter of another bundle to determine whether or not a feature should be enabled. As the core works now, this PR is correct. However, I think the fix *may* be inside Symfony itself. The typical behavior would be that if a third party bundle tags a service for a feature that ultimately does not exist, those tags would effectively be “ignored”. Making each bundle “check” if a feature exists on another bundle and then being smart enough to remove the service I “think” isn’t the right approach (but i certainly could be wrong). Could we open an issue in core to check? I’ll be happy if everyone tells me I’m wrong :)

alanpoulain:
Maybe the removal could be in the `FrameworkBundle`, yes.

DependencyInjection/DoctrineExtension.php
```diff
@@ -813,7 +813,9 @@ private function loadPropertyInfoExtractor($entityManagerName, ContainerBuilder
      */
     private function loadValidatorLoader(string $entityManagerName, ContainerBuilder $container) : void
     {
-        if (! interface_exists(LoaderInterface::class) || ! class_exists(DoctrineLoader::class)) {
+        $autoMapping = $container->hasParameter('validator.auto_mapping') ? $container->getParameter('validator.auto_mapping') : false;
```
weaverryan: Hmm, I don't think this will work here - the `$container` in an extension class only has access to the parameters and services defined IN that class - not the whole container. This would need to be removed via a compiler pass, right?

weaverryan:
@alanpoulain question: does this issue cause some bad behavior? Or is it simply that it's wasteful because a service is in the container that does nothing (that *is* something we should fix - just clarifying). 

Ok, looking at this closer - specifically, the compiler pass - https://github.com/symfony/symfony/blob/835f6b0373b89929e008860985d76317a0040435/src/Symfony/Component/Validator/DependencyInjection/AddAutoMappingConfigurationPass.php#L39-L41

I understand now that the `validator.auto_mapping` parameter is always present (assuming the validator is enabled). But if the feature is, effectively "disabled", then I believe this parameter will be an empty array. 

If I'm right, then the problem is inside that compiler pass... and it's just a matter of *not* adding the tagged services to the validator builder if there is *nothing* auto-mapped. It could even go a step further and ignore adding specific loaders if nothing will use that loader.

So, I think the fix *is* in core (in that compiler pass). The parameter (I believe) is meant to be "private" - just a flag that's set in `FrameworkExtension` as a way to pass info to the compiler pass (but not be used by other bundles).

Cheers!

alcaeus:
I agree with @weaverryan here: if nothing is auto-mapped for a loader, it shouldn't register the loader with the ValidationBuilder. Working around this in every loader over and over again is the wrong solution IMO. I can prepare a PR to the AddAutoMappingConfigurationPass if this sounds right (ping @dunglas).

alcaeus:
@dunglas @alanpoulain is this still relevant?

ostrolucky:
I also agree with @weaverryan and @alcaeus. Let's close. Thank you for contributing.

PR #989, opened 2019-06-29
[EntityListener] Fallback on __invoke method if it exists

fancyweb:
It will allow to fall back on the `__invoke` method by default with no breaking change.

### DIFF
--- a/DependencyInjection/Compiler/EntityListenerPass.php
+++ b/DependencyInjection/Compiler/EntityListenerPass.php
@@ -105,6 +105,8 @@ class EntityListenerPass implements CompilerPassInterface
 
         if (isset($attributes['method'])) {
             $args[] = $attributes['method'];
+        } elseif (! method_exists($serviceDef->getClass(), $attributes['event']) && method_exists($serviceDef->getClass(), '__invoke')) {
+            $args[] = '__invoke';
         }
 
         $container->findDefinition($listenerId)->addMethodCall('addEntityListener', $args);
--- a/Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
+++ b/Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
@@ -856,13 +856,19 @@ abstract class AbstractDoctrineExtensionTest extends TestCase
         $this->compileContainer($container);
 
         $listener = $container->getDefinition('doctrine.orm.em1_entity_listener_resolver');
-        $this->assertDICDefinitionMethodCallOnce($listener, 'registerService', ['EntityListener1', 'entity_listener1']);
+        $this->assertDICDefinitionMethodCallCount($listener, 'registerService', [
+            ['EntityListener1', 'entity_listener1'],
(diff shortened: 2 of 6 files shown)


alcaeus:
Great idea, @fancyweb. Could you add tests please? I'd like to see this in 1.12. Thanks!

fancyweb:
Just added tests and doc.

Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -856,13 +856,19 @@ public function testAttachEntityListenerTag()
         $this->compileContainer($container);
 
         $listener = $container->getDefinition('doctrine.orm.em1_entity_listener_resolver');
-        $this->assertDICDefinitionMethodCallOnce($listener, 'registerService', ['EntityListener1', 'entity_listener1']);
+        $this->assertDICDefinitionMethodCallCount($listener, 'registerService', [
+            ['EntityListener1', 'entity_listener1'],
+            ['Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection\Fixtures\InvokableEntityListener', 'invokable_entity_listener'],
+            ['Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection\Fixtures\InvokableEntityListener', 'invokable_entity_listener'],
+        ], 3);
```
fancyweb: I think tagging mutiple times the same listener should not register the service multiple times. We could avoid it in the `EntityListenerPass`. WDYT ?

alcaeus: If it's the same service and not registering it twice doesn't cause any issues, I'm all for it 👍

fancyweb: I will take a look at it but in another PR because this change is unrelated to the proposed new feature. Is that alright for you?

alcaeus: Yes, that’s alright 👍

alcaeus:
Thanks @fancyweb!

DependencyInjection/Compiler/EntityListenerPass.php
```diff
@@ -105,6 +105,8 @@ private function attachToListener(ContainerBuilder $container, $name, $id, array
 
         if (isset($attributes['method'])) {
             $args[] = $attributes['method'];
+        } elseif (! method_exists($serviceDef->getClass(), $attributes['event']) && method_exists($serviceDef->getClass(), '__invoke')) {
```
nicolas-grekas: this doesn't account for ChildDefinition, I'm sure someone will complain soon :)

nicolas-grekas:
Doesn't this mean `__invoke` will be called with any kind of events? Is this really better? I may really miss some context as I'm not used to these listeners much :)

DependencyInjection/Compiler/EntityListenerPass.php
fancyweb: Should be fixed in https://github.com/doctrine/DoctrineBundle/pull/999

PR #992, opened 2019-07-08
Update year LICENSE

Amin-Hosseini:


alcaeus:
Thanks for your contribution. However, we don't need to update the year every year: Fabien and all other contributors to the bundle retain their copyright depending on the applicable law, regardless of whether they add themselves to the license file or not. The code submitted to this repository is put under the MIT license, which regulates how we can use their contributions.

PR #999, opened 2019-07-31
[EntityListenerPass] Handle ChildDefinition

fancyweb:
Fixes https://github.com/doctrine/DoctrineBundle/pull/989#discussion_r309244615

### DIFF
--- a/DependencyInjection/Compiler/EntityListenerPass.php
+++ b/DependencyInjection/Compiler/EntityListenerPass.php
@@ -4,6 +4,7 @@ namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
 
 use Doctrine\Bundle\DoctrineBundle\Mapping\ContainerEntityListenerResolver;
 use Doctrine\Bundle\DoctrineBundle\Mapping\EntityListenerServiceResolver;
+use Symfony\Component\DependencyInjection\ChildDefinition;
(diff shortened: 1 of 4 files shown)


alcaeus:
Ping @doctrine/team-symfony-integration - I'm not very familiar with child definitions and the changes necessary to support them, so I'd appreciate some input. Thanks!

DependencyInjection/Compiler/EntityListenerPass.php
```diff
@@ -87,40 +88,57 @@ public function process(ContainerBuilder $container)
...
+            $resolverClass = $container->getParameterBag()->resolveValue($resolverClass);
         }
 
         return $resolverClass;
     }
+
+    private function getConcreteDefinitionClass(Definition $definition, ContainerBuilder $container, string $id) : string
+    {
+        $class = $definition->getClass();
+        if (is_string($class)) {
```
nicolas-grekas: we don't usually check with `is_string` in similar situations, it should be just `if ($class) {` to me

SenseException: @fancyweb Do you have this change request on your todo list?

fancyweb: Just done it.

alcaeus:
Thanks @fancyweb!

PR #1007, opened 2019-08-23
Use EXPLAIN QUERY PLAN on SQLite

### LINKED ISSUE #1006: Use EXPLAIN QUERY PLAN when using SQLite
While SQLite does support `EXPLAIN` its return is somewhat cryptic:

explain

Maybe it would be better to use [EXPLAIN QUERY PLAN](https://www.sqlite.org/eqp.html):

explain-query-plan

WDYT?

MatTheCat:
Fixes #1006

### DIFF
--- a/Controller/ProfilerController.php
+++ b/Controller/ProfilerController.php
@@ -3,6 +3,7 @@
 namespace Doctrine\Bundle\DoctrineBundle\Controller;
 
 use Doctrine\DBAL\Connection;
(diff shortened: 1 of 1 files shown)


alcaeus:
Build failure is unrelated, I'll fix that before merging this. I'll add this to the 1.12 milestone 👍

alcaeus:
Thanks @MatTheCat!

Issue #997, opened 2019-07-27
Doctrine Transport for Messenger is only available with ORM


micheh:
Currently, it is not possible to use the Doctrine transport for the Messenger component when only the DBAL is installed without the ORM.

The problem is that the `DoctrineTransportFactory` is not registered, which leads to the following exception:
> No transport supports the given Messenger DSN "doctrine://default".

This is because the Messenger services are only registered when the ORM is available:
https://github.com/doctrine/DoctrineBundle/blob/master/DependencyInjection/DoctrineExtension.php#L403

I think the transport should always be loaded, as it doesn't seem to use the ORM features (only the middlewares use the entity manager).

stof:
Fixed by #1028

PR #1010, opened 2019-08-26
loading messenger services independent from ORM layer

unixslayer:
Due to #997 Doctrine Messenger should not be dependent from ORM layer.

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -62,16 +62,14 @@ class DoctrineExtension extends AbstractDoctrineExtension
             $this->dbalLoad($config['dbal'], $container);
         }
 
+        $this->loadMessengerServices($container);
+
         if (empty($config['orm'])) {
             return;
         }
 
         if (empty($config['dbal'])) {
             throw new LogicException('Configuring the ORM layer requires to configure the DBAL layer as well.');
-        }
-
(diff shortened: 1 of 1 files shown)


DependencyInjection/DoctrineExtension.php
```diff
@@ -62,19 +62,19 @@ public function load(array $configs, ContainerBuilder $container)
             $this->dbalLoad($config['dbal'], $container);
         }
 
-        if (empty($config['orm'])) {
-            return;
+        if (! empty($config['orm'])) {
+            $this->ormLoad($config['orm'], $container);
```
stof: this must happen after the `if (empty($config['dbal'])) {` and after the check that the ORM is installed. Otherwise, you miss the existing validation before loading the ORM stuff (validating after does not perform the same, and may not fail properly when the ORM is missing)

unixslayer: @stof I've moved loading messenger service before ORM which installation check was moved inside ormLoad method. Also fixed dbal configuration check

alcaeus:
@unixslayer this is looking good, can you squash commits before I merge? Thanks!

unixslayer:
@alcaeus done

alcaeus:
Superseded by #1028. Thanks @unixslayer!

PR #1013, opened 2019-08-28
Fix php 7.4 class resolution behaviour change

pounard:
Following Symfony issue https://github.com/symfony/symfony/issues/32995 from PHP 7.4 a difference in PHP behaviour when resolving classes with missing parent classes or interfaces raise a fatal non-recoverable error.

By moving symfony/form related code into their own service definition file and loading it conditionally, it won't fix the PHP behaviour but at least work around it gracefully.

Please note I didn't tested this code with PHP 7.4 yet.

I know that it might be disruptive to explode the xml files into many smaller ones, but I think it have a nice side effect for maintenance: it decouples definitions for different features that today all stick together. For some people, but that's a subjective opinion, it could improve code readability/maintainability.

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -345,6 +345,7 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $loader->load('orm.xml');
 
         if (class_exists(AbstractType::class)) {
+            $loader->load('orm-form.xml');
             $container->getDefinition('form.type.entity')->addTag('kernel.reset', ['method' => 'reset']);
         }
 
--- a/Resources/config/orm-form.xml
+++ b/Resources/config/orm-form.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <parameters>
+        <!-- form field factory guesser -->
+        <parameter key="form.type_guesser.doctrine.class">Symfony\Bridge\Doctrine\Form\DoctrineOrmTypeGuesser</parameter>
+    </parameters>
+
+    <services>
+        <service id="form.type_guesser.doctrine" class="%form.type_guesser.doctrine.class%">
+            <tag name="form.type_guesser" />
+            <argument type="service" id="doctrine" />
+        </service>
+
+        <service id="form.type.entity" class="Symfony\Bridge\Doctrine\Form\Type\EntityType">
+            <tag name="form.type" alias="entity" />
+            <argument type="service" id="doctrine" />
+        </service>
+    </services>
+</container>
--- a/Resources/config/orm.xml
+++ b/Resources/config/orm.xml
@@ -39,9 +39,6 @@
         <!-- cache warmer -->
         <parameter key="doctrine.orm.proxy_cache_warmer.class">Symfony\Bridge\Doctrine\CacheWarmer\ProxyCacheWarmer</parameter>
 
-        <!-- form field factory guesser -->
-        <parameter key="form.type_guesser.doctrine.class">Symfony\Bridge\Doctrine\Form\DoctrineOrmTypeGuesser</parameter>
-
         <!-- validator -->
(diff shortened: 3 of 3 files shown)


pounard:
Failures seem to be unrelated.

nicolas-grekas:
I'm not sure what this fixes: if the form component is not installed, these definitions are removed when compiling the container, because they are unused. This happens without caring about the classes of the services, which means I'm missing the link with  #32995
Do I miss something?

pounard:
@nicolas-grekas I don't know if it really fixes or not, my goal here was to see how it works in the CI, but may I quote this comment from @TerjeBr : https://github.com/symfony/symfony/issues/32995#issuecomment-525713660 I linked the other issue because I did this in this context.

It's more likely linked to #32396 in reality. As for #32995 it's the same PHP behaviour as #32396 that triggers the fatal errors. I think that in the long run, the only solution to avoid any problems is to be more careful about what is or is not left in the container by bundles, it's their responsability, and removing classes from it before compilation should be (but I might be wrong):

 - faster (less services to analyze or remove from the container),
 - much more consistent (no services with incomplete classes or non existing classes within the container) which lead to much less error prone and more resilient code/configuration process.

> When the package symfony/form is not needed, that service is not needed neither. **But the doctrine bridge did not need to do anything special to make that service "optional".** It just expected that when no compiler pass picked up the service, it would just automatically fall out of the compiled container.

Now it does (in this PR I mean).

nicolas-grekas:
The way the definition is currently designed is perfectly fine: it reduces the maintenance burden, by delegating cleanup operations to automated processes. That's the job of computers, letting human brains free from superfluous details.

pounard:
> it reduces the maintenance burden

I am unsure it does actually, there's conditional code in the extension anyway which cannot be removed, and the only thing that PR does in the end, is just segregating form-related configuration into its own file: each file is smaller, has a finer name expressing what it does, it's subjective I admit, but some might find this easier to maintain.

nicolas-grekas:
There is only one condition here that makes those services unused. But in other places, the decision will depend on a matrix of conditions. We're cannot deal with as many files as there are cells in the matric, all multiplied by the number of bundles that share this concern. This would be wrong.

pounard:
> There is only one condition here that makes those services unused. But in other places, the decision will depend on a matrix of conditions. We're cannot deal with as many files as there are cells in the matric, all multiplied by the number of bundles that share this concern. This would be wrong.

Doctrine bundle maintainers are supposed to know what happens, and why they did include the form related code, then even name things "form.*". They're not stupid, they are smart, and they know (at least they should know) why and how this service ends up in the container.

> all multiplied by the number of bundles that share this concern

That's why they are bundles, that's why they are maintained by many individuals, because indeed, you alone cannot guess all the possibilities.

nicolas-grekas:
Maintainers don't want to *have* to be smart to be correct. I'm sorry but the approach you're pushing is creating more issues than it creates.

pounard:
> Maintainers don't want to have to be smart to be correct.

I'm quite sure everyone is smart enough to understand why one would not include a service if the dependencies are not here.

alcaeus:
While well-intended, I don't think opening two discussion threads for the same issue is necessary. I understand where you're coming from @pounard, but as @nicolas-grekas mentioned, this is a simple example, while others are more complicated and heavily depend on the various Symfony component versions that are installed.

That said, I'm waiting for https://github.com/symfony/symfony/issues/32995 to come to a resolution before deciding on further steps. Thanks for understanding.

pounard:
@alcaeus understood, thanks for your time. I'll leave this PoC as-is, it might (or might not) come handy in the future.

DependencyInjection/DoctrineExtension.php
```diff
@@ -345,6 +345,7 @@ protected function ormLoad(array $config, ContainerBuilder $container)
         $loader->load('orm.xml');
 
         if (class_exists(AbstractType::class)) {
+            $loader->load('orm-form.xml');
             $container->getDefinition('form.type.entity')->addTag('kernel.reset', ['method' => 'reset']);
```
derrabus: The tag can be added to `orm-form.xml`. There is no need anymore to add it programmatically.

nicolas-grekas:
Can be closed to me.

alcaeus:
Closing here: since the issue was fixed upstream, we're not changing this at this time. We'll be splitting this bundle into separate parts for DBAL and ORM, at which time we may decide to better handle such dependencies, but for now I'm not very inclined to split this.

PR #1015, opened 2019-09-13
Always register ImportDoctrineCommand

### LINKED ISSUE #1014: ImportDoctrineCommand is only available with ORM
`doctrine:database:import` is registered in [orm.xml](https://github.com/doctrine/DoctrineBundle/blob/1.12.x/Resources/config/orm.xml#L183) but only needs DBAL to work.

Should it be moved to dbal.xml?

MatTheCat:
Fixes #1014

### DIFF
--- a/Resources/config/dbal.xml
+++ b/Resources/config/dbal.xml
@@ -88,6 +88,10 @@
             <tag name="console.command" command="doctrine:database:drop" />
         </service>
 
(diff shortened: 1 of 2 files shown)


alcaeus:
Thanks @MatTheCat! I'll check this out after taking the time to fix the build.

PR #1016, opened 2019-09-13
ensure compatability with Symfony 4.4

xabbuh:


alcaeus:
Thanks @xabbuh! Do you want to take care of the other deprecations (adding missing return types) as well? The last deprecation will have to stay in 1.12.x, but this is one deprecation is ignored in the build itself.

If not, I can always fix this later today.

DependencyInjection/DoctrineExtension.php
```diff
@@ -863,6 +863,11 @@ public function getConfiguration(array $config, ContainerBuilder $container) : C
         return new Configuration($container->getParameter('kernel.debug'));
     }
 
+    protected function getMetadataDriverClass(string $driverType): string
+    {
+        return '%'.$this->getObjectManagerElementName('metadata.'.$driverType.'.class%');
```
greg0ire: I have no idea what this is about, but having one percent sign inside the parenthesis, and the other percent sign outside does not feel right.

greg0ire: I just looked at the methdod definition, don't mind me, it should work.

alcaeus:
@xabbuh thanks for making the changes. I've fixed coding style issues, which should get the build stable. Am I correct to assume that the PHP 7.4 failures are going to be fixed with Symfony 4.4? If so, would it make sense to switch the PHP 7.4 build to use Symfony 4.4 as well, as we already do for one 7.3 build?

xabbuh:
If the PHP 7.4 build failure is related to Symfony, it doesn't really matter which version we use as long as it is still supported.

alcaeus:
I believe it may be related to twig, see https://travis-ci.org/doctrine/DoctrineBundle/jobs/585739666#L382:

>   18x: array_key_exists(): Using array_key_exists() on objects is deprecated. Use isset() or property_exists() instead
18x in ProfilerTest::testRender from Doctrine\Bundle\DoctrineBundle\Tests

Either way, this is good to go. Thanks!

xabbuh:
see twigphp/Twig#3097 then (probably), so it will be fixed in the next Twig release

PR #1019, opened 2019-09-17
Moved the "entity listener" docs to Symfony Docs

javiereguiluz:
We're rewording and revamping Doctrine docs in Symfony Docs, so we've created a new and better article to explain everything about Doctrine events (lifecycle callbacks, lifecycle listeners and subscribers and entity listeners).

That's why I propose to remove these docs from the bundle in favor of the Symfony Docs. If you agree, you must wait and merge this only after this pull request is merged: https://github.com/symfony/symfony-docs/pull/12328

Thanks!


stof:
For now, the symfony-docs PR is applied in 4.3+ while the 3.4 doc are still available as this is the LTS. So these docs are still referenced for now.

javiereguiluz:
OK, let's close this then. Thanks.

Issue #1023, opened 2019-10-01
Remove PHP 7.4 from `allow_failures`


carusogabriel:
PHP 7.4 will be released [soon](https://wiki.php.net/todo/php74) and we need to ensure that `Doctrine` is compatible with it.

All we need to do is to remove these following lines

https://github.com/doctrine/DoctrineBundle/blob/af8ac792c9b970ff2bc25b49ab9b31afd9e03dbf/.travis.yml#L84-L85

 and verify that everything is working.

This is a very good issue for those who want to participate in https://hacktoberfest.digitalocean.com.

greg0ire:
@Klizzy pro-tip for your next contribution: use "Closes #1023" rather than "See #1023" and it will auto-close the issue

Klizzy:
thank you very much for this tipp, didn't know about this!

PR #1024, opened 2019-10-01
Update .travis.yml

Klizzy:
removed PHP-7.4 from `.travis.yml` allow_failures to make it compatible with PHP 7.4.
See #1023

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,6 +80,3 @@ jobs:
         - ./vendor/bin/phpunit -v --coverage-clover ./build/logs/clover.xml
       after_script:
         - php ./vendor/bin/coveralls -v
-
-  allow_failures:
-    - php: 7.4snapshot


greg0ire:
Hi! If we no longer allow 7.4 to fail, we also need to fix the failure described here: https://travis-ci.org/doctrine/DoctrineBundle/jobs/591925826

nimah79:
This problem is related to Twig, not Doctrine. So, this PR can be merged temporarily.

greg0ire:
It really shouldn't IMO. If we merge it, builds that were previously green will become red for no good reason. If something has to be done temporarily, it's listing that 7.4 job as part of `allowed_failures`, and that's what we already do.

xabbuh:
What about using PHP 7.4 for the build that uses not yet released dependencies versions? That should work as Twig already has a patch for this just without a stable release yet, shouldn't it?

greg0ire:
I think it would make a lot of sense!

@nimah79 here is what I think @xabbuh is talking about: https://github.com/doctrine/DoctrineBundle/blob/af8ac792c9b970ff2bc25b49ab9b31afd9e03dbf/.travis.yml#L49-L55

fabpot:
@greg0ire I've just released Twig 2.12 which should unlock this PR.

fabpot:
And tests are green now!

SenseException:
Thank you @fabpot :+1:

greg0ire:
Thanks @fabpot! Merged!

greg0ire:
I think I merged a bit too fast… maybe we should require the new version?

PR #1025, opened 2019-10-06
Require php-7.4-friendly version of Twig

greg0ire:
See https://github.com/doctrine/DoctrineBundle/pull/1024#issuecomment-538668474

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
         "symfony/property-info": "^3.4.30|^4.3.3",
         "symfony/phpunit-bridge": "^4.2",
         "symfony/twig-bridge": "^3.4|^4.1",
-        "twig/twig": "^1.34|^2.4",
+        "twig/twig": "^1.34|^2.12",
         "php-coveralls/php-coveralls": "^2.1",
         "phpunit/phpunit": "^7.5",
         "symfony/web-profiler-bundle": "^3.4.30|^4.3.3",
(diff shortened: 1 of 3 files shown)


greg0ire:
The build fails because of a new deprecation in sf 4.4 . I'm adding a second commit that should fix it.

nicolas-grekas:
could be inside the `try`?
GenerateEntitiesDoctrineCommand also need it, isn't it?

greg0ire:
@nicolas-grekas done and done :)

composer.json
```diff
@@ -42,7 +42,7 @@
         "symfony/property-info": "^3.4.30|^4.3.3",
         "symfony/phpunit-bridge": "^4.2",
         "symfony/twig-bridge": "^3.4|^4.1",
-        "twig/twig": "^1.34|^2.4",
+        "twig/twig": "^1.34|^2.12",
```
derrabus: Is this really necessary? I mean, composer will automatically pick 2.12 since it's the latest stable version.

alcaeus: I can add a `prefer-lowest` build on PHP 7.4 if you'd like. I'm all for this change :)

derrabus: > I can add a `prefer-lowest` build on PHP 7.4 if you'd like.

If you do this, composer will install Twig 1.34 – with or without this change. 😉 Also, you will need to bump quite a few more packages to get the build green.

> I'm all for this change :)

I wouldn't oppose it. All I'm saying is that I don't see the need for this change because this library isn't suddenly incompatible with Twig 2.4 - 2.11.

Running a green build on bleeding-edge php with `prefer-lowest` dependencies shouldn't be your goal.

greg0ire: @derrabus valid points, but IMO, it doesn't hurt, quite the contrary. It's not like I'm forcing a major upgrade.

greg0ire: > I mean, composer will automatically pick 2.12 since it's the latest stable version.

Also, that's true if you run `composer update`, but not if you run `composer doctrine/doctrine-bundle`, and I know some people do that kind of thing.

alcaeus:
Thanks @greg0ire!

PR #1026, opened 2019-10-07
Remove deprecated functionality from DoctrineCommand

derrabus:
Since the master branch is 2.0 now, this code can be removed, right? 😃


greg0ire:
#1025 should fix the build

alcaeus:
Thanks to both of you. I'll take care of #1025 and this later today to fix the build (hopefully once and for all)

derrabus:
@greg0ire Actually, #1016 would fix the red build, but it hasn't been merged up to master yet.

alcaeus:
Thanks @derrabus!

PR #1032, opened 2019-10-14
Remove deprecated features

alcaeus:
This PR removes remaining functionality that was deprecated in 1.x:
* Type commenting features removed; the corresponding config parameter was deprecated and will be dropped in 3.0
* Removed the deprecated `ContainerAwareEntityListenerResolver` class
* Updated the comment regarding the annotation reader service; this alias is needed by the abstract extension inherited from the Doctrine Bridge
* Removed the `minify_query` Twig filter, along with the methods in the Twig extension


DependencyInjection/Configuration.php
```diff
@@ -97,7 +97,7 @@ private function addDbalSection(ArrayNodeDefinition $node)
                             ->end()
                             ->children()
                                 ->scalarNode('class')->isRequired()->end()
-                                ->booleanNode('commented')->defaultNull()->end()
+                                ->booleanNode('commented')->setDeprecated()->end()
```
greg0ire: :thinking: why not deprecate it in 1.x and remove it in this PR?

dmaicher: yeah seems a bit odd to keep this option as its a no-op now, right? Seems not used at all.

alcaeus: This is a complicated one. We deprecated certain usages of the `commented` attribute in #947 - essentially requiring users to either mark their custom types as commented, or explicitly turning it off via the configuration. In order to do this, people still need to pass `commented: false` in some instances until updating to 2.0. If we were to deprecate the attribute in 1.x and drop it in 2.0, they'd have a problem: getting rid of the deprecation notice breaks their functionality, but keeping the attribute in causes a hard BC break when moving to 2.0. That's why we've decided to do this as a 2-step process. [Comment link for history](https://github.com/doctrine/DoctrineBundle/pull/947#issuecomment-480498804).

PR #1035, opened 2019-10-21
[WAIT] Ping command

mcfedr:
This just reflects the command from upstream in the same way the other commands are.

See doctrine/dbal#3697

### DIFF
--- a/Command/Proxy/PingDoctrineCommand.php
+++ b/Command/Proxy/PingDoctrineCommand.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\Command\Proxy;
+
+use Doctrine\DBAL\Tools\Console\Command\PingCommand;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Ping the database to check its accessible.
+ */
+class PingDoctrineCommand extends PingCommand
+{
+    /**
+     * {@inheritDoc}
+     */
+    protected function configure()
+    {
+        parent::configure();
+
+        $this
+            ->setName('doctrine:database:ping')
+            ->addOption('connection', null, InputOption::VALUE_OPTIONAL, 'The connection to use for this command');
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        DoctrineCommandHelper::setApplicationConnection($this->getApplication(), $input->getOption('connection'));
+
+        return parent::execute($input, $output);
+    }
+}
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -9,6 +9,7 @@ use Doctrine\Bundle\DoctrineCacheBundle\DependencyInjection\CacheProviderLoader;
 use Doctrine\Bundle\DoctrineCacheBundle\DependencyInjection\SymfonyBridgeAdapter;
 use Doctrine\Common\Persistence\Mapping\ClassMetadataFactory;
 use Doctrine\ORM\Version;
+use Doctrine\DBAL\Tools\Console\Command\PingCommand;
 use LogicException;
 use Symfony\Bridge\Doctrine\DependencyInjection\AbstractDoctrineExtension;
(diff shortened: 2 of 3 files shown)


stof:
it looks like you submitted your PR to a different branch than the one you used to start your work.

alcaeus:
@mcfedr you've created the branch from the master branch, when it should've been 1.12.x. Please rebase your branch and remove all commits from the master branch that aren't in 1.12 (i.e., all but your last commit). If you need help, let me know.

mcfedr:
> @mcfedr you've created the branch from the master branch, when it should've been 1.12.x. Please rebase your branch and remove all commits from the master branch that aren't in 1.12 (i.e., all but your last commit). If you need help, let me know.

yep, i think i accidently created an extra PR, wasn't clear which branch I should use, but its resolved now.

Command/Proxy/PingDoctrineCommand.php
```diff
@@ -0,0 +1,40 @@
...
+        $this
+            ->addOption('connection', null, InputOption::VALUE_OPTIONAL, 'The connection to use for this command');
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        DoctrineCommandHelper::setApplicationConnection($this->getApplication(), $input->getOption('connection'));
```
greg0ire: I think it would be possible to use [a configurator](https://symfony.com/doc/current/service_container/configurators.html) to achieve the `addOption` call, but I can't think of a way to have the functional equivalent of this line without resorting to a console listener and exposing the input… so I guess extension is the only way here, unless someone has an idea…

Command/Proxy/PingDoctrineCommand.php
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Doctrine\Bundle\DoctrineBundle\Command\Proxy;
+
+use Doctrine\DBAL\Tools\Console\Command\PingCommand;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Ping the database to check its accessible.
```
greg0ire:
```suggestion
 * Ping the database to check if it is accessible.
```

DependencyInjection/DoctrineExtension.php
```diff
@@ -115,6 +115,12 @@ protected function dbalLoad(array $config, ContainerBuilder $container)
         foreach ($config['connections'] as $name => $connection) {
             $this->loadDbalConnection($name, $connection, $container);
         }
+
+        if (class_exists(Doctrine\Bundle\DoctrineBundle\Command\Proxy\PingCommand::class)) {
```
alcaeus: This is never going to return true, as the namespace is relative to the current namespace. Even then, this would never return true as the class in question doesn't exist. You should import the correct class (`use Doctrine\DBAL\Tools\Console\Command\PingCommand;`) at the top of the file, then change this to:
```suggestion
        if (class_exists(PingCommand::class)) {
```

Resources/config/dbal.xml
```diff
@@ -100,5 +100,9 @@
             <tag name="console.command" command="doctrine:query:sql" />
         </service>
 
+        <service id="doctrine.ping_command" class="Doctrine\Bundle\DoctrineBundle\Command\Proxy\PingDoctrineCommand">
+            <tag name="console.command" command="doctrine:ping" />
```
alcaeus: I'd be in favour of naming this `doctrine:database:ping` - opinions?

alcaeus:
FYI, I'll wait until the corresponding DBAL PR is merged before merging this one.

DependencyInjection/DoctrineExtension.php
mcfedr: Yes!

Resources/config/dbal.xml
mcfedr: Sure, i guess everything else has 3 levels so makes sense

DependencyInjection/DoctrineExtension.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection;
 
+use Doctrine\Bundle\DoctrineBundle\Command\Proxy\PingCommand;
```
stof: The class in the bundle will always exist. The class_exists check should be on the DBAL class.

ostrolucky:
At this point this is more than one year old. Please reopen after you address issues in upstream PR.

PR #1036, opened 2019-10-21
Add PingCommand from dbal

mcfedr:
This just reflects the command from upstream in the same way the other commands are.

See https://github.com/doctrine/dbal/pull/3697


alcaeus:
Closing here: #1035 targets the correct branch, you just need to remove all commits from master from it.

PR #1037, opened 2019-10-21
Filter out well-known Symfony tables from DBAL schema migrations

### LINKED ISSUE #987: After migration, doctrine deleted session table
I keep session in database. In my migrations I have code to create table:

```
        $this->addSql('CREATE TABLE sessions (sess_id VARCHAR(128) NOT NULL PRIMARY KEY, sess_data BYTEA NOT NULL, sess_time INTEGER NOT NULL, sess_lifetime INTEGER NOT NULL)
```

The problem is when I want to start another migration. `sessions` table is deleted in that case. Anyone knows why? In next migrations in default `up()` method is line `DROP TABLE sessions`

nicolas-grekas:
This prevents constantly generating `DROP TABLE sessions`, etc. when generating migrations.

Closes #987

### DIFF
--- a/DependencyInjection/Compiler/WellKnownSchemaFilterPass.php
+++ b/DependencyInjection/Compiler/WellKnownSchemaFilterPass.php
@@ -1,0 +1,65 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
+
+use Doctrine\DBAL\Configuration;
+use Symfony\Component\Cache\Adapter\PdoAdapter;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\HttpFoundation\Session\Storage\Handler\PdoSessionHandler;
+use Symfony\Component\Lock\Store\PdoStore;
+use Symfony\Component\Messenger\Transport\Doctrine\Connection;
+
+/**
+ * Blacklist tables used by well-known Symfony classes.
+ */
+class WellKnownSchemaFilterPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (! method_exists(Configuration::class, 'setSchemaAssetsFilter')) {
+            // only supported when using doctrine/dbal 2.9 or higher
+            return;
+        }
+
+        $blacklist = [];
+
+        foreach ($container->getDefinitions() as $definition) {
+            if ($definition->isAbstract() || $definition->isSynthetic()) {
+                continue;
+            }
+
+            switch ($definition->getClass()) {
+                case PdoAdapter::class:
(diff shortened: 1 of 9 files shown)


greg0ire:
I added a closing comment to your original message, I think it's related :)

DependencyInjection/Compiler/WellKnownSchemaFilterPass.php
```diff
@@ -0,0 +1,65 @@
...
+    public function process(ContainerBuilder $container)
+    {
+        if (! method_exists(Configuration::class, 'setSchemaAssetsFilter')) {
+            // only supported when using doctrine/dbal 2.9 or higher
+            return;
+        }
+
+        $blacklist = [];
+
+        foreach ($container->getDefinitions() as $id => $definition) {
```
greg0ire: Looks like `$id` is never used, let's remove it.

```suggestion
        foreach ($container->getDefinitions() as $id => $definition) {
```

DependencyInjection/Compiler/WellKnownSchemaFilterPass.php
```diff
@@ -0,0 +1,65 @@
...
+            if ($definition->isAbstract() || $definition->isSynthetic()) {
+                continue;
+            }
+
+            switch ($definition->getClass()) {
+                case PdoAdapter::class:
+                    $blacklist[] = $definition->getArguments()[3]['db_name'] ?? 'cache_items';
```
greg0ire: Shouldn't it be `db_table`, here and below?

https://github.com/symfony/symfony/blob/3f759d7cf4c00df6694878c1a7db0df1551bd21b/src/Symfony/Component/Cache/Adapter/PdoAdapter.php#L35

https://github.com/symfony/symfony/blob/3f759d7cf4c00df6694878c1a7db0df1551bd21b/src/Symfony/Component/HttpFoundation/Session/Storage/Handler/PdoSessionHandler.php#L159

https://github.com/symfony/symfony/blob/3f759d7cf4c00df6694878c1a7db0df1551bd21b/src/Symfony/Component/Lock/Store/PdoStore.php#L60

DependencyInjection/Compiler/WellKnownSchemaFilterPass.php
nicolas-grekas: removed

DependencyInjection/Compiler/WellKnownSchemaFilterPass.php
nicolas-grekas: indeed, updated

alcaeus:
I like this, as it hopefully solves a lot of problems for people.

Only question I have: would it make sense to require DBAL 2.9? It has been out for almost a year, so I'd hope most people are already running it. We can bump in a separate PR, just thinking that we might as well bump here as this wouldn't work when running anything older than 2.9. WDYT @nicolas-grekas?

alcaeus (approved):

alcaeus:
Thanks @nicolas-grekas! We'll bump doctrine/dbal in a separate PR for 2.0 only.

weaverryan:
Thanks @alcaeus!

PR #1039, opened 2019-10-22
Fix using kernel.root_dir in example

nicolas-grekas:


alcaeus:
Thanks @nicolas-grekas!

PR #1043, opened 2019-10-31
Adding support for DoctrineClearEntityManagerWorkerSubscriber

weaverryan:
Hi!

This wires symfony/symfony#34156 as a service. That PR also removes `DoctrineClearEntityManagerMiddleware`, which was never released in DoctrineBundle, so we can safely remove it without any BC worries.

The idea (validated if/when #34156 is merged) is that we should run `$entityManager->clear()` between *each* message that a worker (`messenger:consume`) handles. I have not made this configurable - I would rather someone approach us with a valid case for disabling this before doing that. If there *is* a valid case for wanting this disabled, I don't know what it is - so this would help us learn about such a case.

Thanks!

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -11,6 +11,7 @@ use Doctrine\Common\Persistence\Mapping\ClassMetadataFactory;
 use Doctrine\ORM\Version;
 use LogicException;
 use Symfony\Bridge\Doctrine\DependencyInjection\AbstractDoctrineExtension;
+use Symfony\Bridge\Doctrine\Messenger\DoctrineClearEntityManagerWorkerSubscriber;
 use Symfony\Bridge\Doctrine\Messenger\DoctrineTransactionMiddleware;
 use Symfony\Bridge\Doctrine\PropertyInfo\DoctrineExtractor;
 use Symfony\Bridge\Doctrine\Validator\DoctrineLoader;
@@ -877,6 +878,10 @@ class DoctrineExtension extends AbstractDoctrineExtension
 
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('messenger.xml');
+
+        if (! class_exists(DoctrineClearEntityManagerWorkerSubscriber::class)) {
+            $container->removeDefinition('doctrine.orm.messenger.event_subscriber.doctrine_clear_entity_manager');
+        }
 
         if (! class_exists(DoctrineTransportFactory::class)) {
             return;
--- a/Resources/config/messenger.xml
+++ b/Resources/config/messenger.xml
@@ -14,14 +14,6 @@
 
         <!--
         The following service isn't prefixed by the "doctrine.orm" namespace in order for end-users to just use
-        the "doctrine_clear_entity_manager" shortcut in message buses middleware config
-        -->
-        <service id="messenger.middleware.doctrine_clear_entity_manager" class="Symfony\Bridge\Doctrine\Messenger\DoctrineClearEntityManagerMiddleware" abstract="true" public="false">
-            <argument type="service" id="doctrine" />
-        </service>
-
-        <!--
-        The following service isn't prefixed by the "doctrine.orm" namespace in order for end-users to just use
         the "doctrine_ping_connection" shortcut in message buses middleware config
         -->
         <service id="messenger.middleware.doctrine_ping_connection" class="Symfony\Bridge\Doctrine\Messenger\DoctrinePingConnectionMiddleware" abstract="true" public="false">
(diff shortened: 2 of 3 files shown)


fabpot (approved):

xabbuh (approved):

weaverryan:
Upstream PR is merged - this should be ready :)

alcaeus (approved):

alcaeus:
Thanks @weaverryan!

byhoratiss:
It's being a while since this was merged, but, I'm facing the same problem using Doctrine ORM & ODM (mongodb + mariadb schema) in the same project (unrelated).
Why is not working with Doctrine ODM? From what I investigated it should work since is using an Event to clear the EM, but somehow, using ODM is not working.
Can you give me a hint @weaverryan, and ofc, thank you for this!

Kalyse:
@weaverryan What's the reason this isn't configurable? 

My use case here is that I have a simple `FlushAfterLimitMessagesStamp::class` which allows me to have each message control when it's flush. The actual implementation here is storing some entities (think of them like log data), they are very lightweight, but I do not want each message to persist, flush, persist flush, persist, flush. 

Instead, I want, persist, persist, persist, persist, flush from my messages... 

For instance, I might publish 20 messages...

```
$stamps = array_merge($stamps, [
    new FlushAfterLimitMessagesStamp(),
]);
$this->loggerBus->dispatch($message, $stamps);
```

My Stamp might look like this:

```
<?php

namespace Lycan\Providers\CoreBundle\Producer\Stamp;

    use Symfony\Component\Messenger\Stamp\StampInterface;

    class FlushAfterLimitMessagesStamp implements StampInterface
    {
        private int $messageCount;
        private int $timeLimit;

        /**
         * FlushAfterLimitMessagesStamp constructor.
         * Purpose here, is to force a Flush of the EM after X messages.
         * Allowing us to batch up messages,.
         *
         * @param int $messageCount
         * @param int $timeLimit
         */
        public function __construct($messageCount = 50, $timeLimit = 10)
        {
            $this->messageCount = $messageCount;
            $this->timeLimit = $timeLimit;
        }

        public function getTimeLimit(): int
        {
            return $this->timeLimit;
        }

        public function getMessageCount(): int
        {
            return $this->messageCount;
        }
    }

```

I then have my messages, being handled as normal, however, my Listener, will then subscribe to three events. 

```

  return [
                WorkerStoppedEvent::class => 'onWorkerStopped',
                WorkerMessageReceivedEvent::class => 'onMessageReceived',
                WorkerMessageHandledEvent::class => 'onMessageHandled',
            ];
```

In full, the class looks like. 

After writing it, I was batting my head against a wall wondering why the UOW was always empty on next run, and I see it's because the middleware was removed and replaced with an always clearing subscriber. For my use-case, which I feel is legitimate, I have no way of controlling the transaction for flushing.

What I wanted, is something like...

If idle for 10 seconds, flush. 
If 20 messages received, flush
If 50 seconds have passed flush. 

However, don't flush after every message. 

```

<?php

namespace Lycan\Providers\CoreBundle\Producer\EventSubscriber;

    use Doctrine\ORM\EntityManagerInterface;
    use Symfony\Component\Messenger\Event\WorkerStoppedEvent;
    use Symfony\Component\EventDispatcher\EventSubscriberInterface;
    use Symfony\Component\Messenger\Event\WorkerMessageHandledEvent;
    use Symfony\Component\Messenger\Event\WorkerMessageReceivedEvent;
    use Lycan\Providers\CoreBundle\Producer\Stamp\FlushAfterLimitMessagesStamp;

    class FlushAfterLimitMessagesListener implements EventSubscriberInterface
    {
        private EntityManagerInterface $em;

        private int $receivedMessages = 0;
        private ?float $nextTick = null;

        /**
         * WorkerStoppedListener constructor.
         */
        public function __construct(EntityManagerInterface $em)
        {
            $this->em = $em;
        }

        /**
         * @return string[]
         */
        public static function getSubscribedEvents(): array
        {
            return [
                WorkerStoppedEvent::class => 'onWorkerStopped',
                WorkerMessageReceivedEvent::class => 'onMessageReceived',
                WorkerMessageHandledEvent::class => 'onMessageHandled',
            ];
        }

        public function onMessageReceived(WorkerMessageReceivedEvent $event): void
        {
            $stamped = $event->getEnvelope()->last(FlushAfterLimitMessagesStamp::class);
            if ($stamped) {
                assert($stamped instanceof FlushAfterLimitMessagesStamp);
                ++$this->receivedMessages;
                $this->nextTick = $this->nextTick ?? microtime(true) + $stamped->getTimeLimit();
            }
        }

        public function onMessageHandled(WorkerMessageHandledEvent $event): void
        {
            $stamped = $event->getEnvelope()->last(FlushAfterLimitMessagesStamp::class);
            if ($stamped) {
                assert($stamped instanceof FlushAfterLimitMessagesStamp);
                if ($this->receivedMessages >= $stamped->getMessageCount()) {
                    $this->receivedMessages = 0;
                    $this->nextTick = null;
                    $this->em->flush();
                }
                if ($this->nextTick <= microtime(true)) {
                    $this->em->flush();
                    $this->receivedMessages = 0;
                    $this->nextTick = null;
                }
                $this->nextTick = $this->nextTick ?? microtime(true) + ($stamped->getTimeLimit());
            }
        }

        public function onWorkerStopped(WorkerStoppedEvent $stoppedEvent): void
        {
            // We've finished, let's just always flush.
            if ($this->nextTick > 0 || $this->receivedMessages > 0) {
                $this->em->flush();
            }
        }
    }

```

PR #1047, opened 2019-11-06
Fix compat with Symfony 5 after a recent deprecation

nicolas-grekas:
This drops support for Symfony 3.4, but the alternative is ugly, so I think we should do it, for v2 of course.

Follows https://github.com/symfony/symfony/pull/34230 and https://github.com/symfony/symfony/pull/34230

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,10 @@ env:
 before_install:
   - mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{,.disabled} || echo "xdebug not available"
   - composer self-update
-  - if [ "$SYMFONY_VERSION" != "" ]; then jq "(.require, .\"require-dev\")|=(with_entries(if .key|test(\"^symfony/\") then .value|=\"${SYMFONY_VERSION}\" else . end))" composer.json|ex -sc 'wq!composer.json' /dev/stdin; fi;
+  - composer global require --no-progress --no-scripts --no-plugins symfony/flex dev-master
 
 install:
-  - travis_retry composer update -n --prefer-dist
+  - travis_retry composer update -n --prefer-dist --prefer-stable
 
 script:
   - ./vendor/bin/phpunit -v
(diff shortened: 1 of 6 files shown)


DataCollector/DoctrineDataCollector.php
```diff
@@ -36,7 +36,7 @@ public function __construct(ManagerRegistry $registry)
     /**
      * {@inheritdoc}
      */
-    public function collect(Request $request, Response $response, Exception $exception = null)
+    public function collect(Request $request, Response $response, Throwable $exception = null)
```
nicolas-grekas: this is the change that triggered all the rest

.travis.yml
```diff
@@ -31,26 +31,21 @@ jobs:
...
-        - travis_retry composer update -n --prefer-dist
+        - travis_retry composer update -n --prefer-dist --prefer-stable
 
     # Test against latest Symfony 4.4 dev
     - php: 7.3
-      env: SYMFONY_VERSION="4.4.*@dev"
+      env: SYMFONY_REQUIRE="4.4.*"
       install:
-        - composer config minimum-stability dev
         - composer require --dev "symfony/messenger:4.4.*@dev" --no-update
```
alcaeus: Does this still need `@dev` if we're removing it above?

```suggestion
        - composer require --dev "symfony/messenger:4.4.*" --no-update
```

composer.json
```diff
@@ -5,6 +5,7 @@
     "keywords": ["DBAL", "ORM", "Database", "Persistence"],
     "homepage": "http://www.doctrine-project.org",
     "license": "MIT",
+    "minimum-stability": "dev",
```
alcaeus: I suppose this is ok to avoid a bunch of `composer config` calls in `.travis.yml` - let's see if we have any adverse side effects.

alcaeus (approved):

alcaeus:
Thanks @nicolas-grekas!

PR #805, opened 2018-04-19
Revert "Remove aliases for common interfaces"

weaverryan:
This reverts #740. Fixes #803

PR #1048, opened 2019-11-11
Remove aliases for common interfaces

### LINKED ISSUE #870: Remove aliases for common interfaces (again)

alcaeus:
Fixes #870. This is a repeat of #740, which was reverted in #805. For 2.0.0, we can do this again.

### DIFF
--- a/Resources/config/orm.xml
+++ b/Resources/config/orm.xml
@@ -75,7 +75,6 @@
     </parameters>
 
     <services>
-        <service id="Doctrine\Common\Persistence\ObjectManager" alias="doctrine.orm.entity_manager" public="false" />
         <service id="Doctrine\ORM\EntityManagerInterface" alias="doctrine.orm.entity_manager" public="false" />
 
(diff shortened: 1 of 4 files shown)


Resources/config/orm.xml
```diff
@@ -75,7 +75,6 @@
     </parameters>
 
     <services>
-        <service id="Doctrine\Common\Persistence\ObjectManager" alias="doctrine.orm.entity_manager" public="false" />
```
maxhelias: Since Symfony 4.3 we can depreciate a service like this : https://symfony.com/blog/new-in-symfony-4-3-deprecating-service-aliases
Would adding them in 1.x cause a BC for a lower application 4.3 ?
It's could be very useful if we can do that

alcaeus: Ah, I didn't know. This could be added to 1.12.x, can you submit a PR?

maxhelias: Of course 😃

PR #1050, opened 2019-11-12
Restore autowiring alias for ManagerRegistry

nicolas-grekas:
Promoting autowiring aliases that target classes is encouraging ppl to create tight coupling with the implementation, while IoC is about coding against abstractions.

The fact that the interface is also implemented by the ODM is not an issue to me: this is just the default autowiring alias for the interface. It should serve the majority. There is no issue keeping it.

If we want two aliases, a latter PR could add named autowiring aliases.

Side note: the alias for `Doctrine\ORM\EntityManagerInterface` is encouraging type hinting against a stateful service. But that's for another PR :)

### DIFF
--- a/Repository/ServiceEntityRepository.php
+++ b/Repository/ServiceEntityRepository.php
@@ -14,7 +14,7 @@ use LogicException;
  *
  * class YourEntityRepository extends ServiceEntityRepository
  * {
- *     public function __construct(Registry $registry)
+ *     public function __construct(ManagerRegistry $registry)
  *     {
  *         parent::__construct($registry, YourEntity::class);
  *     }
--- a/Resources/config/dbal.xml
+++ b/Resources/config/dbal.xml
@@ -64,7 +64,7 @@
             <argument>%doctrine.default_connection%</argument>
             <argument>%doctrine.default_entity_manager%</argument>
         </service>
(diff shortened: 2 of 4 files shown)


fabpot:
Big :+1: for me. Without this, we need to explicitly alias things in a Symfony app. That broke my application I've written for the Symfony 5 book.

alcaeus:
Well, if we revert it once we may as well revert it twice. Sounds like this would be a sensible default for Symfony.

I'm hoping to do away with the concept of having multiple registries when using different mappers (e.g. ORM and MongoDB ODM in the same project), and instead having a shared manager registry that contains all mappers for all managers. In that case, it absolutely makes sense to have this interface autowired.

In the meantime, I'll try to work around this in the MongoDB ODM bundle to allow the same alias in that bundle without having a conflict.

@nicolas-grekas can you run `vendor/bin/phpcbf` to fix cs please? Thanks!

nicolas-grekas:
CS fixed!

alcaeus:
Sorry I have to come back with one more thing: can you update the UPGRADE document: https://github.com/doctrine/DoctrineBundle/blob/master/UPGRADE-2.0.md#service-aliases

nicolas-grekas:
updated

PR #1054, opened 2019-11-17
Fix deprecation failures around doctrine:mapping:import command

ostrolucky:


alcaeus:
Thanks @ostrolucky! Nice try, but I always notice 😉

ostrolucky:
i'll create command deprecations pr after this is upmerged

PR #1055, opened 2019-11-17
[DoctrineDataCollector] Handle runnable flag

fancyweb:
This PR takes into account the new `runnable` flag that was added in collected data in https://github.com/symfony/symfony/pull/34384.


alcaeus:
Thanks @fancyweb!

PR #1056, opened 2019-11-18
[WebProfiler] Add copy button on runnable and formatted queries

fancyweb:
This PR adds a "copy" button on runnable queries.

I also took the opportunity to copy paste all the specific CSS for the Doctrine panel from Symfony core inside this bundle so we can remove it from Symfony core (in 5.0). Let me know if you prefer to have the two things splitted for clarity.

Screenshot:
Screenshot 2019-11-18 at 18 11 31


Resources/views/Collector/db.html.twig
```diff
@@ -144,6 +144,12 @@
         .time-container { position: relative; }
         .time-container .nowrap { position: relative; z-index: 1; text-shadow: 0 0 2px #fff; }
         .time-bar { display: block; position: absolute; top: 0; left: 0; bottom: 0; background: #e0e0e0; }
+        .sql-formatted, .sql-runnable { background: var(--base-1, #f5f5f5); margin: .5em 0; padding: 1em; }
```
fancyweb: `var(--base-1)` for 4.3+, fallback to `#f5f5f5` for 3.4 (that's the value there).

Resources/views/Collector/db.html.twig
```diff
@@ -262,13 +268,15 @@
                                     {% endif %}
                                 </div>
 
-                                <div id="formatted-query-{{ i }}-{{ loop.parent.loop.index }}" class="sql-runnable hidden">
+                                <div id="formatted-query-{{ i }}-{{ loop.parent.loop.index }}" class="sql-formatted hidden">
```
fancyweb: Added this new class to not make this div `display: flex` for nothing.

stof:
I'm wondering whether we should bundle the clipboard.js library here (and maybe again in another bundle wanting to provide such button), or whether we should add a feature in the WebProfilerBundle JS to allow having such copy button easily anywhere in the WebProfilerBundle (no duplicate in each bundle) using native APIs.

Note that clipboard.js does 2 things mainly:

- adding DOM listeners to trigger the copy, which is easy to do in the Sfjs code
- implementing a cross-browser way to write text to the clipboard. But as the profiler UI only cares about modern browsers (not true for the toolbar itself), this is unnecessary. Modern code for that is `navigator.clipboard.writeText()` (inside the click listener, as it requires a user-gesture)

fancyweb:
Having a copy button in Symfony's core totally makes sense indeed 👍 Replacing clipboard.js by custom code is fine as well.

However, as a new feature, this copy button would now be available in Symfony WPB 5.1 while we support 3.4 here. Can we add it like this firstly and if someone ever implements it in core (I personally don't have the skills to do it), we then remove our "own" copy button implementation from the DoctrineBundle?

Tests/ProfilerTest.php
```diff
@@ -42,6 +42,8 @@ public function setUp()
 
         $twigLoaderFilesystem = new FilesystemLoader(__DIR__ . '/../Resources/views/Collector');
         $twigLoaderFilesystem->addPath(__DIR__ . '/../vendor/symfony/web-profiler-bundle/Resources/views', 'WebProfiler');
+        // This is normally added by the TwigBundle TwigExtension class
```
fancyweb: The problem here is that symfony/twig-bundle will only be a true dependency of symfony/web-profiler-bundle in the next 3.4 (and all upper branches) release 😕 Should we bump min required version of WPB or ignore it?

stof:
btw, even if we keep it in DoctrineBundle, we could avoid bundling the library.

Tests/ProfilerTest.php
stof: That's fine to me:

- this test does not rely on TwigBundle as it configures Twig by itself
- projects using WebProfilerBundle will already need TwigBundle anyway to configure twig. That's already the case since SF 2.0.

beberlei:
I want to back @stof here with `navigator.clipboard.writeText`. Users of the profiler have modern browsers, so we can be sure that API exists, otherwise just not show the button is also fine.

ostrolucky:
Agreed. @fancyweb can you please remove bundling of library and replace with native API?

ostrolucky:
I've removed clipboard.js, unrelated changes and implemented copy button for formatted query too
Screenshot 2020-01-04 at 16 35 47

Ready to review

Resources/views/Collector/db.html.twig
```diff
@@ -263,12 +265,16 @@
                                 </div>
 
                                 <div id="formatted-query-{{ i }}-{{ loop.parent.loop.index }}" class="sql-runnable hidden">
-                                    {{ query.sql|doctrine_pretty_query }}
+                                    {% set formatted_sql = query.sql|doctrine_pretty_query %}
+                                    {{ formatted_sql|raw }}
+                                    <button class="btn btn-sm" data-clipboard-text="{{ formatted_sql|striptags|e('html_attr') }}">Copy</button>
```
fancyweb: Nice idea to add this one too but we need to html_entity_decode because htmlentities is applied on the formatted_sql so "<" is "&lt;" for example.

Resources/views/Collector/db.html.twig
```diff
@@ -263,12 +265,16 @@
                                 </div>
 
                                 <div id="formatted-query-{{ i }}-{{ loop.parent.loop.index }}" class="sql-runnable hidden">
-                                    {{ query.sql|doctrine_pretty_query }}
+                                    {% set formatted_sql = query.sql|doctrine_pretty_query %}
+                                    {{ formatted_sql|raw }}
```
fancyweb: Also why do we need `|raw` here now?

ostrolucky: because it's saved to variable now, that's how it works

PR #1070, opened 2019-11-25
Default to real-utf8 when no charset is provided

nicolas-grekas:
Configuring charset in a portable way is a real pain for the community.
See e.g. https://github.com/symfony/recipes/pull/674, which triggered https://github.com/symfony/recipes/issues/689, then https://github.com/symfony/recipes/pull/691/files, but the situation is still not fixed.

This is the only way to fix the issue in a way that would make sense: use real-utf8 by default for all platforms (ie utf8mb4 on Mysql).

Here is the patch doing so, targeting 2.0.1.
I'd happily submit on 1.12 if you think that's possible.

### DIFF
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -6,6 +6,7 @@ use Doctrine\Common\EventManager;
 use Doctrine\DBAL\Configuration;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DBALException;
+use Doctrine\DBAL\Driver\AbstractMySQLDriver;
 use Doctrine\DBAL\DriverManager;
 use Doctrine\DBAL\Exception\DriverException;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
@@ -42,6 +43,22 @@ class ConnectionFactory
         }
 
         $connection = DriverManager::getConnection($params, $config, $eventManager);
+
+        if (! isset($params['pdo']) && ! isset($params['charset'])) {
+            $params            = $connection->getParams();
+            $params['charset'] = 'utf8';
+            $driver            = $connection->getDriver();
+
+            if ($driver instanceof AbstractMySQLDriver) {
+                $params['charset'] = 'utf8mb4';
+
+                if (! isset($params['defaultTableOptions']['collate'])) {
+                    $params['defaultTableOptions']['collate'] = 'utf8mb4_unicode_ci';
+                }
+            }
+
+            $connection = new $connection($params, $driver, $connection->getConfiguration(), $connection->getEventManager());
+        }
 
         if (! empty($mappingTypes)) {
             $platform = $this->getDatabasePlatform($connection);
--- a/Tests/ConnectionFactoryTest.php
+++ b/Tests/ConnectionFactoryTest.php
@@ -49,6 +49,26 @@ class ConnectionFactoryTest extends TestCase
             FakeDriver::$exception = null;
         }
     }
+
+    public function testDefaultCharset()
+    {
+        $factory = new ConnectionFactory([]);
+        $params  = ['driverClass' => FakeDriver::class];
+
+        $connection = $factory->createConnection($params);
+
+        $this->assertSame('utf8', $connection->getParams()['charset']);
+    }
+
+    public function testDefaultCharsetMySql()
+    {
+        $factory = new ConnectionFactory([]);
+        $params  = ['driver' => 'pdo_mysql'];
+
+        $connection = $factory->createConnection($params);
+
+        $this->assertSame('utf8mb4', $connection->getParams()['charset']);
+    }
 }
 
 /**
--- a/UPGRADE-2.0.md
+++ b/UPGRADE-2.0.md
@@ -9,6 +9,8 @@ PHP and Symfony version support
    Symfony 3.4 LTS and 4.1 or newer.
  * Support for Twig 1.34 and below as well as 2.4 and below (for 2.x) releases
    was dropped.
+ * When no charset parameter is defined, it now defaults to `utf8mb4` on the
+   MySQL platform and to `utf8` on all other platforms.
 
 Commands
 --------


nicolas-grekas:
> Shouldn't this be contributed directly in the DBAL, inside DriverManager::getConnection()?

That is not required, and that is a project with policies that are totally decoupled from Symfony.
I agree it would be great to make this the default of DBAL also, but that's not the goal here.

weaverryan:
I would love to have this out of use-land config. For 2.0 would have been better, but 2.0.1 I think works in practice to finally remove the config need.

alcaeus:
I wasn't aware of this problem, so doing this in 2.0.1 makes sense to me. Thanks @nicolas-grekas!

ConnectionFactory.php
```diff
@@ -43,6 +44,22 @@ public function createConnection(array $params, Configuration $config = null, Ev
...
+
+            if ($driver instanceof AbstractMySQLDriver) {
+                $params['charset'] = 'utf8mb4';
+
+                if (! isset($params['defaultTableOptions']['collate'])) {
+                    $params['defaultTableOptions']['collate'] = 'utf8mb4_unicode_ci';
+                }
+            }
+
+            $connection = new $connection($params, $driver, $connection->getConfiguration(), $connection->getEventManager());
```
stof: I'd rather use `DriverManager::getConnection` again rather than assuming the constructor argument here

nicolas-grekas: That's the "contract" of all connection classes: the wrapper_class parameter already requires this signature.

greg0ire:
@ggeorgiev-sfly both things were expected when writing the PR, and approved, because we made the (wrong) assumption that no heavy lifting would take place here… it doesn't look like it's the case: https://github.com/doctrine/dbal/blob/7be002cc478ebf70dc9c261b9e8dec225e9d8c2a/lib/Doctrine/DBAL/Connection.php#L163-L195 :sweat_smile: @nicolas-grekas @alcaeus maybe we should revert this?

nicolas-grekas:
@greg0ire it would be better to understand what happens.

greg0ire:
My understanding is that @ggeorgiev-sfly overrode the wrapper class thanks to this option: https://github.com/doctrine/dbal/blob/7be002cc478ebf70dc9c261b9e8dec225e9d8c2a/lib/Doctrine/DBAL/DriverManager.php#L171-L178

But because of https://github.com/doctrine/DoctrineBundle/pull/1070/files#diff-5b0dbf30dd0e8ce451fd72020914e62cR60 , the constructor is called of the wrapper class is now called twice. While this looks fine for https://github.com/doctrine/dbal/blob/7be002cc478ebf70dc9c261b9e8dec225e9d8c2a/lib/Doctrine/DBAL/Connection.php , it may not always be fine for any user-defined wrapper class.

nicolas-grekas:
Creating a connection to the backend in the constructor of the wrapped class is a critical issue with that constructor. I'm not sure we should consider this as legit.

greg0ire:
@ggeorgiev-sfly can you please give more details about what you are doing here? We suspect there might be a better place for you to do what you need than in the constructor of the connection.

ggeorgiev-sfly:
Thanks for the quick replies, here are some details regarding my case:
We have a large MySQL db with total of 2050 databases - 1 global master, 1 global slave, 1024 shards master, each having a slave replication. As I cannot inject any custom configuration to the doctrine, cause any extra keys throw an exception - I have to generate a connection data for each single shard, resulting over 8000-9000 lines of yaml configuration just for the shard config. And that is not accounting for the replication, just the shards.
Doctrine currently doesn't have master-slave replication on sharded DBs out of the box, so I have to implement that manually.
The heavy lifting is not like I'm creating driver connections in the constructor, but more like processing and validating all that global & shard data - to be stored in the connection class so they can be switched easily when needed.
I'm creating a custom connection wrapper class, based off the `PoolingShardConnection`. I saw there that a lot of configuration copying is taking place - meaning all global & shard configuration is copied over and over again on each of the shards, resulting 1024 shards each having the configuration for all other shards from this line: `$this->connections[$shard['id']] = array_merge($params, $shard);`
I didn't like that big array to be replicated everywhere so I decided to keep only the basic connection data - only what's needed for the connection in the local array - like host, port, user, pass, db, charset.
When I did that - I got an exception that `global` and `shards` are not found - and this is when I found out that the Connection class is instantiated twice. I don't really see a practical need to instantiate a class twice. Why don't you just configure the charset before the first instantiation?
In the constructor I try to validate the all the shard and global data is present, and this means validating more than 1000 entries for valid connection data, which I really like not to do twice.
Of course I can just save what I got into the constructor and validate later on, but still calling the constructor twice just to set the charset the second time seems really unnecessary for me here.
My case may be more extreme. Here are some things I like to be different:
1. Allow custom configuration for the dbal, so we can extend the Doctrine functionality for custom needs more easily.
2. The `$params` array initially passed to the Connection wrapper is stripped from `charset` config - no matter where I try to put it - it never reaches the class initially. Why we're removing that? If we allow to predefine the charset - even with that addition it won't create the connection second time.
3. I think we can avoid instantiating new Connection class just to get the parameters we instantiated it with. Seems illogical to do so. We already have all the data we instantiate the class with.

nicolas-grekas:
Fixed in #1087 

@ggeorgiev-sfly btw, if you have an ideas to improve eithe the bundle or dbal, please send some PRs :)

alcaeus:
@ggeorgiev-sfly can you please confirm the PR fixes your issue? Thanks!

ggeorgiev-sfly:
@alcaeus @nicolas-grekas 
Yes it does fix the double instantiation of the wrapper class. 
It does instantiate the original Connection class and then the wrapper one, which after than uses the original class again, making it execute it __construct twice (the original Connection one). Still I wonder why we cannot set charset before that and avoid that completely?
I mean (as I mentioned above) - I couldn't find a way to pass that 'charset' in the $params that go into `createConnection` initially. Is there a way to do that so I can skip the `if` completely as I have specified the `charset` everywhere?

nicolas-grekas:
The database url can contain the charset too, so we need to parse it before anything.

rosier:
I like to offer an alternative solution #1098 by setting the default charset in the DI

Issue #1069, opened 2019-11-25
Metadata cache no longer clears in dev environment


weaverryan:
Hi!

I believe @alcaeus is alread aware of the issue, but I couldn't find an issue for it yet. This relates also to #1066.

Basically, the new cache changes have a flaw in them: the metadata is being cached in all environments, which requires a cache clear to see Doctrine metadata changes. The easiest path is to "backup" to some sort of an "array" cache for all environments except for "prod".

alcaeus:
Yeah, I was a bit too optimistic when leaving in the default config. By default, the bundle uses the `app` and `system` cache pools. These are configured with sensible defaults, thus also preventing performance issues in dev.

There are a couple of approaches with these defaults:
* In the dev and test environments, metadata is cached aggressively in the `system` cache pool. This requires a complete clear of the system cache pool to reload metadata
* In the test environment, the `app` cache pool is used to cache results. This can create issues between subsequent test runs, as it may reuse result caches from a previous test run.

With that in mind, I would suggest the following:
* For dev and test environments, resort to using an `array` cache adapter for the metadata and result pools. From what I can tell, the query cache may remain in the `system` cache pool, but it can also be changed for consistence
* In the prod environment, use the current defaults (`app` pool for results, `system` for queries and metadata).

dmaicher:
I would propose to restore the previous behavior on DoctrineBundle:

- we won't use system/app cache pools by default
- if no pool is configured we should create new pools that use array adapters (similar to the previously used default `ArrayCache`)

And then adjust the recipe for `prod` to use the system/app pools (similar to how it was done before).

People might have more different environments and I'm not sure we should enable persistent caching by default on those. I would let the users decide to overwrite the default array caching if needed :blush:

alcaeus:
That sounds like a plan. Can you make those changes? I believe 1.12 might be accepted as well if people leave cache config empty.

dmaicher:
I will look into it today :+1:

dmaicher:
New releases 1.12.1 and 2.0.1 are fixing this issue.

Issue #1066, opened 2019-11-24
BC break for cache pool configuration?


dmaicher:
I just tried upgrading a bigger app to 1.12.0 and noticed a few tests were failing.

Some debugging showed that now suddenly there is a persistent result cache used :see_no_evil: 

To me this here seems to be a BC break:

https://github.com/doctrine/DoctrineBundle/compare/1.11.2..1.12.0#diff-8f3578ddca2c59937b37c3331b2ab336R743

The previous behavior was to use an `array` cache in those cases where no cache was configured.

Now we are using the defined `cache.app` pool which by default is using FileSystemAdapter (so persistent data written to disk).

Should we use some new pool here that has an `ArrayAdapter` to keep the previous behavior?

With 1.11.2:
```
$ app/console debug:container doctrine.orm.default_result_cache --show-arguments --env=test

 // This service is a public alias for the service                                                                      
 // doctrine_cache.providers.doctrine.orm.default_result_cache                                                          

Information for Service "doctrine_cache.providers.doctrine.orm.default_result_cache"
====================================================================================

 Array cache driver.

 ---------------- ------------------------------------------------------------ 
  Option           Value                                                       
 ---------------- ------------------------------------------------------------ 
  Service ID       doctrine_cache.providers.doctrine.orm.default_result_cache  
  Class            Doctrine\Common\Cache\ArrayCache                            
  Tags             -                                                           
  Calls            setNamespace                                                
  Public           yes                                                         
  Synthetic        no                                                          
  Lazy             no                                                          
  Shared           yes                                                         
  Abstract         no                                                          
  Autowired        no                                                          
  Autoconfigured   no                                                          
 ---------------- ------------------------------------------------------------ 
```

With 1.12.0:
```
$ app/console debug:container doctrine.orm.default_result_cache --show-arguments --env=test

 // This service is a private alias for the service doctrine.orm.cache.pool.cache.app                                   

Information for Service "doctrine.orm.cache.pool.cache.app"
===========================================================

 ---------------- ------------------------------------------ 
  Option           Value                                     
 ---------------- ------------------------------------------ 
  Service ID       doctrine.orm.cache.pool.cache.app         
  Class            Symfony\Component\Cache\DoctrineProvider  
  Tags             -                                         
  Public           no                                        
  Synthetic        no                                        
  Lazy             no                                        
  Shared           yes                                       
  Abstract         no                                        
  Autowired        no                                        
  Autoconfigured   no                                        
[...]

PR #1073, opened 2019-11-26
Fix: use array adapter cache pools by default

dmaicher:
This fixes https://github.com/doctrine/DoctrineBundle/issues/1069 and https://github.com/doctrine/DoctrineBundle/issues/1066

It restores the previous behavior from 1.11 where we use an array cache by default if no other cache is configured.

The cache pools will also show up in the profiler:

pools

TODOs:
- [x] add more tests
- [x] tested on big work project
- [ ] adjust recipe for 1.12+ to use proper cache for prod environment


DependencyInjection/DoctrineExtension.php
```diff
@@ -906,13 +907,14 @@ private function createPoolCacheDefinition(ContainerBuilder $container, string $
...
     {
-        switch ($driverName) {
-            case 'metadata_cache':
-                return 'cache.system';
-            default:
-                return 'cache.app';
-        }
+        $id = sprintf('cache.doctrine.orm.%s.%s', $objectManagerName, str_replace('_cache', '', $cacheName));
+
+        $poolDefinition = $container->register($id, ArrayAdapter::class);
```
dmaicher: as far as I understand the cache component will always be installed, right?

Since 3.x it seems to be a hard dependency for the framework-bundle: https://github.com/symfony/symfony/blob/3.4/src/Symfony/Bundle/FrameworkBundle/composer.json#L21

DependencyInjection/DoctrineExtension.php
```diff
@@ -906,13 +907,14 @@ private function createPoolCacheDefinition(ContainerBuilder $container, string $
...
-    private function getPoolNameForCacheDriver(string $driverName) : string
+    private function createArrayAdapterCachePool(ContainerBuilder $container, string $objectManagerName, string $cacheName) : string
     {
-        switch ($driverName) {
-            case 'metadata_cache':
-                return 'cache.system';
-            default:
-                return 'cache.app';
-        }
+        $id = sprintf('cache.doctrine.orm.%s.%s', $objectManagerName, str_replace('_cache', '', $cacheName));
```
dmaicher: every object-manager needs separate caches, right? Otherwise there could be weird key collisions I think

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -775,38 +778,38 @@ public static function cacheConfigurationProvider() : array
         return [
             'metadata_cache_default' => [
                 'expectedAliasName' => 'doctrine.orm.default_metadata_cache',
-                'expectedAliasTarget' => 'doctrine.orm.cache.pool.cache.system',
+                'expectedAliasTarget' => 'doctrine.cache.provider.cache.doctrine.orm.default.metadata',
```
dmaicher: Not 100% happy with this service name :confused: But we use the pool name in the id and prefix it...

Any better suggestions?

DependencyInjection/DoctrineExtension.php
```diff
@@ -897,7 +898,7 @@ private function createPoolCacheDefinition(ContainerBuilder $container, string $
             throw new LogicException('Using the "pool" cache type is only supported when symfony/cache is installed.');
         }
 
-        $serviceId = sprintf('doctrine.orm.cache.pool.%s', $poolName);
+        $serviceId = sprintf('doctrine.cache.provider.%s', $poolName);
```
alcaeus:
```suggestion
        $serviceId = sprintf('doctrine.orm.cache.provider.%s', $poolName);
```

DependencyInjection/DoctrineExtension.php
```diff
@@ -906,13 +907,14 @@ private function createPoolCacheDefinition(ContainerBuilder $container, string $
...
-            case 'metadata_cache':
-                return 'cache.system';
-            default:
-                return 'cache.app';
-        }
+        $id = sprintf('cache.doctrine.orm.%s.%s', $objectManagerName, str_replace('_cache', '', $cacheName));
+
+        $poolDefinition = $container->register($id, ArrayAdapter::class);
+        $poolDefinition->addTag('cache.pool');
+        $container->setDefinition($id,  $poolDefinition);
```
alcaeus:
```suggestion
        $container->setDefinition($id, $poolDefinition);
```

DependencyInjection/DoctrineExtension.php
alcaeus: I'm not sure what the EM uses as key for the entry, but this is definitely safer than using shared caches.

DependencyInjection/DoctrineExtension.php
alcaeus: If we do this, we should definitely put a hard requirement on symfony/cache in composer.json. Since it's a direct dependency, we should own it.

dmaicher: done :+1:

greg0ire: I just tried `composer create-project symfony/skeleton my_project`, and it installed the `symfony/cache` indeed.

greg0ire:
Please kindly squash your commits together. If you don't, we'll try to remember to do it for you but it's best if you save us this trouble.

How to do that?

1. `git rebase -i origin/1.12.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.
2. A window will show up with many lines, replace `pick` with `fixup` on every line but the first one
3. Close your editor, git should do its magic, and you should end up with one commit
4. Use `git push --force` to overwrite what you already push. Don't forget the `--force` option otherwise git will try to merge both things together.

DependencyInjection/DoctrineExtension.php
alcaeus: @greg0ire symfony/cache is a transitive dependency through FrameworkBundle. However, I prefer to make such dependencies explicit if we use classes from them. This allows us to be in control of versioning.

greg0ire: Yeah I fully agree, we had issue on Sonata bundles because we didn't spot such transitive dependencies we used ourselves: https://github.com/sonata-project/SonataAdminBundle/pull/5759

I was just confirming that it was not a big deal to make that optional dependency mandatory.

alcaeus:
@dmaicher build is failing due to a deprecation in the 4.4 job. Do you want to take a look at that in a separate PR? Otherwise I'll do so in the afternoon.

dmaicher:
> @dmaicher build is failing due to a deprecation in the 4.4 job. Do you want to take a look at that in a separate PR? Otherwise I'll do so in the afternoon.

@alcaeus yeah I also saw it. Not sure I have time for it today :cry:

alcaeus:
Don't worry: https://github.com/doctrine/DoctrineBundle/pull/1075. Just need to rebase :)

alcaeus:
Thanks David!

dmaicher:
Here is the recipe change: https://github.com/symfony/recipes/pull/695

dmaicher:
@alcaeus @greg0ire @nicolas-grekas when we release this it will possibly cause some issues for users that don't also sync the latest recipe change (https://github.com/symfony/recipes/pull/695)

It will leave them without proper caching for prod environments :cry: Not sure how we can avoid this though...

But maybe chances are high that new projects started with fresh 1.12/2.0 recipes are not running in production yet? :smile: :see_no_evil:

OskarStark:
I think we cannot avoid it, but before it was an implicit dep on my project, so I ended up automatically using 2.0 via composer update. I had to manually require 1.x. But nevermind we need to fix it and help the Community to get a working state back again 👍🏻

alcaeus:
Fixing this is not a question of if, but how. The current upgrade path may cause some pain for people in production if they don't upgrade their flex recipes or don't update their config manually, but I believe we can account for that by communicating the change appropriately.

PR #1077, opened 2019-11-28
Default to real-utf8 when no charset is provided

alcaeus:
This is a back port of #1070 since that was inadvertently merged to master only.

### DIFF
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -6,6 +6,7 @@ use Doctrine\Common\EventManager;
 use Doctrine\DBAL\Configuration;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DBALException;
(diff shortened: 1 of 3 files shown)


alcaeus:
> Let's change the default branch if it's wrong? (on my phone, hard to check)

No, this was just a case of @nicolas-grekas and I not realising that the bug should be fixed in 2.0.1. New features should still go to `master`, so I don't think changing the default branch is a smart idea.

greg0ire:
I didn't realise master was aliased to 2 and not 3. Not a smart idea indeed. Not sure why it's not like in ORM or DBAL though. Also, I see that the alias is 2.0, it should probably be 2.1, right?

alcaeus:
I haven't updated the alias yet. The alias will change to 3.0 once we start working on 3.0. Until then, it can stay at 2.x (pointing to the next minor release): we're not merging PRs with BC breaks..

Issue #1067, opened 2019-11-25
Using the element "deprecated" is deprecated


BlackWiCKED:
Got these 3 deprecation warnings recently:

`Using the element "deprecated" is deprecated for the service "Symfony\Bridge\Doctrine\RegistryInterface" which is defined as an alias in ".../vendor/doctrine/doctrine-bundle/DependencyInjection/../Resources/config/dbal.xml". The XmlFileLoader will raise an exception in Symfony 4.0, instead of silently ignoring unsupported elements.`

`Using the element "deprecated" is deprecated for the service "Doctrine\Bundle\DoctrineBundle\Registry" which is defined as an alias in ".../vendor/doctrine/doctrine-bundle/DependencyInjection/../Resources/config/dbal.xml". The XmlFileLoader will raise an exception in Symfony 4.0, instead of silently ignoring unsupported elements.`

`Using the element "deprecated" is deprecated for the service "Doctrine\Common\Persistence\ObjectManager" which is defined as an alias in ".../vendor/doctrine/doctrine-bundle/DependencyInjection/../Resources/config/orm.xml". The XmlFileLoader will raise an exception in Symfony 4.0, instead of silently ignoring unsupported elements.`

I'm using Symfony 3.4.35.

BlackWiCKED:
Possible solution: I removed the affected lines from the config files. Not sure how they got there, but the suggested replacements were already added so it felt a safe thing to do - and it worked well. Maybe it's just project specific thing, in that case ignore my report.

These were the lines I had to remove:

1. /Resources/config/orm.xml
```
<service id="Doctrine\Common\Persistence\ObjectManager" alias="doctrine.orm.entity_manager" public="false">
    <deprecated>The "%alias_id%" service alias is deprecated, use `Doctrine\ORM\EntityManagerInterface` instead.</deprecated>
</service>
```

2. /Resources/config/dbal.xml
```
<service id="Symfony\Bridge\Doctrine\RegistryInterface" alias="doctrine" public="false">
    <deprecated>The "%alias_id%" service alias is deprecated, use `Doctrine\Common\Persistence\ManagerRegistry` instead.</deprecated>
</service>
<service id="Doctrine\Bundle\DoctrineBundle\Registry" alias="doctrine" public="false">
    <deprecated>The "%alias_id%" service alias is deprecated, use `Doctrine\Common\Persistence\ManagerRegistry` instead.</deprecated>
</service>
```

[...]

PR #1079, opened 2019-11-28
Deprecated alias on DI Extension

maxhelias:
Fix issue #1067

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -92,6 +92,11 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('dbal.xml');
 
+        if (method_exists(Alias::class, 'setDeprecated')) {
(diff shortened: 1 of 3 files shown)


maxhelias:
@stof Can you tell me if it's ok like that ?

DependencyInjection/DoctrineExtension.php
```diff
@@ -92,6 +92,12 @@ protected function dbalLoad(array $config, ContainerBuilder $container)
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('dbal.xml');
 
+        // Deprecated
```
alcaeus: The comment isn't necessary ;)
```suggestion
```

DependencyInjection/DoctrineExtension.php
```diff
@@ -348,6 +354,11 @@ protected function ormLoad(array $config, ContainerBuilder $container)
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('orm.xml');
 
+        // Deprecated
```
alcaeus: The comment isn't necessary ;)
```suggestion
```

alcaeus:
Thanks @maxhelias!

stof:
@alcaeus I suggest making a release for that, to avoid getting lots of bug reports for that from people using stable releases. What do you think ?

alcaeus:
Done.

PR #1082, opened 2019-12-01
Deprecate doctrine:mapping:import

ostrolucky:
Follow-up from https://github.com/doctrine/DoctrineBundle/pull/1054


maxhelias:
A deprecated message could also be trigger in the `execute` method, wdyt ?

ostrolucky:
It could. Feel free to contribute this if you want. I was thinking about it but didn't consider it vital.

Issue #949, opened 2019-04-06
Deprecate class parameters


alcaeus:
With DoctrineBundle 2.0, we'll be dropping the `.class` parameters from the service configuration (see #630). In order for people to be aware of that, we should document the upcoming removal in 1.11.

SenseException:
Do you have a place on the documentation page in mind where this should be mentioned in?

ostrolucky:
Perhaps this is just a question of adding it to upgrade guide?

alcaeus:
Upgrade guide would be a good place. Note that a whole bunch are still in and most likely require cleanup in DoctrineBridge first.

PR #1084, opened 2019-12-07
Deprecate .class parameters

ostrolucky:
Fixes https://github.com/doctrine/DoctrineBundle/issues/949.

Btw I was trying to find how other people deprecated this and found that nobody actually did, not even Symfony. Not even with readme entry, like we are doing here.

### DIFF
--- a/UPGRADE-2.1.md
+++ b/UPGRADE-2.1.md
@@ -5,3 +5,59 @@ Commands
 --------
 
  * `Doctrine\Bundle\DoctrineBundle\Command\ImportMappingDoctrineCommand` has been deprecated  
+
+Parameters
+----------
+
+Following parameters are deprecated and will no longer be defined nor consumed in future.
+Redefine/decorate services where they are used in via DI configuration instead.
+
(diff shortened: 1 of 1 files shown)


alcaeus:
Putting this on hold: there were some class parameters that were necessary due to the Doctrine Bridge, otherwise they would've been removed in 2.0 already. We'll check that out separately, but for now those don't hurt anyways.

ostrolucky:
Doctrine bridge PR was merged right? It's possible something was missed there?

ostrolucky:
@alcaeus Can you elaborate what more needs to be done in Doctrine Bridge, or what is blocking this?

alcaeus:
So the bridge is still using the class parameters for cache, but that code shouldn't be called anymore. I thought there were more but I'm not seeing them, so we could just remove all of them and see if anything breaks 🤷‍♂️

ostrolucky:
So we ecan merge this?

alcaeus:
Thanks @ostrolucky!

PR #1085, opened 2019-12-07
Allow non-scalar DBAL options

### LINKED ISSUE #365: Connection/Driver options does not support arrays
I tried adding "userDefinedFunctions" for a sqlite db via configuration.

While the DoctrineBundle maps the `options` field of the configuration to the correct parameter in the `PDOSqlite\Driver`'s `connect` method, it does not allow arrays therefore it's not useable for my use-case.

I handled this by overwriting `driver_class` and extending the connect method to append the userDefinedFunctions but i still think it should be supported via configuration.

Please don't confuse this with extending dql function which is already supported by the configuration.

ostrolucky:
DBAL actually accepts mixed[], not string[] as we force. This causes problems with some options, such as `userDefinedFunctions`.

Fixes #365

### DIFF
--- a/Tests/DependencyInjection/Fixtures/config/xml/dbal_service_multiple_connections.xml
+++ b/Tests/DependencyInjection/Fixtures/config/xml/dbal_service_multiple_connections.xml
@@ -21,7 +21,9 @@
                 user="sqlite_user"
                 password="sqlite_s3cr3t"
                 path="/tmp/db.sqlite"
-                memory="true" />
+                memory="true">
+                <option key="userDefinedFunctions" />
+            </connection>
             <connection
                 name="oci"
                 driver="oci8"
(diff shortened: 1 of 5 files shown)


alcaeus:
Can you please add a test for this? Thanks!

ostrolucky:
Come on it's just configuration, that would be like testing getter/setter

alcaeus:
This is to prevent regressions in the future, and it's just configuration, so it's easy to write.

ostrolucky:
done

Tests/DependencyInjection/Fixtures/config/xml/dbal_service_multiple_connections.xml
```diff
@@ -21,7 +21,9 @@
                 user="sqlite_user"
                 password="sqlite_s3cr3t"
                 path="/tmp/db.sqlite"
-                memory="true" />
+                memory="true">
+                <option key="userDefinedFunctions" />
```
alcaeus: The XSD currently doesn't allow defining anything other than a scalar value for an option. I'm looking at how to fix this.

alcaeus: I was able to fix this with some help from @dkarlovi and @wouterj. Thanks to both!

ostrolucky:
Thx for fixing XSD @alcaeus. LGTM

alcaeus:
Thanks @ostrolucky!

PR #1086, opened 2019-12-07
Activate selective stale probot

ostrolucky:
I'm strongly opposed to adding bot which automatically closes issues without human involvement, but here it's configured to require certain label first, so it's involvement will be here only in cases we manually mark as such.

Idea here is to post warning in issues labelled as "Status: Waiting feedback"
if there is no activity after 14 days and after no further activity within 7 days happen, closes it.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,8 @@
+onlyLabels: ['Status: Waiting feedback']
+daysUntilStale: 14
+daysUntilClose: 7
+staleLabel: 'Stale'
+markComment: >
+  This issue is still waiting on feedback. Please provide the information requested above to allow us to continue working on this ticket.
+
+  This issue will be closed after 7 days. You can still provide us with the necessary information to continue working on this issue, and we will reopen the ticket for you. Thanks!


alcaeus:
We've used the `stale` app in the mongodb-odm repositories for a while now. You can see the config that we've used [here](https://github.com/doctrine/mongodb-odm/blob/2e1d0bfd4b325449030c0bfde2ebd273cefba10a/.github/stale.yml). We've excluded some labels but left it to mark all issues as stale and later close them.

My only problem is that you can only have a single configuration, so you can't set different waiting times for questions and bug reports, etc. There is a new GitHub action to close stale issues that we could check out: https://github.com/marketplace/actions/close-stale-issues. Since this is configured as a job, there can be multiple jobs of the same type with different configurations. So the first step would be to ping back once on questions and then close them if there is no feedback. For bugs, you want to be less aggressive and only ask back if a "Waiting for feedback" label is applied. This would allow us to be more selective about how we treat issues.

Would you like to give this a look, @ostrolucky?

ostrolucky:
I don't want to mark issues as stale automatically, that's pretty obvious from last update time, so label doesn't provide value. All we need is to warn users in issues with certain label that if they keep ignoring request to provide more information, issue will be closed and then close it.

On first look, it doesn't look like referenced GitHub action can make this work. It looks like it considers issues with certain labels as stale, while I want it to consider issues with no response after certain time of applying label as stale.

alcaeus:
> On first look, it doesn't look like referenced GitHub action can make this work. It looks like it considers issues with certain labels as stale, while I want it to consider issues with no response after certain time of applying label as stale.

From the description:

> Warns and then closes issues and PRs that have had no activity for a specified amount of time.

After a certain time with no response, the bot posts a warning and adds a label to the issue/PR. After another amount of time, the issue is closed. You can also close before marking as stale and be done with it. Or you can only mark as stale and not close automatically. I'm not sure if you can avoid adding a label, but the label to be added is configurable (so "Waiting for feedback" could be added automatically after waiting for a response)

ostrolucky:
Again, intend is to do this after certain amount of applying specific label only. GitHub action doesn't seem to have label filter. I need it to be inactive for issues not marked as feedback requested.

.github/stale.yml
```diff
@@ -0,0 +1,9 @@
+onlyLabels: ['Status: Waiting feedback']
+daysUntilStale: 14
+daysUntilClose: 7
+staleLabel: 'Won''t Fix'
+markComment: >
+  We requested feedback a while ago and as of now it's still not clear to us what needs to be done.
```
alcaeus: Suggested rewording:

> This issue is still waiting on feedback. Please provide the information requested above to allow us to continue working on this ticket.

This issue will be closed after 7 days. You will still be able to reopen the issue and provide us with the information needed to work on this issue.

ostrolucky: I don't think they will be able to reopen the issue. I will rather leave out that part.

alcaeus: They will be, and they should if they want to follow up with the information they requested.

stof:
I agree with @ostrolucky about not wanting this to be applied to all issues.
As someone reporting issues, I hate it when a bot marks issues as stale not because they are awaiting feedback from me, but because the maintainer never looked at them.

.github/stale.yml
ostrolucky: IIRC they can open new issue, not reopen their issue when somebody else closes it

alcaeus: They can reopen their issue just fine. Issue has to be locked to prevent reopening, which we generally don’t do. With history already in the ticket, I’d advise against requesting OP open a new ticket for the same issue but rather reopen if new information comes up.

stof: IIRC, when an issue is closed by a collaborator, only collaborators can reopen them, not the issue author. They can reopen only what they closed themselves (unless Github changed that in the past few months, as I haven't tried it again recently).

.github/stale.yml
```diff
@@ -0,0 +1,9 @@
+onlyLabels: ['Status: Waiting feedback']
+daysUntilStale: 14
+daysUntilClose: 7
+staleLabel: 'Won''t Fix'
```
stof: I'd rather use a `stale` label here to identify a stale issue awaiting feedback than `Won't fix`, which means something else.

.github/stale.yml
ostrolucky: This. Users can not really reopen issue when somebody else closes it, that's misconception. Plenty times my issue was closed and I had to open new issue.

.github/stale.yml
ostrolucky: Fine by me

.github/stale.yml
alcaeus: Welp, I was sure this would work.

> This issue is still waiting on feedback. Please provide the information requested above to allow us to continue working on this ticket.
>
> This issue will be closed after 7 days. You can still provide us with the necessary information to continue working on this issue, and we will reopen the ticket for you. Thanks!

ostrolucky:
Suggested changed applied.

alcaeus:
Thanks @ostrolucky!

PR #1087, opened 2019-12-10
Fix creating wrapped connection twice

nicolas-grekas:
As explained in https://github.com/doctrine/DoctrineBundle/pull/1070#issuecomment-563804621

### DIFF
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -11,6 +11,7 @@ use Doctrine\DBAL\DriverManager;
 use Doctrine\DBAL\Exception\DriverException;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\Type;
+use function is_subclass_of;
 
 class ConnectionFactory
 {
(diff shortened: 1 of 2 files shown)


ConnectionFactory.php
```diff
@@ -42,22 +43,34 @@ public function createConnection(array $params, Configuration $config = null, Ev
...
         }
 
-        $connection = DriverManager::getConnection($params, $config, $eventManager);
-
         if (! isset($params['pdo']) && ! isset($params['charset'])) {
-            $params            = $connection->getParams();
-            $params['charset'] = 'utf8';
-            $driver            = $connection->getDriver();
+            $wrapperClass = Connection::class;
+            if (isset($params['wrapperClass'])) {
```
ostrolucky: Not a fan of inlining this logic here, that's what's DriverManager for. Other places using it are still having old behaviour and in case we update DriverManager, it can break ConnectionFactory, since it unsets wrapperClass option.

ostrolucky (changes requested):
Seems this doesn't fully solve original problem. Also I would say tests are required here.

nicolas-grekas:
Test added.

About your other comments @ostrolucky I didn't understand what you're suggesting instead. To me, this effectively solves the reported issue. And I can't think of any alternative implementation. Do you?

ostrolucky:
My point was it should be fixed in `DriverManager::getConnection`, hence in DBAL. I agree with @greg0ire. #1070 should not have been merged, I would prefer that reverted. DoctrineBundle is a glue and shouldn't act as workaround for issues in DBAL. Problems should not be removed for Symfony users only.

Besides this being ugly fix (34, poorly tested LOC with half code copy pasted; which could be < 10 in DBAL), it still instantiates original Connection class twice so by avoiding fixing it in DBAL we are implementing subpar solutions.

I believe #1070 causes other unreported issues as well. Since we are no longer touching recipe which is generated only at beginning of installing bundle, it changes charset for everybody not specifying one from usually default `latin1` for MySQL < 8 into `utf8mb4` which might corrupt new data. 

Perhaps we can also just ignore this problem for a little while longer, since MySQL 8 fixed this itself by defaulting to `utf8mb4` and nobody should start new project with MySQL 5/

ostrolucky (dismissed):
Test case is incorrect. It tests that driver was instantiated once, while it should test connection wrapper is instantiated once only. Driver was never a problem.

nicolas-grekas:
> It tests that driver was instantiated once, while it should test connection wrapper is instantiated once only.

Indeed, fixed.

ConnectionFactory.php
alcaeus: I agree this should be ported to DriverManager, but fixing this here was the faster choice for Symfony users. Once this is fixed in DBAL 2.9, we can remove it here and bump the doctrine/dbal dependency.

alcaeus (approved):

alcaeus:
Thanks @nicolas-grekas!

PR #1095, opened 2019-12-14
Fix doctrine deprecations

ostrolucky:
Rel: https://github.com/doctrine/orm/pull/7953, https://github.com/doctrine/persistence/pull/71

This will make our build after upmerging it clean.

Other than new Persistance deprecations, I have also fixed some extra ones I noticed, such as deprecated `Version` class (we always require ORM in require-dev, so in test cases I just removed its usage, otherwise I used UnitOfWork - EntityManagerInterface triggers deprecation)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,7 @@
         "php": "^7.1",
         "doctrine/dbal": "^2.5.12",
         "doctrine/doctrine-cache-bundle": "~1.2",
+        "doctrine/persistence": "^1.3.3",
         "jdorn/sql-formatter": "^1.2.16",
         "symfony/cache": "^3.4.30|^4.3.3",
         "symfony/config": "^3.4.30|^4.3.3",
--- a/Command/DoctrineCommand.php
+++ b/Command/DoctrineCommand.php
@@ -2,11 +2,11 @@
 
 namespace Doctrine\Bundle\DoctrineBundle\Command;
 
-use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Sharding\PoolingShardConnection;
 use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\Tools\EntityGenerator;
+use Doctrine\Persistence\ManagerRegistry;
 use LogicException;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\DependencyInjection\ContainerInterface;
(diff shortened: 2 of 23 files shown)


composer.json
```diff
@@ -39,6 +39,7 @@
     "require-dev": {
         "doctrine/coding-standard": "^6.0",
         "doctrine/orm": "^2.6",
+        "doctrine/persistence": "^1.3.3",
```
alcaeus: This is not sufficient: people are still able to install a new version of this package with doctrine/persistence 1.2, causing fatal errors. Please move this requirement up to the `require` stage. Thanks!

ostrolucky: It's not ideal though, because orm is not requirement of this bundle and this package makes sense for orm only.

alcaeus: Then you need to add a `conflict` entry to avoid installing the new version with persistence 1.3.2 or older.

ostrolucky: Which will not work for orm < 2.7, as only 2.7 requires it. Are you up for bumping orm conflict version?

alcaeus: No. We have a conflict with doctrine/persistence 1.3 and lower because we use new class names that aren't in 1.2. Due to issues with 1.3 releases before 1.3.3, we should conflict with anything < 1.3.3. ORM has nothing to do with this. In fact, ORM 2.7 will not require persistence 1.3, so you'd effectively end up with no installable versions.

ostrolucky: That's my point. If we don't bump to ORM 2.7 and don't require doctrine/persistence ourselves, it will not be installed, even though we are using it in bundle.

ostrolucky: That's my point. If we don't bump to ORM 2.7 and don't require doctrine/persistence ourselves, it will not be installed, even though we are using it in bundle. In such case it doesn't matter if we add it to conflict section if it's not required by anything in the first place.

alcaeus: Except that when people install ORM 2.7, they might not have upgraded to persistence 1.3 yet, which breaks the logic contained in this PR. Hence either requiring persistence directly or adding a conflict.

TBH, I have no issues adding doctrine/persistence to this: it's lightweight enough that we can add it, and once we are done extracting DBALBundle, this bundle will be for ORM only and should have a hard requirement anyways.

ostrolucky: > Except that when people install ORM 2.7, they might not have upgraded to persistence 1.3 yet, which breaks the logic contained in this PR. 

I was suggesting to both bump orm conflict and add persistence conflict. But since we can't/don't want bump, I will do a require.

alcaeus: As long as there is no stable release with deprecations fixed, we can't bump the ORM conflict. https://github.com/doctrine/orm/pull/7953 will be merged to 2.8 only, so we need to bide our time here.

alcaeus:
Thanks @ostrolucky!

PR #1098, opened 2019-12-17
Default to charset utf8mb4 for mysql and utf8 otherwise

rosier:
I like to purpose an alternative for #1070 and fix of #1087

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -27,6 +27,7 @@ use Symfony\Component\Messenger\MessageBusInterface;
 use Symfony\Component\Messenger\Transport\Doctrine\DoctrineTransportFactory;
 use Symfony\Component\PropertyInfo\PropertyInfoExtractorInterface;
 use function class_exists;
+use function in_array;
 use function sprintf;
 
 /**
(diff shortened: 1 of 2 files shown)


DependencyInjection/DoctrineExtension.php
```diff
@@ -199,6 +200,14 @@ protected function getConnectionOptions($connection)
     {
         $options = $connection;
 
+        if (! isset($options['charset'])) {
+            if (isset($options['url'])) {
+                $options['charset'] = strpos($options['url'], 'mysql') === 0 ? 'utf8mb4' : 'utf8';
```
stof: the scheme might also be `pdo_mysql` or `mysqli` as that's the name of the DBAL drivers, which are also supported as schemes.

stof:
This PR should also remove the code introduced by #1070.

DependencyInjection/DoctrineExtension.php
rosier: The scheme mysqli also starts with mysql, so that should be fine.

I didn't account for pdo_mysql, I will look into that.

ostrolucky:
I still think we should just revert #1070 as this is no longer a problem in newer mysql versions.

nicolas-grekas:
I did not implement the charset defaults this way because it harcodes the list and is thus impermeable to future prefixes that DBAL might handle.
The current list is way longuer than what is proposed here, see:
https://github.com/doctrine/dbal/blob/f6312cb97c50eaa35999dc73894bd3fd23d7f60b/lib/Doctrine/DBAL/DriverManager.php#L50-L77

Maintaining this list in the bundle didn't look like a good idea to me.

nicolas-grekas:
OH, last but not least, this doesn't work with database DSN. That's the definitive decider that made me go for the runtime logic. This just cannot work.

alcaeus:
Agree with @nicolas-grekas here. Closing in favour of the other solution.

nicolas-grekas:
Thanks for giving it a try @rosier!

DependencyInjection/DoctrineExtension.php
rosier: I don't think pdo_mysql can be use as a scheme, parse_url doesn't work with it. So that leaves just mysql, mysql2 and mysqli as a scheme.
```
$url = 'pdo_mysql://dbuser:dbpassword@127.0.0.1:3306/dbname?charset=utf8';
var_dump(parse_url($url));

array(2) {
  ["path"]=>  string(51) "pdo_mysql://dbuser:dbpassword@127.0.0.1:3306/dbname"
  ["query"]=>  string(12) "charset=utf8"
}
```

PR #1099, opened 2019-12-17
Make manager registry resettable

### LINKED ISSUE #1083: ManagerRegistery should reset entity managers on kernel.reset
Follow up of https://github.com/symfony/symfony/issues/34720

[`ManagerRegistry`](https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/Doctrine/ManagerRegistry.php) is able to reset entity managers, but I think we missed wiring this capability with the `kernel.reset` tag so that EM are reset between requests in long-running loops.

Given the dependency on `LazyLoadingInterface`, this should be done only when the interface is found of course.

alcaeus:
Fixes #1083.

This also adds a dependency to the Symfony Service contracts to ensure we've got the ResettableInterface.

### DIFF
--- a/Registry.php
+++ b/Registry.php
@@ -7,11 +7,12 @@ use Doctrine\ORM\ORMException;
 use Psr\Container\ContainerInterface;
 use Symfony\Bridge\Doctrine\ManagerRegistry;
 use Symfony\Bridge\Doctrine\RegistryInterface;
+use Symfony\Contracts\Service\ResetInterface;
 
 /**
  * References all Doctrine connections and entity managers in a given Container.
  */
-class Registry extends ManagerRegistry implements RegistryInterface
+class Registry extends ManagerRegistry implements RegistryInterface, ResetInterface
 {
     /**
      * @param string[] $connections
@@ -163,5 +164,12 @@ class Registry extends ManagerRegistry implements RegistryInterface
         @trigger_error('getEntityManagerForClass is deprecated since Symfony 2.1. Use getManagerForClass instead', E_USER_DEPRECATED);
 
         return $this->getManagerForClass($class);
+    }
+
+    public function reset() : void
+    {
+        foreach ($this->getManagerNames() as $managerName) {
+            $this->resetManager($managerName);
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


Registry.php
```diff
@@ -164,4 +165,11 @@ public function getEntityManagerForClass($class)
 
         return $this->getManagerForClass($class);
     }
+
+    public function reset() : void
+    {
+        foreach ($this->getManagerNames() as $managerName) {
```
alexander-schranz: @alcaeus My array looks like this in `1.12.3`:

```
  ["default"]=>
  string(35) "doctrine.orm.default_entity_manager"
```

and ends so with:

> Doctrine %s Manager named "doctrine.orm.default_entity_manager" does not exist.

Exception.

any hint how this can happen?

alexander-schranz: the `bin/console debug:container --parameter doctrine.entity_managers` does also return it this way

alcaeus: See #1103.

stof:
should we force resetting the entity manager (which will unload all mapping metadata), or would clearing it be enough when it is still open (resetting is indeed necessary if the EM got closed by an error during the flush) ?

hugochinchilla:
Since the inclusion of this PR in the master branch I'm getting this error when running my tests:

```
LogicException: Resetting a non-lazy manager service is not supported. Declare the "doctrine.orm.default_entity_manager" service as lazy.
```

alcaeus:
@hugochinchilla can you try the latest 1.12.x branch and report back? That should fix your issue. I’ll merge up to 2.0.x later today.

hugochinchilla:
@alcaeus I can confirm than cherrypicking #1108 on top o 2.0.5 fixes my issue.

Also requiring `symfony/proxy-manager-bridge` solves the issue, I don't know how it does it. I just tried the solution to a [similar issue posted a year ago on StackOverflow](https://stackoverflow.com/questions/53857220/how-to-mark-a-doctrine-entity-manager-as-lazy-in-symfony-4-2).

I will lock to 2.0.2 until #1108 is merged on 2.0.x branch.

Thanks for your help.

alcaeus:
Thanks for confirming! I've released 1.12.6 and 2.0.6 containing the fix.

The reason why installing the proxy-manager-bridge works is because that makes your entity managers lazy, which avoids the exception that you ran into before. With the release, this is fixed without having lazy entity managers.

PR #1100, opened 2019-12-18
fix well known schema filter blacklist tables

bendavies:
These tables were backwards.


alcaeus:
@bendavies can you please take a look at the failing tests? Thanks!

bendavies:
@alcaeus done. it would be better if the tests tested the services one by one, which would have caught this bug, but there we go.

alcaeus:
> @alcaeus done. it would be better if the tests tested the services one by one, which would have caught this bug, but there we go.

Absolutely - do you want to make the change?

bendavies:
sure

bendavies:
@alcaeus done

alcaeus:
Thanks @bendavies!

PR #1102, opened 2019-12-18
Fix resetting of managers on kernel reset

alexander-schranz:
Not sure about this change but `bin/console debug:container --parameter doctrine.entity_managers`

does return in my case:

```php
    'default' => 'doctrine.orm.default_entity_manager',
```

and if the value instead of key is used it will end up in a 

> Doctrine ORM Manager named "doctrine.orm.default_entity_manager'" does not exist.

Exception

### DIFF
--- a/Registry.php
+++ b/Registry.php
@@ -168,7 +168,7 @@ class Registry extends ManagerRegistry implements RegistryInterface, ResetInterf
 
     public function reset() : void
     {
-        foreach ($this->getManagerNames() as $managerName) {
+        foreach ($this->getManagerNames() as $managerName => $serviceId) {
             $this->resetManager($managerName);
         }
     }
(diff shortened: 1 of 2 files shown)


alcaeus:
Thanks for the PR. I was working on a fix to be merged to the 1.12.x as that's the lowest supported branch where the issue appears. See #1103.

alexander-schranz:
@alcaeus oh sorry forget to change the target branch. Thank you for working on it 👍

alcaeus:
No, thank YOU for working on a fix. I was writing the test when I saw your PR, but it lacked tests so I continued. Had I known you were working on a test as well I'd have stopped, but as is rebasing would've taken more time. We'll have this fixed in a few minutes.

alexander-schranz:
@alcaeus Thx for the fast release :)

PR #1106, opened 2019-12-18
Skip resetting managers when proxy-manager is not installed

nicolas-grekas:
Fixes #1105

PR #1108, opened 2019-12-19
Fix resetting non-lazy managers

alcaeus:
See https://github.com/doctrine/DoctrineBundle/pull/1106#issuecomment-567178319.

This refines the logic for resetting managers on `kernel.reset`:
* Lazy managers will be reset through `resetManager`
* Non-lazy managers will be cleared using `EntityManager::clear`
* Uninitialised managers will be ignored

Also adds tests for all use-cases.

@ro0NL can you give this a test and see if it works? As much as I love releasing with our new tools, I'm afraid I'll run out of release numbers by the time this is fixed 😂

### DIFF
--- a/Registry.php
+++ b/Registry.php
@@ -9,7 +9,6 @@ use Psr\Container\ContainerInterface;
 use Symfony\Bridge\Doctrine\ManagerRegistry;
 use Symfony\Bridge\Doctrine\RegistryInterface;
 use Symfony\Contracts\Service\ResetInterface;
-use function interface_exists;
 
 /**
  * References all Doctrine connections and entity managers in a given Container.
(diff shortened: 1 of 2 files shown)


Registry.php
```diff
@@ -170,12 +169,25 @@ public function getEntityManagerForClass($class)
...
-        if (! interface_exists(LazyLoadingInterface::class)) {
+        foreach ($this->getManagerNames() as $managerName => $serviceId) {
+            $this->resetOrClearManager($managerName, $serviceId);
+        }
+    }
+
+    private function resetOrClearManager(string $managerName, string $serviceId) : void
+    {
+        if (! $this->container->initialized($serviceId)) {
             return;
```
nicolas-grekas: The method is not in the interface so it needs an extra check before calling it.

alcaeus: The type hint for the `$container` property in `ManagerRegistry` specifies this as a `Container`, not `ContainerInterface`. Similarly, the `resetService` method there doesn't check for this either. We should either fix the type hint and calls in `ManagerRegistry`, or restrict the constructor type hint in `Registry` to no longer accept any `ContainerInterface`.

Given the BC implications of changing the constructor signature in the bundle, I'd rather fix this in the bridge. I can create the PR if you agree.

nicolas-grekas: Ok thanks for double checking. No need for more on my side.

alcaeus:
Release will happen later today. I'll let some folks test this before doing it 😉

PR #1118, opened 2020-01-05
Fix desynchronized entity managers after kernel reset

### LINKED ISSUE #1114: Weird behaviour due to EM resets after update to 2.0.6
Hi,

PHP-PM used to manually reset or clear the entity manager after every request (https://github.com/php-pm/php-pm-httpkernel/blob/master/Bootstraps/Symfony.php#L168). Now it seems like Symfony is able to automatically do it thanks to the `kernel.reset` tag but we're seeing a very strange behaviour.

On the first request all is fine,
On the second request, the EntityManager is reset at the beginning of the request, but the first ORM query is still going to the old EntityManager. The second ORM query sees that the manager has been reset, and instantiates another EM/UnitOfWork, effectively clearing the entityStates cache, which is bad, because in my code I have an entity created by the first query that the newly created UnitOfWork doesn't know anything about. If it is persisted, it will try an INSERT, instead of an UPDATE :(

To be honest, I don't really understand why the first ORM query after the reset is still using the old EM/UnitOfWork.

In any case, the EM shouldn't be reset after every request, a simple clear should be enough unless the EM hasn't been closed due to an Exception. In that case, yes, the manager should be reset. We've been working with that code in PHP-PM for a very long time and it works pretty well.

```php
            foreach ($doctrineRegistry->getManagers() as $curManagerName => $curManager) {
                if (!$curManager->isOpen()) {
                    $doctrineRegistry->resetManager($curManagerName);
                } else {
                    $curManager->clear();
                }
            }
```

We've reverted to 2.0.2 in the meantime, but I'm not entirely sure how to solve this issue going forward.

ostrolucky:
Fixes #1114

We can't reset because of this https://github.com/symfony/symfony/issues/35216 + performance issues, so let's just do clear() so it works for long running processes (same thing php-pm does for years).

### DIFF
--- a/Tests/RegistryTest.php
+++ b/Tests/RegistryTest.php
@@ -4,8 +4,12 @@ namespace Doctrine\Bundle\DoctrineBundle\Tests;
 
 use Closure;
 use Doctrine\Bundle\DoctrineBundle\Registry;
+use Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection\Fixtures\TestKernel;
 use Doctrine\ORM\EntityManagerInterface;
+use Fixtures\Bundles\RepositoryServiceBundle\Entity\TestCustomClassRepoEntity;
+use Fixtures\Bundles\RepositoryServiceBundle\Repository\TestCustomClassRepoRepository;
 use ProxyManager\Proxy\LazyLoadingInterface;
+use ProxyManager\Proxy\ProxyInterface;
 use stdClass;
 
 class RegistryTest extends TestCase
@@ -132,7 +136,7 @@ class RegistryTest extends TestCase
         $noProxyManager->expects($this->once())
             ->method('clear');
 
-        $proxyManager = $this->getMockBuilder(LazyLoadingInterface::class)->getMock();
+        $proxyManager = $this->getMockBuilder([LazyLoadingInterface::class, EntityManagerInterface::class])->getMock();
         $proxyManager->expects($this->once())
             ->method('setProxyInitializer')
             ->with($this->isInstanceOf(Closure::class));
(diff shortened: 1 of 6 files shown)


bendavies:
This is what I was going to propose.
just clearing seems enough. resetting in case of an error if the EM has become closed in still necessary.

This also fixes a large performance regression seen in tests and probably `php-pm`, whereby resetting the entity managers every request unloads all the mapping metadata, which means it is loaded every request.

ostrolucky:
I'm not entirely happy with this because reset semantics are in bridge and we are now working around that here. If these semantics are wrong, it should be changed there. All we should ideally do is just call resetService, without conditionals on our side. That said, it's our fault for causing breaks and symfony team is ignoring the issue, so we must do it on our side, at least until they fix it.

Tests/RegistryTest.php
```diff
@@ -157,4 +161,33 @@ public function testReset()
         $registry = new Registry($container, [], $entityManagers, 'default', 'default');
         $registry->reset();
     }
+
+    public function testIdentityMapsStayConsistentAfterReset()
```
greg0ire:
```suggestion
    public function testIdentityMapsStayConsistentAfterReset() : void
```

greg0ire: Nevermind, this is 1.12.x

alcaeus:
Merging this as it definitely fixes the duplicated service issue we've been experiencing. We're still discussing how to proceed with other breaks introduced by resetting the manager.

Thanks @ostrolucky!

acasademont:
Thanks @ostrolucky, this should be enough :)

We also merged a check in php-pm to not clear the EM if the Registry already implements the `ResetInterface`

Issue #821, opened 2018-05-21
[FEATURE] EntityListenerInterface for autowiring


DonCallisto:
As symfony >= 3.3 brings autowiring function, I think it could be somehow useful to have an "empty" interface called `EntityListenerInterface` to take advantage of the autowiring features.
This way we can auto-tag all services that implements the interface with `doctrine.entity_listener` instead of doing it manually.

So, something like

```
// Doctrine\Bundle\DoctrineBundle\DoctrineBundle.php
/**
 * {@inheritDoc}
 */
public function build(ContainerBuilder $container)
{
    [...]
    $container
        ->registerForAutoconfiguration(EntityListenerInterface::class)
        ->addTag('doctrine.entity_listener');
}
```

Of course this will not be compatible with all sf supported version so we should think how include it properly (checking sf version, or checking method definition, and so on) or leave it for a release were sf minimum supported version is compliant with this feature.

WDYT?

weaverryan:
These empty interfaces isn’t ideal. But, the entity listeners are already kinda magic, where iirc Doctrine looks for specific methods by name. So, a marker interface like this wouldn’t bother me.

However, there is a problem if you want to specify this on a different entity manager other than default. In that case, you would have to tag the service manually. That’s ok, but thanks to autoconfigure, you would now have 2 tags. I think this could be fixed by adding some special attribute to the autoconfgure tag, then removing that tag in the compiler pass, if there are multiple.

If you want to open a PR, you can. I like the idea, but I’m not in charge here :).

DonCallisto:
Ok, I'll wait for other to check and validate our idea and then try to write an implementation.
Thanks for reviewing this.

kimhemsoe:
We can not use the shared interfaces for autowiring as they conflict with ODM and etc. It is a mistake that we already accept some of them and i want to kill them off as soon as i can.

I agree with @weaverryan that a empty interface is not something we should do. If we had a subscriber  interface for ORM only and we could solve the "which entity manager" issue  i would be quite open for the feature.

weaverryan:
> and we could solve the "which entity manager"

I think we can solve this

> If we had a subscriber interface for ORM only

Yea, that would be best. So, if you *really* want this feature (it *would* be cool), we would need to propose that to the ORM first I think.

Btw, there is something easier that can be done now. Doctrine *does* have normal event subscribers (with an interface). But these do not play well with autoconfiguration yet. That could be done now, entirely in this bundle.

[...]

PR #1119, opened 2020-01-06
Added EventSubscriberInterface marker interface to support autoconfiguration

### LINKED ISSUE #674: Autoconfiguration
Can we have the `EventSubscriber` autoconfigured?
And it would be also nice if entity listeners can be autoconfigured, maybe by creating a marker interface?

lyrixx:
fixes #674

### DIFF
--- a/EventSubscriber/EventSubscriberInterface.php
+++ b/EventSubscriber/EventSubscriberInterface.php
@@ -1,0 +1,9 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\EventSubscriber;
+
+use Doctrine\Common\EventSubscriber;
+
+interface EventSubscriberInterface extends EventSubscriber
+{
+}
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -4,6 +4,7 @@ namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection;
 
 use Doctrine\Bundle\DoctrineBundle\Dbal\RegexSchemaAssetFilter;
 use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\ServiceRepositoryCompilerPass;
+use Doctrine\Bundle\DoctrineBundle\EventSubscriber\EventSubscriberInterface;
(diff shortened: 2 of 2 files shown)


ostrolucky:
This is solution for https://github.com/doctrine/DoctrineBundle/issues/821 right? But interface proposed in this PR looks nothing like proposals in issue.

lyrixx:
Nope, it's for https://github.com/doctrine/DoctrineBundle/issues/674 as stated in the PR description.

EventSubscriber/EventSubscriberInterface.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\EventSubscriber;
+
+interface EventSubscriberInterface
```
greg0ire: Shouldn't it extend another interface? That would make it easier on the user.

lyrixx: Good idea. Fixed.

alcaeus:
Thanks @lyrixx! 🎉

rvanlaak:
Brilliant in its simplicity! 👏 

Would it be possible to on userland let DI alias the old interface for the new one, so without making code changes to PHP this autowiring could get used (read: tested) on our application?

lyrixx:
@rvanlaak thanks.

If you are using **only** the ORM (and not the ORM) (see explaination [here](https://github.com/doctrine/DoctrineBundle/issues/674#issuecomment-349347553) you can do [that](https://github.com/doctrine/DoctrineBundle/issues/674#issuecomment-503519110):

```
services:
    _instanceof:
        Doctrine\Common\EventSubscriber:
            tags: [ doctrine.event_subscriber ]
```

rvanlaak:
Had that one in mind, which probably is sufficient. Was more thinking of something conceptual like:

```yaml
Doctrine\Common\EventSubscriber:
    alias: Doctrine\Bundle\DoctrineBundle\EventSubscriber\EventSubscriberInterface 
```

This in order to gain advantage of all functionality the doctrine bundle in the possible future would add via DI (e.g. tags, compiler passes) based on `EventSubscriberInterface`.

PR #1120, opened 2020-01-06
Add a feature switch to trigger SchemaValidator in Collector

### LINKED ISSUE #1115: [performance] DoctrineDataCollector SchemaValidator could be optional
I love the details and the validation given by the DoctrineDataCollector SchemaValidator. But on a serious project, it's useful only a couple of times, and cost a lot at each dev request. We already have a command for that (`bin console doctrine:schema:validate`) that we can launch manually and put in CI.

I'm talking about that: https://github.com/doctrine/DoctrineBundle/blob/master/DataCollector/DoctrineDataCollector.php#L61

It forces to get a loooot of data about any entity type, seems to be more than useful for the job being made to comply with the request.

I want to put that behind a feature switch, same as the `profiling_collect_backtrace`, but before creating a PR, I want to know the opinion of the maintainers about that. Is it something than you can consider merging? 

This feature switch could be set as true by default to provide the full DX by default bu be disable by power user if needed. I know I will on ALL my project: we are talking about minutes / hours cpu by each member of my team per day.

image

ps: -21% means -160ms.

bastnic:
Fixes #1115


DataCollector/DoctrineDataCollector.php
```diff
@@ -26,9 +26,16 @@ class DoctrineDataCollector extends BaseCollector
     /** @var string[] */
     private $groupedQueries;
 
-    public function __construct(ManagerRegistry $registry)
+    /** @var bool */
+    private $shouldValidateSchemaClass;
```
stof: I would name it `shouldValidateSchema`. I don't see why there is `Class` in the name.

DataCollector/DoctrineDataCollector.php
```diff
@@ -26,9 +26,16 @@ class DoctrineDataCollector extends BaseCollector
...
+
+    /**
+     * @param bool $shouldValidateSchemaClass
```
stof: please use a typehint instead, as the master branch requires PHP 7.1+

bastnic: done (I had 1.12.x in ming but @ostrolucky confirmed it's master)

DataCollector/DoctrineDataCollector.php
bastnic: done

bastnic:
@stof all done.

Tests added.

bastnic:
As it's targeted a big release, should we put it to false by default?

greg0ire:
Please add a summary of #1115 to the body of your commit message (but keep the `Fixes #1115` though)

DependencyInjection/Configuration.php
```diff
@@ -136,6 +136,10 @@ private function getDbalConnectionsNode()
                     ->defaultValue(false)
                     ->info('Enables collecting backtraces when profiling is enabled')
                 ->end()
+                ->booleanNode('profiling_collect_schema_errors')
```
greg0ire: There is [an ongoing discussion to rename the "profiler" as "inspector"](https://github.com/symfony/symfony/issues/34042), so I think we should avoid that vocabulary, but given there are already `profiling`, and `profiling_collect_backtrace` options, I'm torn…

Tests/DataCollector/DoctrineDataCollectorTest.php
```diff
@@ -49,6 +49,32 @@ public function testCollectEntities()
         $this->assertCount(2, $entities['default']);
     }
 
+    public function testDoNotCollectEntities()
```
greg0ire:
```suggestion
    public function testDoNotCollectEntities() : void
```

Tests/DataCollector/DoctrineDataCollectorTest.php
```diff
@@ -49,6 +49,32 @@ public function testCollectEntities()
...
+    {
+        $manager   = $this->getMockBuilder('Doctrine\ORM\EntityManager')->disableOriginalConstructor()->getMock();
+        $config    = $this->getMockBuilder('Doctrine\ORM\Configuration')->getMock();
```
greg0ire: I think you should use `createMock` in at least one of those case, probably the first one…

Tests/DataCollector/DoctrineDataCollectorTest.php
```diff
@@ -49,6 +49,32 @@ public function testCollectEntities()
...
+        $factory   = $this->getMockBuilder('Doctrine\Common\Persistence\Mapping\AbstractClassMetadataFactory')
+            ->setMethods(['getLoadedMetadata'])->getMockForAbstractClass();
+        $collector = $this->createCollector(['default' => $manager], false);
+
+        $manager->expects($this->never())
+            ->method('getMetadataFactory');
+        $manager->expects($this->any())
```
greg0ire: IIRC `->expects($this->any())` can be removed, might depend on the version of PHPUnit

DependencyInjection/Configuration.php
ostrolucky: We shouldn't rename it until symfony does. And it won't happen until at least Symfony 6.

Tests/DataCollector/DoctrineDataCollectorTest.php
bastnic: I tend to follow the lead of the current file, and no other method as return type.

bastnic:
@ostrolucky @greg0ire all done.

Tests/DataCollector/DoctrineDataCollectorTest.php
bastnic: it seems to be a required parameter. `getConfiguration` is called outside of the if I just added, so I must give something.

Tests/DataCollector/DoctrineDataCollectorTest.php
bastnic: more than one in fact ;).

Tests/DataCollector/DoctrineDataCollectorTest.php
greg0ire: I was not talking about the parameter `$this->any()`, but about the entire call `->expects($this->any())`, what about you?

greg0ire: So 
```suggestion
        $manager
```

bastnic: done, thanks

Tests/DataCollector/DoctrineDataCollectorTest.php
```diff
@@ -49,6 +49,22 @@ public function testCollectEntities()
...
+        $collector = $this->createCollector(['default' => $manager], false);
+
+        $manager->expects($this->never())
+            ->method('getMetadataFactory');
+        $manager->method('getConfiguration')
+            ->will($this->returnValue($config));
+
+        $collector->collect(new Request(), new Response());
+
+        $this->assertEmpty($collector->getEntities());
```
ostrolucky: Asserting this doesn't make sense to me, any particular reason to do it? Using `getMappingErrors` would at least make *some* sense

Tests/DataCollector/DoctrineDataCollectorTest.php
```diff
@@ -49,6 +49,22 @@ public function testCollectEntities()
         $this->assertCount(2, $entities['default']);
     }
 
+    public function testDoNotCollectEntities() : void
```
ostrolucky:
```suggestion
    public function testDoesNotCollectEntities() : void
```

Tests/DataCollector/DoctrineDataCollectorTest.php
bastnic: if `entities` is empty, it's because we never enter in the if condition, otherwise it would be `['default' => []]`. It's also the condition asked by @stof, to not display anything in the template.

You are right, I will add a test on `getMappingErrors`.

bastnic: done

Tests/DataCollector/DoctrineDataCollectorTest.php
bastnic: done

PR #1121, opened 2020-01-06
Add void return type declaration to tests

greg0ire:
As a side effect, this makes the code a bit more forward-compatible with
recent versions of PHPUnit, that have that return type declaration on
setUp().


alcaeus:
Same as #1122, this should not target a stable release branch but `master`.

PR #1122, opened 2020-01-06
Add type declarations where backwards compatible

greg0ire:
That means:

- methods in final classes
- final methods (none of those found)
- private methods
- methods in Tests (that excludes `setUp()`, `tearDown()` and `test*`, which
  are handled with an automated tool in #1121 )

### DIFF
--- a/Controller/ProfilerController.php
+++ b/Controller/ProfilerController.php
@@ -74,7 +74,10 @@ class ProfilerController implements ContainerAwareInterface
         ]));
     }
 
-    private function explainSQLitePlatform(Connection $connection, $query)
+    /**
+     * @param mixed[] $query
+     */
+    private function explainSQLitePlatform(Connection $connection, array $query)
     {
         $params = $query['params'];
 
--- a/ManagerConfigurator.php
+++ b/ManagerConfigurator.php
@@ -13,12 +13,12 @@ class ManagerConfigurator
     /** @var string[] */
     private $enabledFilters = [];
 
-    /** @var string[] */
+    /** @var array<string,array<string,string>> */
     private $filtersParameters = [];
 
     /**
-     * @param string[] $enabledFilters
-     * @param string[] $filtersParameters
+     * @param string[]                           $enabledFilters
+     * @param array<string,array<string,string>> $filtersParameters
      */
     public function __construct(array $enabledFilters, array $filtersParameters)
     {
@@ -37,7 +37,7 @@ class ManagerConfigurator
     /**
      * Enables filters for a given entity manager
      */
-    private function enableFilters(EntityManagerInterface $entityManager)
+    private function enableFilters(EntityManagerInterface $entityManager) : void
     {
         if (empty($this->enabledFilters)) {
             return;
@@ -56,11 +56,8 @@ class ManagerConfigurator
 
     /**
      * Sets default parameters for a given filter
-     *
-     * @param string    $name   Filter name
-     * @param SQLFilter $filter Filter object
      */
-    private function setFilterParameters($name, SQLFilter $filter)
+    private function setFilterParameters(string $name, SQLFilter $filter) : void
     {
         if (empty($this->filtersParameters[$name])) {
             return;
--- a/Repository/ContainerRepositoryFactory.php
+++ b/Repository/ContainerRepositoryFactory.php
@@ -39,7 +39,7 @@ final class ContainerRepositoryFactory implements RepositoryFactory
     /**
      * {@inheritdoc}
      */
-    public function getRepository(EntityManagerInterface $entityManager, $entityName)
+    public function getRepository(EntityManagerInterface $entityManager, $entityName) : ObjectRepository
     {
         $metadata            = $entityManager->getClassMetadata($entityName);
         $repositoryServiceId = $metadata->customRepositoryClassName;
@@ -72,8 +72,10 @@ final class ContainerRepositoryFactory implements RepositoryFactory
         return $this->getOrCreateRepository($entityManager, $metadata);
     }
 
-    private function getOrCreateRepository(EntityManagerInterface $entityManager, ClassMetadata $metadata)
-    {
+    private function getOrCreateRepository(
+        EntityManagerInterface $entityManager,
+        ClassMetadata $metadata
+    ) : ObjectRepository {
         $repositoryHash = $metadata->getName() . spl_object_hash($entityManager);
         if (isset($this->managedRepositories[$repositoryHash])) {
             return $this->managedRepositories[$repositoryHash];
--- a/Tests/Command/CreateDatabaseDoctrineTest.php
+++ b/Tests/Command/CreateDatabaseDoctrineTest.php
@@ -3,8 +3,8 @@
 namespace Doctrine\Bundle\DoctrineBundle\Tests\Command;
 
 use Doctrine\Bundle\DoctrineBundle\Command\CreateDatabaseDoctrineCommand;
+use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
-use PHPUnit_Framework_MockObject_MockObject;
 use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Tester\CommandTester;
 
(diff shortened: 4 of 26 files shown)


Controller/ProfilerController.php
```diff
@@ -74,7 +74,10 @@ public function explainAction($token, $connectionName, $query)
         ]));
     }
 
-    private function explainSQLitePlatform(Connection $connection, $query)
+    /**
+     * @param mixed[] $query
```
ostrolucky: This doesn't look useful

ManagerConfigurator.php
```diff
@@ -13,12 +13,12 @@ class ManagerConfigurator
...
 
-    /** @var string[] */
+    /** @var array<array<string,string>> */
     private $filtersParameters = [];
 
     /**
-     * @param string[] $enabledFilters
-     * @param string[] $filtersParameters
+     * @param string[]                    $enabledFilters
+     * @param array<array<string,string>> $filtersParameters
```
ostrolucky:
```suggestion
     * @param array<string,array<string,string>> $filtersParameters
```

Tests/Command/DropDatabaseDoctrineTest.php
```diff
@@ -75,12 +75,9 @@ public function testExecuteWithoutOptionForceWillFailWithAttentionMessage() : vo
...
 
     /**
-     * @param string     $connectionName Connection name
-     * @param array|null $params         Connection parameters
-     *
-     * @return PHPUnit_Framework_MockObject_MockObject
+     * @param array|null $params Connection parameters
      */
-    private function getMockContainer($connectionName, $params = null)
+    private function getMockContainer(string $connectionName, $params = null) : MockObject
```
ostrolucky:
```suggestion
    private function getMockContainer(string $connectionName, array $params) : MockObject
```

Tests/Command/CreateDatabaseDoctrineTest.php
```diff
@@ -86,12 +86,9 @@ public function testExecuteWithShardOption() : void
...
 
     /**
-     * @param string       $connectionName Connection name
-     * @param mixed[]|null $params         Connection parameters
-     *
-     * @return PHPUnit_Framework_MockObject_MockObject
+     * @param mixed[]|null $params Connection parameters
      */
-    private function getMockContainer($connectionName, $params = null)
+    private function getMockContainer(string $connectionName, $params = null) : MockObject
```
ostrolucky:
```suggestion
    private function getMockContainer(string $connectionName, array $params) : MockObject
```

greg0ire:
@ostrolucky here is why we need to use `mixed[]`: https://travis-ci.org/doctrine/DoctrineBundle/jobs/633955021#L328-L33

I initially did it to convey the fact that I did my due diligence and looked into the type of the items only to find it's mixed instead of just being lazy and use `array`.

alcaeus:
Please don't target stable releases with such changes - only bug fixes should hit a stable branch with all other changes going to master.

greg0ire:
Sorry, my bad!

ostrolucky:
These are not features either though. It would make solving conflicts much easier to target lowest affected branch.

greg0ire:
> It would make solving conflicts much easier to target lowest affected branch.

That's a good point, and IMO it would be more coherent with our stance on docs fixes that don't affect the software either but that should be contributed to the lowest affected branch.

alcaeus:
I'm not going to revert it, but we may want to limit such changes to upcoming releases to avoid issues in patch releases.

ostrolucky:
I think these are very safe changes, I don't see any way to break anything

greg0ire:
I'm torn: on one hand it will make things more complicated for us with the upmerges, especially for that kind of change that touch a lot of files, on the other hand it's hard to say that every pedantic change will always be risk-free, and it's equally hard to ask maintainers to use their best judgment and handle this on a case by case basis. Maybe we should try it for a while and see if it does result in an instability that we are not willing to accept, but maybe some of you already did that in the past, I don't know.

alcaeus:
I've generally avoided touching the type system in patch releases since the breaks can be very subtle, even for private methods. Think about `@param string $foo` vs. `string $foo` - these behave very differently with stringables when `strict_types` is enabled. Introducing such breaks in patch releases is unnecessary, which is why I normally avoid it.

Repository/ContainerRepositoryFactory.php
```diff
@@ -39,7 +39,7 @@ public function __construct(ContainerInterface $container = null)
     /**
      * {@inheritdoc}
      */
-    public function getRepository(EntityManagerInterface $entityManager, $entityName)
+    public function getRepository(EntityManagerInterface $entityManager, $entityName) : ObjectRepository
```
Majkl578: This change is in fact a BC break because before repositories were not enforced to be subclass of ObjectRepository.

andrew-demb: But `ContainerRepositoryFactory` is final and users can't extends from it. Isn't it?

alcaeus: > But `ContainerRepositoryFactory` is final and users can't extends from it. Isn't it?

Precisely.

Majkl578: @andrew-demb Yes but in Symfony ContainerRepositoryFactory overrides default RepositoryFactory from Doctrine, making it a BC break. Not a BC break in the contract of this class but a BC break in the bundle itself. It's no longer possible to use custom repositories not extending ObjectRepository and doing so makes the app crash due to this added typehint.

greg0ire: It's been documented as supposed to be `ObjectRepository` since the creation of `RepositoryFactory` 7 years ago (!) : https://github.com/doctrine/orm/commit/7eb744126ba5081ea51441a893dd950046777383

repositories were not enforced to be subclass of ObjectRepository because type declarations did not exist back then, but even if the contract is not enforced, it's still a contract. Call me heartless, but I can't say I feel sorry for apps experiencing this crash, they kind of had it coming, don't you agree?

Majkl578: I don't agree, this kind of stuff does not belong to a patch version.

Majkl578: Requirement for the ObjectRepository/EntityRepository was never documented: https://github.com/doctrine/orm/blob/v2.6.0/lib/Doctrine/ORM/EntityManager.php#L712
It was also never validated: https://github.com/doctrine/orm/blob/v2.7.1/docs/en/reference/working-with-objects.rst#by-native-queries
There is return annotation for EntityManager::getRepository(): https://github.com/doctrine/orm/blob/v2.6.0/lib/Doctrine/ORM/EntityManager.php#L712 but that is the only public API and has never been enforced anyway.

It's not nice "feature" and I don't like it either but it was working and documented behaviour (_" You can overwrite this behaviour by specifying the class name of your own Entity Repository in the Annotation, XML or YAML metadata."_).

alcaeus: > I don't personally need this "feature" but I don't think this should've been merged as is.

You are, as always, entitled to your opinion. To look at the facts:
`ContainerRepositoryFactory` has always implemented `RepositoryFactory` from ORM with an `@inheritdoc` docblock.
`RepositoryFactory` in ORM has had an `@return ObjectRepository` on its `getRepository` method since this commit in 2013, where it was widened from ORM's `EntityRepository` to the persistence library's `ObjectRepository`: https://github.com/doctrine/orm/commit/3488049c18849a0cb942f4fea00501fa11727a83.

I don't know where you see the BC break, but if we treat docblocks as a contract (which we always have and always should), this change is perfectly fine. You are always welcome to improve documentation which is clearly wrong in this case, as the code contains the absolute truth, which is that the contract always specified an `ObjectRepository` return type for as long as the class has existed.

Majkl578: > I don't know where you see the BC break

The BC break I see is that upgrading doctrine/doctrine-bundle from 2.0.6 to 2.0.7 leads to fatal errors.

> as the code contains the absolute truth

The code can actually contain bugs.

alcaeus: > The BC break I see is that upgrading doctrine/doctrine-bundle from 2.0.6 to 2.0.7 leads to fatal errors.

Please post an example that passes on 2.0.6 and produces a fatal error on 2.0.7 and I'll fix it. Thanks!

PR #1125, opened 2020-01-09
Make console commands lazy

andrew-demb:
Link to doc: https://symfony.com/doc/current/console/commands_as_services.html#lazy-loading (since symfony 3.4)

Will be helpful for https://github.com/doctrine/orm/issues/7972

### DIFF
--- a/Command/CreateDatabaseDoctrineCommand.php
+++ b/Command/CreateDatabaseDoctrineCommand.php
@@ -16,13 +16,15 @@ use Symfony\Component\Console\Output\OutputInterface;
  */
 class CreateDatabaseDoctrineCommand extends DoctrineCommand
 {
+    /** @var string */
+    protected static $defaultName = 'doctrine:database:create';
+
     /**
      * {@inheritDoc}
      */
     protected function configure()
     {
         $this
-            ->setName('doctrine:database:create')
             ->setDescription('Creates the configured database')
             ->addOption('shard', null, InputOption::VALUE_REQUIRED, 'The shard connection to use for this command')
             ->addOption('connection', null, InputOption::VALUE_OPTIONAL, 'The connection to use for this command')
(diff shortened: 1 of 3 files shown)


andrew-demb:
Rebased to master

stof:
All commands are *already* lazy-loaded, because we always specify the `command` attribute in the tag in the service definition. So this does not change anything.

stof:
and this does not help for the ORM issue at all btw. The commands are not the places where we use the platform in class constructors.

alcaeus:
Closing as per @stof's comment.

ostrolucky:
Is this a bug in debug command?

```
bin/console debug:container doctrine.database_create_command

Information for Service "doctrine.database_create_command"
==========================================================

 ---------------- ----------------------------------------------------------------------
  Option           Value
 ---------------- ----------------------------------------------------------------------
  Service ID       doctrine.database_create_command
  Class            Doctrine\Bundle\DoctrineBundle\Command\CreateDatabaseDoctrineCommand
  Tags             console.command (command: doctrine:database:create)
  Calls            setName
  Public           no
  Synthetic        no
  Lazy             no
  Shared           yes
  Abstract         no
  Autowired        no
  Autoconfigured   no
 ---------------- ----------------------------------------------------------------------
```

stof:
`lazy` is `debug:container` shows whether the service is a lazy one, wrapped in a ProxyManager proxy.

The lazy-loading of commands does not rely on lazy services at all (which could not work as getting the name of a command would then still need to call a method on the command object and so initialize the proxy).

stof:
Note that we have many more places in Symfony performing lazy-loading of some dependencies without using lazy services (lazy services are not lazy-loading the dependencies, but the service itself).

ostrolucky:
Perhaps we should remove setName calls to not confuse people. I can easily see myself making a mistake of changing setName call instead of a tag.

PR #1131, opened 2020-01-24
Updated service definition to support Messenger 5.1

Nyholm:
I have a PR open to Symfony to move the Doctrine transport to a new package. See https://github.com/symfony/symfony/pull/35422

If my Symfony PR is merged, this PR makes sure that DoctrineBundle is compatible with Symfony Messenger 5.1 without any deprecation notices. 

This PR could also be cherry-picked to 1.x branch.

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -25,7 +25,7 @@ use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Messenger\MessageBusInterface;
-use Symfony\Component\Messenger\Transport\Doctrine\DoctrineTransportFactory;
+use Symfony\Component\Messenger\Bridge\Doctrine\Transport\DoctrineTransportFactory;
 use Symfony\Component\PropertyInfo\PropertyInfoExtractorInterface;
 use function class_exists;
 use function sprintf;
@@ -842,11 +842,17 @@ class DoctrineExtension extends AbstractDoctrineExtension
             $container->removeDefinition('doctrine.orm.messenger.event_subscriber.doctrine_clear_entity_manager');
         }
 
+        $transportFactoryDefinition = $container->getDefinition('messenger.transport.doctrine.factory');
         if (! class_exists(DoctrineTransportFactory::class)) {
-            return;
+            // If symfony/messenger < 5.1
+            if (class_exists(\Symfony\Component\Messenger\Transport\Doctrine\DoctrineTransportFactory::class)) {
+                $transportFactoryDefinition->setClass(\Symfony\Component\Messenger\Transport\Doctrine\DoctrineTransportFactory::class);
+            } else {
+                // Dont add the tag
+                return;
+            }
         }
 
-        $transportFactoryDefinition = $container->getDefinition('messenger.transport.doctrine.factory');
         $transportFactoryDefinition->addTag('messenger.transport_factory');
     }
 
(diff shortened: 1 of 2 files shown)


DependencyInjection/DoctrineExtension.php
```diff
@@ -842,11 +842,17 @@ private function loadMessengerServices(ContainerBuilder $container) : void
...
         }
 
+        $transportFactoryDefinition = $container->getDefinition('messenger.transport.doctrine.factory');
         if (! class_exists(DoctrineTransportFactory::class)) {
-            return;
+            // If symfony/messenger < 5.1
+            if (!class_exists(\Symfony\Component\Messenger\Transport\Doctrine\DoctrineTransportFactory::class)) {
+                // Dont add the tag
+                return;
+            } else {
```
SenseException: Because the if-case contains a `return`, the logic inside this `else` doesn't need to be in an else-block.

Nyholm: Thank you!

Nyholm:
Thank you for the review. 

My PR got merged in Symfony so I removed WIP on this PR.

SenseException:
Nice. After adding (a) testcase(s), this should be fine from my point of view.

Nyholm:
Thank you. 

Could you give me some guidance how I can test both outcomes of `if (! class_exists(DoctrineTransportFactory::class))`?

SenseException:
The build process is installing composer packages differently for many use cases. One is a PHP 7.1 using `--prefer-lowest`. If there's an installation without a Messenger component, another path will be covered. You'll find an example for the Messenger component here:
https://github.com/doctrine/DoctrineBundle/blob/master/Tests/DependencyInjection/DoctrineExtensionTest.php#L660

DependencyInjection/DoctrineExtension.php
```diff
@@ -842,11 +842,17 @@ private function loadMessengerServices(ContainerBuilder $container) : void
             $container->removeDefinition('doctrine.orm.messenger.event_subscriber.doctrine_clear_entity_manager');
         }
 
+        $transportFactoryDefinition = $container->getDefinition('messenger.transport.doctrine.factory');
         if (! class_exists(DoctrineTransportFactory::class)) {
-            return;
+            // If symfony/messenger < 5.1
+            if (! class_exists(\Symfony\Component\Messenger\Transport\Doctrine\DoctrineTransportFactory::class)) {
```
alcaeus: You could use a `use` statement with an alias, like `LegacyDoctrineTransportFactory`. Makes this look just a little nicer ;)

PReimers: @alcaeus I'm not sure, but what about use statements to nonexisting classes?

In this situation, I would prefer to drop the use statements and use the FQCN in code. 🤔

Resources/config/messenger.xml
```diff
@@ -32,7 +32,7 @@
         The following service isn't tagged as transport factory because the class may not exist.
         The tag is added conditionally in DoctrineExtension.
         -->
-        <service id="messenger.transport.doctrine.factory" class="Symfony\Component\Messenger\Transport\Doctrine\DoctrineTransportFactory" public="false">
+        <service id="messenger.transport.doctrine.factory" class="Symfony\Component\Messenger\Bridge\Doctrine\Transport\DoctrineTransportFactory" public="false">
```
ostrolucky: This FCQN doesn't exist in Messenger 5.0, hence BC layer you implemented is broken

DependencyInjection/DoctrineExtension.php
ostrolucky: I slightly prefer @alcaeus suggestion just because it's used on two lines, but this is really petty discussion, so let's not block merge because of it.

alcaeus: Yep, this can always be fixed later.

alcaeus:
Thanks @Nyholm!

alcaeus:
And I just realised I merged this without squashing commits. Oh well 🤷‍♂️

PR #1135, opened 2020-02-03
Add priority support for tagged entity listeners

### LINKED ISSUE #991: Add priority support for tagged entity listeners
I was wondering if priority support could be added to entity listeners tags in the compiler pass as it's supported for doctrine `event_subscriber` and `event_listener` tags in doctrine symfony bridge.

https://github.com/doctrine/DoctrineBundle/blob/ead3f067e82b4f5ff91642cd86112401c6e5c6a8/DependencyInjection/Compiler/EntityListenerPass.php#L24

ostrolucky:
Fixes #991

### DIFF
--- a/DependencyInjection/Compiler/EntityListenerPass.php
+++ b/DependencyInjection/Compiler/EntityListenerPass.php
@@ -6,6 +6,7 @@ use Doctrine\Bundle\DoctrineBundle\Mapping\ContainerEntityListenerResolver;
 use Doctrine\Bundle\DoctrineBundle\Mapping\EntityListenerServiceResolver;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\Compiler\PriorityTaggedServiceTrait;
 use Symfony\Component\DependencyInjection\Compiler\ServiceLocatorTagPass;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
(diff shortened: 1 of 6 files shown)


Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -997,16 +997,16 @@ public function testPrivateLazyEntityListener() : void
 
     /**
      * @expectedException \InvalidArgumentException
-     * @expectedExceptionMessageRegExp /The service ".*" must not be abstract as this entity listener is lazy-loaded/
+     * @expectedExceptionMessageRegExp /The service ".*" must not be abstract\./
      */
-    public function testAbstractLazyEntityListener() : void
+    public function testAbstractEntityListener() : void
```
ostrolucky: This is not a BC break, merely misleading test case. It DID throw even for non-lazy abstract listeners before.

DependencyInjection/Compiler/EntityListenerPass.php
```diff
@@ -62,10 +66,6 @@ public function process(ContainerBuilder $container)
                 if (! isset($attributes['lazy']) && $resolverSupportsLazyListeners || $lazyByAttribute) {
                     $listener = $container->findDefinition($id);
 
-                    if ($listener->isAbstract()) {
-                        throw new InvalidArgumentException(sprintf('The service "%s" must not be abstract as this entity listener is lazy-loaded.', $id));
```
ostrolucky: This check is done from within `findAndSortTaggedServices` now, so no need for it here anymore

alcaeus:
Thanks @ostrolucky!

PR #1139, opened 2020-02-18
Add missing return type to fixture file

DurandSacha:


alcaeus:
Closing here: Adding a return type to a non-final, non-private method in a non-final class is a BC break and not suitable for a new minor release. We'll be adding return type hints in 3.0.

ostrolucky:
This is ok to do for tests in same fashion as https://github.com/doctrine/DoctrineBundle/pull/1121 did

alcaeus:
Sorry, didn’t see this was a test fixture 👍

ostrolucky:
🤦‍♂ indeed

PR #1141, opened 2020-02-20
Add void return declaration for RegistryTest.php

DurandSacha:


nicolas-grekas:
Thanks but the is not needed.

Issue #1143, opened 2020-02-22
Profiler give a non-executable query when using empty array.


VincentLanglet:
When writing the following where clause
```
$query->andWhere('o.id not in (:q)');
$query->setParameter('q', []);
```
The query is executed without errors. (And return no result)

I was looking for the real executed query, looked at the profiler and saw this.
![image](https://user-images.githubusercontent.com/9052536/75091861-44f8a500-5572-11ea-8cb9-70e9d41f818e.png)

But 
```
SELECT count(DISTINCT a0_.id) AS sclr_0 FROM asv_client a0_ WHERE a0_.id NOT IN ();
```
Is not an executable query.

So, something else is executed in database ; and should be displayed in the profiler.

The actual code to display the query in the profiler is the following
```
# Ressources/views/Collector/db.html.twig
 {{ (query.sql ~ ';')|doctrine_replace_query_parameters(query.params)|doctrine_pretty_query(highlight_only = true) }}
```

VincentLanglet:
After some investigation, I found that when you write
```
$query->andWhere("$alias.foo IN (:bar)");
$query->setParameter('bar', []);
```
Or
```
$query->andWhere("$alias.foo NOT IN (:bar)");
$query->setParameter('bar', []);
```

The params are processed by the method `expandListParameters` of the class `Doctrine\DBAL\SQLParserUtils` by the following line.
$expandStr = $count > 0 ? implode(', ', array_fill(0, $count, '?')) : 'NULL';

So the real executed query is
```
SELECT count(DISTINCT a0_.id) AS sclr_0 FROM asv_client a0_ WHERE a0_.id NOT IN (NULL);
```

PR #1144, opened 2020-02-22
Improve replaceQueryParameters function for empty array

VincentLanglet:
Fix https://github.com/doctrine/DoctrineBundle/issues/1143


ostrolucky:
I think we shouldn't fix such superficial issue, but fix underlying problem - For some reason, bundle is not using `SQLParserUtils::expandListParameters`

greg0ire:
Thanks @VincentLanglet !

PR #1142, opened 2020-02-21
Update missing void return

DurandSacha:

PR #1145, opened 2020-02-26
Update void return in function

DurandSacha:


dmaicher:
See https://github.com/doctrine/DoctrineBundle/pull/1142#issuecomment-589691846

PR #1147, opened 2020-02-29
Refactor

DurandSacha:
Refactor DoctrineCommand.php with light code


greg0ire:
The interface is not fluent this is never going to work. Plus it's a parse error, you would have to add parenthesis around the `new EntityGenerator()` for this to be syntactically valid.

But let's zoom out. Can we talk about your recent attempts at contributing? Why did you ignore https://github.com/doctrine/DoctrineBundle/pull/1142#issuecomment-589691846 ?

PR #1150, opened 2020-03-17
Fix render controller when explaining a query

azzra:
When trying to explain a query in the profiler, I had an exception `An exception has been thrown during the rendering of a template ("The controller for URI "/_fragment" is not callable. Controller "DoctrineBundle:Profiler:explain" does neither exist as service nor as class").`

Using the controller method FQCN did solve the problem

### DIFF
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -112,7 +112,7 @@
     {% set profiler_markup_version = profiler_markup_version|default(1) %}
 
     {% if 'explain' == page %}
-        {{ render(controller('DoctrineBundle:Profiler:explain', {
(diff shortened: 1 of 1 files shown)


ostrolucky:
Can you please tell us how to reproduce your problem, so we are sure this solves it?

greg0ire:
I did it for you @azzra :)

greg0ire:
Looks like the 1.12.x pipeline is broken :/

stof:
Well, the 1.12 pipeline should not have a job trying to force using Symfony 5.0, as we don't have a release of DoctrineCacheBundle supporting SF 5.0 (on purpose)

greg0ire:
It doesn't actually, I think I did the things in the wrong order, I should probably have changed base before force pushing.

greg0ire:
There. New pipeline, no sf5 job. Thanks @stof !

PR #1153, opened 2020-04-01
fix symfony/config deprecation

jrushlow:
Starting in Symfony/Config 5.1, calling `NodeDefinition::setDeprecated()` with no arguments is deprecated. See https://github.com/symfony/symfony/pull/35871 

This addresses the deprecation by providing the soon to be required parameters.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -699,16 +699,16 @@ class Configuration implements ConfigurationInterface
      * introduced at the same time. By checking if getDefinition() exists,
      * we can determine the correct param count to use when calling setDeprecated.
      */
-    private function getCommentedParamDeprecationMsg(): array
+    private function getCommentedParamDeprecationMsg() : array
     {
         if (method_exists(BaseNode::class, 'getDefinition')) {
             return [
                 'doctrine/doctrine-bundle',
                 '2.0',
-                'Type commenting features removed; the corresponding config parameter was deprecated and will be dropped in 3.0.'
+                'Type commenting features removed; the corresponding config parameter was deprecated and will be dropped in 3.0.',
             ];
         }
 
-        return [];
+        return [null];
     }
 }
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -695,15 +695,15 @@ class Configuration implements ConfigurationInterface
      * Returns the correct deprecation param's as an array for setDeprecated.
      *
      * Symfony/Config v5.1 introduces a deprecation notice when calling
-     * setDeprecation() with less than 3 args and the getDefinition method was
-     * introduced at the same time. By checking if getDefinition() exists,
+     * setDeprecation() with less than 3 args and the getDeprecation method was
+     * introduced at the same time. By checking if getDeprecation() exists,
      * we can determine the correct param count to use when calling setDeprecated.
      */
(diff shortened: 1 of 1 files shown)


DependencyInjection/Configuration.php
```diff
@@ -91,7 +91,11 @@ private function addDbalSection(ArrayNodeDefinition $node) : void
                             ->end()
                             ->children()
                                 ->scalarNode('class')->isRequired()->end()
-                                ->booleanNode('commented')->setDeprecated()->end()
+                                ->booleanNode('commented')->setDeprecated(
+                                    'doctrine/DoctrineBundle',
```
alcaeus, line 95: We should use the packagist name:
```suggestion
                                    'doctrine/doctrine-bundle',
```

alcaeus:
Changes look good, just waiting to see if travis-ci likes this version better than the previous one. Breakage is most likely unrelated, but I'd like to investigate the failure if it occurs. I'll be merging this once the build is stable. Thanks @jrushlow!

DependencyInjection/Configuration.php
```diff
@@ -91,7 +91,11 @@ private function addDbalSection(ArrayNodeDefinition $node) : void
                             ->end()
                             ->children()
                                 ->scalarNode('class')->isRequired()->end()
-                                ->booleanNode('commented')->setDeprecated()->end()
+                                ->booleanNode('commented')->setDeprecated(
+                                    'doctrine/doctrine-bundle',
```
xabbuh, line 95: With `symfony/config` < 5.1 this change would now mean that the message displayed to the user would be `doctrine/doctrine-bundle` as the only argument of `setDeprecated()` used to be the message.

jrushlow: hmm, would something along the lines of 
```
$messageArray = $this-getMessage();
private func getMessage(): array
{
 if config < 5.1 $message = [] else $message ['doctrine/bundle', 2.0, 'this is why';
}

....->setDeprecated(...$message)->end()
```
do the trick?

xabbuh: Yes, I think so.

DependencyInjection/Configuration.php
```diff
@@ -91,7 +92,7 @@ private function addDbalSection(ArrayNodeDefinition $node) : void
                             ->end()
                             ->children()
                                 ->scalarNode('class')->isRequired()->end()
-                                ->booleanNode('commented')->setDeprecated()->end()
+                                ->booleanNode('commented')->setDeprecated(...$this->getCommentedParamDeprecationMsg())->end()
```
jrushlow, line 95: if return value of `$this->getCommentedParamDeprecationMsg()` is `[]`, I believe using `...` will cause `ArgumentCountError`

jrushlow: fixed - for config versions before `5.1`, a message is provided. Attempted to return `null` for  < 5.1 but that triggered a `TypeError` in an even earlier version before `setDeprecated()` allowed `?string`

DependencyInjection/Configuration.php
jrushlow: i believe this is now taken care of. See my comment below...

alcaeus:
@jrushlow thanks for the contribution! Merging this as the coverage build is also failing in master. travis-ci not reporting back its result is an issue they've been having that we're trying to work around separately.

PR #1160, opened 2020-04-29
Fix CI when testing with Symfony 5.1

nicolas-grekas:


ostrolucky:
Can you provide some backstory please? Why are we not asserting now that tags contain `container.no_preload` *only*, but instead trying to hiding this fact?

nicolas-grekas:
`container.no_preload` is added because the only services that reference the tested listener are commands, and commands propagate this tag to their deps.
Testing this is unrelated to the bundle, so it should just be ignored here.

greg0ire:
The CI still fails, I will push a commit that fixes it shortly.

nicolas-grekas:
The fix for deprecations should be borrowed from master

greg0ire:
Oh didn't know, I'll fix this.

greg0ire (approved):

SenseException (approved):

ostrolucky:
You ask for my approval, but I am still uncomfortable with this unset thing. We should change assertions to use `$definition->getTags()['doctrine.event_subscriber']` / `$definition->getTags()['doctrine.event_listener']`, instead of coupling these test cases more tightly to Symfony DI by hardcoding their internal tags. In the end these asertions were wrong. These test cases shouldn't care about other tags, just tags we are touching - and adding every other tag in future to unset() later on just doesn't scale.

nicolas-grekas:
@ostrolucky thanks for the suggestion, I updated the PR accordingly. I hope it's OK for everyone else also.

ostrolucky (approved):

Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -655,7 +654,7 @@ public function testAttachEntityListeners() : void
         $methodCalls = $definition->getMethodCalls();
 
         $this->assertDICDefinitionMethodCallCount($definition, 'addEntityListener', [], 6);
-        $this->assertEquals(['doctrine.event_listener' => [ ['event' => 'loadClassMetadata'] ] ], $definition->getTags());
+        $this->assertSame([ ['event' => 'loadClassMetadata'] ], $definition->getTags()['doctrine.event_subscriber']);
```
ostrolucky, line 657:
```suggestion
        $this->assertSame([ ['event' => 'loadClassMetadata'] ], $definition->getTags()['doctrine.event_listener']);
```

ostrolucky (changes requested):

ostrolucky (approved):
Thanks for bearing with me!

PR #1161, opened 2020-04-30
Update Entity Listener Docs to New Service Style

huebs:
Switching to Symfony 4's new container style: classes for service names.

### DIFF
--- a/Resources/doc/entity-listeners.rst
+++ b/Resources/doc/entity-listeners.rst
@@ -31,8 +31,7 @@ Full example:
     .. code-block:: yaml
 
         services:
-            user_listener:
-                class: \UserListener
+            UserListener:
                 tags:
                     # Minimal configuration below
                     - { name: doctrine.orm.entity_listener }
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
This style is recommended only for applications. For bundles, decoupling the class and the id is still recommended. Also, big :-1: as this is a BC break.

nicolas-grekas:
Aaaand, this is for the doc, so my comment is void... sorry :D

SenseException:
@nicolas-grekas Is this recommendation somewhere in the Symfony docs? Maybe this could be added as a link since the docs just explain how to use the bundle.

Resources/doc/entity-listeners.rst
```diff
@@ -46,7 +45,7 @@ Full example:
             xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
 
             <services>
-                <service id="user_listener" class="UserListener">
+                <service id="UserListener">
```
ostrolucky, line 48: This will be confusing for unaware users, because this doesn't look like FCQN anymore. We should change it to eg. `App\UserListener`

PR #1163, opened 2020-05-02
Add support for new MessengerTransportDoctrineSchemaSubscriber

weaverryan:
For https://github.com/symfony/symfony/pull/36655

This adds service integration for 2 new schema subscribers. They allow `doctrine:schema:update` to automatically see tables needed by Messenger & cache.

I've tested this in a real project and it works beautifully: the table show up only when activated (i.e. when using a `doctrine` transport in messenger or configuring a real cache pool that uses the `pdo` adapter.

### DIFF
--- a/Tests/CacheSchemaSubscriberTest.php
+++ b/Tests/CacheSchemaSubscriberTest.php
@@ -1,0 +1,71 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\Tests;
+
+use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\CacheSchemaSubscriberPass;
+use Doctrine\Bundle\DoctrineBundle\DependencyInjection\DoctrineExtension;
+use Symfony\Bridge\Doctrine\SchemaListener\PdoCacheAdapterDoctrineSchemaSubscriber;
+use Symfony\Bundle\FrameworkBundle\DependencyInjection\FrameworkExtension;
+use Symfony\Component\DependencyInjection\Alias;
+use Symfony\Component\DependencyInjection\Compiler\PassConfig;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\ParameterBag\ParameterBag;
+use Symfony\Component\DependencyInjection\Reference;
+
+class CacheSchemaSubscriberTest extends TestCase
+{
+    public function testSchemaSubscriberWiring() : void
+    {
+        if (! class_exists(PdoCacheAdapterDoctrineSchemaSubscriber::class)) {
+            $this->markTestSkipped('This test requires Symfony 5.1 or higher');
+        }
+
+        $container = new ContainerBuilder(new ParameterBag([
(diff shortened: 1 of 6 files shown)


DependencyInjection/Compiler/WellKnownSchemaFilterPass.php
```diff
@@ -42,7 +42,10 @@ public function process(ContainerBuilder $container)
 
                 case LegacyConnection::class:
                 case Connection::class:
-                    $blacklist[] = $definition->getArguments()[0]['table_name'] ?? 'messenger_messages';
+                    // add the blacklist *unless* the new schema subscriber service exists
+                    if (!$container->getDefinition('doctrine.orm.messenger.doctrine_schema_subscriber')) {
+                        $blacklist[] = $definition->getArguments()[0]['table_name'] ?? 'messenger_messages';
```
weaverryan, line 47: The idea is that we would transition away from this. The new correct solution would be that the subscriber always adds the table to the schema for you - so it should never be in the schema_filter.

ostrolucky:
Have you considered just adding composer conflict rule for messenger < 5.1? I wouldn't mind doing that personally, as 2.1 is exclusively targetted to bring support for symfony 5.1 components and it's release date will be synced with symfony 5.1 release.

stof:
Well, that would mean that any project using the 4.4 LTS won't be able to upgrade to DoctrineBundle 2.1, and so Doctrine will have to provide LTS support for 2.0.x of the bundle. I'd rather keep support for 4.4 and 5.x of Symfony in the same maintained version of the bundle (meaning that we don't have to support 2 branches, and Symfony 4.4 users still benefit from future improvements of the bundle).

DependencyInjection/Compiler/WellKnownSchemaFilterPass.php
nicolas-grekas: Is this needed for the PR? If not, I'd suggest leaving this as is - it's unrelated and still works.

weaverryan:
Pending tests, this is ready! It works well on a real project and I managed to get a nice "integration" test on the cache compiler pass, which is a bit more complex.

weaverryan:
Tests are green :)

Tests/CacheSchemaSubscriberTest.php
```diff
@@ -0,0 +1,71 @@
...
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\ParameterBag\ParameterBag;
+use Symfony\Component\DependencyInjection\Reference;
+
+class CacheSchemaSubscriberTest extends TestCase
+{
+    public function testSchemaSubscriberWiring() : void
+    {
+        if (! class_exists(PdoCacheAdapterDoctrineSchemaSubscriber::class)) {
+            $this->markTestSkipped();
```
greg0ire, line 20:
```suggestion
            $this->markTestSkipped('This test requires Symfony 5.1 or higher');
```

alcaeus:
Thanks @weaverryan!

PR #1165, opened 2020-05-07
Update comment on the schema_filter key

Gwalchmei:
See https://github.com/doctrine/dbal/issues/3986

The `schema_filter` key does not filter only tables.

### DIFF
--- a/Resources/doc/configuration.rst
+++ b/Resources/doc/configuration.rst
@@ -96,8 +96,9 @@ Configuration Reference
                         platform_service:     ~
                         auto_commit:          ~
 
-                        # If set to "/^sf2_/" all tables not prefixed with "sf2_" will be ignored by the schema
-                        # tool. This is for custom tables which should not be altered automatically.
+                        # If set to "/^sf2_/" all tables, and any named objects such as sequences
+                        # not prefixed with "sf2_" will be ignored by the schema tool.
+                        # This is for custom tables which should not be altered automatically.
                         schema_filter:        ~
 
                         # When true, queries are logged to a "doctrine" monolog channel


Resources/doc/configuration.rst
```diff
@@ -95,7 +95,7 @@ Configuration Reference
                         platform_service:     ~
                         auto_commit:          ~
 
-                        # If set to "/^sf2_/" all tables not prefixed with "sf2_" will be ignored by the schema
+                        # If set to "/^sf2_/" all tables, and any named objects such as sequences, not prefixed with "sf2_" will be ignored by the schema
```
greg0ire, line 98: Please wrap the comment consistently with the previous version

ostrolucky: WDYM? In both versions it's single line so no wrapping, isn't it?

greg0ire: No, the end of the sentence is on a new line (`tool.` is on line 99), so there is wrapping. The goal here is to avoid horizontal in the docs, let's not make the docs harder to read.

Resources/doc/configuration.rst
```diff
@@ -95,8 +95,8 @@ Configuration Reference
                         platform_service:     ~
                         auto_commit:          ~
 
-                        # If set to "/^sf2_/" all tables, and any named objects such as sequences, not prefixed with "sf2_" will be ignored by the schema
-                        # tool. This is for custom tables which should not be altered automatically.
+                        # If set to "/^sf2_/" all tables, and any named objects such as sequences, not prefixed with "sf2_" will be ignored by the schema tool.
```
greg0ire, line 98: What I mean by wrapping is avoiding long lines by splitting them when reaching a certain length

greg0ire:
> Can't really imagine how would we split this to more lines, split sentence at random point?

Either that or you can use https://sembr.org/ if you want. My point is, this is going to look really bad here: 
Peek 2020-05-21 13-25

You should not have to scroll horizontally here, one dimension to explore should be enough.

Resources/doc/configuration.rst
```diff
@@ -95,8 +95,8 @@ Configuration Reference
                         platform_service:     ~
                         auto_commit:          ~
 
-                        # If set to "/^sf2_/" all tables not prefixed with "sf2_" will be ignored by the schema
-                        # tool. This is for custom tables which should not be altered automatically.
+                        # If set to "/^sf2_/" all tables, and any named objects such as sequences, not prefixed with "sf2_" will be ignored by the schema tool.
```
greg0ire, line 98:
```suggestion
                        # If set to "/^sf2_/" all tables, and any named objects such as sequences not prefixed with "sf2_"
                        # will be ignored by the schema tool.
```

ostrolucky:
Looks like there is an auto wrapping. As you can see we don't have current version split to more lines, but it's still split in docs output.

Screenshot 2020-05-21 at 13 38 39

Besides, as you shown in GIF users have to scroll already, so I don't see what this improves, users will still have to scroll.

greg0ire:
But it is currently split, look: https://github.com/doctrine/DoctrineBundle/blob/0ea57dc43c8919a0cea2604d6d309676d7ce317b/Resources/doc/configuration.rst#L98-L100

> Besides, as you shown in GIF users have to scroll already, so I don't see what this improves, users will still have to scroll.

IIRC someone asked how we could improve this, and we didn't find a solution, but let's not make it worse, ok?

PR #1167, opened 2020-05-15
Preload Autoloader class when the ORM is used

fancyweb:
The `Autoloader` class is always used on bundle boot when the ORM is used.

### DIFF
--- a/Tests/DependencyInjection/Fixtures/config/yml/orm_service_single_entity_manager.yml
+++ b/Tests/DependencyInjection/Fixtures/config/yml/orm_service_single_entity_manager.yml
@@ -10,7 +10,7 @@ doctrine:
                 memory: true
 
     orm:
-        default_entity_manager: dm2
+        default_entity_manager: default
         proxy_namespace: Proxies
         auto_generate_proxy_classes: true
         entity_managers:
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
What about using a class_exists at the top of the DoctrineBundle file?
Should we list `ClassUtils` also?

fancyweb:
> What about using a class_exists() at the top of the DoctrineBundle file?

That's what I did first until I realized there is a condition on boot: `$this->container->hasParameter('doctrine.orm.proxy_namespace')`

The class is hit only when the ORM is used.

alcaeus:
Does this fix a bug in 2.0.x? Otherwise I'd opt to introduce this in 2.1.0 which will be released this week.

fancyweb:
Technically it does not a fix a bug, it's an improvement. But I need to take care of tests.

alcaeus:
In that case, let's push this into master. I didn't change the base (yet) as master seems to be missing an up merge from 2.0.x. Either way, let's try to get this into 2.1. Let me know if you need help

Tests/DependencyInjection/Fixtures/config/yml/orm_service_single_entity_manager.yml
```diff
@@ -10,7 +10,7 @@ doctrine:
                 memory: true
 
     orm:
-        default_entity_manager: dm2
+        default_entity_manager: default
```
alcaeus, line 13: The new logic above uncovered this error in the configuration. Looks like we don't check whether the default entity manager even exists 🤷‍♂️

alcaeus:
Thanks @fancyweb!

PR #1171, opened 2020-05-22
Replace unmaintained dependency

### LINKED ISSUE #785: Replace outdated jdorn/sql-formatter dependency with doctrine/sql-formatter
Hello,

I can see that DoctrineBundle uses the [jdorn/sql-formatter](https://github.com/jdorn/sql-formatter) package.

If you check the travis file of the latest tagged version the latest version of the PHP tested against is PHP5.4

Do you believe it would be better to copy this library under the doctrine/common package and drop the dependency to this package?

greg0ire:
Closes #785

### DIFF
--- a/Tests/ProfilerTest.php
+++ b/Tests/ProfilerTest.php
@@ -87,14 +87,18 @@ class ProfilerTest extends BaseTestCase
             'queries' => $this->logger->queries,
         ]);
 
-        $output = str_replace(["\e[37m", "\e[0m", "\e[32;1m", "\e[34;1m"], '', $output);
-        $this->assertContains("SELECT * FROM foo WHERE bar IN ('foo', 'bar') AND \"\" >= \"\";", $output);
-
-        $expectedEscapedSql = 'SELECT&#x20;&#x0A;&#x20;&#x20;&#x2A;&#x20;&#x0A;FROM&#x20;&#x0A;&#x20;&#x20;foo&#x20;&#x0A;WHERE&#x20;&#x0A;&#x20;&#x20;bar&#x20;IN&#x20;&#x28;&#x3F;,&#x20;&#x3F;&#x29;&#x20;&#x0A;&#x20;&#x20;AND&#x20;&quot;&quot;&#x20;&gt;&#x3D;&#x20;&quot;&quot;';
-        $this->assertContains($expectedEscapedSql, $output);
+        $expectedEscapedSql = 'SELECT&#x0A;&#x20;&#x20;&#x2A;&#x0A;FROM&#x0A;&#x20;&#x20;foo&#x0A;WHERE&#x0A;&#x20;&#x20;bar&#x20;IN&#x20;&#x28;&#x3F;,&#x20;&#x3F;&#x29;&#x0A;&#x20;&#x20;AND&#x20;&quot;&quot;&#x20;&gt;&#x3D;&#x20;&quot;&quot;';
         $this->assertSame(
-            "SELECT \n  * \nFROM \n  foo \nWHERE \n  bar IN (?, ?) \n  AND \"\" >= \"\"",
+            "SELECT\n  *\nFROM\n  foo\nWHERE\n  bar IN (?, ?)\n  AND \"\" >= \"\"",
             html_entity_decode($expectedEscapedSql)
         );
+
+        $this->assertContains($expectedEscapedSql, $output);
+
+        $this->assertSame(1, preg_match('/' . str_replace(
+            ' ',
+            '.*',
+            preg_quote('SELECT * FROM foo WHERE bar IN ( ? , ? )')
+        ) . '/', $output));
     }
 }
--- a/Twig/DoctrineExtension.php
+++ b/Twig/DoctrineExtension.php
@@ -2,7 +2,9 @@
 
 namespace Doctrine\Bundle\DoctrineBundle\Twig;
 
-use SqlFormatter;
+use Doctrine\SqlFormatter\HtmlHighlighter;
+use Doctrine\SqlFormatter\NullHighlighter;
+use Doctrine\SqlFormatter\SqlFormatter;
 use Symfony\Component\VarDumper\Cloner\Data;
 use Twig\Extension\AbstractExtension;
 use Twig\TwigFilter;
@@ -12,6 +14,9 @@ use Twig\TwigFilter;
  */
 class DoctrineExtension extends AbstractExtension
 {
+    /** @var SqlFormatter */
+    private $sqlFormatter;
+
     /**
      * Define our functions
      *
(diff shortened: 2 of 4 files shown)


Twig/DoctrineExtension.php
```diff
@@ -315,26 +316,26 @@ static function ($matches) use ($parameters, &$i) {
...
+            HtmlHighlighter::HIGHLIGHT_QUOTE          => 'class="string"',
+            HtmlHighlighter::HIGHLIGHT_BACKTICK_QUOTE => 'class="string"',
+            HtmlHighlighter::HIGHLIGHT_RESERVED       => 'class="keyword"',
+            HtmlHighlighter::HIGHLIGHT_BOUNDARY       => 'class="symbol"',
+            HtmlHighlighter::HIGHLIGHT_NUMBER         => 'class="number"',
+            HtmlHighlighter::HIGHLIGHT_WORD           => 'class="word"',
+            HtmlHighlighter::HIGHLIGHT_ERROR          => 'class="error"',
+            HtmlHighlighter::HIGHLIGHT_COMMENT        => 'class="comment"',
+            HtmlHighlighter::HIGHLIGHT_VARIABLE       => 'class="variable"',
+        ], false));
```
greg0ire, line 329: With this change, I make the assumption that this isn't used in the CLI context. This allows us to add some tests.

ostrolucky:
I'm not sure it's ok to merge this into 1.12. I think at least @alcaeus would mind that. And I'm also leaning to merge this for master only, it might be too risky to do it in 1.12 while we don't have problems with current formatter. If we had some problems, eg. with php 8 compatibility, then I would be fully on board with this.

alcaeus:
I'd advocate to only merge this into 2.1 for the upcoming release.

greg0ire:
Fine by me.

Tests/ProfilerTest.php
```diff
@@ -87,14 +87,10 @@ public function testRender() : void
...
-        $output = str_replace(["\e[37m", "\e[0m", "\e[32;1m", "\e[34;1m"], '', $output);
-        $this->assertContains("SELECT * FROM foo WHERE bar IN ('foo', 'bar') AND \"\" >= \"\";", $output);
-
-        $expectedEscapedSql = 'SELECT&#x20;&#x0A;&#x20;&#x20;&#x2A;&#x20;&#x0A;FROM&#x20;&#x0A;&#x20;&#x20;foo&#x20;&#x0A;WHERE&#x20;&#x0A;&#x20;&#x20;bar&#x20;IN&#x20;&#x28;&#x3F;,&#x20;&#x3F;&#x29;&#x20;&#x0A;&#x20;&#x20;AND&#x20;&quot;&quot;&#x20;&gt;&#x3D;&#x20;&quot;&quot;';
-        $this->assertContains($expectedEscapedSql, $output);
-        $this->assertSame(
-            "SELECT \n  * \nFROM \n  foo \nWHERE \n  bar IN (?, ?) \n  AND \"\" >= \"\"",
-            html_entity_decode($expectedEscapedSql)
-        );
+        $this->assertSame(1, preg_match('/' . str_replace(
```
ostrolucky, line 90: Those assertions you removed before were important, they were testing escaping functionality that was broken at one point. Important to watch for regressions. Perhaps we should go with true flag as we used to so it's formatted in CLI too after all. Those unit tests are not super important since we have this integration test, but if you want those, you can make assertions for them in similar fashion as was done here.

greg0ire: Oh right I think it is indeed messed up… is this assertion  about the `data-clipboard-text`? It looks like this: `data-clipboard-text="&lt;pre&#x20;class&#x3D;&quot;highlight&#x20;highlight-sql&quot;&gt;&lt;span&#x20;class&#x3D;&quot;keyword&quot;&gt;SELECT&lt;&#x2F;span&gt;&#x0A;&#x20;&#x20;&lt;span&#x20;class&#x3D;&quot;symbol&quot;&gt;&#x2A;&lt;&#x2F;span&gt;&#x0A;&lt;span&#x20;class&#x3D;&quot;keyword&quot;&gt;FROM&lt;&#x2F;span&gt;&#x0A;&#x20;&#x20;&lt;span&#x20;class&#x3D;&quot;word&quot;&gt;foo&lt;&#x2F;span&gt;&#x0A;&lt;span&#x20;class&#x3D;&quot;keyword&quot;&gt;WHERE&lt;&#x2F;span&gt;&#x0A;&#x20;&#x20;&lt;span&#x20;class&#x3D;&quot;word&quot;&gt;bar&lt;&#x2F;span&gt;&#x20;&lt;span&#x20;class&#x3D;&quot;keyword&quot;&gt;IN&lt;&#x2F;span&gt;&#x20;&#x28;&lt;span&#x20;class&#x3D;&quot;word&quot;&gt;&#x3F;&lt;&#x2F;span&gt;&lt;span&#x20;class&#x3D;&quot;symbol&quot;&gt;,&lt;&#x2F;span&gt;&#x20;&lt;span&#x20;class&#x3D;&quot;word&quot;&gt;&#x3F;&lt;&#x2F;span&gt;&#x29;&#x0A;&#x20;&#x20;&lt;span&#x20;class&#x3D;&quot;keyword&quot;&gt;AND&lt;&#x2F;span&gt;&#x20;&lt;span&#x20;class&#x3D;&quot;string&quot;&gt;&amp;quot&#x3B;&amp;quot&#x3B;&lt;&#x2F;span&gt;&#x20;&lt;span&#x20;class&#x3D;&quot;symbol&quot;&gt;&amp;gt&#x3B;&lt;&#x2F;span&gt;&lt;span&#x20;class&#x3D;&quot;symbol&quot;&gt;&#x3D;&lt;&#x2F;span&gt;&#x20;&lt;span&#x20;class&#x3D;&quot;string&quot;&gt;&amp;quot&#x3B;&amp;quot&#x3B;&lt;&#x2F;span&gt;&lt;&#x2F;pre&gt;"`

greg0ire: I have found an error in this PR that, when fixed, leads to `data-clipboard-text="SELECT&#x20;&#x2A;&#x20;FROM&#x20;foo&#x20;WHERE&#x20;bar&#x20;IN&#x20;&#x28;&#x27;foo&#x27;,&#x20;&#x27;bar&#x27;&#x29;&#x20;AND&#x20;&quot;&quot;&#x20;&gt;&#x3D;&#x20;&quot;&quot;&#x3B;"`

greg0ire: It would make sense to use CLI formatting for `data-clipboard-text`, I'm going to try that.

greg0ire: Since highlighting is turned off, it shouldn't matter. the only difference here are the linebreaks.

greg0ire: Okay I finally understood. The test can stay, but has to be changed because I fixed https://github.com/doctrine/sql-formatter/issues/29

ostrolucky: Yeah I don't care about highlighting, but having something that tests for escaping regressions. And yeah I think it's about data-clipboard-text

greg0ire: I think it is fixed now :) Thanks for spotting this, you rock!

PR #1173, opened 2020-05-25
Allow DBAL 3.0 and ignore ORM tests on php 8

derrabus:
On order to run at least a few tests on php 8, this PR…

* Allows DBAL 3.0 to be installed because DBAL 2 will trigger fatal errors on php 8.
* Uninstalls ORM on the php 8 build because there currently is no ORM version that allows DBAL 3.
* Skips all tests that require ORM if it's not there.
* Ignores deprecation warnings because PHPUnit will trigger lots of them. We need to upgrade PHPUnit to fix this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,8 +65,10 @@ jobs:
 
     # Test against PHP 8
     - php: nightly
+      env: SYMFONY_DEPRECATIONS_HELPER=weak
       before_install:
         - composer config platform.php 7.4.99
+        - composer remove --dev doctrine/orm --no-update
       install:
         - travis_retry composer update -n --prefer-dist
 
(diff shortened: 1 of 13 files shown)


.travis.yml
```diff
@@ -65,8 +65,10 @@ jobs:
 
     # Test against PHP 8
     - php: nightly
+      env: SYMFONY_DEPRECATIONS_HELPER=weak
       before_install:
         - composer config platform.php 7.4.99
+        - composer remove --dev doctrine/orm
```
alcaeus, line 71: You could either use `--no-update` or remove the `update` line below since it is superfluous.

derrabus: You're right! Done.

PR #1175, opened 2020-05-29
Bugfix: drop database command was not working with shard argument.

ribeiropaulor:
The driver param is stored in global params, so the drop database for shards was throwing an exception.


ostrolucky:
If you check https://github.com/doctrine/DoctrineBundle/blame/master/Command/CreateDatabaseDoctrineCommand.php#L72, there are lot of options unset there. Don't we need to do the same here?

ribeiropaulor:
I don't think so. It's working as I left.

Below that line the dbname/url is already being unset. That's the most important thing as some vendors do not allow dropping the database connected to.

I could unset the same params, but I don't see a reason for it.

PR #1177, opened 2020-06-08
[Bug] Run CacheSchemaSubscriberPass after CacheCollectorPass

sstok:
The `CacheCollectorPass` of the Cache component decorates the cache adapters with
a `TracingAdapter` which breaks the expectation that the provided Adapter provided to
the `CacheSchemaSubscriber` is a `PdoAdapter` with the `configureSchema` method.

By running the `CacheSchemaSubscriberPass` after the `CacheCollectorPass` only
the actual Adapters (not decorated one) are correctly provided.

Note: This bug only happens when debugging *and* the profiler is enabled.

/cc @weaverryan

### DIFF
--- a/DoctrineBundle.php
+++ b/DoctrineBundle.php
@@ -41,7 +41,7 @@ class DoctrineBundle extends Bundle
         $container->addCompilerPass(new ServiceRepositoryCompilerPass());
         $container->addCompilerPass(new WellKnownSchemaFilterPass());
         $container->addCompilerPass(new DbalSchemaFilterPass());
-        $container->addCompilerPass(new CacheSchemaSubscriberPass(), PassConfig::TYPE_OPTIMIZE, -10);
+        $container->addCompilerPass(new CacheSchemaSubscriberPass(), PassConfig::TYPE_BEFORE_REMOVING, -10);
     }
 
     /**
--- a/Tests/CacheSchemaSubscriberTest.php
+++ b/Tests/CacheSchemaSubscriberTest.php
@@ -65,7 +65,7 @@ class CacheSchemaSubscriberTest extends TestCase
         $container->register('uses_my_cache_adapter', 'stdClass')
             ->addArgument(new Reference('my_cache_adapter'))
             ->setPublic(true);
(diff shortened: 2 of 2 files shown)


ostrolucky:
It looks to me like your description conflicts itself:

1. > CacheCollectorPass of the Cache component decorates the cache adapters with
a TracingAdapter
1. > By running the CacheSchemaSubscriberPass after the CacheCollectorPass only
the actual Adapters (not decorated one) are correctly provided.

If `CacheCollectorPass` causes issue, how could this problem be solved by executing `CacheSchemaSubscriberPass` *after* it?

And by the way what's the actual issue from user's perspective?

sstok:
The actual problem is this:

```
Database
--------


In PdoCacheAdapterDoctrineSchemaSubscriber.php line 40:

  Attempted to call an undefined method named "configureSchema" of class "Symfony\Component\Cache\Adapter\TraceableAdapter".


doctrine:schema:validate [--skip-mapping] [--skip-sync] [--em [EM]] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-e|--env ENV] [--no-debug] [--] <command>

bin/console doctrine:schema:update --force
```

> how could this problem be solved by executing CacheSchemaSubscriberPass after it?

Because the adapter is decorated, the original definition is replaced with the `TraceableAdapter`, but if you execute `CacheSchemaSubscriberPass` after this process it will collect the original Adapter (which is still available under a different service-id). See https://github.com/symfony/symfony/blob/be8adfaf999b96c48e5543a6b72bd4f83e7f3b9a/src/Symfony/Component/Cache/DependencyInjection/CacheCollectorPass.php#L76-L77 for details.

ostrolucky:
Please also update this test case https://github.com/doctrine/DoctrineBundle/blob/d7b61d6a6cfb50bb5162cc51e3cbb243ce88f626/Tests/CacheSchemaSubscriberTest.php#L64

sstok:
> Please also update this test case 

Done 👍

acasademont:
Thanks @sstok I was just hit by this when updating to 2.1, hope it can be merged soon, I'm reverting back to 2.0.x meanwhile

PR #1180, opened 2020-06-23
Allow passing a custom connection factory for DBAL

### LINKED ISSUE #1178: Dynamic connection parameters
I need the ability to pass the hostname, database name and port dynamically (i.e. I don't know those until runtime). I'm happy to send a PR that would allow that, though I'd like to discuss the approach first.
There are two solutions that come to my mind:
1. Create a new option for DBAL that would allow passing a factory (a service from DI) and a method that would just return a DBAL connection.
2. Create an extension point that would allow dynamically generating the connection parameters, but would still use the current instantiation logic.

I'm open to other ideas (or maybe there's a way to do it with the bundle as it is right now that I missed?).

Agares:
This resolves #1178. My goal is to allow resolving connection string and other connection parameters at runtime.
Happy to discuss different approaches, if this is not the right solution.

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -118,6 +118,20 @@ class DoctrineExtension extends AbstractDoctrineExtension
     protected function loadDbalConnection($name, array $connection, ContainerBuilder $container)
     {
         $configuration = $container->setDefinition(sprintf('doctrine.dbal.%s_connection.configuration', $name), new ChildDefinition('doctrine.dbal.connection.configuration'));
+        $connectionServiceId = sprintf('doctrine.dbal.%s_connection', $name);
+
+        if (isset($connection['factory'])) {
+            $definition = new Definition();
+            $definition
+                ->setPublic(true)
+                ->setFactory(new Reference($connection['factory']))
+                ->setArguments($connection['factory_arguments']);
+
+            $container->setDefinition($connectionServiceId, $definition);
+
+            return;
+        }
+
         $logger        = null;
         if ($connection['logging']) {
             $logger = new Reference('doctrine.dbal.logger');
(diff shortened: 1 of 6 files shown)


Agares:
I'm not quite sure why the coverage build is failing, the tests themselves seem to be passing

ostrolucky:
This is not extendable. I was expecting `factory` will be just service id.

Agares:
It is a service id, as you can see in the config file here: https://github.com/doctrine/DoctrineBundle/pull/1180/files#diff-736234b5eafd2fee078df0773ac8cfaeR10. In the test it's defined as a Clousre: https://github.com/doctrine/DoctrineBundle/pull/1180/files#diff-f8d40d5cb350684762a2ac727468690fR1042-R1047 but it could also be a class that has the __invoke magic method (or anything else that is invokable).

DependencyInjection/DoctrineExtension.php
```diff
@@ -117,8 +117,22 @@ protected function dbalLoad(array $config, ContainerBuilder $container)
...
+        if (isset($connection['factory'])) {
+            $definition = new Definition();
+            $definition
+                ->setPublic(true)
+                ->setFactory(new Reference($connection['factory']))
+                ->setArguments($connection['factory_arguments']);
+
+            $container->setDefinition($connectionServiceId, $definition);
+
+            return;
```
ostrolucky, line 132: This is why I hesitated to accept this feature request. Custom factory on config level like this means ignoring all the other config options. This is not desirable. It's not expected by user that when they see config of connection1 composed of 10 keys, only 1 is applied and rest are ignored. If this feature is accepted, I think it should either:
- Not be part of the config, only generate `doctrine.dbal.%_connection.factory` definitions for each connection - then users can just redefine it. This makes it obvious that since extension point is on DI level, fault cannot be in doctrine-bundle config. But I actually don't see why can't you just define your `doctrine.dbal.%_connection` service, this is almost same thing.
- Be part of a config, but all existing config options need to be in effect. That would probably mean there needs to be prescribed certain format of factory which would be compatible with our existing factory - so that all values that are generated here can be passed to it.

Agares: 🤦  for reasons which will forever remain a mystery, I was under the impression that to create my own connection, I have to first create it in the `dbal` section and only then override it. But that's of course not the case and it works.
Thanks for your help! Feel free to close this PR.

ostrolucky:
Perhaps fault was from my side too, since I didn't really understand your use case so I was unnecessarily waiting for PR just so I could see what you need and could tell you easier way. But yeah, let's close this.

PR #1184, opened 2020-07-02
Deprecate ImportDoctrineCommand

seferov:
As base command `Doctrine\DBAL\Tools\Console\Command\ImportCommand` had been deprecated in 2.x and removed 3.x, this command should be deprecated as well.


seferov:
@ostrolucky actually the command is deprecated in dbal 2.9 (https://github.com/doctrine/dbal/commit/f242a9db3acbc2d1779bbb09982a4d6e3bf6c1e6) and released 2 years ago.
I think it should be on branch starting 2.0.x since it depends on [doctrine/dbal: ^2.9.0](https://github.com/doctrine/DoctrineBundle/blob/2.0.x/composer.json)

ostrolucky:
Ah, my bad. Then this is good to go :)

PR #1195, opened 2020-08-18
Address deprecation of Definition::setPrivate()

greg0ire:
~The call might be useless in many versions of Symfony, not sure it's
useless for all of them though. setPrivate is deprecated since 5.2, in
favor of setPublic which is available since 2.0.~

Services are private by default since symfony/dependency-injection 4.0,
which means the deprecated call can safely be dropped.


ostrolucky:
Services are private by default since Symfony 4.0, I don't see why we need this call a anymore

dmaicher (approved):

PR #1196, opened 2020-08-20
Add PhpArrayAdapter to cache metadata

### LINKED ISSUE #1186: Add PhpArrayAdapter to cache metadata
I propose to cache metadata using PhpArrayAdapter the same way as serializer or validator in Symfony.
I've added it in my project and got 1-2% speed up on production.

If the communtity approve this, I will create a PR.

ossinkine:
Closes #1186

cc @ostrolucky @dmaicher @nicolas-grekas 

I see two weak places in this implementation:
1. `DoctrineMetadataCacheWarmer` extends `AbstractPhpFileCacheWarmer` which also extended by `ValidatorCacheWarmer`, `SerializerCacheWarmer` and `AnnotationsCacheWarmer` (everywhere `PhpArrayAdapter` used) but it's marked as internal, so BC breaks possible here. I can avoid extending this class but a lot of code will be copy-pasted.
2. Since `PhpArrayAdapter` requires Symfony cache I wrap Doctrine cache with `DoctrineAdapter`. If Doctrine cache is `DoctrineProvider` which wrap a Symfony cache we get a redudant chain of adapters. For example the chain `PhpArrayAdapter -> DoctrineAdapter -> DoctrineProvider -> RedisAdapter` can be simplified by removing `DoctrineAdapter` and `DoctrineProvider`: `PhpArrayAdapter -> RedisAdapter`. I've tried to create a compiler pass which remove this redundancy but get list of entity managers in compiler pass a little bit difficult. And maybe we can implement this simplification in Symfony framework bundle globally for all such cache chains.

What is your opinion about this?

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -688,6 +688,13 @@ class Configuration implements ConfigurationInterface
                 ->scalarNode('pool')->end()
             ->end();
 
+        if ($name === 'metadata_cache_driver') {
+            $node->setDeprecated(...$this->getDeprecationMsg(
+                'The "metadata_cache_driver" configuration key is deprecated. PHP Array cache is now automatically registered when %kernel.debug% is false.',
+                '2.2'
+            ));
+        }
+
         return $node;
     }
 
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection;
 
+use Doctrine\Bundle\DoctrineBundle\CacheWarmer\DoctrineMetadataCacheWarmer;
 use Doctrine\Bundle\DoctrineBundle\Dbal\ManagerRegistryAwareConnectionProvider;
 use Doctrine\Bundle\DoctrineBundle\Dbal\RegexSchemaAssetFilter;
 use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\ServiceRepositoryCompilerPass;
@@ -21,6 +22,8 @@ use Symfony\Bridge\Doctrine\SchemaListener\MessengerTransportDoctrineSchemaSubsc
 use Symfony\Bridge\Doctrine\SchemaListener\PdoCacheAdapterDoctrineSchemaSubscriber;
 use Symfony\Bridge\Doctrine\Validator\DoctrineLoader;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
+use Symfony\Component\Cache\Adapter\DoctrineAdapter;
+use Symfony\Component\Cache\Adapter\PhpArrayAdapter;
 use Symfony\Component\Cache\DoctrineProvider;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Alias;
@@ -814,6 +817,11 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $this->loadCacheDriver('metadata_cache', $entityManager['name'], $entityManager['metadata_cache_driver'], $container);
         $this->loadCacheDriver('result_cache', $entityManager['name'], $entityManager['result_cache_driver'], $container);
         $this->loadCacheDriver('query_cache', $entityManager['name'], $entityManager['query_cache_driver'], $container);
+
+        if (! $container->getParameter('kernel.debug')) {
+            $this->registerMetadataPhpArrayCacheWarmer($entityManager['name'], $container);
+            $this->registerMetadataPhpArrayCache($entityManager['name'], $container);
+        }
     }
 
     /**
@@ -926,5 +934,37 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $container->setDefinition($id, $poolDefinition);
 
         return $id;
+    }
+
+    private function registerMetadataPhpArrayCacheWarmer(string $entityManagerName, ContainerBuilder $container): void
+    {
+        $cacheWarmerDefinition = $container->register(sprintf('doctrine.orm.%s_metadata_cache.php_array_warmer', $entityManagerName), DoctrineMetadataCacheWarmer::class);
+        $cacheWarmerDefinition->setArguments([
+            new Reference(sprintf('doctrine.orm.%s_entity_manager', $entityManagerName)),
+            $this->getPhpArrayFile($entityManagerName),
+        ]);
+        $cacheWarmerDefinition->addTag('kernel.cache_warmer');
+    }
+
+    private function registerMetadataPhpArrayCache(string $entityManagerName, ContainerBuilder $container): void
+    {
+        $aliasId   = sprintf('doctrine.orm.%s_metadata_cache', $entityManagerName);
+        $serviceId = (string) $container->getAlias($aliasId);
+
+        $phpArrayAdapterDefinition = new Definition(PhpArrayAdapter::class, [
+            $this->getPhpArrayFile($entityManagerName),
+            new Definition(DoctrineAdapter::class, [new Reference($serviceId)]),
+        ]);
+
+        $serviceId .= '.php_array';
+
+        $doctrineCacheDefinition = $container->register($serviceId, DoctrineProvider::class);
+        $doctrineCacheDefinition->addArgument($phpArrayAdapterDefinition);
+        $container->setAlias($aliasId, $serviceId);
+    }
+
+    private function getPhpArrayFile(string $entityManagerName): string
+    {
+        return '%kernel.cache_dir%' . sprintf('/doctrine/orm/%s_metadata.php', $entityManagerName);
     }
 }
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -818,10 +818,11 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $this->loadCacheDriver('result_cache', $entityManager['name'], $entityManager['result_cache_driver'], $container);
         $this->loadCacheDriver('query_cache', $entityManager['name'], $entityManager['query_cache_driver'], $container);
 
-        if (! $container->getParameter('kernel.debug')) {
-            $this->registerMetadataPhpArrayCacheWarmer($entityManager['name'], $container);
-            $this->registerMetadataPhpArrayCache($entityManager['name'], $container);
+        if ($container->getParameter('kernel.debug')) {
+            return;
         }
+
+        $this->registerMetadataPhpArrayCaching($entityManager['name'], $container);
     }
 
     /**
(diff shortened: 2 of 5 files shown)


nicolas-grekas:
> `AbstractPhpFileCacheWarmer` is internal

That can be changed, PR welcome on symfony/symfony to discuss it there.

> DoctrineAdapter -> DoctrineProvider

This would be better implemented on the bundle imho, as that would allow releasing the logic in sync with the feature provided here.

ostrolucky:
Tests are DI only, very simplistic. Can we have test which asserts that cache is hit when loading metadata and not generated again? Also, can you provide some numbers so everyone can see what's the expected improvement?

DependencyInjection/DoctrineExtension.php
```diff
@@ -898,4 +906,34 @@ private function createArrayAdapterCachePool(ContainerBuilder $container, string
...
+    {
+        $cacheWarmerDefinition = $container->register(sprintf('doctrine.orm.%s_metadata_cache.php_array_warmer', $entityManagerName), DoctrineMetadataCacheWarmer::class);
+        $cacheWarmerDefinition->setArguments([
+            new Reference(sprintf('doctrine.orm.%s_entity_manager', $entityManagerName)),
+            $this->getPhpArrayFile($entityManagerName)
+        ]);
+        $cacheWarmerDefinition->addTag('kernel.cache_warmer');
+    }
+
+    private function registerMetadataPhpArrayCache(string $entityManagerName, ContainerBuilder $container): void
```
ostrolucky, line 920: I would personally prefer to not unnecessarily register new service if its only purpose is to inject it to warmer. Service should be instead just inlined, IMHO. Otherwise we are unnecessarily increasing public API surface.

DependencyInjection/Compiler/PhpArrayCachePass.php
```diff
@@ -0,0 +1,44 @@
...
+use Symfony\Component\Cache\Adapter\ProxyAdapter;
+use Symfony\Component\Cache\DoctrineProvider;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+
+/**
+ * Removes redundant decorators such DoctrineAdapter -> DoctrineProvider -> AdapterInterface.
+ */
+class PhpArrayCachePass implements CompilerPassInterface
```
ostrolucky, line 15: We would prefer not having to maintain this in bundle forever. Can you please contribute this to Symfony core too? That way you will also get better feedback, as I'm not exactly confident about these internals.

ostrolucky (changes requested):
Tests are DI only, very simplistic. Can we have test which asserts that cache is hit when loading metadata and not generated again? Also, can you provide some numbers so everyone can see what's the expected improvement?

Also, we probably need to mark a test as legacy so CI passes.
And of course, needs to be rebased as it has conflicts.

DependencyInjection/DoctrineExtension.php
ossinkine: `doctrine.orm.default_metadata_cache` is the alias which refers to the cache service. How can I create inline service and refer the alias to it?

DependencyInjection/Compiler/PhpArrayCachePass.php
ossinkine: As I understand @nicolas-grekas has another opinion. Since this is just optimization we can merge without it.

ossinkine:
Tests don't passed because AbstractPhpFileCacheWarmer marked as internal, `@group legacy` does not hide the errors

nicolas-grekas:
Hmm, I didn't consider this aspect.
I'd be fine removing the annotation on branch 3.4 on Symfony.

DependencyInjection/Compiler/PhpArrayCachePass.php
ostrolucky: Nicolas didn't mean it doesn't make sense to contribute to Symfony, he said it needs to be merged to doctrine bundle anyways, as otherwise this feature would be available in next minor Symfony version only. But yeah I am also ok with just removing it for now, would make it easier to merge

ostrolucky:
Any chance for finishing this @ossinkine?

ossinkine:
@ostrolucky I can remove `PhpArrayCachePass` which optimizes cache adapter chaining right now and later publish PR to Symfony but I don't what tests you requested and I don't how much time it will take

ostrolucky:
Sounds good enough to me!

ossinkine:
@ostrolucky I've removed the commit with optimizations, rebased the branch from the master and fixed code style issue except the one, I think proposed fix is a bad idea

ostrolucky:
How about this. Does that work for you @ossinkine?

stof:
> get list of entity managers in compiler pass a little bit difficult.

this could actually be simple. We can add a tag on the services and use that tag in the compiler pass

DependencyInjection/Configuration.php
```diff
@@ -735,7 +741,7 @@ private function getDeprecationMsg(string $message): array
         if (method_exists(BaseNode::class, 'getDeprecation')) {
             return [
                 'doctrine/doctrine-bundle',
-                '2.0',
+                '2.2',
```
stof, line 744: This change is wrong. If different methods need different versions, you need to pass that as an argument, not changing the version for previous deprecations

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -776,6 +793,7 @@ public function testInvalidCacheConfiguration(): void
      * @param array|string $cacheConfig
      *
      * @dataProvider cacheConfigurationProvider
+     * @group legacy
```
stof, line 796: The legacy cases must be split into separate tests. Not everything about cache configuration is legacy.

DependencyInjection/Configuration.php
ostrolucky: it's already wrong in master, will fix in separate commit

Tests/DependencyInjection/DoctrineExtensionTest.php
ostrolucky: i don't this it's worth it

DependencyInjection/Configuration.php
```diff
@@ -688,6 +688,13 @@ private function getOrmCacheDriverNode(string $name): ArrayNodeDefinition
                 ->scalarNode('pool')->end()
             ->end();
 
+        if ($name === 'metadata_cache_driver') {
```
dmaicher, line 691: this means we also need to create a new recipe for 2.2?

See https://github.com/symfony/recipes/blob/master/doctrine/doctrine-bundle/1.12/config/packages/prod/doctrine.yaml#L4 (it's used for 2.x currently as well via symlink)

ostrolucky: should yes

ossinkine:
@ostrolucky since PHP Array Cache adapter is a decorator for another adapter I'm not sure `metadata_cache_driver` should be depreacated. Or you suppose there can't be cache misses?

ostrolucky:
If PHP Array Cache adapter is a decorator for another adapter, what happens right now when `metadata_cache_driver` is not set?

Tests/DependencyInjection/DoctrineExtensionTest.php
stof: to me it is, for 2 reasons:

- when you remove a deprecated feature, you can remove the associated legacy tests. But here, that would remove the coverage for other cache configurations too
- marking tests covering other (non-deprecated) features as `@legacy` can hide accidental usage of deprecated APIs (which is precisely the reason to make test fail on deprecations in non-legacy tests)

DependencyInjection/Configuration.php
```diff
@@ -688,6 +688,13 @@ private function getOrmCacheDriverNode(string $name): ArrayNodeDefinition
...
+            $node->setDeprecated(...$this->getDeprecationMsg(
+                'The "metadata_cache_driver" configuration key is deprecated. PHP Array cache is now automatically registered when %kernel.debug% is false.',
```
greg0ire, line 693: Why are we deprecating this configuration key? There should be an explanation in the commit message.

DependencyInjection/DoctrineExtension.php
```diff
@@ -927,4 +936,25 @@ private function createArrayAdapterCachePool(ContainerBuilder $container, string
 
         return $id;
     }
+
+    private function registerMetadataPhpArrayCaching(string $entityManagerName, ContainerBuilder $container): void
+    {
+        $metadataCacheAlias              = $this->getObjectManagerElementName($entityManagerName . '_metadata_cache');
+        $decoratedMetadataCacheServiceId = (string) $container->getAlias($metadataCacheAlias);
+        $phpArrayCacheDecoratorServiceId = $decoratedMetadataCacheServiceId . '.php_array';
```
greg0ire, line 944: Why is there decoration involved?

greg0ire (commented):
Please state the obvious. It would be great not to have to read several issues and PRs on different repositories to understand this.

DependencyInjection/Configuration.php
ostrolucky: We are deprecating it because it doesn't make sense to specify any other cache than what we automatically configure now. If we are leaving it, then it would make more sense to change flex recipe instead.

DependencyInjection/DoctrineExtension.php
ostrolucky: PhpArrayAdapter requires it https://github.com/symfony/symfony/blob/57e39b41b94dd9d0beb04c7f88fc0955a5bbe605/src/Symfony/Component/Cache/Adapter/PhpArrayAdapter.php#L48. Also for BC, so current configuration is not completely ignored until we remove it.

greg0ire (approved):
I pushed new commits with explanations that might be completely wrong, please review and amend them if necessary.

stephanvierkant:
It isn't clear to me what I've to change in my application and I'm not the only one: https://stackoverflow.com/questions/64741823/symfony-4-4-metadata-cache-driver-configuration-key-deprecation-notice

Can someone explain why this change was needed and how we should replace `metadata_cache_driver`? If I understand, I can update the Symfony docs (if needed).

ostrolucky:
metadata_cache_driver configuration needs to be removed. Change is needed because defining own metadata_cache_driver is useless from now on.

DependencyInjection/DoctrineExtension.php
```diff
@@ -927,4 +936,25 @@ private function createArrayAdapterCachePool(ContainerBuilder $container, string
...
+        $phpArrayFile                    = '%kernel.cache_dir%' . sprintf('/doctrine/orm/%s_metadata.php', $entityManagerName);
+
+        $container->register(DoctrineMetadataCacheWarmer::class)
```
dmaicher, line 947: @ossinkine actually this is not working properly for multiple entity managers. I just tried it on one of my bigger projects that uses 2 managers.

Since `DoctrineMetadataCacheWarmer::class` is the service id it will only work for the last manager and overwrite any previous definitions

Issue #1201, opened 2020-08-25
BC break at 2.1.1


KartaviK:
At version 2.1.1 i got this

```bash
!!  
!!  In DefinitionErrorExceptionPass.php line 54:
!!                                                                                 
!!    Cannot autowire service "App\Repository\UserRepository": argument "$regi  
!!    stry" of method "__construct()" references interface "Doctrine\Common\Persi  
!!    stence\ManagerRegistry" but no such service exists. Did you create a class   
!!    that implements this interface?                                              
!!              
```

Restoring at 2.1.0 and all works fine

bastien70:
Same problem

`Cannot autowire service "App\Repository\App\AbsenceRepository": argument "$registry" of method "__construct()" references interface "Doctrine\Common\Persistence\ManagerRegistry" but no such service exists. Did you create a class that implements this interface?`

dmaicher:
I guess we need to restore this service alias?

https://github.com/doctrine/DoctrineBundle/compare/2.1.0..2.1.1#diff-526de6d56db54965998fa70108bbe787L70

KartaviK:
The main problem that there was removed that alias and old usages (who already use this class) was broken.
In current version ServiceEntityRepository uses `Doctrine\Persistence\ManagerRegistry` that does not have alias to `Doctrine\Common\Persistence\ManagerRegistry`. In new automatically generated repositories there are correct classes

KartaviK:
You can totally rename `use Doctrine\Common\Persistence\ManagerRegistry` to `use Doctrine\Persistence\ManagerRegistry` in project and all will work fine

dmaicher:
Release `2.1.2` fixes this BC break.

PR #1202, opened 2020-08-25
fix BC break for removed service alias

dmaicher:
Fixes https://github.com/doctrine/DoctrineBundle/issues/1201

### DIFF
--- a/Resources/config/dbal.xml
+++ b/Resources/config/dbal.xml
@@ -67,6 +67,7 @@
             <tag name="kernel.reset" method="reset" />
         </service>
         <service id="Doctrine\Persistence\ManagerRegistry" alias="doctrine" public="false" />
+        <service id="Doctrine\Common\Persistence\ManagerRegistry" alias="doctrine" public="false" />
 
         <service id="doctrine.twig.doctrine_extension" class="Doctrine\Bundle\DoctrineBundle\Twig\DoctrineExtension" public="false">
             <tag name="twig.extension" />


NicoHaase:
Out of curiosity: is this a long-term fix, as in: we can expect that this service is kept longer, or will this be removed again in nearer future?

dmaicher:
> Out of curiosity: is this a long-term fix, as in: we can expect that this service is kept longer, or will this be removed again in nearer future?

I think we should keep this alias as long as we support `doctrine/persistence < 2.0`. What do you think @ostrolucky ?

maxhelias (approved):

nicolas-grekas:
A quick release would be much appreciated, we're getting the same report on many repos.

alcaeus (approved):

alcaeus:
Merged and released. Thanks @dmaicher.

NicoHaase:
> Should this "just fix everything that broke" ? Im between a rock and a hard place at the moment and cannot resolve it.

What does that mean? Is there any specific problem you are facing?

PhilETaylor:
Sorry, should have put more work in before opening my mouth, Im debugging in smaller steps now to see if I can actually find anything helpful or if Im just an idiot... large project... 

I think my issue is to do with doctrine/persistence upgrade to 2.0.0 form 1.3.8 ... so ignore me totally for now :) sorry.

PhilETaylor:
ok my problem is that FOSUser is showing its age, can't wait to dump it, and its not compatible with doctrine/persistence: 2.0.0 or  doctrine/common: 3.0.2 (not sure which yet) - so best just ignore me here :) sorry.

>Argument 3 passed to FOS\UserBundle\Doctrine\UserManager::__construct() must be an instance of Doctrine\Common\Persistence\ObjectManager, instance of Doctrine\ORM\EntityManager given, called in /tmp/app/cache/Container0xAwrXe/configApp_KernelDevDebugContainer.php on line 3439

ostrolucky:
> I  think we should keep this alias as long as we support doctrine/persistence < 2.0. What do you think @ostrolucky ?

Yep

greg0ire:
I guess we should [deprecate one in favor of the other](https://symfony.com/blog/new-in-symfony-4-3-deprecating-service-aliases). Not sure under which conditions exactly the new service is available/usable. It seems to be defined all the time, so… no condition?

ostrolucky:
Users get deprecations when using common\persistence classes already, I don't think we need to do more there. They know they shoiuld not use them. In some future version of bundle that will drop support doctrine/common 1.x we will just also drop old aliases. By requiring doctrine/common 2.x we will guarantee nobody can use old classes anymore, hence also service aliases.

PR #1207, opened 2020-09-04
Add symfony uuid to a list of custom types as default uuid/ulid type

gennadigennadigennadi:
As a result from https://github.com/symfony/symfony/pull/37678, this PR is the last missing piece to implement symfony/uid as DoctrineTypes.


ostrolucky (changes requested):
This won't work with Symfony 4.4 that we need to support.

gennadigennadigennadi:
What about adding a check if the Pass is available then only add it if it is so?

ostrolucky:
Well of course. After that, you also need to revert composer.json changes

greg0ire (approved):

DavidBadura:
I am wondering if the automatic type registration can also be deactivated / prevented.

The background is that we use the Ramsey UUID Lib in many projects and want to stay with it. The doctrine type "uuid" is already defined by Ramsey Doctrine Extension. So you should have a conflict here, right?

PR #1211, opened 2020-09-19
#1210: add different statements metric to the Web Profiler Toolbar

### LINKED ISSUE #1210: Add "different statements" field to Symfony Web Debug Toolbar
Copy of: https://github.com/symfony/symfony/issues/38158

**Description**  
The Symfony Web Debug Toolbar is a powerful tool. Whenever a dev sees a colored section it starts to fear what it has potentially broken.

The Doctrine section is very useful as it shows the number of queries and the query time. If something goes wrong it became red/yellow and the dev is ready to detect what/where is the problem.
This section could be better if it also takes into account the __different statements__ count (that is already shown in the dedicated Symfony profiler page).

This information is useful to detect repeated queries that potentially impact performances. It could be also a detector for N+1 queries created due to a Doctrine bug.

**Example**  
The "different statements" information could be placed into the div that appears on mouse hover (as in the image below)
proposal

To have a fast detection of repeated queries, the Doctrine section of the toolbar could be colored with orange/yellow (the alert color) when the count of "different statements" is less than the count of "database queries".

lomby92:
Resolve #1210

### DIFF
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -46,6 +46,10 @@
                 <span class="sf-toolbar-status {{ collector.querycount > 50 ? 'sf-toolbar-status-yellow' : '' }}">{{ collector.querycount }}</span>
             </div>
             <div class="sf-toolbar-info-piece">
+                <b>Different statements</b>
+                <span class="sf-toolbar-status">{{ collector.groupedQueryCount }}</span>
+            </div>
+            <div class="sf-toolbar-info-piece">
                 <b>Query time</b>
                 <span>{{ '%0.2f'|format(collector.time * 1000) }} ms</span>
             </div>


ostrolucky (approved):

greg0ire:
#1216 has been merged, let's restart the build.

ostrolucky:
you would have to rebase

greg0ire:
I don't understand why but it looks like you're right… I mean the new build should be run on the result of the merge of this commit and the current master, shouldn't it?

ostrolucky:
nope

greg0ire:
Weird… I will rebase.

greg0ire:
Actually, it's not a normal checkout, the build is run on ae1a88587befd8880bc5cbdf19b7f12389244b85

greg0ire:
Oh but the issue is not the same as before!

greg0ire:
https://github.com/doctrine/DoctrineBundle/pull/1216 was not enough to fix the build, I will work on that.

greg0ire:
Please associate your email address with your Github account, or change the
email in your commits to an address already associated with it. If you do not
want to expose your personal email address, you may use
`lomby92@users.noreply.github.com`, that way we can still reach you through Github.
Additionally, you will get credit for this contribution on your Github profile.

greg0ire (approved):

greg0ire:
Thanks @lomby92 !

lomby92:
Thank you too!

PR #1213, opened 2020-09-23
Add config parameter for read-only SQLite databases

Ma27:
See https://github.com/doctrine/dbal/pull/4292

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -201,6 +201,7 @@ class Configuration implements ConfigurationInterface
                 ->scalarNode('application_name')->end()
                 ->scalarNode('charset')->end()
                 ->scalarNode('path')->end()
+                ->booleanNode('read_only')->info('Whether to open a sqlite database in read-only mode')->end()
                 ->booleanNode('memory')->end()
                 ->scalarNode('unix_socket')->info('The unix socket to use for MySQL')->end()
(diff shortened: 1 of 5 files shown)


DependencyInjection/Configuration.php
```diff
@@ -201,6 +201,7 @@ private function configureDbalDriverNode(ArrayNodeDefinition $node) : void
                 ->scalarNode('application_name')->end()
                 ->scalarNode('charset')->end()
                 ->scalarNode('path')->end()
+                ->booleanNode('read_only')->end()
```
ostrolucky, line 204: should define info explaining this, similarly as is done for `persistent` option

ostrolucky (changes requested):
Requires documentation, xsd updates

Ma27:
@ostrolucky thanks, fixed :)

The build failure appears unrelated as it's also broken on `master`.

Ma27:
Just pushed an update: the DBAL maintainers preferred to not have this kind of logic in their library however I think that it's still useful to have this setting accessible and exposed in a config layer like this bundle. For further references to the original discussion, please take a look at the commit message :)

ostrolucky:
Sorry, I agree with @morozov. This adds just yet another abstraction level to real driver option that doesn't work for any other drivers. Rather confusing if users can just use existing constant for a driver option for same thing. Also, we generally do not add driver options here that do not exist in DBAL, I would recommend to wait until feature is accepted in DBAL first before working on its integration here. Thanks for proposing though!

Issue #1214, opened 2020-09-25
doctrine:database:create and doctrine:database:drop not working with Slaves configuration and Doctrine/DBAL 2.11


Dodenis:
Since Doctrine/dbal 2.11 master/slave connection options have been replaced by primary/replica.

The commands `doctrine:database:create` and `doctrine:database:drop` displays the following error:
`Connection does not contain a 'path' or 'dbname' parameter and cannot be dropped.`

ostrolucky:
Should be solved in next release

PR #1215, opened 2020-09-25
Fix master-slaves deprecation from dbal 2.11

Dodenis:
Since dbal 2.11 MasterSlaveConnection is deprecated and the `doctrine:database:create` and` doctrine:database:drop` commands no longer work with the "slaves" option. https://github.com/doctrine/dbal/pull/4054

Issue : #1214

### DIFF
--- a/Command/CreateDatabaseDoctrineCommand.php
+++ b/Command/CreateDatabaseDoctrineCommand.php
@@ -57,6 +57,11 @@ EOT
             $params = $params['master'];
         }
 
+        // Since doctrine/dbal 2.11 master has been replaced by primary
+        if (isset($params['primary'])) {
+            $params = $params['primary'];
+        }
+
         // Cannot inject `shard` option in parent::getDoctrineConnection
         // cause it will try to connect to a non-existing database
         if (isset($params['shards'])) {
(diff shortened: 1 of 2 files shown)


ostrolucky:
You should rather report a BC break in DBAL repo.

stof:
I'm quite sure this PR makes a BC break in the bundle config by removing support for `slaves` keys as soon as the new version of DBAL is installed.

What we need is a proper migration to the new names, where the bundle config also uses them and deprecates `master/slave`. Then, the bundle would use the right name on the DBAL side based on the DBAL version (whatever the user-facing name is).

My proposal is to fix only the commands in that PR (to fix compat with 2.11), and perform the master/slave to primary/replicas in the config in a separate PR.

@ostrolucky no. The 2 commands are using `Connection::getParams` which is tagged as `@internal` so that cannot be considered as a regression in DBAL.

Dodenis:
@stof I'm agree with you.

I modify this PR to fix only the commands

kralos:
> You should rather report a BC break in DBAL repo.

See https://github.com/doctrine/dbal/issues/4295

stof (approved):

greg0ire (approved):

ostrolucky:
Let's merge after https://github.com/doctrine/dbal/pull/4308 is merged. If we merge this first, there will no longer be a pressure to fix a BC break in DBAL. This is not a pressing issue, people can just hold off from upgrading dbal.

PR #1220, opened 2020-09-26
Make PHP 8 build green

greg0ire:


Tests/RegistryTest.php
```diff
@@ -140,7 +138,7 @@ public function testReset() : void
         $noProxyManager->expects($this->once())
             ->method('clear');
 
-        $proxyManager = $this->getMockBuilder([LazyLoadingInterface::class, EntityManagerInterface::class])->getMock();
+        $proxyManager = $this->createMock(LazyLoadingEntityManagerInterface::class);
```
SenseException, line 141: Why did you create `LazyLoadingEntityManagerInterface`? Isn't the previous multi-interface mock creation not working anymore in PHP 8?

greg0ire: It does not have to do with PHP 8, but with PHPUnit 9, which no longer supports it because it's a code smell IIRC.

greg0ire: See https://github.com/sebastianbergmann/phpunit/issues/3955

SenseException: Guess I have to remove another one of my old code snippets :smile:

SenseException (approved):

Issue #1179, opened 2020-06-09
Console commands: compatiblity with upcoming doctrine/dbal 3.0


dmaicher:
Due to some deprecations and the new `ConnectionProvider` interface we need to fix a few things to be compatible with `doctrine/dbal` 3.0.

See https://github.com/doctrine/dbal/pull/3956, https://github.com/doctrine/dbal/pull/4059

Also related was this fix: https://github.com/doctrine/DoctrineBundle/pull/1170

Some thoughts: 
- we will need an adapter from `Symfony\Bridge\Doctrine\ManagerRegistry` to `Doctrine\DBAL\Tools\Console\ConnectionProvider` and inject that into the 2 commands `RunSqlDoctrineCommand` and `ImportDoctrineCommand` (can we use the dbal commands directly and deprecate the proxy commands?)
- remove usages of `Doctrine\DBAL\Tools\Console\Helper\ConnectionHelper`

ostrolucky:
Yes, I concluded same. But like I mentioned in #1170 I don't want to change DoctrineBundle only to find dbal changed their API before release.

dmaicher:
@ostrolucky I think now we could work on this?

Since release [2.11](https://github.com/doctrine/dbal/releases/tag/2.11.0) I now have some deprecations related to this on some projects:

```
Other deprecation notices (1)

  1x: Not passing a connection provider as the first constructor argument is deprecated
    1x in Application::run from Symfony\Bundle\FrameworkBundle\Console

**Comment by ostrolucky** [comment]
Yep

**Comment by Adirelle** [comment]
> Since release [2.11](https://github.com/doctrine/dbal/releases/tag/2.11.0) I now have some deprecations related to this

It also triggers an error when using the `--connection` parameter:

```
$ bin/console doctrine:query:sql --connection=someconnection 'CREATE SCHEMA IF NOT EXISTS someschema'

In RunSqlCommand.php line 108:

   Specifying a connection is only supported when a ConnectionProvider is used
```

With these versions:
* doctrine/dbal 2.11.1
* doctrine/doctrine-bundle 2.1.2

dmaicher:
@Adirelle indeed. I proposed a fix for that issue here: https://github.com/doctrine/DoctrineBundle/pull/1226

ostrolucky:
Fixed in master

PR #1226, opened 2020-10-07
fix issue with doctrine/dbal 2.11 and connection option

dmaicher:
See https://github.com/doctrine/DoctrineBundle/issues/1179#issuecomment-704930637

### DIFF
--- a/Command/Proxy/RunSqlDoctrineCommand.php
+++ b/Command/Proxy/RunSqlDoctrineCommand.php
@@ -43,6 +43,10 @@ EOT
     {
         DoctrineCommandHelper::setApplicationConnection($this->getApplication(), $input->getOption('connection'));
 
+        // compatibility with doctrine/dbal 2.11+
+        // where this option is also present and unsupported before we are not switching to use a ConnectionProvider
+        $input->setOption('connection', null);
+
         return parent::execute($input, $output);
     }
 }


stof:
The proper fix would be to wire the ConnectionProvider instead.

dmaicher:
> The proper fix would be to wire the ConnectionProvider instead.

This is more a quick fix to make sure it works with dbal 2.11 until we have a proper fix including `ConnectionProvider`.

If you think we should not release this fix first then I can also close here. I don't mind :blush:

ostrolucky:
Why are we again fixing doctrinebundle instead of fixing DBAL? DBAL broke this in in minor version, issue is on their side.

dmaicher:
@ostrolucky not sure how to fix this on DBAL's side :confused: 

This fix is similar to your fix here https://github.com/doctrine/DoctrineBundle/commit/86d2469d6be06d55ad7b9e2f076f6942476f2e87

stof:
@ostrolucky the issue here is that we were adding new options in a child class. This broke when upstreaming the option.
But I don't think DBAL can be considered responsible for that. If you want to allow child classes to add new command options in a BC way, you have to forbid DBAL from ever changing the available options (the same would be true regarding adding methods in a child class, and that's why Symfony excludes such possibility from safe changes in downstream code).

stof:
And as DBAL 2.11 is released now, we should not do a weird hack to keep the DoctrineBundle feature working. We should use the new upstream feature instead.

ostrolucky:
Yeah, after some more investigation I don't think we should hold DBAL accountable for this.

> If you want to allow child classes to add new command options in a BC way, you have to forbid DBAL from ever changing the available options

You are talking about what happened previously, in linked commit. What happened now is that DBAL added exception throw when connection is specified. But I don't personally stand behind opinion that adding exception throws is a BC break in general either.


> And as DBAL 2.11 is released now, we should not do a weird hack to keep the DoctrineBundle feature working. We should use the new upstream feature instead.

We will do that, but in 2.2. This is non trivial change which will also include deprecating our commands.

ostrolucky (approved):

PR #1234, opened 2020-10-22
Allow twig ^3.0

schobner:
Hallo,

the DoctrineBundle prevent update twig 2 to twig 3. With this tiny change, it will be works. I have testet it local. All tests are green.

Greatings
Matthias Schobner

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,7 +49,7 @@
         "symfony/validator": "^3.4.30|^4.3.3|^5.0",
         "symfony/web-profiler-bundle": "^3.4.30|^4.3.3|^5.0",
         "symfony/yaml": "^3.4.30|^4.3.3|^5.0",
-        "twig/twig": "^1.34|^2.12"
+        "twig/twig": "^1.34|^2.12|^3.0"
     },
     "config": {
         "sort-packages": true


stof:
This change is good to actually run tests against Twig 3. But note that `require-dev` is a root-only field, so it does **not** prevent updating Twig in your projects.

stof (approved):

schobner:
The test fails because of symfony/http-kernel update at 5.2. 
I will add a pull request at DoctrineBundle:2.0. to solve this backwards.

`    "conflict": {
        ....
        "symfony/http-kernel": ">=5.2"`

Are you fine with this fixing idea?

schobner:
The parameter `kernel.runtime_environment` was added in 5.2
[https://github.com/symfony/http-kernel/blob/5.x/CHANGELOG.md](https://github.com/symfony/http-kernel/blob/5.x/CHANGELOG.md)

stof:
Conflicting with the next version of Symfony is not a fix. The testsuite should be fixed instead.

schobner:
Ok, i will show if i can fix this

schobner:
Can you please tag this branch/commit with 2.1.3

ostrolucky:
Will do that later together with 2.2.0

stof:
@schobner as I said above, this PR has **strictly no impact** on your ability to use Twig 3 in your own project

schobner:
> @schobner as I said above, this PR has **strictly no impact** on your ability to use Twig 3 in your own project

Thank you for the explanation, i have understand it! Next tim i will know it.

Issue #1235, opened 2020-10-22
Use PrimaryReadReplicaConnection instead of MasterSlaveConnection


zazoomauro:
In order to avoid deprecation message:

`DEPRECATION: Doctrine\DBAL\Connections\MasterSlaveConnection is deprecated since doctrine/dbal 2.11 and will be removed in 3.0, use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection instead. /var/www/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php:98
`

Seldaek:
Related to this, primary/replica parameters are impossible to set, so even if one sets `                wrapper_class: 'Doctrine\DBAL\Connections\PrimaryReadReplicaConnection'` in the doctrine bundle config, it fixes the deprecation warning about the class, but then PrimaryReadReplicaConnection fails hard with `InvalidArgumentException("primary or replica configuration missing")`

ostrolucky:
See https://github.com/doctrine/DoctrineBundle/pull/1237. Does that solve issue for all of you?

ostrolucky:
merged in master/2.2

PR #1237, opened 2020-10-25
#1235 Resolve DBAL 2.11 master/slave deprecations

ostrolucky:
One problem of this is is that I could not follow DBALs slaves -> replica exactly. Users have `replicas` in YAML instead. This appears to be a limitation of symfony config component, see https://github.com/symfony/symfony/pull/14053#issuecomment-86105385.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -93,7 +93,11 @@ class Configuration implements ConfigurationInterface
                             ->end()
                             ->children()
                                 ->scalarNode('class')->isRequired()->end()
-                                ->booleanNode('commented')->setDeprecated(...$this->getCommentedParamDeprecationMsg())->end()
+                                ->booleanNode('commented')
+                                    ->setDeprecated(
+                                        ...$this->getDeprecationMsg('The doctrine-bundle type commenting features were removed; the corresponding config parameter was deprecated in 2.0 and will be dropped in 3.0.')
+                                    )
+                                ->end()
                             ->end()
                         ->end()
                     ->end()
@@ -123,6 +127,7 @@ class Configuration implements ConfigurationInterface
             ->fixXmlConfig('option')
             ->fixXmlConfig('mapping_type')
             ->fixXmlConfig('slave')
+            ->fixXmlConfig('replica')
             ->fixXmlConfig('shard')
             ->fixXmlConfig('default_table_option')
             ->children()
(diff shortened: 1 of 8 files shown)


DependencyInjection/Configuration.php
```diff
@@ -146,7 +151,12 @@ private function getDbalConnectionsNode(): ArrayNodeDefinition
                 ->scalarNode('shard_manager_class')->end()
                 ->scalarNode('shard_choser')->end()
                 ->scalarNode('shard_choser_service')->end()
-                ->booleanNode('keep_slave')->end()
+                ->booleanNode('keep_slave')
+                    ->setDeprecated(
+                        ...$this->getDeprecationMsg('"keep_slave" option is deprecated since doctrine-bundle 2.2. Use keep_replica option instead.')
```
greg0ire, line 156:
```suggestion
                        ...$this->getDeprecationMsg('The "keep_slave" configuration key is deprecated since doctrine-bundle 2.2. Use the "keep_replica" configuration key instead.')
```

DependencyInjection/Configuration.php
```diff
@@ -162,13 +172,25 @@ private function getDbalConnectionsNode(): ArrayNodeDefinition
                 ->end()
             ->end();
 
+        // dbal < 2.11
         $slaveNode = $connectionNode
             ->children()
                 ->arrayNode('slaves')
+                    ->setDeprecated(
+                        ...$this->getDeprecationMsg('"slaves" config key will be renamed to "replicas" in doctrine-bundle 3.0. "slaves" is deprecated since doctrine-bundle 2.2.')
```
greg0ire, line 180:
```suggestion
                        ...$this->getDeprecationMsg('The "slaves" configuration key will be renamed to "replicas" in doctrine-bundle 3.0. "slaves" is deprecated since doctrine-bundle 2.2.')
```

Webonaute:
am not sure on which branch 2.1.2 is but it actually does not have getDeprecationMsg method which seem to have in this PR.

Webonaute:
it also look to have issue in DriverManager. where replica without s is expected when we configure replicas in Doctrine bundle? did I miss something?

ostrolucky:
> am not sure on which branch 2.1.2 is but it actually does not have getDeprecationMsg method which seem to have in this PR.

Why are you looking 2.1 branch? This PR points to master, master has it. Target here is unreleased 2.2.

> it also look to have issue in DriverManager. where replica without s is expected when we configure replicas in Doctrine bundle? did I miss something?

Indeed, that was an oversight. I've addressed that now. 👍

PR #1240, opened 2020-10-25
Fix #1230 - Console commands do not respect driverOptions when master/slave configuration is used

### LINKED ISSUE #1230: Console Commands: Possible bug - Using MasterSlaveConnection does not respect driverOptions
We're using `MasterSlaveConnection` to allow us to make use of read replicas in production.

I'm trying to set up an SSL certificate with the connections as we'd like to enforce TLS on our connections. I'm using the below config:

```yaml
doctrine:
  dbal:
    driver: pdo_mysql
    host: "%env(DATABASE_WRITER_HOST)%"
    port: "%env(DATABASE_WRITER_PORT)%"
    dbname: "%env(DATABASE_NAME)%"
    user: "%env(DATABASE_USER)%"
    password: "%env(DATABASE_PASSWORD)%"
    slaves:
      slave1:
        host: "%env(DATABASE_READER_HOST)%"
        port: "%env(DATABASE_READER_PORT)%"
        dbname: "%env(DATABASE_NAME)%"
        user: "%env(DATABASE_USER)%"
        password: "%env(DATABASE_PASSWORD)%"
    options:
      !php/const:PDO::MYSQL_ATTR_SSL_CA: "%kernel.project_dir%/certs/ca.crt"
```

Symfony responds to this config fine, and a TLS connection seems to be established with the MySQL server. However, when I try to run a CLI command:

```sh
bin/console doctrine:database:drop --force --env test

In AbstractMySQLDriver.php line 106:
                                                                                                                       
  An exception occurred in driver: SQLSTATE[HY000] [3159] Connections using insecure transport are prohibited while -  
  -require_secure_transport=ON.  
```

When I inspect what's going on in the command:

```php
$params = $connection->getParams();
if (isset($params['master'])) {
    $params = $params['master'];
}
// Yields
/*
array(6) {
  ["host"]=>
  string(9) "op_api_db"
  ["port"]=>
  string(4) "3306"
  ["dbname"]=>
  string(14) "local"
  ["user"]=>
  string(4) "root"
  ["password"]=>
  string(12) "insecurepass"
  ["driver"]=>
  string(9) "pdo_mysql"
}
*/
```

The `driverOptions` key is present in the initial `$params` array, but is then lost by this condition. If I replace with:

```php
$params = $connection->getParams();
$driverOptions = $params['driverOptions'] ?? [];
if (isset($params['master'])) {
    $params = $params['master'];
}
$params['driverOptions'] = $driverOptions;
```

It works. Is this a bug, or have I misconfigured?

ostrolucky:
@stof could you please check this, you were the one adding driverOptions to this blacklist, so there is a tiny chance you would know if that was on purpose

### DIFF
--- a/Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
+++ b/Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
@@ -126,6 +126,7 @@ abstract class AbstractDoctrineExtensionTest extends TestCase
                 'dbname' => 'mysql_db',
                 'host' => 'localhost',
                 'unix_socket' => '/path/to/mysqld.sock',
+                'driverOptions' => [],
             ],
             $param['master']
         );
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -261,7 +261,6 @@ class DoctrineExtension extends AbstractDoctrineExtension
         if (! empty($options['slaves'])) {
             $nonRewrittenKeys = [
                 'driver' => true,
-                'driverOptions' => true,
                 'driverClass' => true,
                 'wrapperClass' => true,
                 'keepSlave' => true,


Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -126,6 +126,7 @@ public function testDbalLoadSingleMasterSlaveConnection()
                 'dbname' => 'mysql_db',
                 'host' => 'localhost',
                 'unix_socket' => '/path/to/mysqld.sock',
+                'driverOptions' => [],
```
greg0ire, line 129: Should the file on line 114 be modified to add options in both the master and the slave, to make things more interesting? Without your change, what happens? Are the master's options copied to the slave?

ostrolucky: There is actually no such option as "master" configuration key/option. What's defined one level above slaves is what's master. And yes that's how this patch works. It copies everything from master to slaves except the things in the list.

greg0ire: Indeed, I read the issue backwards. What if there is already an option at the slave level? Will it be overwritten? Should the configuration at https://github.com/doctrine/DoctrineBundle/blob/6f3f4d9e27327cff20c584b0818b78c406325d72/Tests/DependencyInjection/Fixtures/config/yml/dbal_service_single_master_slave_connection.yml#L2-L15 be modified to test what happens when there are options in both levels? I'd expect some kind of merge to happen, but I'm not sure which level should get the priority. Maybe the slave one, because otherwise, there would be a BC-break?

ostrolucky: I don't think that's necessary. I didn't change merging logic. Whatever slave specifies is kept for that slave.

greg0ire (approved):

PR #1242, opened 2020-10-27
Fix profiler explain for Oracle DB

derflocki:


greg0ire:
`vendor/bin/phpcbf` to fix the cs issue

ostrolucky:
also, this should ideally point to 1.12.x

derflocki:
> `vendor/bin/phpcbf` to fix the cs issue
> also, this should ideally point to 1.12.x

Thanks for the feedback. I will update the PR accordingly.

Controller/ProfilerController.php
```diff
@@ -117,4 +119,15 @@ private function explainOtherPlatform(Connection $connection, $query)
...
     }
+
+    /**
+     * @param mixed[] $query
+     */
+    private function explainOraclePlatform(Connection $connection, array $query)
+    {
+        $connection->executeQuery('EXPLAIN PLAN FOR ' . $query['sql']);
+
+        return $connection->executeQuery('SELECT * FROM TABLE(DBMS_XPLAN.DISPLAY)')
```
ostrolucky, line 130: - [Manual](https://docs.oracle.com/cd/B19306_01/server.102/b14211/ex_plan.htm#i26038) uses `DBMS_XPLAN.DISPLAY()` with brackets, not `DBMS_XPLAN.DISPLAY`. Why did you leave out brackets?
- Manual also uses `SELECT PLAN_TABLE_OUTPUT`, you do `SELECT *`, why?

Controller/ProfilerController.php
```diff
@@ -117,4 +119,15 @@ private function explainOtherPlatform(Connection $connection, $query)
         return $connection->executeQuery('EXPLAIN ' . $query['sql'], $params, $query['types'])
             ->fetchAll(PDO::FETCH_ASSOC);
     }
+
+    /**
+     * @param mixed[] $query
+     */
+    private function explainOraclePlatform(Connection $connection, array $query)
+    {
+        $connection->executeQuery('EXPLAIN PLAN FOR ' . $query['sql']);
```
ostrolucky, line 128: Why isn't this using `SET statement_id = ` like oracle manual specifies? Doesn't your current implementation pull out records for every statement recorded?

ostrolucky (changes requested):
Thanks for this, but as I don't use Oracle, some stuff need to be clarified

Controller/ProfilerController.php
derflocki: I have been following [DBMS_XPLAN](https://docs.oracle.com/database/121/ARPLS/d_xplan.htm#ARPLS70125), where `SELECT * FROM table(DBMS_XPLAN.DISPLAY)` is used. Parentheses can be omitted in Oracle if no parameter is given. I have no link to an official oracle documentation, but there is [some info about this on stackoverflow](https://stackoverflow.com/a/56511361). Also selecting all columns seems like the failsafe thing todo.

Controller/ProfilerController.php
derflocki: I did not know about the `SET statement_id = ` variant, but `DBMS_XPLAN.DISPLAY` without parameters will return the plan for the most recent explained query.

derflocki: I have investigated further: The results of `EXPLAIN PLAN` are inserted into a table. Thus they are properly isolated on the transaction level.

derflocki: [From Ask Tom](https://asktom.oracle.com/pls/apex/f?p=100:11:0::::P11_QUESTION_ID:9535917800346105419) 
> Note that plan_table is a synonym to sys.plan_table$ (unless you've created your own...). This is a session-level temporary table (on commit preserve rows). So your data is wiped from this when you close the session regardless of whether you commit or rollback.

So another sessions can't interfere with "our" explain plan.

Controller/ProfilerController.php
ostrolucky: So according that SO post it works without brackets by accident and it's not consistent across different functions. I would sleep better if we used documented version

ostrolucky (approved):

Controller/ProfilerController.php
derflocki: I have updated the PR to use `DBMS_XPLAN.DISPLAY()`, and fixed the missing `use 
 Doctrine\DBAL\Platforms\OraclePlatform`. To be fair to Oracle, it's inconsistent across different **types** of functions if parentheses are optional, mandatory or have to be omitted. It's not "by accident", or undocumented. See the docs for [User-Defined Functions](https://docs.oracle.com/cd/E11882_01/server.112/e41084/functions256.htm#SQLRF06181), [Single-Row Function SYSDATE](https://docs.oracle.com/cd/E11882_01/server.112/e41084/functions191.htm#SQLRF06124) or [Analytic Function ROW_NUMBER](https://docs.oracle.com/cd/E11882_01/server.112/e41084/functions156.htm#SQLRF06100)

PR #1248, opened 2020-11-09
Fix incorrect assertions

### LINKED ISSUE #1247: Using DoctrineDataCollector with EntityManagerDecorator fails
version: 2.2.x 
symfony environment: DEV or TEST with profiler enabled

This assert fails when I use custom EntityManagerDecorator:

https://github.com/doctrine/DoctrineBundle/blob/2aacd8380030fb2497a896fe33aaa5b1052feccc/DataCollector/DoctrineDataCollector.php#L67

After changing to:

`assert($em instanceof EntityManagerInterface);`

everything seems to be ok.

pkly:
Fixes #1247

### DIFF
--- a/DataCollector/DoctrineDataCollector.php
+++ b/DataCollector/DoctrineDataCollector.php
@@ -6,7 +6,7 @@ use Doctrine\ORM\Cache\CacheConfiguration;
 use Doctrine\ORM\Cache\Logging\CacheLoggerChain;
 use Doctrine\ORM\Cache\Logging\StatisticsCacheLogger;
 use Doctrine\ORM\Configuration;
-use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\ORM\Tools\SchemaValidator;
 use Doctrine\Persistence\ManagerRegistry;
@@ -64,7 +64,7 @@ class DoctrineDataCollector extends BaseCollector
         ];
 
         foreach ($this->registry->getManagers() as $name => $em) {
-            assert($em instanceof EntityManager);
+            assert($em instanceof EntityManagerInterface);
             if ($this->shouldValidateSchema) {
                 $entities[$name] = [];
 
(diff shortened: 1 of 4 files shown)


DoctrineBundle.php
```diff
@@ -78,7 +77,7 @@ public function boot()
                 assert($registry instanceof Registry);
 
                 foreach ($registry->getManagers() as $em) {
-                    assert($em instanceof EntityManager);
+                    assert($em instanceof EntityManagerInterface);
```
dmaicher, line 80: I don't see any `use` statement for this one?

pkly: Yep, my bad. Storm decided not to remind me.

SenseException (approved):

DataCollector/DoctrineDataCollector.php
```diff
@@ -101,7 +101,7 @@ public function collect(Request $request, Response $response, Throwable $excepti
             $cacheConfiguration = $emConfig->getSecondLevelCacheConfiguration();
             assert($cacheConfiguration instanceof CacheConfiguration);
             $cacheLoggerChain = $cacheConfiguration->getCacheLogger();
-            assert($cacheLoggerChain instanceof CacheLoggerChain);
+            assert($cacheLoggerChain instanceof CacheLoggerChain || $cacheConfiguration === null);
```
stof, line 104: Instead of the assert, the `if (! $cacheLoggerChain)` below could be changed to `! $cacheLoggerChain instanceof CacheLoggerChain)` (which would also cover a case where a custom CacheLogger is passed without using a chain)

ostrolucky: I don't like that because it would silence the error in case of unexpected instance.

stof: The thing is, the data collector is not in charge of configuring the cache logger. If the debugging tool breaks entirely because you customized the config in a way that they did not expect, that's quite bad. It seems better to skip the SLC stats in such case.

ostrolucky: I'm more proponent of [offensive programming](https://en.wikipedia.org/wiki/Offensive_programming) and there it's better to crash than display incorrect information.

PR #1254, opened 2020-11-10
fix service ids for DoctrineMetadataCacheWarmer

dmaicher:
See https://github.com/doctrine/DoctrineBundle/pull/1196#discussion_r520782503

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -942,7 +942,7 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $decoratedMetadataCacheServiceId = (string) $container->getAlias($metadataCacheAlias);
         $phpArrayCacheDecoratorServiceId = $decoratedMetadataCacheServiceId . '.php_array';
         $phpArrayFile                    = '%kernel.cache_dir%' . sprintf('/doctrine/orm/%s_metadata.php', $entityManagerName);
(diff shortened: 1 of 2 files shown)


DependencyInjection/DoctrineExtension.php
```diff
@@ -942,8 +942,9 @@ private function registerMetadataPhpArrayCaching(string $entityManagerName, Cont
         $decoratedMetadataCacheServiceId = (string) $container->getAlias($metadataCacheAlias);
         $phpArrayCacheDecoratorServiceId = $decoratedMetadataCacheServiceId . '.php_array';
         $phpArrayFile                    = '%kernel.cache_dir%' . sprintf('/doctrine/orm/%s_metadata.php', $entityManagerName);
+        $cacheWarmerServiceId            = sprintf('doctrine.orm.metadata_cache_warmer.%s', $entityManagerName);
```
ostrolucky, line 945: I might be nitpicking, but can we please use similar approach here as on line 941? Meaning, this would be 
```suggestion
        $cacheWarmerServiceId            = $this->getObjectManagerElementName($entityManagerName . '_metadata_cache_warmer');
```

This would follow same naming convention

ostrolucky:
Can we update tests?

dmaicher:
See https://github.com/doctrine/DoctrineBundle/pull/1255

Issue #1257, opened 2020-11-12
Console commands prefer 'master' over ' primary' when using Slaves


prengerjarno:
Following #1215 and doctrine/dbal#4295, the console commands now support the new 'primary' config while the MasterSlaveConnection also has BC support for 'master'. This now leads to the console commands using the 'master' config which is missing the driver option.

I think it can simple prefer the 'primary' config.

ostrolucky:
This looks like same issue as here https://github.com/doctrine/DoctrineBundle/issues/1252 

What version are you using?

prengerjarno:
@ostrolucky I thought that at first as well, but after the fix in #1252 i still am missing the 'driver' param when using the create and drop commands, which results in not being able to create the connection. With the PrimaryReadReplicaConnection this is explicitly added to the primary/replica params.

I am using 1.12.12

PR #1258, opened 2020-11-12
fix: Prefer primary config over master.

prengerjarno:
Make sure the primary config gets used over the master so it uses the added options when PrimaryReadReplicaConnection is used. #1257

### DIFF
--- a/Command/CreateDatabaseDoctrineCommand.php
+++ b/Command/CreateDatabaseDoctrineCommand.php
@@ -59,14 +59,14 @@ EOT
             $driverOptions = $params['driverOptions'];
         }
 
-        if (isset($params['master'])) {
-            $params                  = $params['master'];
-            $params['driverOptions'] = $driverOptions;
-        }
-
         // Since doctrine/dbal 2.11 master has been replaced by primary
         if (isset($params['primary'])) {
             $params                  = $params['primary'];
+            $params['driverOptions'] = $driverOptions;
+        }
+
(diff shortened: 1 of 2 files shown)


ostrolucky:
New primary config was added in 2.2, so this is the branch this should point to

prengerjarno:
@ostrolucky I'm sorry, its not the config, its the params. Both 'master' and 'primary' params are present to because of the BC support in doctrine/dbal's `MasterSlaveConnection`. In this case it should use the 'primary' params, but is using the 'master' which is missing the 'driver' option (because it only gets added to 'primary' in `PrimaryReadReplicaConnection::__construct`).

stof (approved):

ostrolucky:
If DBAL is adding `driver` to primary only and not to master, that's a bug in DBAL's BC layer and you need to report issue there.

stof:
@ostrolucky `->getParams` is an API marked as internal.

ostrolucky:
Up to dbal 2.11 initiating connection with `master` would result in `getParams` giving you back array including `master['driver']` and then it was changed because in 2.11 it was marked internal and that's ok? That's not how BC works. Method cannot be broken in minor release like that and claim it's ok because it's internal now. It can be broken in 3.x, not in 2.x.

ostrolucky (approved):

greg0ire:
I managed to trigger a build.

greg0ire:
Thanks @prengerjarno

Issue #1021, opened 2019-09-18
Allow wiring specific entity managers and repositories into services


alcaeus:
When injecting an entity manager into a service, auto wiring the service by using the `EntityManagerInterface` type will always wire the default entity manager. This makes it complicated to inject a custom entity manager for users that have more than one manager configured.

To select a correct entity manager, users would have to create a specific entity manager for each entity that decorates the entity manager service created. This could be done automatically in a compiler pass by using a service tag which specifies which entity manager to use.

The syntax would be simple:
```xml
<tag name="doctrine.entity_manager" entity="App\Entity\MyEntity" />
```

Same goes for repositories. Currently, users have to create an empty repository class if they want to use service repositories that they can inject into other services. To avoid this boilerplate code, we can leverage the same technique:

```xml
<tag name="doctrine.repository" entity="App\Entity\MyEntity" />
```

This is similar to what the MonologBundle does when configuring a channel using the `monolog.logger` tag.

stof:
> To select a correct entity manager, users would have to create a specific entity manager for each entity that decorates the entity manager service created. This could be done automatically in a compiler pass by using a service tag which specifies which entity manager to use.
> 
> The syntax would be simple:
> 
> ```
> <tag name="doctrine.entity_manager" entity="App\Entity\MyEntity" />
> ```

I don't understand what you mean here. You talk about creating specific entity managers, but your issue is about *injecting* managers.

alcaeus:
Okay, maybe I worded the paragraph wrong.

Currently, users would have to create a specific entity manager class that decorates the original entity manager to select the proper one. This however needs a bunch of boilerplate code that shouldn't be necessary.

The proposed tag would be a different solution for the problem, allowing users to apply the tag and thus no longer having to redeclare a specific entity manager to allow autowiring.

ostrolucky:
I'm not a fan of this idea. I really dislike monologbundle way of injecting loggers and I say we should not take it as example. Pretty much same thing can be achieved nowadays with [binding](https://symfony.com/blog/new-in-symfony-4-1-getting-container-parameters-as-a-service)

alcaeus:
Binding works for me, but we can also use more complex service aliases to also use the variable name. This would also help solve a problem we have when people type hint the `ObjectManager` interface but have ORM and ODM installed. We can configure service aliases like this:
```
<service id="Doctrine\Common\Persistence\ManagerRegistry $entityManagerRegistry" alias="doctrine" />

[...]

PR #1259, opened 2020-11-12
Autowire entity managers by argument type and name

HypeMC:
This will allow autowiring entity managers by argument type and name which is useful when you have multiple entity managers:

```yaml
orm:
    entity_managers:
        my_name:
            mappings:
                MyBundle: []
```

```php
use Doctrine\ORM\EntityManagerInterface;

class Foo
{
    public function __construct(EntityManagerInterface $myNameEntityManager) {}
}
```

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -11,6 +11,7 @@ use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepositoryInterface;
 use Doctrine\DBAL\Connections\MasterSlaveConnection;
 use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection;
 use Doctrine\DBAL\Tools\Console\ConnectionProvider;
+use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Proxy\Autoloader;
 use Doctrine\ORM\UnitOfWork;
 use LogicException;
@@ -544,14 +545,20 @@ class DoctrineExtension extends AbstractDoctrineExtension
             $entityManager['connection'] = $this->defaultConnection;
         }
 
+        $entityManagerId = sprintf('doctrine.orm.%s_entity_manager', $entityManager['name']);
+
         $container
-            ->setDefinition(sprintf('doctrine.orm.%s_entity_manager', $entityManager['name']), new ChildDefinition('doctrine.orm.entity_manager.abstract'))
+            ->setDefinition($entityManagerId, new ChildDefinition('doctrine.orm.entity_manager.abstract'))
             ->setPublic(true)
             ->setArguments([
                 new Reference(sprintf('doctrine.dbal.%s_connection', $entityManager['connection'])),
                 new Reference(sprintf('doctrine.orm.%s_configuration', $entityManager['name'])),
             ])
             ->setConfigurator([new Reference($managerConfiguratorName), 'configure']);
+
+        $container
+            ->registerAliasForArgument($entityManagerId, EntityManagerInterface::class, sprintf('%sEntityManager', $entityManager['name']))
+            ->setPublic(false);
 
         $container->setAlias(
             sprintf('doctrine.orm.%s_entity_manager.event_manager', $entityManager['name']),
--- a/Tests/DependencyInjection/DoctrineExtensionTest.php
+++ b/Tests/DependencyInjection/DoctrineExtensionTest.php
@@ -52,7 +52,49 @@ class DoctrineExtensionTest extends TestCase
 
             $alias = $container->getAlias($id);
             $this->assertEquals($target, (string) $alias, sprintf('The autowiring for `%s` should use `%s`.', $id, $target));
-            $this->assertFalse($alias->isPublic(), sprintf('The autowiring alias for `%s` should be private.', $id, $target));
+            $this->assertFalse($alias->isPublic(), sprintf('The autowiring alias for `%s` should be private.', $id));
+        }
+    }
+
+    public function testEntityManagerAutowiringAlias()
+    {
+        if (! interface_exists(EntityManagerInterface::class)) {
+            self::markTestSkipped('This test requires ORM');
+        }
+
+        $container = $this->getContainer([
+            'YamlBundle',
+            'XmlBundle',
+        ]);
+        $extension = new DoctrineExtension();
+
+        $config = BundleConfigurationBuilder::createBuilder()
+            ->addBaseConnection()
+            ->addEntityManager([
+                'entity_managers' => [
+                    'default' => [
+                        'mappings' => ['YamlBundle' => []],
+                    ],
+                    'purchase_logs' => [
+                        'mappings' => ['XmlBundle' => []],
+                    ],
+                ],
+            ])
+            ->build();
+
+        $extension->load([$config], $container);
+
+        $expectedAliases = [
+            EntityManagerInterface::class . ' $defaultEntityManager' => 'doctrine.orm.default_entity_manager',
+            EntityManagerInterface::class . ' $purchaseLogsEntityManager' => 'doctrine.orm.purchase_logs_entity_manager',
+        ];
+
+        foreach ($expectedAliases as $id => $target) {
+            $this->assertTrue($container->hasAlias($id), sprintf('The container should have a `%s` alias for autowiring support.', $id));
+
+            $alias = $container->getAlias($id);
+            $this->assertEquals($target, (string) $alias, sprintf('The autowiring for `%s` should use `%s`.', $id, $target));
+            $this->assertFalse($alias->isPublic(), sprintf('The autowiring alias for `%s` should be private.', $id));
         }
     }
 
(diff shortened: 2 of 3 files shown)


ostrolucky:
This can be solved by one line in user's service declaration, hence was also rejected in past https://github.com/doctrine/DoctrineBundle/issues/1021. This is not worth to do in a bundle, people wouldn't understand how is symfony wiring their managers.

HypeMC:
@ostrolucky I'm not sure what you mean by "people wouldn't understand", if someone is not familiar with how Symfony's container works, then yes, there might be some confusion, but that's true for any feature including basic autowiring by type hinting. Lucky, all this is [documented](https://symfony.com/doc/4.4/service_container/autowiring.html#dealing-with-multiple-implementations-of-the-same-type), so it shouldn't be a problem for people to learn.

Also, this wouldn't be the first bundle to implement such a feature, eg [MonologBundle has already implemented it the same way](https://github.com/symfony/monolog-bundle/pull/315/files).

X-Coder264:
I agree with @HypeMC. At the very least it's confusing/inconsistent from a user perspective that the same thing is enabled in one widely used Symfony Bundle (Monolog bundle) while in Doctrine bundle that's not the case.

ostrolucky:
That's not what was documented. What was documented is having this in your application config

```
App\Util\TransformerInterface $shoutyTransformer: '@App\Util\UppercaseTransformer
```

This part is missing from your PR and is done implicitly instead. People wouldn't know where `$fooEntityManager` wiring comes from. It actually breaks expectation of people who read the documentation. People expect searching for `$fooEntityManager` would yield a result.

And luckily we are completely independent projects, so similarly as we didn't follow MonologBundle with their mess of using tags for injecting logger with specific channel, we are also not following their implicit autowiring strategy. Using one line code as is in the provided documentation is simple enough as it is.

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -54,7 +54,51 @@ public function testAutowiringAlias()
...
+        $extension = new DoctrineExtension();
+
+        $config = BundleConfigurationBuilder::createBuilder()
+            ->addBaseConnection()
+            ->addEntityManager([
+                'entity_managers' => $entityManagers = [
+                    'default' => [
+                        'mappings' => ['YamlBundle' => []],
+                    ],
+                    'logs' => [
```
alcaeus, line 84: Could you please use a snake_cased_name here? I'm curious how this plays out, as I'd expect a name like `purchase_logs` being mapped to `$purchaseLogsEntityManager`, so I think it's worth confirming this.

alcaeus:
> This can be solved by one line in user's service declaration, hence was also rejected in past #1021. This is not worth to do in a bundle, people wouldn't understand how is symfony wiring their managers.

The linked PR was about something different. It suggested adding syntax to XML mappings in the form of tags to achieve something that can already be done with service aliases, hence the rejection. Typed/named aliases are indeed the way to solve this, which is exactly what this PR proposes. The advantage is that it doesn't require users to manually specify this for each entity manager they define, instead doing this automatically when creating the service definitions in the first place.

Whatever people's opinion on autowiring is, we shouldn't forget that this library is designed to provide integration into the Symfony framework, which includes following the core principles and developer experience that Symfony users have come to expect. That's why there's a significant amount of glue code to make ORM play nice with Symfony, like the ServiceRepositoryFactory for example.

I'm reopening this PR, if only for the sake of having an open discussion. I'd also like to invite the @doctrine/team-symfony-integration team to comment on this from a Symfony perspective.

@HypeMC: I'll provide a review on the PR in a separate comment.

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -54,7 +54,51 @@ public function testAutowiringAlias()
...
+    public function testEntityManagerAutowiringAlias()
+    {
+        if (! interface_exists(EntityManagerInterface::class)) {
+            self::markTestSkipped('This test requires ORM');
+        }
+
+        if (! method_exists(ContainerBuilder::class, 'registerAliasForArgument')) {
+            self::markTestSkipped('Test requires Symfony 4.2 or higher');
+        }
+
```
alcaeus, line 70: This is unnecessary as we require Symfony 4.3.3 as lowest supported version.

```suggestion
```

DependencyInjection/DoctrineExtension.php
```diff
@@ -511,15 +512,24 @@ protected function loadOrmEntityManager(array $entityManager, ContainerBuilder $
...
                 new Reference(sprintf('doctrine.orm.%s_configuration', $entityManager['name'])),
             ])
             ->setConfigurator([new Reference($managerConfiguratorName), 'configure']);
 
+        // Symfony 4.2+
+        if (method_exists($container, 'registerAliasForArgument')) {
+            $container
+                ->registerAliasForArgument($entityManagerId, EntityManagerInterface::class, sprintf('%sEntityManager', $entityManager['name']))
+                ->setPublic(false);
+        }
```
alcaeus, line 531: This is unnecessary as we require Symfony 4.3.3 as lowest supported version.

```suggestion
        $container
            ->registerAliasForArgument($entityManagerId, EntityManagerInterface::class, sprintf('%sEntityManager', $entityManager['name']))
            ->setPublic(false);
```

nicolas-grekas:
That'd really make sense to provide these named aliases to me. :+1:

X-Coder264:
>  People wouldn't know where `$fooEntityManager` wiring comes from. It actually breaks expectation of people who read the documentation. People expect searching for `$fooEntityManager` would yield a result.

This is just your assumption which is IMO wrong. Both Monolog channel autowiring and Symfony HTTP client scoped clients autowiring for example are clearly documented (https://symfony.com/doc/4.4/logging/channels_handlers.html#how-to-autowire-logger-channels and https://symfony.com/doc/current/http_client.html#scoping-client ) and people are using that feature all the time (myself included).

> And luckily we are completely independent projects, so similarly as we didn't follow MonologBundle with their mess of using tags for injecting logger with specific channel, we are also not following their implicit autowiring strategy. Using one line code as is in the provided documentation is simple enough as it is.

Of course this project is independent and that is great, but IMO further discussion is needed as just declining a PR based on an assumption that users are dumb and wouldn't know how to use this is not really great.

alcaeus (changes requested):
Thanks for the contribution, @HypeMC!

First of all, the default branch in this repository targets the oldest version supported for bug fixes. New features need to go to the "master" branch. This also changes the minimum Symfony version to 4.3, which makes the checks for `registerAliasForArgument` obsolete. I've added suggestions below to remove these checks.

One additional thing I'd like to see is documentation. I believe we should show examples how to use this feature and how the entity manager name plays into it (e.g. when using snake_case for entity manager names, but camelCase for variable names).

ostrolucky:
> this library is designed to provide integration into the Symfony framework, which includes following the core principles and developer experience that Symfony users have come to expect.

I disagree that this is a core principle and developer experience. But if that's so, then Symfony should just register alias `$bazBar` for any service using `foo.bar.baz` service id convention. Similarly as you want to save work of users having to register aliases for entity managers, similarly it doesn't make sense to suddenly request from every bundle to implement this feature for consistency.

nicolas-grekas:
> Symfony should just register alias $barBaz for any service using foo.bar.baz

Named autowiring aliases are designed to solve the case where multiple services implement the same interface.

> it doesn't make sense to suddenly request from every bundle to implement this feature for consistency.

Every bundle that has collections of user-facing services that implement the same interface would benefit from using named autowiring aliases yes. That'd improve the overall experience of using Symfony.

ostrolucky:
> Named autowiring aliases are designed to solve the case where multiple services implement the same interface.

Then alias it only when there are multiple services implementing same interface, I still fail to see why it cannot be done in generic way but every bundle is supposed to implement it separately. It also doesn't solve the problem for cases when I'm defining multiple services implementing same interface in my own application, not a bundle. It would only improve consistency and hence DX.

nicolas-grekas:
> I still fail to see why it cannot be done in generic way

The reason is naming: the full-service id cannot be inflected into an argument name without local knowledge. Also, there is no way to identify collection of services like the one computed here. Last but not least, because only bundles know which services are "the useful ones", aka the one that the users are supposed to use - vs the others that are declared for internal wiring reasons.

Of course, the future might provide us with solutions to these reasons. But we don't have any right now (and I personally think there is none, but I'd be happy to be proved wrong.)

Tests/DependencyInjection/DoctrineExtensionTest.php
HypeMC: @alcaeus Done, works as you expected it would :smile:

Tests/DependencyInjection/DoctrineExtensionTest.php
HypeMC: Done

DependencyInjection/DoctrineExtension.php
HypeMC: Done

ostrolucky:
Alright, feel free to continue. But review, merge and support will need to be done by someone else.

alcaeus (approved):

Issue #1251, opened 2020-11-10
Can't clear metadata with DoctrineBundle 2.2.0 (prod)


stephanvierkant:
As of DoctrineBundle 2.2, I get an error message when running `bin/console doctrine:cache:clear-metadata` in `prod` environment. Must be related to #1196, but I don't know what's going wrong, since the error message isn't very helpfull:

>  [ERROR] No cache entries were deleted.                                                                                 

The problem is that `\Symfony\Component\Cache\Adapter\PhpArrayAdapter::save` returns false, but I don't understand why.

ostrolucky:
Cannot reproduce :/

```
root@6639ad92298f:/var/www# php -m | grep OPcache
Zend OPcache
Zend OPcache
root@6639ad92298f:/var/www# bin/console doctrine:cache:clear-metadata --env=prod

 // Clearing all Metadata cache entries


 [OK] Successfully deleted cache entries.


root@6639ad92298f:/var/www# php -m | grep OPcache
Zend OPcache
Zend OPcache
root@6639ad92298f:/var/www# composer show doctrine/doctrine-bundle | grep versions
versions : * 2.2.0
root@6639ad92298f:/var/www#
```

ostrolucky:
Alright to reproduce this cache warmup needs to be run too

ostrolucky:
We decided to rollback the change which caused this issue. We released 2.2.1 which should fix your issue.

ossinkine:
@stephanvierkant Since we started using PhpArray cache adapter and it stored in opcache we can't clear metadata cache the same as using for example APCu cache. Therefore `PhpArrayAdapter::save` returns `false`.

ostrolucky:
We shouldn't ship it if it breaks cache clearing though, no matter what internal details prevent it. I think wiping the cache might be important for users.

ossinkine:
@ostrolucky Actually clearing matadata cache already has some restrictions, not every cache adapter can clear it: https://github.com/doctrine/orm/blob/2.7/lib/Doctrine/ORM/Tools/Console/Command/ClearCache/MetadataCommand.php#L84-L90
Metadata on the production environment should not be changed as well as all other code, so IMO clearing metadata cache does not make sence.
Since we deprecated `metadata_cache_driver` configuration I think we also should deprecate this command.
[...]

Issue #1249, opened 2020-11-10
EntityListener not called since 2.2.0 under prod env only


PierrickMartos:
Hello,

Since the release of `doctrine-bundle:2.2.0`, we are facing a strange issue with our EntityListeners **not called only when symfony run under prod env**. Note, It works normally on `dev` and `test` env.

To give you more context, our EntityListener is registered on a MappedSuperClasss and not directly on the final Entity class. But we never had an issue with this before the 2.2.0 release.

Here is how we register our EntityListener (the service is registered inside a vendor managed by us):

```yaml
  SAM\DealFlowBundle\EventListener\DealFlowListener:
    arguments:
      - '%sam_entities%'
    tags:
      - { name: doctrine.orm.entity_listener, event: prePersist, entity: SAM\DealFlowBundle\Entity\DealFlow, method: prePersist, lazy: true, priority: 100 }
```

Here is the `DealFlowListener`:

```php 
<?php

namespace SAM\DealFlowBundle\EventListener;

use Doctrine\ORM\EntityManagerInterface;
use Psr\Log\LoggerInterface;
use SAM\CommonBundle\Utils\LoggerUtils;
use SAM\DealFlowBundle\Entity\DealFlow;
use Doctrine\ORM\Event\LifecycleEventArgs;
use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;

/**
 * class DealFlowListener.
 */
class DealFlowListener
{
    /**
     * @var TokenStorageInterface
     */
    private $tokenStorage;

    /**
     * @var array
     */
    protected $entities;

    /**
     * @var EntityManagerInterface
     */
    protected $em;

    /**
     * @var LoggerInterface
     */
    private $logger;

    /**
     * DealFlowListener constructor.
     *
     * @param array                 $entities
     * @param TokenStorageInterface $tokenStorage
     */
    public function __construct($entities, TokenStorageInterface $tokenStorage, LoggerInterface $logger)
    {
        $this->tokenStorage = $tokenStorage;
        $this->entities = $entities;
        $this->logger = $logger;
    }

    /**
     * @param DealFlow                 $deal
     * @param LifecycleEventArgs $args
     *
     * @throws \Exception
     */
    public function prePersist(DealFlow $deal, LifecycleEventArgs $args)
    {
        $this->logger->info('[DealFlowListener] prePersist called');

        // Some code here removed from this snippet to keep it simple
}
```

[...]

PR #1260, opened 2020-11-13
Fix bugs in PhpArrayAdapter introduction

ossinkine:
Second attempt at doing #1196

Fixes:
- #1251 by deprecating clear metadata cache command
- #1249 by adding priority for cache warmer
- #1254 thank you @dmaicher

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection;
 
+use Doctrine\Bundle\DoctrineBundle\CacheWarmer\DoctrineMetadataCacheWarmer;
 use Doctrine\Bundle\DoctrineBundle\Command\Proxy\ImportDoctrineCommand;
 use Doctrine\Bundle\DoctrineBundle\Dbal\ManagerRegistryAwareConnectionProvider;
 use Doctrine\Bundle\DoctrineBundle\Dbal\RegexSchemaAssetFilter;
@@ -31,6 +32,8 @@ use Symfony\Bridge\Doctrine\SchemaListener\MessengerTransportDoctrineSchemaSubsc
 use Symfony\Bridge\Doctrine\SchemaListener\PdoCacheAdapterDoctrineSchemaSubscriber;
 use Symfony\Bridge\Doctrine\Validator\DoctrineLoader;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
+use Symfony\Component\Cache\Adapter\DoctrineAdapter;
+use Symfony\Component\Cache\Adapter\PhpArrayAdapter;
 use Symfony\Component\Cache\DoctrineProvider;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Alias;
@@ -869,6 +872,12 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $this->loadCacheDriver('metadata_cache', $entityManager['name'], $entityManager['metadata_cache_driver'], $container);
         $this->loadCacheDriver('result_cache', $entityManager['name'], $entityManager['result_cache_driver'], $container);
         $this->loadCacheDriver('query_cache', $entityManager['name'], $entityManager['query_cache_driver'], $container);
+
+        if ($container->getParameter('kernel.debug')) {
+            return;
+        }
+
+        $this->registerMetadataPhpArrayCaching($entityManager['name'], $container);
     }
 
     /**
@@ -981,5 +990,27 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $container->setDefinition($id, $poolDefinition);
 
         return $id;
+    }
+
+    private function registerMetadataPhpArrayCaching(string $entityManagerName, ContainerBuilder $container): void
+    {
+        $metadataCacheAlias              = $this->getObjectManagerElementName($entityManagerName . '_metadata_cache');
+        $decoratedMetadataCacheServiceId = (string) $container->getAlias($metadataCacheAlias);
+        $phpArrayCacheDecoratorServiceId = $decoratedMetadataCacheServiceId . '.php_array';
+        $phpArrayFile                    = '%kernel.cache_dir%' . sprintf('/doctrine/orm/%s_metadata.php', $entityManagerName);
+        $cacheWarmerServiceId            = $this->getObjectManagerElementName($entityManagerName . '_metadata_cache_warmer');
+
+        $container->register($cacheWarmerServiceId, DoctrineMetadataCacheWarmer::class)
+            ->setArguments([new Reference(sprintf('doctrine.orm.%s_entity_manager', $entityManagerName)), $phpArrayFile])
+            ->addTag('kernel.cache_warmer', ['priority' => 1000]); // priority should be higher than ProxyCacheWarmer
+
+        $container->setAlias($metadataCacheAlias, $phpArrayCacheDecoratorServiceId);
+        $container->register($phpArrayCacheDecoratorServiceId, DoctrineProvider::class)
+            ->addArgument(
+                new Definition(PhpArrayAdapter::class, [
+                    $phpArrayFile,
+                    new Definition(DoctrineAdapter::class, [new Reference($decoratedMetadataCacheServiceId)]),
+                ])
+            );
     }
 }
--- a/CacheWarmer/DoctrineMetadataCacheWarmer.php
+++ b/CacheWarmer/DoctrineMetadataCacheWarmer.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\CacheWarmer;
+
+use Doctrine\ORM\EntityManagerInterface;
+use LogicException;
+use Symfony\Bundle\FrameworkBundle\CacheWarmer\AbstractPhpFileCacheWarmer;
+use Symfony\Component\Cache\Adapter\ArrayAdapter;
+use Symfony\Component\Cache\DoctrineProvider;
+
+class DoctrineMetadataCacheWarmer extends AbstractPhpFileCacheWarmer
+{
+    /** @var EntityManagerInterface */
+    private $entityManager;
+
+    /** @var string */
+    private $phpArrayFile;
+
+    public function __construct(EntityManagerInterface $entityManager, string $phpArrayFile)
+    {
+        $this->entityManager = $entityManager;
+        $this->phpArrayFile  = $phpArrayFile;
+
+        parent::__construct($phpArrayFile);
+    }
+
+    /**
+     * It must not be optional because it should be called before ProxyCacheWarmer which is not optional.
+     */
+    public function isOptional(): bool
+    {
+        return false;
+    }
+
+    /**
+     * @param string $cacheDir
+     */
+    protected function doWarmUp($cacheDir, ArrayAdapter $arrayAdapter): bool
+    {
+        // cache already warmed up, no needs to do it again
+        if (is_file($this->phpArrayFile)) {
+            return false;
+        }
+
+        $metadataFactory = $this->entityManager->getMetadataFactory();
+        if (count($metadataFactory->getLoadedMetadata()) > 0) {
+            throw new LogicException('DoctrineMetadataCacheWarmer must load metadata first, check priority of your warmers.');
+        }
+
+        $metadataFactory->setCacheDriver(new DoctrineProvider($arrayAdapter));
+        $metadataFactory->getAllMetadata();
+
+        return true;
+    }
+}
--- a/Command/Proxy/ClearMetadataCacheDoctrineCommand.php
+++ b/Command/Proxy/ClearMetadataCacheDoctrineCommand.php
@@ -9,6 +9,8 @@ use Symfony\Component\Console\Output\OutputInterface;
 
 /**
  * Command to clear the metadata cache of the various cache drivers.
+ *
+ * @deprecated
  */
 class ClearMetadataCacheDoctrineCommand extends MetadataCommand
 {
@@ -30,6 +32,8 @@ class ClearMetadataCacheDoctrineCommand extends MetadataCommand
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        @trigger_error(sprintf('The "%s" (doctrine:cache:clear-metadata) is deprecated, metadata cache now uses PHP Array cache which can not be cleared.', self::class), E_USER_DEPRECATED);
+
         DoctrineCommandHelper::setApplicationEntityManager($this->getApplication(), $input->getOption('em'));
 
         return parent::execute($input, $output);
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -688,6 +688,13 @@ class Configuration implements ConfigurationInterface
                 ->scalarNode('pool')->end()
             ->end();
 
+        if ($name === 'metadata_cache_driver') {
+            $node->setDeprecated(...$this->getDeprecationMsg(
+                'The "metadata_cache_driver" configuration key is deprecated. PHP Array cache is now automatically registered when %kernel.debug% is false.',
+                '2.3'
+            ));
+        }
+
         return $node;
     }
 
--- a/Tests/DependencyInjection/DoctrineExtensionTest.php
+++ b/Tests/DependencyInjection/DoctrineExtensionTest.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection;
 
+use Doctrine\Bundle\DoctrineBundle\CacheWarmer\DoctrineMetadataCacheWarmer;
 use Doctrine\Bundle\DoctrineBundle\DependencyInjection\DoctrineExtension;
 use Doctrine\Bundle\DoctrineBundle\Tests\Builder\BundleConfigurationBuilder;
 use Doctrine\DBAL\Connection;
@@ -12,6 +13,8 @@ use LogicException;
 use PHPUnit\Framework\TestCase;
 use Symfony\Bridge\Doctrine\Messenger\DoctrineClearEntityManagerWorkerSubscriber;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
+use Symfony\Component\Cache\Adapter\DoctrineAdapter;
+use Symfony\Component\Cache\Adapter\PhpArrayAdapter;
 use Symfony\Component\Cache\DoctrineProvider;
 use Symfony\Component\DependencyInjection\Compiler\ResolveChildDefinitionsPass;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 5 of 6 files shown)


ossinkine:
When I deprecate the command by adding `<deprecated>` tag to XML-config, the tests are failing with an error `Since symfony/dependency-injection 5.1: Not setting the attribute "package" of the node "deprecated" in "/home/travis/build/doctrine/DoctrineBundle/DependencyInjection/../Resources/config/orm.xml" is deprecated.
` (https://travis-ci.com/github/doctrine/DoctrineBundle/builds/201245230)
When I add this attributes, other tests are failing with an error `Symfony\Component\DependencyInjection\Exception\InvalidArgumentException: Unable to parse file "/home/travis/build/doctrine/DoctrineBundle/DependencyInjection/../Resources/config/orm.xml": [ERROR 1827] Element '{http://symfony.com/schema/dic/services}deprecated', attribute 'package': The attribute 'package' is not allowed. (in /home/travis/build/doctrine/DoctrineBundle/ - line 156, column 0)
` (https://travis-ci.com/github/doctrine/DoctrineBundle/builds/201255247)

ostrolucky:
Yeah, this is why we don't usually deprecate commands in xml, just put `@deprecated` and trigger_error inside command, similarly as is done in other commands

ostrolucky:
I won't merge this again on my own, so please other doctrine members kindly review as well.

greg0ire:
Please do one commit per bug fixed, and for each bugfix, summarize what's broken, and explain why your changes fix what is broken.

ossinkine:
@greg0ire I've updated the commits. I've put the links to fixed issues in started post, they explaining what's broken and how it proposed to fix.

greg0ire:
>  I've put the links to fixed issues in started post, they explaining what's broken and how it proposed to fix.

That's my point, you shouldn't have to read lengthy discussions. Since you insist on not doing this, then I'm going to do it myself for the next reviewer, or for the person who is going to read the history if these commits somehow cause an issue. This is going to take me some time, but I can't give a good review if I fully understand what this is about first.

greg0ire:
The first commit is labeled "Add PhpArrayAdapter to cache metadata", but that's the title of #1196 … you are not doing something twice, are you?

ossinkine:
> That's my point, you shouldn't have to read lengthy discussions. Since you insist on not doing this, then I'm going to do it myself for the next reviewer, or for the person who is going to read the history if these commits somehow cause an issue. This is going to take me some time, but I can't give a good review if I fully understand what this is about first.

IMO if you want to figure out why changes was made you whatever will read all related to understand the root of the issue, what initial state produce the issue and so on, even if the problem is described here.

ossinkine:
> The first commit is labeled "Add PhpArrayAdapter to cache metadata", but that's the title of #1196 … you are not doing something twice, are you?

#1196 was reverted in #1255
I just commited it again

greg0ire:
Shouldn't the first commit be squashed with the first one then? What the point of introducing something broken then fixing it in the commit right after?

ossinkine:
> Please do one commit per bug fixed

🤷‍♂️

greg0ire:
All the bugs this PR are pointing to are fixed by the revert, aren't they? Or can some of the bugs still be experienced with the latest version?

greg0ire:
About the cache metadata clearing command, it's deprecated, but running it will still cause an issue, won't it? In short, if we merge and release this, can everybody just run `composer update` and their deploy scripts will still work, or not?

ossinkine:
> All the bugs this PR are pointing to are fixed by the revert, aren't they? Or can some of the bugs still be experienced with the latest version?

Yes, all bugs was fixed by the revert, because they was introduced by inital PR.
This PR contains the initial commit and the bug fixes which the inital commit has.

ossinkine:
> About the cache metadata clearing command, it's deprecated, but running it will still cause an issue, won't it? In short, if we merge and release this, can everybody just run `composer update` and their deploy scripts will still work, or not?

`composer update` does not execute the metadata cache clearing by default, so it's safe. If somebody has the metadata cache clearing in his post-install scripts, he should remove it because it does not make sence now.

greg0ire:
> he should remove it because it does not make sence now.

Ok so it's a BC-break I'm afraid. You can be sure you will get bug reports from people about this. But since this is only an issue with the new cache system you are introducing, maybe you can somehow make the command always succeed in addition to the deprecation?

greg0ire:
> UPD: This only applies to the no debug mode (usually only prod environment)

That makes the BC-break far worse because it cannot be detected in dev.

greg0ire:
> Yes, all bugs was fixed by the revert, because they was introduced by inital PR. This PR contains the initial commit and the bug fixes which the inital commit has.

Then it's simple: do one commit. If you have a lot of explaining to do, do it in the body of the commit message, or with code comments.

ossinkine:
> Ok so it's a BC-break I'm afraid. You can be sure you will get bug reports from people about this. But since this is only an issue with the new cache system you are introducing, maybe you can somehow make the command always succeed in addition to the deprecation?

I can add a check to the command if PHP Array cache is enabled we just return a success result with the appropriate message

UPGRADE-2.2.md
```diff
@@ -13,3 +13,4 @@ Configuration
     * `doctrine.dbal.keep_slave`. Use `doctrine.dbal.keep_replica`
     
     Similarly, if you use XML configuration, please replace `<slave>` with `<replica>`.
+ * The `metadata_cache_driver` configuration key has been deprecated. PHP Array cache is now automatically registered when `%kernel.debug%` is false.
```
greg0ire: I think the command deprecation should be mentioned here too.

DependencyInjection/DoctrineExtension.php
```diff
@@ -927,4 +936,26 @@ private function createArrayAdapterCachePool(ContainerBuilder $container, string
...
+    {
+        $metadataCacheAlias              = $this->getObjectManagerElementName($entityManagerName . '_metadata_cache');
+        $decoratedMetadataCacheServiceId = (string) $container->getAlias($metadataCacheAlias);
+        $phpArrayCacheDecoratorServiceId = $decoratedMetadataCacheServiceId . '.php_array';
+        $phpArrayFile                    = '%kernel.cache_dir%' . sprintf('/doctrine/orm/%s_metadata.php', $entityManagerName);
+        $cacheWarmerServiceId            = $this->getObjectManagerElementName($entityManagerName . '_metadata_cache_warmer');
+
+        $container->register($cacheWarmerServiceId, DoctrineMetadataCacheWarmer::class)
+            ->setArguments([new Reference(sprintf('doctrine.orm.%s_entity_manager', $entityManagerName)), $phpArrayFile])
+            ->addTag('kernel.cache_warmer', ['priority' => 1000]);
```
greg0ire: That's a magic number, I think a comment would be welcome here, saying what other warmer(s) we want to get priority on (all? or just the one you mentioned in the issue you linked?)

dmaicher:
I think we should do some more benchmarks and make sure this change is actually worth it?

I just tried it on one of my apps and I can actually not measure any real performance improvement.

I changed my front controller to just this:

```php
require __DIR__.'/../vendor/autoload.php';

$kernel = new AppKernel('prod', false);
$kernel->boot();

/** @var \Doctrine\ORM\EntityManagerInterface $em */
$em = $kernel->getContainer()->get('doctrine.orm.default_entity_manager');
$em->getMetadataFactory()->getAllMetadata();
``` 

I made sure to clear the cache and re-start php-fpm in between the runs.

I got those results on 3 runs with 10000 requests each using apache-bench

```
ab -n 100000 -c 1 "http://my.app"
``` 

And I'm taking `Time per request (mean)` as the benchmark here.

| cache type | run 1 | run 2 | run 3 | 
|---|---|---|---|
| old meta data caching using `cache.system` | 0.203ms  | 0.206ms | 0.204ms |  
| this PR | 0.202ms  | 0.205ms  | 0.203ms  |  


Some details on my app and system:
- Symfony 4.4
- PHP 7.4.12
- 116 Entity classes on default EM. So `getAllMetadata()` loaded data for all 116 (checked that).

@ossinkine can you share some details on the benchmarks you did?

dmaicher:
I just realized my first benchmark numbers shared above were wrong. I had Varnish in front of nginx and php-fpm and it was just returning cached responses. :man_facepalming:

Now the numbers are updated and correct.

ossinkine:
Unfortunatelly I did not do any synthetic tests but I just apply this PR on my prod app and I get the following results in the monitoring:
Screenshot_1

Could you please try to benchmark concurrent requests, for example using `-c 10` flag?

ossinkine:
Median response time with APCu cache is about 84ms but with PHP array cache is about 80ms.
The system configuration like you: PHP 7.4.12, Symfony 4.4, about 80 entities.

dmaicher:
@ossinkine interesting. And what cache configuration did you use exactly before the deploy of the change? 

I used this "before" config: https://github.com/symfony/recipes/blob/master/doctrine/doctrine-bundle/1.12/config/packages/prod/doctrine.yaml#L6

Let me try to increase the concurrency and do another benchmark but I don't see how this would affect the results

ostrolucky:
you need to be sure you have opcache correctly configured as well

ossinkine:
> And what cache configuration did you use exactly before the deploy of the change?
```yaml
doctrine:
    orm:
        auto_generate_proxy_classes: false
        metadata_cache_driver:
            type: pool
            pool: doctrine.system_cache_pool
        query_cache_driver:
            type: pool
            pool: doctrine.system_cache_pool
        result_cache_driver:
            type: pool
            pool: doctrine.result_cache_pool

framework:
    cache:
        pools:
            doctrine.result_cache_pool:
                adapter: cache.app
            doctrine.system_cache_pool:
                adapter: cache.system

        app: cache.adapter.redis
        system: cache.apcu_redis
        pools:
            cache.apcu_redis:
                adapters:
                    - cache.adapter.apcu
                    - cache.adapter.redis
```

dmaicher:
I tried it with concurrency 4 (`-c 4`) as my laptop only has 4 cores.

Only did one run:

| cache type | run 1 |
|---|---|
| old meta data caching using `cache.system` | 17.562ms  |
| this PR | 17.978ms |

Which shows the same trend as the benchmark above.

my opcache settings:

```
opcache.enable = 1
opcache.max_accelerated_files = 40000
opcache.revalidate_freq = 3
```

ossinkine:
My opcache settings:
```
opcache.blacklist_filename => no value => no value
opcache.consistency_checks => 0 => 0
opcache.dups_fix => Off => Off
opcache.enable => On => On
opcache.enable_cli => Off => Off
opcache.enable_file_override => On => On
opcache.error_log => no value => no value
opcache.file_cache => no value => no value
opcache.file_cache_consistency_checks => On => On
opcache.file_cache_only => Off => Off
opcache.file_update_protection => 2 => 2
opcache.force_restart_timeout => 180 => 180
opcache.huge_code_pages => Off => Off
opcache.interned_strings_buffer => 32 => 32
opcache.lockfile_path => /tmp => /tmp
opcache.log_verbosity_level => 2 => 2
opcache.max_accelerated_files => 20000 => 20000
opcache.max_file_size => 0 => 0
opcache.max_wasted_percentage => 5 => 5
opcache.memory_consumption => 256 => 256
opcache.opt_debug_level => 0 => 0
opcache.optimization_level => 0x7FFEBFFF => 0x7FFEBFFF
opcache.preferred_memory_model => no value => no value
opcache.preload => no value => no value
opcache.preload_user => no value => no value
opcache.protect_memory => Off => Off
opcache.restrict_api => no value => no value
opcache.revalidate_freq => 2 => 2
opcache.revalidate_path => Off => Off
opcache.save_comments => On => On
opcache.use_cwd => On => On
opcache.validate_permission => Off => Off
opcache.validate_root => Off => Off
opcache.validate_timestamps => Off => Off
```

ossinkine:
@dmaicher Do you use APCu or this results using PHP Files Adapter?

dmaicher:
> @dmaicher Do you use APCu or this results using PHP Files Adapter?

as mentioned I use the "`cache.system` based cache pool" vs your PR. I did not customize anything regarding APCu.

Here another benchmark on a real controller action on my app using `ab -n 1000 -c 1 "http://my.app/...`

| cache type | run 1 |
|---|---|
| old meta data caching using `cache.system` | 40.108ms |
| this PR | 41.754ms |

So at least on my machine its performing slightly worse.

dmaicher:
> > And what cache configuration did you use exactly before the deploy of the change?
> 
> ```yaml
> doctrine:
>     orm:
>         auto_generate_proxy_classes: false
>         metadata_cache_driver:
>             type: pool
>             pool: doctrine.system_cache_pool
>         query_cache_driver:
>             type: pool
>             pool: doctrine.system_cache_pool
>         result_cache_driver:
>             type: pool
>             pool: doctrine.result_cache_pool
> 
> framework:
>     cache:
>         pools:
>             doctrine.result_cache_pool:
>                 adapter: cache.app
>             doctrine.system_cache_pool:
>                 adapter: cache.system
> 
>         app: cache.adapter.redis
>         system: cache.apcu_redis
>         pools:
>             cache.apcu_redis:
>                 adapters:
>                     - cache.adapter.apcu
>                     - cache.adapter.redis
> ```

Ok I see @ossinkine you are not using the default config as in the recipe here :wink: 

> `system: cache.apcu_redis`

ossinkine:
@dmaicher `cache.system` creates cache using `PhpFilesAdapter` if you have no APCu enabled. If APCu enabled it creates chaining cache with `ApcuAdapter` and `PhpFilesApapter`.
I have APCu enabled so `cache.system` uses `ApcuAdapter` exactly like my configuration.
That is why I asked if you have APCu enabled.

dmaicher:
@ossinkine this is my metadata driver with the default `cache.system` based pool. So yes I have APCu enabled.

Screenshot at 2020-11-16 14-06-00

ossinkine:
@ostrolucky Which feedback we are waiting? I've created the PR, described benefit I got. @dmaicher made his tests and can't approve my experience. Also we've fixed all reproducable bugs, and right now I know about one which I couldn't reproduce #1249. So what's the next step?

ostrolucky:
What's missing is reproducer for your results. There is a report from @dmaicher that speed up is not there, but there wasn't a reply to that. We need to be able to check claimed speed up ourselves to properly assess the situation, otherwise we should just close here. This would affect lot of appplications and deprecating stuff like `doctrine:cache:clear-metadata` without alternative is not great. Gotta measure good numbers here for it to be worth it.

ossinkine:
Ok, I made exactly the same benchmark on my localhost app as @dmaicher described and there is my results below:

cache type | run 1 | run 2 | run 3
-- | -- | -- | --
old meta data caching using cache.system | 16.619ms | 16.891ms | 15.958ms
this PR | 14.576ms | 14.535ms | 14.747ms

- Symfony 4.4.18
- PHP 7.4.13
- 87 entities

ossinkine:
@ostrolucky @dmaicher Hi guys, hope your Christmas holidays are good!

I've created the reproducer to easily test performance of proposed behavior: https://github.com/ossinkine/doctrine-bundle-issue-1260
I've created a new Symfony 4.4 project with default configuration, added 120 test entities, create docker-compose with production configuration and create two app environments `prod` and `prod_new` with old and this PR behaviors.
So to test just clone my repo, install composer dependencies and follow README instructions:
```bash
docker-compose up -d --build
echo APP_ENV=prod > .env.local # or prod_env
echo APP_DEBUG=false >> .env.local
rm -rf var # ensure there is no cache
docker-compose exec php bin/console cache:warmup
docker-compose exec ab wget -O - http://nginx/ # check we are ready to benchmark, there are no errors
docker-compose exec ab ab -n 10000 -c 1 http://nginx/
```
I've benched both environment and get this results:

env | run 1 | run 2 | run 3
-- | -- | -- | --
prod | 8.194ms | 8.453ms | 8.562ms
prod_new | 6.187ms | 6.286ms | 6.246ms

Looking forward to your results.

dmaicher:
@ossinkine thanks a lot for the reproducible benchmark :blush: 

here are my results:

env | run 1 | run 2 | run 3
-- | -- | -- | --
prod | 11.146ms | 11.171ms | 11.539ms
prod_new | 8.502ms | 8.473ms | 8.696ms

So indeed your approach is faster!

That made me wonder whats happening on my app... so I checked again with your changes ported to my app (compiler pass + cache warmer).

Now when I clear the cache I get this error:

```
DoctrineMetadataCacheWarmer must load metadata first, check priority of your warmers.
``` 

The only way I can get it to work is by running `cache:warmup --env=prod --no-debug` twice :confused: Then I can also confirm the speed-up on my app!

So I think the last time I benchmarked (where the exception was not thrown yet in the cache warmer) I ended up with an empty meta data cache file. So the warm-up did not work.

I can actually also reproduce it on your repo:

```
$ sudo rm -rf var/*
$ docker-compose exec php bin/console cache:clear

 // Clearing the cache for the prod_new environment with debug false                                                    


In DoctrineMetadataCacheWarmer.php line 33:
                                                                                         
  DoctrineMetadataCacheWarmer must load metadata first, check priority of your warmers.  
                                                                                         

cache:clear [--no-warmup] [--no-optional-warmers] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-e|--env ENV] [--no-debug] [--] <command>
```

So something is still a bit broken there.

ossinkine:
@dmaicher thank you for your results.

You are right, the issue was the warmers are called twice when container does not exist (first on container building and second on command executing). I've made the following changes to fix:
1. made the warmer not optional because not optional warmers is called on container building and proxy cache warmer loads metadata before the cache warmer;
2. check the cache file existance and don't warm the cache if the cache file exists because not optional warmers is called twice, so the second time metadata is already loaded and an exception is throwed.

Also I've rebased current branch from the target branch and updated the reproducer. So could you please check again?

dmaicher:
I can confirm it works now without any issues on my app :+1:

tested a few scenarios
- no cache yet + cache:clear
- no cache yet + cache:warmup
- existing cache + cache:clear
- existing cache + cache:warmup

Also when loading all meta data (118 entities on my app) there is a 2ms performance increase :rocket: 

I'm still a bit hesitant to activate this new caching strategy by default for all users of the bundle though... I can foresee there might be issues for people currently depending on the `doctrine:cache:clear-metadata` command that is executed during deployment. 

~~For my apps there is no issue as there are deployed in a clean environment every time but I guess there will be some people out there that are just uploading new files to a server (same path on the filesystem) and then clearing cache etc. For those people it will break without clearing opcache during deploy, right?~~ Actually this should not be an issue really.

ostrolucky:
> I guess there will be some people out there that are just uploading new files to a server (same path on the filesystem) and then clearing cache etc. For those people it will break without clearing opcache during deploy, right?

Actually I think if they are not clearing opcache and are using it without timestamp validation, those people are in deep trouble already, because it would mean any changes in php files won't take effect, including their changes in business logic.

dmaicher:
> Actually I think if they are not clearing opcache and are using it without timestamp validation, those people are in deep trouble already, because it would mean any changes in php files won't take effect, including their changes in business logic.

Yeah exactly. I edited my comment above :blush:

ossinkine:
So what the next out step, gentlemen?
We can deprecate command and `metadata_cache_driver` option but don't enable php array cache by default.
Or we can even don't deprecate anything and just add an option to enable php array cache.
But it just won't be consistent with Symfony caches.

ostrolucky:
> I'm still a bit hesitant to activate this new caching strategy by default for all users of the bundle though... I can foresee there might be issues for people currently depending on the doctrine:cache:clear-metadata command that is executed during deployment.

What do you suggest? IMHO BC break is very minor - is not an API change and neither runtime (command is meant to be executed during deployment only). Such things are not covered in Symfony BC promise either. In my opinion we should just document it as pointed out by @greg0ire and go with it. We got report about this from single user at https://github.com/doctrine/DoctrineBundle/issues/1251 and even there they admitted this was a legacy thing.

However, there is also possibility to change cache clear command so it just outputs warning message that nothing has actually been cleared (because it's not necessary) when kernel.debug is false, instead of error.

dmaicher:
I just checked the `doctrine:cache:clear-metadata` command and actually it does something interesting when called with `--flush`. It actually deletes the php cache file :see_no_evil: 

So if people are calling `doctrine:cache:clear-metadata --flush` during deployment it does not fail and would result in a performance loss due to additional cache lookups/misses :confused: 

Not sure how to approach this to be honest.

What do you think @stof ?

ostrolucky:
@ossinkine Can you please resolve last remaining comment from @greg0ire (adding changelog entry about deprecating the command)? After that I will merge this, you did fine work and I'm not willing to let this rot longer.

dmaicher:
Ok let's get this merged :+1: :sunglasses: Actually also want the performance gain for my apps :smile:

ossinkine:
> Base automatically changed from master to 2.3.x on 18 Jan

@ostrolucky @dmaicher Please tell me which targed branch should be specified?

ostrolucky:
2.3.x, it's correct

ostrolucky (approved):

ossinkine:
Godspeed 🙏

dmaicher (approved):

ostrolucky:
Thanks for your strongwill for pushing this through! I hope there won't be more unexpected issues after we release this second time 🙏

PR #1261, opened 2020-11-14
Implement unit tests with Github Actions

greg0ire:


greg0ire:
@ostrolucky as promised, here is a PR that shows Github Actions and Travis side-by-side. Github Actions manages to finish before even though coverage is computed and uploaded for every job.

I don't know why there is no status report from Codecov, but it seems to occur for all Doctrine repositories, and is under investigation at https://community.codecov.io/t/codecov-commit-status-check-missing-for-pr-when-using-codecov-action/1813/14

ostrolucky:
CS is fixed already in 2.2. This was done because it's not worth doing it in 1.12.x, as maintenance of this branch will be stopped this month. Hence, I would like to remove CS related changes from this PR. We already have GH action for this in 2.2+. Also, why is this draft?

ostrolucky:
Times:

<table>
<tr>
	<td></td>
	<th>GH action</th>
	<th>Travis</th>
</tr>
<tr>
	<th>PHP 7.1</th>
	<td>2m</td>
	<td>44s</td>
</tr>
<tr>
	<th>PHP 7.2</th>
	<td>1m</td>
	<td>39s</td>
</tr>
<tr>
	<th>PHP 7.3</th>
	<td>1m</td>
	<td>53s</td>
</tr>
<tr>
	<th>PHP 7.4</th>
	<td>51s</td>
	<td>56s</td>
</tr>
<tr>
	<th>lowest</th>
	<td>2m</td>
	<td>50s</td>
</tr>
</table>

Is this because you run tests with coverage? Why would you do that and hence increase the times of every job? Also, you are using pcov instead of xdebug for almost all jobs, so this is really not a fair benchmark.

greg0ire:
CS is fixed already in 2.2. This was done because it's not worth doing it in 1.12.x, as maintenance of this branch will be stopped this month. Hence, I would like to remove CS related changes from this PR.

Ok, I think I will be able to remove them by allowing `doctrine/coding-standard` 8 but making sure the job is run with 6.

We already have GH action for this in 2.2+.

IMO things related to CI should always always be the same on every branch (i.e. contributed to the lowest supported branch), otherwise this makes merges up painful because the rules are slightly different.


Also, why is this draft?

It's a draft because I have to `git rm -f .travis.yml`, but I promised you I'd let you have a look at a comparison between the 2 first.


> Also, you are using pcov instead of xdebug for almost all jobs unlike Travis which uses xdebug everywhere, so this is really not a fair benchmark.

:thinking: XDebug is only enable once in Travis, right? So the benchmark is unfair, but to Github Actions, not to Travis.

Also, the benchmark must be wrong because although Travis advertises faster times, it finishes after. Maybe more jobs are run in parallelm, or the timings are not the same? Aboe you can see that Travis as a whole is successful in 3m (Travis itself says 3m 40s)

greg0ire:
> Why would you do that and hence increase the times of every job?

I think it's really cheap and that way, codepaths that are specific to a specific Symfony version or PHP version are counted.

greg0ire:
I kust noticed that I can make things faster with `--prefer-stable` and `--prefer-dist` here and there. That got lost in translation.

ostrolucky:
> IMO things related to CI should always always be the same on every branch (i.e. contributed to the lowest supported branch), otherwise this makes merges up painful because the rules are slightly different.

Yes, so point this to 2.2 only, so we don't have to merge up. As I mentioned 1.12 will go ouf ot maintenance soon.

> It's a draft because I have to git rm -f .travis.yml, but I promised you I'd let you have a look at a comparison between the 2 first.

No you don't have to, we can do it in separate PR. Might be good idea to run these concurrently for some time.

> XDebug is only enable once in Travis, right? So the benchmark is unfair, but to Github Actions, not to Travis.

Yes, hence why the times are better on Travis in table I shown. I would like to see comparison when conditions are the same, otherwise I see Travis is better from this comparison (see table I have shown).

> Also, the benchmark must be wrong because although Travis advertises faster times, it finishes after. Maybe more jobs are run in parallelm, or the timings are not the same? Aboe you can see that Travis as a whole is successful in 3m (Travis itself says 3m 40s)

Travis has only one job for coverage, which takes most time. This is why I asked why did you make it different in GH action and hence made every job slower than Travis, rather than having one coverage job.

> I think it's really cheap and that way, codepaths that are specific to a specific Symfony version or PHP version are counted.

Not worth it IMHO, nobody from us is looking at coverage reports. I would rather I get test failures feedback from CI faster.

greg0ire:
> Yes, so point this to 2.2 only

Ok I'll rebase.

> Not worth it IMHO, nobody from us is looking at coverage reports. I would rather I get test failures feedback from CI faster.

I don't think you will see much difference with PCOV vs without PCOV. Also, once [this](https://community.codecov.io/t/codecov-commit-status-check-missing-for-pr-when-using-codecov-action/1813/14) is fixed, you will get a status report right in this UI, so I think you might look more at it.

greg0ire:
I'm going to try without PCOV in case it makes a big difference at the install step (which is indeed longer when there is PCOV, but that might be because there also is `pdo_sqlite`.

greg0ire:
@ostrolucky I pushed a version with just one job for PCOV, and it does not seem to make a noticeable difference. As I'm writing this lines, the Travis build is still running.

ostrolucky:
Travis coverage job is using xdebug, of course it's slower than one with pcov that GH action is using. 

I wanted to give benefit of doubt to GH actions and try to rationalize why it's slower in most jobs (because it's doing coverage in each job unlike Travis), but you changed it now so coverage is not runnig in each build and it's still slower than Travis in most jobs.

So this indeed confirms my suspicions that GH actions are slower than Travis.

greg0ire:
@ostrolucky reverted to coverage for each job

PR #1263, opened 2020-11-16
Fixes for compatibility with DBAL 3

### LINKED ISSUE #1262: Call to undefined method Doctrine\DBAL\Logging\LoggerChain::addLogger() with DBAL 3
We advertise compatibility with `doctrine/dbal` 3, but still call that method, which was removed in favor of calling the constructor here:

https://github.com/doctrine/DoctrineBundle/blob/f5a5a3dd1351818080b609b544608b472b07029f/Resources/config/dbal.xml#L27

and here:

https://github.com/doctrine/DoctrineBundle/blob/5a94ec5c9081ddf10061b95e0c140dc260d6b6d3/DependencyInjection/DoctrineExtension.php#L156

I might fix that later, feel free to beat me to it, anyone.

### LINKED ISSUE #1264: Attempted to load class "ImportCommand" from namespace "Doctrine\DBAL\Tools\Console\Command" with DBAL 3
After fixing #1262 , I'm getting this, I suppose there is more work to be done for DBAL 3 compatibility.

```
bin/console doctrine:database:import ./myfile.sql
Symfony\Component\ErrorHandler\Error\ClassNotFoundError^ {#99
  #message: """
    Attempted to load class "ImportCommand" from namespace "Doctrine\DBAL\Tools\Console\Command".\n
    Did you forget a "use" statement for another namespace?
    """
  #code: 0
  #file: "./vendor/doctrine/doctrine-bundle/Command/Proxy/ImportDoctrineCommand.php"
  #line: 15
  trace: {
    ./vendor/doctrine/doctrine-bundle/Command/Proxy/ImportDoctrineCommand.php:15 { …}
    ./vendor/composer/ClassLoader.php:444 { …}
    ./vendor/composer/ClassLoader.php:322 { …}
    ./vendor/symfony/error-handler/DebugClassLoader.php:349 { …}
    Symfony\Component\ErrorHandler\DebugClassLoader->loadClass() {}
    ./var/cache/dev/ContainerUi1zBM0/getDoctrine_DatabaseImportCommandService.php:20 {
      ContainerUi1zBM0\getDoctrine_DatabaseImportCommandService::do($container, $lazyLoad = true)^
      › {
      ›     $container->privates['doctrine.database_import_command'] = $instance = new \Doctrine\Bundle\DoctrineBundle\Command\Proxy\ImportDoctrineCommand();
      › 
      arguments: {
        $class_name: "Doctrine\Bundle\DoctrineBundle\Command\Proxy\ImportDoctrineCommand"
      }
    }
    ./var/cache/dev/ContainerUi1zBM0/App_KernelDevDebugContainer.php:351 { …}
    ./vendor/symfony/dependency-injection/Container.php:441 { …}
    ./vendor/symfony/dependency-injection/Argument/ServiceLocator.php:40 { …}
    ./vendor/symfony/console/CommandLoader/ContainerCommandLoader.php:45 { …}
    ./vendor/symfony/console/Application.php:527 { …}
    ./vendor/symfony/console/Application.php:616 { …}
    ./vendor/symfony/framework-bundle/Console/Application.php:116 { …}
    ./vendor/symfony/console/Application.php:230 { …}
    ./vendor/symfony/framework-bundle/Console/Application.php:82 { …}
    ./vendor/symfony/console/Application.php:142 { …}
    ./bin/console:49 { …}
  }
}
```

`ImportCommand` is deprecated with `@deprecated Use a database client application instead.`, not sure how we should handle this in the bundle…

greg0ire:
I tested this on a real project, it allows to install the project without errors.

Fixes #1262 , fixes #1264 

# How can I test this?

```shell
composer config repositories.greg0ire vcs https://github.com/greg0ire/DoctrineBundle
composer require doctrine/doctrine-bundle "dev-dbal-3-compat as 2.2.1"
```

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -9,6 +9,8 @@ use Doctrine\Bundle\DoctrineBundle\EventSubscriber\EventSubscriberInterface;
 use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepositoryInterface;
 use Doctrine\DBAL\Connections\MasterSlaveConnection;
 use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection;
+use Doctrine\DBAL\Logging\LoggerChain;
+use Doctrine\DBAL\SQLParserUtils;
 use Doctrine\DBAL\Tools\Console\ConnectionProvider;
 use Doctrine\ORM\Proxy\Autoloader;
 use Doctrine\ORM\UnitOfWork;
@@ -85,6 +87,14 @@ class DoctrineExtension extends AbstractDoctrineExtension
     {
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('dbal.xml');
+        $chainLogger = $container->getDefinition('doctrine.dbal.logger.chain');
+        $logger      = new Reference('doctrine.dbal.logger');
+        if (! method_exists(SQLParserUtils::class, 'getPositionalPlaceholderPositions') && method_exists(LoggerChain::class, 'addLogger')) {
+            // doctrine/dbal < 2.10.0
+            $chainLogger->addMethodCall('addLogger', [$logger]);
+        } else {
+            $chainLogger->addArgument([$logger]);
+        }
 
         if (empty($config['default_connection'])) {
             $keys                         = array_keys($config['connections']);
@@ -143,8 +153,16 @@ class DoctrineExtension extends AbstractDoctrineExtension
                 ->replaceArgument(1, $connection['profiling_collect_schema_errors']);
 
             if ($logger !== null) {
-                $chainLogger = new ChildDefinition('doctrine.dbal.logger.chain');
-                $chainLogger->addMethodCall('addLogger', [$profilingLogger]);
+                $chainLogger = $container->register(
+                    'doctrine.dbal.logger.chain',
+                    LoggerChain::class
+                );
+                if (! method_exists(SQLParserUtils::class, 'getPositionalPlaceholderPositions') && method_exists(LoggerChain::class, 'addLogger')) {
+                    // doctrine/dbal < 2.10.0
+                    $chainLogger->addMethodCall('addLogger', [$profilingLogger]);
+                } else {
+                    $chainLogger->addArgument([$logger, $profilingLogger]);
+                }
 
                 $loggerId = 'doctrine.dbal.logger.chain.' . $name;
                 $container->setDefinition($loggerId, $chainLogger);
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection;
 
+use Doctrine\Bundle\DoctrineBundle\Command\Proxy\ImportDoctrineCommand;
 use Doctrine\Bundle\DoctrineBundle\Dbal\ManagerRegistryAwareConnectionProvider;
 use Doctrine\Bundle\DoctrineBundle\Dbal\RegexSchemaAssetFilter;
 use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\ServiceRepositoryCompilerPass;
@@ -11,6 +12,7 @@ use Doctrine\DBAL\Connections\MasterSlaveConnection;
 use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection;
 use Doctrine\DBAL\Logging\LoggerChain;
 use Doctrine\DBAL\SQLParserUtils;
+use Doctrine\DBAL\Tools\Console\Command\ImportCommand;
 use Doctrine\DBAL\Tools\Console\ConnectionProvider;
 use Doctrine\ORM\Proxy\Autoloader;
 use Doctrine\ORM\UnitOfWork;
@@ -94,6 +96,11 @@ class DoctrineExtension extends AbstractDoctrineExtension
             $chainLogger->addMethodCall('addLogger', [$logger]);
         } else {
             $chainLogger->addArgument([$logger]);
+        }
+
+        if (class_exists(ImportCommand::class)) {
+            $container->register('doctrine.database_import_command', ImportDoctrineCommand::class)
+                ->addTag('console.command', ['command' => 'doctrine:database:import']);
         }
 
         if (empty($config['default_connection'])) {
--- a/Resources/config/dbal.xml
+++ b/Resources/config/dbal.xml
@@ -23,10 +23,8 @@
         <service id="Doctrine\DBAL\Driver\Connection" alias="database_connection" public="false" />
         <service id="Doctrine\DBAL\Connection" alias="database_connection" public="false" />
 
-        <service id="doctrine.dbal.logger.chain" class="%doctrine.dbal.logger.chain.class%" public="false" abstract="true">
-            <call method="addLogger">
-                <argument type="service" id="doctrine.dbal.logger" />
-            </call>
+        <service id="doctrine.dbal.logger.chain" class="%doctrine.dbal.logger.chain.class%" public="false" abstract="true" >
+            <!-- TODO: deprecate this service in 2.3.0 -->
         </service>
 
         <service id="doctrine.dbal.logger.profiling" class="%doctrine.dbal.logger.profiling.class%" public="false" abstract="true" />
--- a/Resources/config/dbal.xml
+++ b/Resources/config/dbal.xml
@@ -92,10 +92,6 @@
             <tag name="console.command" command="doctrine:database:drop" />
         </service>
 
-        <service id="doctrine.database_import_command" class="Doctrine\Bundle\DoctrineBundle\Command\Proxy\ImportDoctrineCommand">
-            <tag name="console.command" command="doctrine:database:import" />
-        </service>
-
         <service id="doctrine.query_sql_command" class="Doctrine\Bundle\DoctrineBundle\Command\Proxy\RunSqlDoctrineCommand">
             <tag name="console.command" command="doctrine:query:sql" />
         </service>
--- a/Tests/ContainerTest.php
+++ b/Tests/ContainerTest.php
@@ -8,6 +8,7 @@ use Doctrine\Common\EventManager;
 use Doctrine\DBAL\Configuration as DBALConfiguration;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Logging\LoggerChain;
+use Doctrine\DBAL\Tools\Console\Command\ImportCommand;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\Configuration;
 use Doctrine\ORM\EntityManager;
@@ -34,6 +35,11 @@ class ContainerTest extends TestCase
             LoggerChain::class,
             $container->get('doctrine.dbal.logger.chain.default')
         );
+        if (class_exists(ImportCommand::class)) {
+            self::assertTrue($container->has('doctrine.database_import_command'));
+        } else {
+            self::assertFalse($container->has('doctrine.database_import_command'));
+        }
     }
 
     /**
--- a/Tests/ContainerTest.php
+++ b/Tests/ContainerTest.php
@@ -2,10 +2,12 @@
 
 namespace Doctrine\Bundle\DoctrineBundle\Tests;
 
+use Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection\Fixtures\DbalTestKernel;
 use Doctrine\Common\Annotations\Reader;
 use Doctrine\Common\EventManager;
 use Doctrine\DBAL\Configuration as DBALConfiguration;
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Logging\LoggerChain;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\Configuration;
 use Doctrine\ORM\EntityManager;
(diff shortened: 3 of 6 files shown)


ostrolucky:
Looks good so far, ask me to review after it's no longer draft please.

greg0ire:
> Looks good so far, ask me to review after it's no longer draft please.

I need to revert the `composer.json` change and detect 2.10.0, will work on it this evening but feel fre to do it and push to my branch 👍

Resources/config/dbal.xml
```diff
@@ -23,11 +23,7 @@
         <service id="Doctrine\DBAL\Driver\Connection" alias="database_connection" public="false" />
         <service id="Doctrine\DBAL\Connection" alias="database_connection" public="false" />
 
-        <service id="doctrine.dbal.logger.chain" class="%doctrine.dbal.logger.chain.class%" public="false" abstract="true">
-            <call method="addLogger">
-                <argument type="service" id="doctrine.dbal.logger" />
```
stof, line 28: wasn't this call needed ? I don't see its replacement

greg0ire: The replacement is in PHP, it's the very first lines of this PR

DependencyInjection/DoctrineExtension.php
```diff
@@ -85,6 +87,14 @@ protected function dbalLoad(array $config, ContainerBuilder $container)
...
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('dbal.xml');
+        $chainLogger     = $container->getDefinition('doctrine.dbal.logger.chain');
+        $profilingLogger = new Reference('doctrine.dbal.logger');
+        if (! method_exists(SQLParserUtils::class, 'getPositionalPlaceholderPositions') && method_exists(LoggerChain::class, 'addLogger')) {
+            // doctrine/dbal < 2.10.0
+            $chainLogger->addMethodCall('addLogger', [$profilingLogger]);
+        } else {
+            $chainLogger->addArgument([$profilingLogger]);
+        }
```
greg0ire, line 97: @stof here is the replacement you asked for in https://github.com/doctrine/DoctrineBundle/pull/1263/files#r524450538 (I think it's equivalent)

stof: this is not equivalent. As you need to pass an array of loggers to the constructor rather than multiple calls to `addLogger`, you need to build an array containing them all.

DependencyInjection/DoctrineExtension.php
```diff
@@ -85,6 +87,14 @@ protected function dbalLoad(array $config, ContainerBuilder $container)
     {
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('dbal.xml');
+        $chainLogger     = $container->getDefinition('doctrine.dbal.logger.chain');
+        $profilingLogger = new Reference('doctrine.dbal.logger');
```
stof, line 91: This variable name is really confusing, as `doctrine.dbal.logger` is *not* the profiling logger, but the logger logging your queries in the Symfony logs through PSR-3.

DependencyInjection/DoctrineExtension.php
greg0ire: But there is only one logger before this PR, right?

greg0ire: What are you suggesting exactly? This?

```
            $chainLogger->replaceArgument(
                $loggers,
                array_merge($chainLogger->getArgument($loggers), $logger)
            );
```

stof: no. Cases where the chain logger are used are precisely cases where we have 2 loggers:

- `doctrine.dbal.logger` logging to your app PSR-3 logger (generally Monolog)
- the profiling logger which logs query executions for the profiler

Your suggestion won't work if the reference to `doctrine.dbal.logger` is added to the constructor of the `doctrine.dbal.logger.chain` abstract definition, because the child definition won't have anything yet for the argument. Instead, you should just use `[$logger, $profilingLogger]` as the value of the argument

DependencyInjection/DoctrineExtension.php
greg0ire: Fixed.

DependencyInjection/DoctrineExtension.php
stof: In the existing code, the call done in XML is done on the abstract service definition which is extended by the definitions in each of the DBAL connections.

so after resolving the definition inheritance, each chain logger will have 2 `addLogger` calls:

- the one inherited from the abstract definition for `doctrine.dbal.logger`
- the one set on themselves for the connection-specific profiling logger.

Tests/ContainerTest.php
```diff
@@ -22,13 +23,13 @@
...
     {
-        if (interface_exists(EntityManagerInterface::class)) {
-            return;
-        }
-
-        self::markTestSkipped('This test requires ORM');
+        $container = $this->createDbalOnlyTestContainer();
+        $this->assertInstanceOf(
+            LoggerChain::class,
+            $container->get('doctrine.dbal.logger.chain.default')
```
greg0ire, line 31: I must have messed up somewhere because although the service definition has 2 loggers, that object has only one when I `dump` it.

greg0ire: Oh but when I remove the first `else` branch in this PR, it works.

DependencyInjection/DoctrineExtension.php
```diff
@@ -85,6 +87,14 @@ protected function dbalLoad(array $config, ContainerBuilder $container)
     {
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('dbal.xml');
+        $chainLogger = $container->getDefinition('doctrine.dbal.logger.chain');
+        $logger      = new Reference('doctrine.dbal.logger');
+        if (! method_exists(SQLParserUtils::class, 'getPositionalPlaceholderPositions') && method_exists(LoggerChain::class, 'addLogger')) {
+            // doctrine/dbal < 2.10.0
+            $chainLogger->addMethodCall('addLogger', [$logger]);
+        } else {
+            $chainLogger->addArgument([$logger]);
```
stof, line 96: this must be removed. You are adding an argument here with an array with 1 element, and then later adding a second argument with an array with 2 elements. But the ChainLogger expects the list of loggers as its first argument so that's why things don't work.

Tests/ContainerTest.php
stof: See my comment on the broken code.

DependencyInjection/DoctrineExtension.php
stof: an alternative would be to use `replaceArgument` in the ChildDefinition

Tests/ContainerTest.php
greg0ire: I just pushed, does it look correct or do I still misunderstood? Sorry, I'm quite tired today :sweat_smile:

DependencyInjection/DoctrineExtension.php
greg0ire: That might be better :thinking: in case people are using the abstract service definition to build their custom service.

Tests/ContainerTest.php
stof: This now looks good.

DependencyInjection/DoctrineExtension.php
greg0ire: Pushed again, this time with `replaceArgument`

stof: if they use it to build their own custom service, they will need to change their code for the DBAL 3 way anyway.

Tests/ContainerTest.php
greg0ire: [Pushed again](https://github.com/doctrine/DoctrineBundle/pull/1263#discussion_r524490742)

DependencyInjection/DoctrineExtension.php
greg0ire: Yes, but I'm switching to the constructor as soon as possible, not as late as possible. This means people using DBAL >=2.10 and building their own custom service should need this.

stof: > Pushed again, this time with `replaceArgument`

We could even go a cleaner way to prepare for the day we will require DBAL 2.10+ in the bundle (hopefully soon, if the PR adding back PHP 7.1 support in 2.12 is accepted):
- stop using `ChildDefinition` in `loadDbalConnection` but a normal `Definition` using the chainlogger class (this requires adding the `addLogger` call for `$logger` in the branch dealing with DBAL 2.9, but that's fine)
- mark the `doctrine.dbal.logger.chain` abstract service as deprecated (in case someone was using it for their own service)

greg0ire: That does seem more straightforward, let me try 👍

Resources/config/dbal.xml
```diff
@@ -23,10 +23,8 @@
...
-            </call>
+        <service id="doctrine.dbal.logger.chain" class="%doctrine.dbal.logger.chain.class%" public="false" abstract="true" >
+            <!-- TODO: deprecate this service in 2.3.0 -->
```
greg0ire, line 27: @stof introducing a new deprecation is minor, so I should do it in a subsequent PR to master I think.

Tests/ContainerTest.php
```diff
@@ -22,13 +24,18 @@
 
 class ContainerTest extends TestCase
 {
-    public static function setUpBeforeClass(): void
+    public function testContainerWithDbalOnly(): void
```
ostrolucky, line 27: How come is this not covered by existing tests already?

Tests/ContainerTest.php
```diff
@@ -38,6 +45,12 @@ public static function setUpBeforeClass(): void
      */
     public function testContainer(): void
     {
+        if (interface_exists(EntityManagerInterface::class)) {
+            return;
```
ostrolucky, line 49: You can't copy paste this block of code from previous method, it needs to be adjusted. Currently rest of the body is skipped for both conditions.

Tests/TestCase.php
```diff
@@ -15,6 +15,61 @@
 
 class TestCase extends BaseTestCase
 {
+    public function createDbalOnlyTestContainer(): ContainerBuilder
+    {
+        $container = new ContainerBuilder(new ParameterBag([
+            'kernel.name' => 'app',
```
ostrolucky, line 21: Are you replicating what FrameworkBundle is doing here? This is prone to breakages in my experience. Why not having DBALTestKernel, similarly as we have [TestKernel](https://github.com/doctrine/DoctrineBundle/blob/ec69ba812cbf768c0fdabfc1cff663d14ae7109c/Tests/DependencyInjection/Fixtures/TestKernel.php)? Or even just injection of argument via constructor of TestKernel which enables/disables orm.

DependencyInjection/DoctrineExtension.php
```diff
@@ -85,6 +89,19 @@ protected function dbalLoad(array $config, ContainerBuilder $container)
     {
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('dbal.xml');
+        $chainLogger = $container->getDefinition('doctrine.dbal.logger.chain');
+        $logger      = new Reference('doctrine.dbal.logger');
+        if (! method_exists(SQLParserUtils::class, 'getPositionalPlaceholderPositions') && method_exists(LoggerChain::class, 'addLogger')) {
```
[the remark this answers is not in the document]
greg0ire: You mean "why there are 2 calls to `method_exists`"? The first one is about telling 2.10 apart from 2.9, but it also matches for 3.0.0 (the class `SQLParserUtils` does not even exist in 3.0). The second call allows to make sure we are not in `3.0` or above.

Tests/ContainerTest.php
greg0ire: Testing with only the DBAL? You tell me I don't know :sweat_smile:

Tests/ContainerTest.php
greg0ire: I don't understand your comment, or why this block was in `setUpBeforeClass()` in the first place: there is only one method, the one I am adding should not be skipped (and isn't)

Tests/TestCase.php
greg0ire: I must admit I copy pasted a lot of what I found in `createXmlBundleTestContainer` here… maybe I can reduce this to something more simple.

Tests/ContainerTest.php
ostrolucky: I realized now that ORM doesn't suppport DBAL 3.x yet, so this is why DBAL 3.x integration cannot be tested with ORM combination by existing tests currently.

Tests/ContainerTest.php
ostrolucky:
```php
         self::markTestSkipped('This test requires ORM');
```
is early return, same as `return;` I am commenting here. So if EntityManagerInterface doesn't exist, return. Otherwise, skip (=return). That's what code in this PR is currently doing (nothing).

It was fine to use it like that in `setUpBeforeClass()` because if that method just returns null without doing null, testcase would still be executed.

Tests/TestCase.php
greg0ire: Addressed, it's indeed better like this I think :)

Tests/ContainerTest.php
greg0ire: Oh right, how did I not spot this, that's embarassing :sweat_smile:

ostrolucky (approved):

SenseException (approved):

greg0ire:
@stof can you please give this another review?

greg0ire:
I suppose the 2 existing reviews are enough already :)

dmaicher:
@greg0ire I think we need some of those changes also on 1.12.x, or?

As it also allows installing dbal 3: https://github.com/doctrine/DoctrineBundle/blob/1.12.x/composer.json#L29

greg0ire:
@dmaicher you're right, I had no idea dbal 3 was supported on 1.12.x (or why, it sounds weird, since that version of the bundle is supposed to be left unmaintained soon).

dmaicher:
> @dmaicher you're right, I had no idea dbal 3 was supported on 1.12.x (or why, it sounds weird, since that version of the bundle is supposed to be left unmaintained soon).

Yeah I also just found out while debugging some travis CI failures while working on https://github.com/dmaicher/doctrine-test-bundle/pull/136 :confused:

ostrolucky:
with symfony 3.4 out of maintenance, 1.12.x is pretty much dead now, all that is left is just officially mark it as unmaintained on website config

PR #1273, opened 2020-12-23
Run all tests on PHP 8

alcaeus:
This PR replaces the ocramius/proxy-manager dev dependency with friendsofphp/proxy-manager-lts, which provides support for a wider range of PHP versions. With ORM 2.8 supporting PHP 8 as well, we should now be able to run all tests on PHP 8.


nicolas-grekas (approved):
Thanks!

PR #1283, opened 2021-01-27
Propagate template from EntityRepository

VincentLanglet:
Hi @greg0ire, seems like there is no psalm on this project
I could try to add it during some of my free time.

This PR avoid to a `TooManyTemplateParams` error from psalm and allow to benefit from the `ObjectRepository<T>` interface.

### DIFF
--- a/Repository/ServiceEntityRepository.php
+++ b/Repository/ServiceEntityRepository.php
@@ -19,6 +19,9 @@ use LogicException;
  *         parent::__construct($registry, YourEntity::class);
  *     }
  * }
+ *
+ * @template T
+ * @template-extends EntityRepository<T>
  */
 class ServiceEntityRepository extends EntityRepository implements ServiceEntityRepositoryInterface
 {
--- a/Repository/ServiceEntityRepository.php
+++ b/Repository/ServiceEntityRepository.php
@@ -27,6 +27,8 @@ class ServiceEntityRepository extends EntityRepository implements ServiceEntityR
 {
     /**
      * @param string $entityClass The class name of the entity this repository manages
+     *
+     * @psalm-param class-string<T> $entityClass
      */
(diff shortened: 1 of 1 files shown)


ostrolucky (approved):

dmaicher (approved):

Repository/ServiceEntityRepository.php
```diff
@@ -19,6 +19,9 @@
  *         parent::__construct($registry, YourEntity::class);
  *     }
  * }
+ *
+ * @template T
+ * @template-extends EntityRepository<T>
```
stof, line 24: you might want to add `@psalm-param class-string<T> $entityClass` on the constructor (check the exact name of that special type. I may have made a mistake in it as I haven't checked the doc again about it when writing this comment).

VincentLanglet: Nice idea, I added it. I checked and I think that's the right name of the type.

greg0ire (approved):
> I could try to add it during some of my free time.

That would be much appreciated! There is a template for that [here](https://github.com/doctrine/.github/blob/2c60b2f943cf08df21be3dc733f1b92db8d24beb/workflow-templates/static-analysis.yml#L40) in case you're not already aware of that.

UPD: you must be aware already since you created https://github.com/doctrine/common/pull/920

greg0ire:
Thanks @VincentLanglet !

VincentLanglet:
Hi @greg0ire, what is needed for the next release ? :)

ostrolucky:
I want to squeeze into 2.3.0 https://github.com/doctrine/DoctrineBundle/issues/1300 and ideally https://github.com/doctrine/DoctrineBundle/pull/1297 too. But there was no work on #1300 done yet. Feel free to help implement it if you would like 2.3.0 out of the door faster.

PR #1284, opened 2021-01-28
Enable `@CustomIdGenerator()` to reference services tagged as "doctrine.id_generator"

### LINKED ISSUE #759: [Request] Service IDs in CustomIdGenerator
Allow passing of service ids to CustomIdGenerator annotation.

Would love to be able to do 

```php
/**
 * @ORM\CustomIdGenerator(service="my.id.generator")
 */
public $id;
```

Or something along those lines.

nicolas-grekas:
Fixes #759

This PR enables referencing services when using the `@CustomIdGenerator()` annotation.

It works by creating a service locator of all services tagged as `doctrine.id_generator`, and checking that locator when a custom id generator is found. Usage:

```php
/**
 * @ORM\GeneratedValue(strategy="CUSTOM")
 * @ORM\CustomIdGenerator('doctrine.uuid_generator')
 */
```

### DIFF
--- a/DependencyInjection/Compiler/IdGeneratorPass.php
+++ b/DependencyInjection/Compiler/IdGeneratorPass.php
@@ -1,0 +1,75 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
+
+use Doctrine\Bundle\DoctrineBundle\Mapping\ClassMetadataFactory;
+use Doctrine\Bundle\DoctrineBundle\Mapping\MappingDriver;
+use Doctrine\ORM\Mapping\ClassMetadataFactory as ORMClassMetadataFactory;
+use Symfony\Component\DependencyInjection\Alias;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\Compiler\ServiceLocatorTagPass;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+final class IdGeneratorPass implements CompilerPassInterface
+{
+    const ID_GENERATOR_TAG  = 'doctrine.id_generator';
+    const CONFIGURATION_TAG = 'doctrine.orm.configuration';
+
+    public function process(ContainerBuilder $container): void
+    {
+        $generatorIds = array_keys($container->findTaggedServiceIds(self::ID_GENERATOR_TAG));
+
+        // when ORM is not enabled
+        if (! $container->hasDefinition('doctrine.orm.configuration') || ! $generatorIds) {
+            return;
+        }
+
+        $generatorRefs = array_map(static function ($id) {
+            return new Reference($id);
+        }, $generatorIds);
+
+        $ref = ServiceLocatorTagPass::register($container, array_combine($generatorIds, $generatorRefs));
+        $container->setAlias('doctrine.id_generator_locator', new Alias((string) $ref, false));
+
+        foreach ($container->findTaggedServiceIds(self::CONFIGURATION_TAG) as $id => $tags) {
+            $configurationDef   = $container->getDefinition($id);
+            $methodCalls        = $configurationDef->getMethodCalls();
+            $metadataDriverImpl = null;
+
+            foreach ($methodCalls as $i => [$method, $arguments]) {
+                if ($method === 'setMetadataDriverImpl') {
+                    $metadataDriverImpl = (string) $arguments[0];
+                }
+
+                if ($method !== 'setClassMetadataFactoryName') {
+                    continue;
+                }
+
+                if ($arguments[0] !== ORMClassMetadataFactory::class && $arguments[0] !== ClassMetadataFactory::class) {
+                    $class = $container->getReflectionClass($arguments[0]);
+
+                    if ($class && $class->isSubclassOf(ClassMetadataFactory::class)) {
+                        break;
+                    }
+
+                    continue 2;
+                }
+
+                $methodCalls[$i] = ['setClassMetadataFactoryName', [ClassMetadataFactory::class]];
+            }
+
+            if ($metadataDriverImpl === null) {
+                continue;
+            }
+
+            $configurationDef->setMethodCalls($methodCalls);
+            $container->register('.' . $metadataDriverImpl, MappingDriver::class)
+                ->setDecoratedService($metadataDriverImpl)
+                ->setArguments([
+                    new Reference(sprintf('.%s.inner', $metadataDriverImpl)),
+                    new Reference('doctrine.id_generator_locator'),
+                ]);
+        }
+    }
+}
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -5,6 +5,7 @@ namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection;
 use Doctrine\Bundle\DoctrineBundle\Command\Proxy\ImportDoctrineCommand;
 use Doctrine\Bundle\DoctrineBundle\Dbal\ManagerRegistryAwareConnectionProvider;
 use Doctrine\Bundle\DoctrineBundle\Dbal\RegexSchemaAssetFilter;
+use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\IdGeneratorPass;
 use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\ServiceRepositoryCompilerPass;
 use Doctrine\Bundle\DoctrineBundle\EventSubscriber\EventSubscriberInterface;
 use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepositoryInterface;
@@ -15,10 +16,13 @@ use Doctrine\DBAL\SQLParserUtils;
 use Doctrine\DBAL\Tools\Console\Command\ImportCommand;
 use Doctrine\DBAL\Tools\Console\ConnectionProvider;
 use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Id\AbstractIdGenerator;
 use Doctrine\ORM\Proxy\Autoloader;
 use Doctrine\ORM\UnitOfWork;
 use LogicException;
 use Symfony\Bridge\Doctrine\DependencyInjection\AbstractDoctrineExtension;
+use Symfony\Bridge\Doctrine\IdGenerator\UlidGenerator;
+use Symfony\Bridge\Doctrine\IdGenerator\UuidGenerator;
 use Symfony\Bridge\Doctrine\Messenger\DoctrineClearEntityManagerWorkerSubscriber;
 use Symfony\Bridge\Doctrine\Messenger\DoctrineTransactionMiddleware;
 use Symfony\Bridge\Doctrine\PropertyInfo\DoctrineExtractor;
@@ -405,6 +409,14 @@ class DoctrineExtension extends AbstractDoctrineExtension
             $container->removeDefinition('doctrine.orm.listeners.pdo_cache_adapter_doctrine_schema_subscriber');
         }
 
+        if (! class_exists(UlidGenerator::class)) {
+            $container->removeDefinition('doctrine.ulid_generator');
+        }
+
+        if (! class_exists(UuidGenerator::class)) {
+            $container->removeDefinition('doctrine.uuid_generator');
+        }
+
         $entityManagers = [];
         foreach (array_keys($config['entity_managers']) as $name) {
             $entityManagers[$name] = sprintf('doctrine.orm.%s_entity_manager', $name);
@@ -459,6 +471,9 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $container->registerForAutoconfiguration(EventSubscriberInterface::class)
             ->addTag('doctrine.event_subscriber');
 
+        $container->registerForAutoconfiguration(AbstractIdGenerator::class)
+            ->addTag(IdGeneratorPass::ID_GENERATOR_TAG);
+
         /**
          * @see DoctrineBundle::boot()
          */
@@ -477,6 +492,7 @@ class DoctrineExtension extends AbstractDoctrineExtension
     protected function loadOrmEntityManager(array $entityManager, ContainerBuilder $container)
     {
         $ormConfigDef = $container->setDefinition(sprintf('doctrine.orm.%s_configuration', $entityManager['name']), new ChildDefinition('doctrine.orm.configuration'));
+        $ormConfigDef->addTag(IdGeneratorPass::CONFIGURATION_TAG);
 
         $this->loadOrmEntityManagerMappingInformation($entityManager, $ormConfigDef, $container);
         $this->loadOrmCacheDrivers($entityManager, $container);
--- a/Mapping/MappingDriver.php
+++ b/Mapping/MappingDriver.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\Mapping;
+
+use Doctrine\ORM\Mapping\ClassMetadataInfo;
+use Doctrine\Persistence\Mapping\ClassMetadata;
+use Doctrine\Persistence\Mapping\Driver\MappingDriver as MappingDriverInterface;
+use Psr\Container\ContainerInterface;
+
+class MappingDriver implements MappingDriverInterface
+{
+    /** @var MappingDriverInterface */
+    private $driver;
+
+    /** @var ContainerInterface */
+    private $idGeneratorLocator;
+
+    public function __construct(MappingDriverInterface $driver, ContainerInterface $idGeneratorLocator)
+    {
+        $this->driver             = $driver;
+        $this->idGeneratorLocator = $idGeneratorLocator;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getAllClassNames()
+    {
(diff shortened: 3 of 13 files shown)


ostrolucky:
You are changing this a lot. We will review after you are done. We will also need tests in the end. But feature is welcome.

nicolas-grekas:
I was fighting the CS checker ;)
Should be good for a first round!

Dbal/Configuration.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\Dbal;
```
stof: should not be in a `Dbal` namespace as it extends the *ORM* configuration class.

stof (changes requested):

Dbal/Configuration.php
nicolas-grekas: moved to a new `ORM` namespace

nicolas-grekas:
PR is now tested and ready for review.

DependencyInjection/DoctrineExtension.php
```diff
@@ -477,6 +492,7 @@ protected function ormLoad(array $config, ContainerBuilder $container)
     protected function loadOrmEntityManager(array $entityManager, ContainerBuilder $container)
     {
         $ormConfigDef = $container->setDefinition(sprintf('doctrine.orm.%s_configuration', $entityManager['name']), new ChildDefinition('doctrine.orm.configuration'));
+        $ormConfigDef->addTag(IdGeneratorPass::CONFIGURATION_TAG);
```
ostrolucky, line 495: This should be in doctrine.orm.configuration xml definition preferably. Are you trying to make this internal? Then I suggest to just add comment that it is internal inside xml.

Mapping/MappingDriver.php
```diff
@@ -0,0 +1,73 @@
...
+
+        if (
+            ! $this->acceptIdGeneratorInstance()
+            || $metadata->generatorType !== ClassMetadataInfo::GENERATOR_TYPE_CUSTOM
+            || ! $this->idGeneratorLocator->has($metadata->customGeneratorDefinition['class'] ?? '')
+        ) {
+            return;
+        }
+
+        $metadata->customGeneratorDefinition['instance'] = $this->idGeneratorLocator->get($metadata->customGeneratorDefinition['class'] ?? '');
```
ostrolucky: null coalescence operator is unneeded here

Resources/config/orm.xml
```diff
@@ -213,5 +213,18 @@
...
+            <argument type="service" id="doctrine.ulid_factory" on-invalid="ignore" />
+            <tag name="doctrine.id_generator" />
+        </service>
+
+        <service id="doctrine.uuid_generator" class="Symfony\Bridge\Doctrine\IdGenerator\UuidGenerator" public="false">
+            <argument type="service" id="doctrine.uuid_factory" on-invalid="ignore" />
+            <tag name="doctrine.id_generator" />
+        </service>
+
+        <service id="doctrine.ulid_factory" alias="ulid.factory" public="false" />
```
ostrolucky: Where are definitions for these services? Where are their classes defined?

Mapping/MappingDriver.php
```diff
@@ -0,0 +1,73 @@
...
+    /** @var bool */
+    private $acceptIdGeneratorInstance = false;
+
+    public function __construct(MappingDriverInterface $driver, ContainerInterface $idGeneratorLocator)
+    {
+        $this->driver             = $driver;
+        $this->idGeneratorLocator = $idGeneratorLocator;
+    }
+
+    public function acceptIdGeneratorInstance(?bool $accept = null): bool
```
ostrolucky: Performance improvement you are going for is not worth it, drop the method please. Checking if uuid type is custom and container has the service is completely fine on its own without this brittle stateful mechanism.

Mapping/MappingDriver.php
```diff
@@ -0,0 +1,73 @@
...
+     * {@inheritDoc}
+     */
+    public function loadMetadataForClass($className, ClassMetadata $metadata): void
+    {
+        $this->driver->loadMetadataForClass($className, $metadata);
+
+        if (
+            ! $this->acceptIdGeneratorInstance()
+            || $metadata->generatorType !== ClassMetadataInfo::GENERATOR_TYPE_CUSTOM
+            || ! $this->idGeneratorLocator->has($metadata->customGeneratorDefinition['class'] ?? '')
```
ostrolucky: It's impossible to define service with blank id, so passing empty string to has method is weird. Instead of being elegant this will mislead people. Use separate isset call please. That better represents intention.

Mapping/MappingDriver.php
ostrolucky: We are breaking the interface here. It's explicitly marked as a read only field. ORM should be changed first to allow this.

DependencyInjection/Compiler/IdGeneratorPass.php
```diff
@@ -0,0 +1,75 @@
...
+    const ID_GENERATOR_TAG  = 'doctrine.id_generator';
+    const CONFIGURATION_TAG = 'doctrine.orm.configuration';
+
+    public function process(ContainerBuilder $container): void
+    {
+        $generatorIds     = array_keys($container->findTaggedServiceIds(self::ID_GENERATOR_TAG));
+        $configurationIds = array_keys($container->findTaggedServiceIds(self::CONFIGURATION_TAG));
+
+        foreach ($configurationIds as $id) {
+            $container->getDefinition($id)->clearTag(self::CONFIGURATION_TAG);
```
ostrolucky: Not worth confusion. When dumping container, devs expect to see tags that were defined.

DependencyInjection/Compiler/IdGeneratorPass.php
```diff
@@ -0,0 +1,75 @@
...
+
+        $generatorRefs = array_map(static function ($id) {
+            return new Reference($id);
+        }, $generatorIds);
+
+        $ref = ServiceLocatorTagPass::register($container, array_combine($generatorIds, $generatorRefs));
+        $container->setAlias('doctrine.id_generator_locator', new Alias((string) $ref, false));
+
+        foreach ($configurationIds as $id) {
+            $methodCalls        = $container->getDefinition($id)->getMethodCalls();
```
ostrolucky: We need result of `$container->getDefinition($id)` later. Instead of calling it again, let's extract it to variable.

DependencyInjection/Compiler/IdGeneratorPass.php
```diff
@@ -0,0 +1,75 @@
...
+            foreach ($methodCalls as $i => [$method, $arguments]) {
+                if ($method === 'setMetadataDriverImpl') {
+                    $metadataDriverImpl = (string) $arguments[0];
+                }
+
+                if ($method !== 'setClassMetadataFactoryName') {
+                    continue;
+                }
+
+                if ($arguments[0] !== ORMClassMetadataFactory::class && $arguments[0] !== ClassMetadataFactory::class) {
```
ostrolucky, line 49: What's the purpose of this block? Body looks fishy to me and it's not covered by tests.

DependencyInjection/Compiler/IdGeneratorPass.php
```diff
@@ -0,0 +1,75 @@
...
+
+                    continue 2;
+                }
+
+                $methodCalls[$i] = ['setClassMetadataFactoryName', [ClassMetadataFactory::class]];
+            }
+
+            $container->getDefinition($id)->setMethodCalls($methodCalls);
+            $container->register('.' . $metadataDriverImpl, MappingDriver::class)
+                ->setDecoratedService($metadataDriverImpl)
```
ostrolucky, line 68: What if `$metadataDriverImpl` is null?

Tests/CustomIdGeneratorTest.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        if (interface_exists(EntityManagerInterface::class)) {
+            return;
+        }
+
+        self::markTestSkipped('This test requires ORM');
+    }
+
+    /**
+     * https://github.com/doctrine/orm/pull/7953 needed, otherwise ORM classes we define services for trigger deprecations
```
ostrolucky: C&P? Referenced PR was merged already.

Tests/CustomIdGeneratorTest.php
```diff
@@ -0,0 +1,90 @@
...
+                            'dir' => __DIR__ . '/DependencyInjection/Fixtures/Bundles/AnnotationsBundle/Entity',
+                            'prefix' => 'Fixtures\Bundles\AnnotationsBundle\Entity',
+                        ],
+                    ],
+                ],
+            ],
+        ], $container);
+
+        $def = $container->register('my_id_generator', CustomIdGenerator::class)
+            ->setPublic(false);
```
ostrolucky: Why is this call needed? It should be private by default already.

Tests/CustomIdGeneratorTest.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * https://github.com/doctrine/orm/pull/7953 needed, otherwise ORM classes we define services for trigger deprecations
+     *
+     * @group legacy
+     */
+    public function testRepositoryServiceWiring(): void
+    {
+        $container = new ContainerBuilder(new ParameterBag([
```
ostrolucky, line 30: Please use TestKernel instead

Resources/config/orm.xml
```diff
@@ -213,5 +213,18 @@
 
             <tag name="console.command" command="doctrine:mapping:import" />
         </service>
+
+        <service id="doctrine.ulid_generator" class="Symfony\Bridge\Doctrine\IdGenerator\UlidGenerator" public="false">
+            <argument type="service" id="doctrine.ulid_factory" on-invalid="ignore" />
```
ostrolucky: I'm confused, there is no constructor in UlidGenerator, so why are we defining arguments? Same for UuidGenerator

ostrolucky (changes requested):
So far looks like a can of worms to me without having https://github.com/doctrine/orm/issues/7134

DependencyInjection/DoctrineExtension.php
nicolas-grekas: Tags are not inherited so adding it to the XML won't help here. Also yes, this is mostly internal, flagging some services so that IdGeneratorPass can collect them easily.

Tests/CustomIdGeneratorTest.php
nicolas-grekas: the bundle still supports 3.4, where services are not yet private by default

nicolas-grekas:
Thanks for the review @ostrolucky. I fixed the most technical comments. I'll let you know when I'm done with the remaining ones.

DependencyInjection/Compiler/IdGeneratorPass.php
nicolas-grekas: The purpose of this block is to *not* wire the `MappingDriver` decorators if the `ClassMetadataFactory` doesn't know how to handle the `instance` key in the `$customGeneratorDefinition`. But I'm fine removing the check if you prefer.

Resources/config/orm.xml
nicolas-grekas: I'm removing this and will submit as a separate PR, when the corresponding factories are merged into Symfony.

Tests/CustomIdGeneratorTest.php
nicolas-grekas: I'm not sure I get what you mean here. Note that I borrowed this from `ServiceRepositoryTest`.

Tests/CustomIdGeneratorTest.php
nicolas-grekas: removed

Tests/CustomIdGeneratorTest.php
ostrolucky: I meant to use this https://github.com/doctrine/DoctrineBundle/blob/3a88af21cda5796fa252d4e94ee286ab3da9051b/Tests/DependencyInjection/Fixtures/TestKernel.php#L12-L12 But I noticed now you are registering custom service here, so further adjustments would be need. Ok, leave it.

nicolas-grekas: OK, thanks. Review addressed then. Let me know if you have any other comment.

nicolas-grekas: OK, thanks. Review addressed then. Let me know if you have any other comments.

nicolas-grekas:
Now with a doc page. Ready for another review.

ostrolucky (approved):

Resources/config/orm.xml
```diff
@@ -150,6 +150,17 @@
...
+        <!-- custom id generators -->
+        <service id="doctrine.ulid_generator" class="Symfony\Bridge\Doctrine\IdGenerator\UlidGenerator" public="false">
+            <argument type="service" id="ulid.factory" on-invalid="ignore" />
+            <tag name="doctrine.id_generator" />
+        </service>
+
+        <service id="doctrine.uuid_generator" class="Symfony\Bridge\Doctrine\IdGenerator\UuidGenerator" public="false">
+            <argument type="service" id="uuid.factory" on-invalid="ignore" />
+            <tag name="doctrine.id_generator" />
+        </service>
```
nicolas-grekas, line 162: added back now that https://github.com/symfony/symfony/pull/39507 is merged

Resources/doc/custom-id-generators.rts
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * @ORM\Entity
+     */
+    class User
+    {
+        /**
+         * @Id
+         * @Column(type="integer")
+         * @ORM\GeneratedValue(strategy="CUSTOM")
+         * @ORM\CustomIdGenerator('my_id_generator')
```
alcaeus: I may be missing something, but `my_id_generator` does not match the config below, which uses the FQCN as identifier for the service.
```suggestion
         * @ORM\CustomIdGenerator(MyIdGenerator::class)
```

Resources/doc/custom-id-generators.rts
```diff
@@ -0,0 +1,64 @@
...
+
+Starting with Doctrine bundle 2.3, the ``CustomIdGenerator`` annotation can be
+used to reference any services tagged with the ``doctrine.id_generator`` tag.
+
+.. code-block:: php
+
+    <?php
+    // User.php
+
+    use Doctrine\ORM\Mapping as ORM;
```
alcaeus: Only relevant if this should indeed use the FQCN as service name:

```suggestion
    use App\MyIdGenerator;
    use Doctrine\ORM\Mapping as ORM;
```

Resources/doc/custom-id-generators.rts
```diff
@@ -0,0 +1,64 @@
+Custom ID Generators
```
alcaeus: The file has the extension `rts`, where it should be `rst`.

Tests/CustomIdGeneratorTest.php
alcaeus: symfony/dependency-injection 3.4 is not supported, so I believe this could indeed be dropped. That said, we do support 3.4 for some other components, but that should not interfere here.

DependencyInjection/Compiler/IdGeneratorPass.php
```diff
@@ -0,0 +1,75 @@
...
+
+                $methodCalls[$i] = ['setClassMetadataFactoryName', [ClassMetadataFactory::class]];
+            }
+
+            if ($metadataDriverImpl === null) {
+                continue;
+            }
+
+            $configurationDef->setMethodCalls($methodCalls);
+            $container->register('.' . $metadataDriverImpl, MappingDriver::class)
```
alcaeus, line 67: Is prefixing with a single dot a convention? I'm wondering if there could be conflicts if other bundles use the same naming logic.

alcaeus (changes requested):
This is a great addition that makes working with ID generators in Symfony so much easier, thank you very much @nicolas-grekas!

I've left some comments, but in general this LGTM!

Resources/doc/custom-id-generators.rts
nicolas-grekas: I was updating the doc while you were reviewing :)

Resources/doc/custom-id-generators.rts
nicolas-grekas: I removed this example actually now :)

DependencyInjection/Compiler/IdGeneratorPass.php
nicolas-grekas: The `.` prefix makes the service "hidden". I don't think this can collide...

alcaeus: Gotcha 👍

Resources/doc/custom-id-generators.rts
alcaeus: So I noticed - I actually liked the example with configuration.

Resources/doc/custom-id-generators.rts
nicolas-grekas: good catch!

Resources/doc/custom-id-generators.rts
nicolas-grekas: I don't think the job of the bundle's doc is to tell how to add a tag, especially when the tag doesn't need attribute and autoconfiguration is good enough (eg for event listeners, the example make sense, because of the attributes)

alcaeus: Sounds good 👍

alcaeus (approved):

nicolas-grekas:
Am I allowed to merge now that the two of you approved the PR?

Issue #1282, opened 2021-01-27
doctrine:query:sql broken after update from 2.2.2 to 2.2.3


bicpi:
Hi,
our config defines multiple database connections and we have a script in our CI that uses 

    $ bin/console doctrine:query:sql --connection $CON "SELECT 1"

to check the availability of the database before starting some tests.

This is working with version 2.2.2 but is broken with 2.2.3, we see this error:

```
ERROR     [console] Error thrown while running command "doctrine:query:sql --connection conn_name 'SELECT 1'". Message: "An exception occurred in driver: SQLSTATE[HY000] [2002] No such file or directory" ["exception" => Doctrine\DBAL\Exception\ConnectionException^ { …},"command" => "doctrine:query:sql --connection affiliate 'SELECT 1'","message" => "An exception occurred in driver: SQLSTATE[HY000] [2002] No such file or directory"] ["version" => "{VERSION}","source_env" => "local-dev","process_id" => 365]

In AbstractMySQLDriver.php line 112:

  An exception occurred in driver: SQLSTATE[HY000] [2002] No such file or directory


In Exception.php line 18:

  SQLSTATE[HY000] [2002] No such file or directory


In PDOConnection.php line 38:

  SQLSTATE[HY000] [2002] No such file or directory


doctrine:query:sql [--connection CONNECTION] [--depth DEPTH] [--force-fetch] [--] <sql>
```

It seems we cannot connect to the database any more (but it's there). Downgrading to 2.2.2 fixes the problem so I assume this to be a bug that was introduced with the fix from https://github.com/doctrine/DoctrineBundle/issues/1271.

As the command is deprecated we were able to improve and fix our concrete use case by using `doctrine-dbal:query` instead while still using 2.2.3. But it seems like a BC break in the patch version?

ostrolucky:
Hmm yeah this most likely breaks for users using dbal:>=2.11,<3.0

bicpi:
Our `doctrine/dbal` version is 2.12.1 still, yes.

ostrolucky:
https://github.com/doctrine/DoctrineBundle/pull/1285

ostrolucky:
Fix merged to 2.2.x branch

PR #1285, opened 2021-01-30
#1282 Fix DBAL < 2.11 BC layer in doctrine:query:sql

ostrolucky:
This BC layer didn't work correctly with dbal:^2.11,<3.0 when --connection option was used


dmaicher (approved):
tested on my app with dbal 2.10 and 2.11+. Fixes the issue :+1:

Issue #1176, opened 2020-06-07
Rename "BlacklistSchemaAssetFilter" to match "WellKnownSchemaFilterPass"


albe:
It just makes more sense. I suggest using "UnknownSchemaAssetFilter" instead - or probably even better "WellKnownSchemaAssetFilter", since it actually filters out well known assets. In case the filter itself should be more generic, "ExcludedSchemaAssetFilter" could work well, too.

TBD: Only change this in master or target lower. AFAIS it would only be breaking for 3rd parties that actually extend `BlacklistSchemaAssetFilter` or directly use it (could possibly be leviated by keeping an alias class around), but then this is a "glue" bundle, not a framework, so is this even API?

greg0ire:
The list is passed dynamically rather than being harcoded, so the implementation could be used for something else entirely, which would make me lean towards `ExcludedSchemaAssetFilter`… but it does not convey the meaning as well as the original term IMO. How about `ExcludeListSchemaAssetFilter`?

ostrolucky:
I am not sure if this feature is even used, I don't see this connected to anything. This comment kinda confirms it https://github.com/symfony/symfony/pull/36655#issuecomment-622959966. @weaverryan do you know about anything?

greg0ire:
e4ca3d2c7f5f9076b228d0762ee48ae2b1090340

ostrolucky:
Alright, it's needed then. But I actually like current name. As @greg0ire mentioned already, it doesn't filter out well known stuff, it's pretty generic and can be used for anything. Suggested alternative `ExcludeListSchemaAssetFilter` is just awkward. So I would say no to this proposal.

greg0ire:
> But I actually like current name. 

You might not have seen [this tweet](https://twitter.com/leahculver/status/1269109776983547904) that shows that some other people don't like it at all, probably because it strongly connects the color "black" with exclusion, which, when you think of it, doesn't make much sense. Maybe we can find a word that makes everyone happier _and_ makes things clearer in the process.

albe:
> lean towards ExcludedSchemaAssetFilter… but it does not convey the meaning as well as the original term IMO. How about ExcludeListSchemaAssetFilter?

Hmm. Why does it not convey the meaning as well for you? Is it due to the active terminology? `ExcludeSchemaAssetFilter` could work in that case.

Looking at what it does, I would phrase it something like this: "A filter that skips specific assets for a given schema". `SkipSchemaAssetFilter`?

[...]

PR #1286, opened 2021-01-30
#1176 Deprecate (Blacklist|WellKnown) schema filters 

ostrolucky:


nicolas-grekas:
Should we unregister the pass in DoctrineBundle?
I don't think it provides many benefits now, isn't it?

ostrolucky:
In linked issue @greg0ire pointed out this is still being used by eg. Lock component. Is that no longer the case and there is no longer any reliance from Symfony on these? If no, only then I'm ok to unregister the pass.

ostrolucky:
Alright, enough time passed. I even pinged @weaverryan on Slack. Merging.

weaverryan:
Sorry for not being on time. Here's the related PR: https://github.com/symfony/symfony/pull/36655

So Messenger & Cache is no longer needed. For Lock & Session, those still need more work. We *should* really implement those correctly (so that they can be fully removed here)... but it's not done yet.

Btw, in this PR, we deprecate `WellKnownSchemaFilterPass`. Will this trigger deprecations in all projects since it's still being used in DoctrineBundle? And if I do see this deprecation, how should I update my code to remove it?

Thanks!

ostrolucky:
> Btw, in this PR, we deprecate WellKnownSchemaFilterPass. Will this trigger deprecations in all projects since it's still being used in DoctrineBundle? And if I do see this deprecation, how should I update my code to remove it?

I don't know. But if it does, I don't see how can we fix this here in doctrine-bundle. We need to mark those classes as deprecated, but at the same time we should internally keep compiler pass registered for BC. In such case I think it would be best if symfony didn't report such deprecations, only if users outside of this bundle are instantiating deprecated compiler pass class.

nicolas-grekas:
Symfony won't report it because there is no runtime deprecation. A static analyzer will of course.

PR #1287, opened 2021-01-31
Use ramsey/composer-install

greg0ire:
It makes for shorter workflow files.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -25,30 +25,36 @@ jobs:
           - "7.2"
           - "7.3"
           - "7.4"
-        deps:
-          - "normal"
+        dependencies:
+          - "highest"
+        stability:
+          - "stable"
         symfony-require:
           - ""
         symfony-deprecations-helper:
           - ""
         include:
           # Tests the lowest set of dependencies
-          - deps: "low"
+          - dependencies: "lowest"
+            stability: "stable"
             php-version: "7.1"
             symfony-deprecations-helper: "weak"
 
           # Test against latest Symfony 4.4 dev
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -75,31 +78,22 @@ jobs:
...
+      - name: "Enforce using stable dependencies"
+        run: "composer config minimum-stability stable"
+        if: "${{ matrix.stability == 'stable' }}"
 
-      - name: "Install dev dependencies with composer"
-        run: "composer update --no-interaction --prefer-dist"
-        if: "${{ matrix.deps == 'dev' }}"
-
-      - name: "Install lowest possible dependencies with composer"
-        run: "composer update --no-interaction --prefer-dist --prefer-stable --prefer-lowest"
```
ostrolucky, line 101: --prefer-lowest job needs to exist

ostrolucky (changes requested):
CI is failing

.github/workflows/continuous-integration.yml
greg0ire: It does exist :)

.github/workflows/continuous-integration.yml
```diff
@@ -25,30 +25,36 @@ jobs:
...
+        stability:
+          - "stable"
         symfony-require:
           - ""
         symfony-deprecations-helper:
           - ""
         include:
           # Tests the lowest set of dependencies
-          - deps: "low"
+          - dependencies: "lowest"
```
greg0ire, line 38: @ostrolucky here is why it exists

ostrolucky (approved):

PR #1288, opened 2021-02-01
add ability to override url dbname

jrushlow:
Use case:

If you define a connection `url` in Symfony using the `DATABASE_URL` env var for development, and you use the same server, but different database for tests - you have to define the entire `DATABASE_URL` in your `config/packages/test/doctrine.yaml`

This PR would allow you to override the database used in the connection `url` by setting `dbname` in your `config/packages/{ENV}/doctrine.yaml`. To preserve BC, we would add a `doctrine.dbal.override_url` `bool` configuration param that defaults to `false`

Example:

```
// config/packages/doctrine.yaml

doctrine:
    dbal:
        url: mysql://user:pass@host/main_database
```

```
// config/packages/test/doctrine.yaml

doctrine:
    dbal:
        override_url: true
        dbname: main_test
```

In the `test` environment:
`doctrine.dbal.url === mysql://user:pass@host/main_test`

else

`doctrine.dbal.url === mysql://user:pass@host/main_database`

### DIFF
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -43,10 +43,6 @@ class ConnectionFactory
             $this->initializeTypes();
         }
 
-        if (isset($params['override_url'], $params['url']) && $params['override_url']) {
-            $params['url'] = $this->overrideUrl($params);
-        }
-
         if (! isset($params['pdo']) && ! isset($params['charset'])) {
             $wrapperClass = null;
             if (isset($params['wrapperClass'])) {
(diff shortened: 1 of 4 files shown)


stof:
@jrushlow why was this closed without any comment ?

jrushlow:
@stof replaced w/ #1290 - i didnt have the implementation quite right and made a git mess locally.. started off fresh to avoid headaches

PR #1290, opened 2021-02-04
add ability to override url with dsn params

jrushlow:
Use case:

If you define a connection `url` in Symfony using a real `DATABASE_URL` env var for development, and you use the same server, but different database for tests - you have to override the *entire* `DATABASE_URL` in your `config/packages/test/doctrine.yaml` file.

This PR would allow you to override part of the `url` using any combination of `dbname`, `host`, `port`, `user`, and/or `password`. To preserve BC, we would add a `doctrine.dbal.override_url` `bool` configuration param. 

Example:

```
// config/packages/doctrine.yaml

doctrine:
    dbal:
        url: '%env(DATABASE_URL)%'
```

```
// config/packages/test/doctrine.yaml

doctrine:
    dbal:
        override_url: true
        dbname: main_test
```

Assuming there is a `DATABASE_DSN` env var set to `mysql://user:pass@host/main_database`, then in the `test` environment it would still use `mysql://user:pass@host` but with the `main_test` database.

### DIFF
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -45,10 +45,10 @@ class ConnectionFactory
             $this->initializeTypes();
         }
 
-        $connectionOverrideOptions = $params['connection_override_options'] ?? [];
+        $overriddenOptions = $params['connection_override_options'] ?? [];
         unset($params['connection_override_options']);
 
-        if (! isset($params['pdo']) && (! isset($params['charset']) || (! empty($connectionOverrideOptions) && isset($params['url'])))) {
+        if (! isset($params['pdo']) && (! isset($params['charset']) || $overriddenOptions)) {
             $wrapperClass = null;
 
             if (isset($params['wrapperClass'])) {
@@ -65,7 +65,7 @@ class ConnectionFactory
             }
 
             $connection = DriverManager::getConnection($params, $config, $eventManager);
-            $params     = array_merge($connection->getParams(), $connectionOverrideOptions);
+            $params     = array_merge($connection->getParams(), $overriddenOptions);
             $driver     = $connection->getDriver();
 
             if ($driver instanceof AbstractMySQLDriver) {
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -45,8 +45,12 @@ class ConnectionFactory
             $this->initializeTypes();
         }
 
-        if (! isset($params['pdo']) && ! isset($params['charset'])) {
+        $connectionOverrideOptions = $params['connection_override_options'] ?? [];
+        unset($params['connection_override_options']);
+
+        if (! isset($params['pdo']) && (! isset($params['charset']) || (! empty($connectionOverrideOptions) && isset($params['url'])))) {
             $wrapperClass = null;
+
             if (isset($params['wrapperClass'])) {
                 if (! is_subclass_of($params['wrapperClass'], Connection::class)) {
                     if (class_exists(DBALException::class)) {
@@ -61,7 +65,7 @@ class ConnectionFactory
             }
 
             $connection = DriverManager::getConnection($params, $config, $eventManager);
-            $params     = $connection->getParams();
+            $params     = array_merge($connection->getParams(), $connectionOverrideOptions);
             $driver     = $connection->getDriver();
 
             if ($driver instanceof AbstractMySQLDriver) {
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -319,10 +319,10 @@ class Configuration implements ConfigurationInterface
             ->end()
             ->beforeNormalization()
                 ->ifTrue(static function ($v) {
-                    return ! ($v['override_url'] ?? false) && isset($v['url']);
+                    return empty($v['override_url']) && isset($v['url']);
                 })
                 ->then(static function ($v) {
-                    trigger_error('Not setting doctrine.dbal.override_url to true is deprecated. True is the only value that will be supported in doctrine-bundle 3.0.', E_USER_DEPRECATED);
+                    @trigger_error('Not setting doctrine.dbal.override_url to true is deprecated. True is the only value that will be supported in doctrine-bundle 3.0.', E_USER_DEPRECATED);
 
                     return $v;
                 })
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -217,10 +217,11 @@ class Configuration implements ConfigurationInterface
             ->children()
                 ->scalarNode('url')->info('A URL with connection information; any parameter value parsed from this string will override explicitly set parameters')->end()
                 ->scalarNode('dbname')->end()
-                ->scalarNode('host')->defaultValue('localhost')->end()
-                ->scalarNode('port')->defaultNull()->end()
-                ->scalarNode('user')->defaultValue('root')->end()
-                ->scalarNode('password')->defaultNull()->end()
+                ->scalarNode('host')->info('Defaults to "localhost" at runtime.')->end()
+                ->scalarNode('port')->info('Defaults to null at runtime.')->end()
+                ->scalarNode('user')->info('Defaults to "root" at runtime.')->end()
+                ->scalarNode('password')->info('Defaults to null at runtime.')->end()
+                ->booleanNode('override_url')->defaultValue(false)->info('Allows overriding parts of the "url" parameter with dbname, host, port, user, and/or password parameters.')->end()
                 ->scalarNode('application_name')->end()
                 ->scalarNode('charset')->end()
                 ->scalarNode('path')->end()
@@ -312,6 +313,16 @@ class Configuration implements ConfigurationInterface
                 ->then(static function ($v) {
                     $v['MultipleActiveResultSets'] = $v['multiple_active_result_sets'];
                     unset($v['multiple_active_result_sets']);
+
+                    return $v;
+                })
+            ->end()
+            ->beforeNormalization()
+                ->ifTrue(static function ($v) {
+                    return ! ($v['override_url'] ?? false) && isset($v['url']);
+                })
+                ->then(static function ($v) {
+                    trigger_error('Not setting doctrine.dbal.override_url to true is deprecated. True is the only value that will be supported in doctrine-bundle 3.0.', E_USER_DEPRECATED);
 
                     return $v;
                 })
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -262,35 +262,26 @@ class DoctrineExtension extends AbstractDoctrineExtension
 
     protected function getConnectionOptions($connection)
     {
-        $options = $connection;
-
-        $options['connection_override_options'] = [];
+        $options = ['connection_override_options' => []] + $connection;
 
         $connectionDefaults = [
-            'dbname' => null,
             'host' => 'localhost',
             'port' => null,
             'user' => 'root',
             'password' => null,
         ];
 
-        if (($options['override_url'] ?? false)) {
-            $options['connection_override_options'] = array_intersect_key($options, $connectionDefaults);
+        if ($options['override_url']) {
+            $options['connection_override_options'] = array_intersect_key($options, ['dbname' => null] + $connectionDefaults);
         }
 
-        unset($options['override_url'], $connectionDefaults['dbname']);
+        unset($options['override_url']);
 
-        foreach ($connectionDefaults as $defaultKey => $defaultValue) {
-            if (! array_key_exists($defaultKey, $options)) {
-                $options[$defaultKey] = $defaultValue;
-            }
+        $options += $connectionDefaults;
 
-            foreach (['shards', 'replicas', 'slaves'] as $connectionKey) {
-                if (! array_key_exists($connectionKey, $options)) {
-                    continue;
-                }
-
-                $options[$connectionKey] = $this->setDefaultOnNestedConnectionOption($options[$connectionKey], $defaultKey, $defaultValue);
+        foreach (['shards', 'replicas', 'slaves'] as $connectionKey) {
+            foreach (array_keys($options[$connectionKey]) as $name) {
+                $options[$connectionKey][$name] += $connectionDefaults;
             }
         }
 
@@ -1020,22 +1011,6 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $container->setDefinition($id, $poolDefinition);
 
         return $id;
-    }
-
-    /**
-     * @return array<string|int, array<string|int, mixed>>
-     */
-    private function setDefaultOnNestedConnectionOption(array $connectionOptions, string $paramName, $defaultValue): array
-    {
-        foreach ($connectionOptions as $connectionKey => $connection) {
-            if (array_key_exists($paramName, $connection)) {
-                continue;
-            }
-
-            $connectionOptions[$connectionKey][$paramName] = $defaultValue;
-        }
-
-        return $connectionOptions;
     }
 
     private function registerMetadataPhpArrayCaching(string $entityManagerName, ContainerBuilder $container): void
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -264,6 +264,36 @@ class DoctrineExtension extends AbstractDoctrineExtension
     {
         $options = $connection;
 
+        $options['connection_override_options'] = [];
+
+        $connectionDefaults = [
+            'dbname' => null,
+            'host' => 'localhost',
+            'port' => null,
+            'user' => 'root',
+            'password' => null,
+        ];
+
+        if (($options['override_url'] ?? false)) {
+            $options['connection_override_options'] = array_intersect_key($options, $connectionDefaults);
+        }
+
+        unset($options['override_url'], $connectionDefaults['dbname']);
+
+        foreach ($connectionDefaults as $defaultKey => $defaultValue) {
+            if (! array_key_exists($defaultKey, $options)) {
+                $options[$defaultKey] = $defaultValue;
+            }
+
+            foreach (['shards', 'replicas', 'slaves'] as $connectionKey) {
+                if (! array_key_exists($connectionKey, $options)) {
+                    continue;
+                }
+
+                $options[$connectionKey] = $this->setDefaultOnNestedConnectionOption($options[$connectionKey], $defaultKey, $defaultValue);
+            }
+        }
+
         if (isset($options['platform_service'])) {
             $options['platform'] = new Reference($options['platform_service']);
             unset($options['platform_service']);
@@ -990,6 +1020,22 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $container->setDefinition($id, $poolDefinition);
 
         return $id;
+    }
+
+    /**
+     * @return array<string|int, array<string|int, mixed>>
+     */
+    private function setDefaultOnNestedConnectionOption(array $connectionOptions, string $paramName, $defaultValue): array
+    {
+        foreach ($connectionOptions as $connectionKey => $connection) {
+            if (array_key_exists($paramName, $connection)) {
+                continue;
+            }
+
+            $connectionOptions[$connectionKey][$paramName] = $defaultValue;
+        }
+
+        return $connectionOptions;
     }
 
     private function registerMetadataPhpArrayCaching(string $entityManagerName, ContainerBuilder $container): void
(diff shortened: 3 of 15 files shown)


weaverryan (approved):
Hi!

👍 from me for the use-case.

This could also be used to power[ParaTest](https://github.com/paratestphp/paratest): you could set the `dbname`  to `main_test_%env(TEST_TOKEN)%` in `config/packages/test/doctrine.yaml`. That would allow you to use your normal database connection params, but toggle the database name for parallel tests.

Cheers!

ostrolucky:
I'm also fine with idea, but I think we should push for this to be a default. Perhaps we should make it deprecated to not set this option to true and in doctrine bundle 4 remove it and just make this behavior default. I don't see a use case currently where it makes sense to define both ways to define database connection options. But that changes with this pr, that's why we should introduce such deprecation together with introducing this.

stof:
or maybe this should be changed in DBAL so that specifying both an option explicitly and the URL makes the option win instead of the URL, removing the need to do such hack in the bundle.

ostrolucky:
yep

weaverryan:
>  but I think we should push for this to be a default. Perhaps we should make it deprecated to not set this option to true and in doctrine bundle 4 remove it and just make this behavior default

I was wondering about this too. Let's add those deprecations

> or maybe this should be changed in DBAL so that specifying both an option explicitly and the URL makes the option win instead of the URL, removing the need to do such hack in the bundle

That would also be awesome. We could do this PR, then see how this could be moved upstream so that - someday - the hackery in the `ConnectionFactory` isn't needed.

stof:
I think the test failures are happening for some cases where the `url` option is not used at all

ConnectionFactory.php
```diff
@@ -45,8 +45,21 @@ public function createConnection(array $params, Configuration $config = null, Ev
             $this->initializeTypes();
         }
 
-        if (! isset($params['pdo']) && ! isset($params['charset'])) {
-            $wrapperClass = null;
+        $allowOverrideUrl = $params['override_url'] ?? false;
+        unset($params['override_url']);
+
+        if (! $allowOverrideUrl) {
```
stof: I would add `&& isset($params['url'])`. there is no point forcing to configure `override_url` if `url` is not used at all.

ConnectionFactory.php
```diff
@@ -45,8 +45,21 @@ public function createConnection(array $params, Configuration $config = null, Ev
...
+            trigger_deprecation(
+                'doctrine/DoctrineBundle',
+                '2.3.0',
+                'Setting doctrine.dbal.override_url to true allows you to override the dsn parameters set in doctrine.dbal.url. This will be the default behavior in future versions.'
+            );
+        }
+
+        if (! isset($params['pdo']) && (! isset($params['charset']) || $allowOverrideUrl)) {
```
stof: the whole `$allowOverrideUrl` should probably be executed only if `$params['url']` is set.

jrushlow: good catch & thank you!

ConnectionFactory.php
jrushlow: > I'm also fine with idea, but I think we should push for this to be a default. Perhaps we should make it deprecated to not set this option to true and in doctrine bundle 4 remove it and just make this behavior default. ...

we would still need to trigger the deprecation here regardless if `$params['url']` is set or not wouldn't we? based off of @ostrolucky idea that is.

jrushlow:
the remaining test failures do not appear to be caused by this PR as I see them in PR #1295 as well

ConnectionFactory.php
stof: well, forcing them to add an option explicitly so that we can change the default does not make sense if they don't use url at all: they would not be affected by the change of default

jrushlow: that makes sense and it's now implemented. thanks

composer.json
```diff
@@ -33,6 +33,7 @@
         "symfony/config": "^4.3.3|^5.0",
         "symfony/console": "^3.4.30|^4.3.3|^5.0",
         "symfony/dependency-injection": "^4.3.3|^5.0",
+        "symfony/deprecation-contracts": "^2.2",
```
ostrolucky: This (out of scope change) is controversial change, as there also exist https://github.com/doctrine/deprecations. Hence I would need to ask you to do the deprecation current way please, until we decide internally which of these 2 libraries to use.

ConnectionFactory.php
```diff
@@ -45,8 +45,21 @@ public function createConnection(array $params, Configuration $config = null, Ev
...
-        if (! isset($params['pdo']) && ! isset($params['charset'])) {
-            $wrapperClass = null;
+        $allowOverrideUrl = $params['override_url'] ?? false;
+        unset($params['override_url']);
+
+        if (! $allowOverrideUrl && isset($params['url'])) {
+            trigger_deprecation(
+                'doctrine/DoctrineBundle',
+                '2.3.0',
+                'Setting doctrine.dbal.override_url to true allows you to override the dsn parameters set in doctrine.dbal.url. This will be the default behavior in future versions.'
```
ostrolucky:
```suggestion
                'Not setting doctrine.dbal.override_url to true is deprecated. True is the only value that will be supported in doctrine-bundle 3.0.'
```

ConnectionFactory.php
```diff
@@ -64,6 +77,24 @@ public function createConnection(array $params, Configuration $config = null, Ev
...
+                $supportedOverrideParams = [
+                    'dbname',
+                    'host',
+                    'port',
+                    'user',
+                    'password',
+                ];
+
+                foreach ($supportedOverrideParams as $paramKey) {
+                    if (! isset($originalParams[$paramKey])) {
```
ostrolucky: This simple algorithm doesn't count with options that don't default to NULL, like `host`.

DependencyInjection/Configuration.php
```diff
@@ -221,6 +221,7 @@ private function configureDbalDriverNode(ArrayNodeDefinition $node): void
                 ->scalarNode('port')->defaultNull()->end()
                 ->scalarNode('user')->defaultValue('root')->end()
                 ->scalarNode('password')->defaultNull()->end()
+                ->booleanNode('override_url')->end()
```
ostrolucky: Weirdly, you added docs in deprecation message instead of here as `->info()`. Please add info here instead.

Tests/DependencyInjection/Fixtures/config/yml/dbal_allow_url_override.yml
```diff
@@ -0,0 +1,3 @@
+doctrine:
+    dbal:
+        override_url: true
```
ostrolucky, line 3: This config is not very useful. At least also add `url` and few other options. You would perhaps find that enabling this option overrides host with `localhost` without specifying `host` option

ostrolucky (changes requested):
- Adding note in UPGRADE-2.3.md required
- Adding docs in configuration.rst required
- Removal of deprecation-contracts required
- Fix for override_url: true in combination with URL and unspecified options that don't default to NULL required

Tests/DependencyInjection/Fixtures/config/yml/dbal_allow_url_override.yml
jrushlow: > You would perhaps find that enabling this option overrides host with `localhost` without specifying `host` option

Yes I just discovered that earlier today and working up a fix now for it.

composer.json
jrushlow: Sure thing. 

For context, I added the `symfony/deprecation-contracts` is its already available as a dependency of one of the other `require`'s

ConnectionFactory.php
jrushlow: fixed

DependencyInjection/Configuration.php
jrushlow: fixed

jrushlow:
this should be ready for another review

ConnectionFactory.php
```diff
@@ -45,8 +45,16 @@ public function createConnection(array $params, Configuration $config = null, Ev
             $this->initializeTypes();
         }
 
-        if (! isset($params['pdo']) && ! isset($params['charset'])) {
+        $connectionOverrideOptions = [];
+
+        if (isset($params['connection_override_options'])) {
+            $connectionOverrideOptions = $params['connection_override_options'];
+            unset($params['connection_override_options']);
```
weaverryan: Could we do:

```php
$connectionOverrideOptions = $params['connection_override_options'] ?? [];
unset($params['connection_override_options']);
```

It's harmless to unset a key that exists.

ConnectionFactory.php
```diff
@@ -45,8 +45,16 @@ public function createConnection(array $params, Configuration $config = null, Ev
...
+        }
+
+        if (! isset($params['pdo']) && (! isset($params['charset']) || (! empty($connectionOverrideOptions) && isset($params['url'])))) {
```
weaverryan: Is that last `isset($params['url'])` needed? I mean, I get why you put it there... but I'm not sure it's strictly necessary.

DependencyInjection/Configuration.php
```diff
@@ -217,10 +217,11 @@ private function configureDbalDriverNode(ArrayNodeDefinition $node): void
...
                 ->scalarNode('url')->info('A URL with connection information; any parameter value parsed from this string will override explicitly set parameters')->end()
                 ->scalarNode('dbname')->end()
-                ->scalarNode('host')->defaultValue('localhost')->end()
-                ->scalarNode('port')->defaultNull()->end()
-                ->scalarNode('user')->defaultValue('root')->end()
-                ->scalarNode('password')->defaultNull()->end()
+                ->scalarNode('host')->info('Defaults to "localhost" at runtime.')->end()
+                ->scalarNode('port')->info('Default to null at runtime.')->end()
+                ->scalarNode('user')->info('Default to "root" at runtime.')->end()
+                ->scalarNode('password')->info('Default to null at runtime.')->end()
```
weaverryan: 3x need "Defaults" with an "s"

DependencyInjection/Configuration.php
```diff
@@ -217,10 +217,11 @@ private function configureDbalDriverNode(ArrayNodeDefinition $node): void
...
+                ->booleanNode('override_url')->info('Allows overriding URL parameters with dbname, host, port, user, and/or password parameters.')->end()
```
weaverryan:
```suggestion
                ->booleanNode('override_url')->info('Allows overriding parts of the "url" parameter with dbname, host, port, user, and/or password parameters.')->end()
```

weaverryan (commented):
Very minor comments!

DependencyInjection/Configuration.php
jrushlow: done

DependencyInjection/Configuration.php
jrushlow: done

ConnectionFactory.php
jrushlow: done

ConnectionFactory.php
jrushlow: Added based on @stof feedback in a previous iteration https://github.com/doctrine/DoctrineBundle/pull/1290#discussion_r578595522

weaverryan (approved):
This looks ready to me 👍

DependencyInjection/DoctrineExtension.php
```diff
@@ -254,6 +255,54 @@ protected function getConnectionOptions($connection)
...
+
+                $options['connection_override_options'][$paramName] = $options[$paramName];
+            }
+        }
+
+        unset($options['override_url'], $connectionDefaults['dbname']);
+
+        $nestedConnections = ['shards', 'replicas', 'slaves'];
+
+        foreach ($connectionDefaults as $defaultKey => $defaultValue) {
```
ostrolucky: I don't understand why is this done. If I remove this block and revert removal of default values in Configuration.php, all tests pass. It should be clear from code why was this done this way.

DependencyInjection/DoctrineExtension.php
```diff
@@ -254,6 +255,54 @@ protected function getConnectionOptions($connection)
...
+                    continue;
+                }
+
+                $options['connection_override_options'][$paramName] = $options[$paramName];
+            }
+        }
+
+        unset($options['override_url'], $connectionDefaults['dbname']);
+
+        $nestedConnections = ['shards', 'replicas', 'slaves'];
```
ostrolucky: When variable is not used more than once, it should be inlined. But hopefully we can remove whole thing because of comment above.

ConnectionFactory.php
```diff
@@ -64,6 +68,24 @@ public function createConnection(array $params, Configuration $config = null, Ev
             $params     = $connection->getParams();
```
ostrolucky: I don't know why doing it in such convoluted way below. Isn't all we need to do instead is change this line to 
```suggestion
            $params     = $connectionOverrideOptions + $connection->getParams();
```
?

Tests/DependencyInjection/Fixtures/config/yml/dbal_allow_url_override.yml
```diff
@@ -0,0 +1,9 @@
...
+        url: 'mysql://root:password@database:3306/main?serverVersion=mariadb-10.5.8'
+        dbname: main_test
+        user: tester
+        password: wordpass
+        host: docker
```
ostrolucky: I suggest to change one of these to its default value to prove overwriting url components work even in case when option value matches the value it would have when url is not specified. And most importantly, to guard this behaviour for regressions. I would pick `host` option in this case and set its value to `localhost`

ConnectionFactory.php
```diff
@@ -45,8 +45,12 @@ public function createConnection(array $params, Configuration $config = null, Ev
             $this->initializeTypes();
         }
 
-        if (! isset($params['pdo']) && ! isset($params['charset'])) {
+        $connectionOverrideOptions = $params['connection_override_options'] ?? [];
+        unset($params['connection_override_options']);
+
+        if (! isset($params['pdo']) && (! isset($params['charset']) || (! empty($connectionOverrideOptions) && isset($params['url'])))) {
```
ostrolucky: Harmful use of `empty` detected. Doesn't bring anything, only disadvantages, like silencing error when declaration of variable disappears. I suggest simply
```suggestion
        if (! isset($params['pdo']) && (! isset($params['charset']) || ($connectionOverrideOptions && isset($params['url'])))) {
```

DependencyInjection/DoctrineExtension.php
```diff
@@ -254,6 +255,54 @@ protected function getConnectionOptions($connection)
...
+            'host' => 'localhost',
+            'port' => null,
+            'user' => 'root',
+            'password' => null,
+        ];
+
+        $allowOverrideUrl = $options['override_url'] ?? false;
+
+        if (! $allowOverrideUrl && isset($options['url'])) {
+            Deprecation::trigger(
```
ostrolucky: We don't require this package at the moment. And it is not installed via transitive dependencies in case of non-dev stability either. Will be done later as part of https://github.com/doctrine/DoctrineBundle/issues/1300. If you want to use it already, adding requirement to composer.json is needed.

DependencyInjection/Configuration.php
```diff
@@ -217,10 +217,11 @@ private function configureDbalDriverNode(ArrayNodeDefinition $node): void
...
                 ->scalarNode('dbname')->end()
-                ->scalarNode('host')->defaultValue('localhost')->end()
-                ->scalarNode('port')->defaultNull()->end()
-                ->scalarNode('user')->defaultValue('root')->end()
-                ->scalarNode('password')->defaultNull()->end()
+                ->scalarNode('host')->info('Defaults to "localhost" at runtime.')->end()
+                ->scalarNode('port')->info('Defaults to null at runtime.')->end()
+                ->scalarNode('user')->info('Defaults to "root" at runtime.')->end()
+                ->scalarNode('password')->info('Defaults to null at runtime.')->end()
+                ->booleanNode('override_url')->info('Allows overriding parts of the "url" parameter with dbname, host, port, user, and/or password parameters.')->end()
```
ostrolucky:
```suggestion
                ->booleanNode('override_url')->defaultValue(false)->info('Allows overriding parts of the "url" parameter with dbname, host, port, user, and/or password parameters.')->end()
```

This should default to false and deprecation should be triggered from Configuration, via something like this
```php
->beforeNormalization()
    ->ifTrue(static function ($v) {
        return ! $v['override_url'] && isset($v['url']);
    })
    ->then(static function ($v) {
        Deprecation::trigger(...);

        return $v;
    })
->end();
```

DependencyInjection/DoctrineExtension.php
```diff
@@ -254,6 +255,54 @@ protected function getConnectionOptions($connection)
...
+                'doctrine/doctrineBundle',
+                'https://github.com/doctrine/DoctrineBundle/pull/1290',
+                'Not setting doctrine.dbal.override_url to true is deprecated. True is the only value that will be supported in doctrine-bundle 3.0.'
+            );
+        }
+
+        if ($allowOverrideUrl) {
+            foreach (array_keys($connectionDefaults) as $paramName) {
```
ostrolucky: Can be done simpler. Remember, PHP standard library is powerful.
```suggestion
            $options['connection_override_options'] = array_intersect_key($options, $connectionDefaults);
```

ostrolucky (changes requested):

DependencyInjection/DoctrineExtension.php
jrushlow: good catch!

DependencyInjection/DoctrineExtension.php
jrushlow: done

ostrolucky:
@jrushlow Hey thanks for starting to fix this, but if you would like, I could fix all of the comments for you. Just please let me know in my comments if I am wrong in some of them. Especially removal of default values is questionable for me, but maybe you had a reason for that - in that case I would like to know about it.

jrushlow:
Thanks @ostrolucky - im going through the rest of this to try and get it wrapped up for today, but ill let you know if i run out of time (kids have a soccer game in a couple hours) / get stuck on something

> Especially removal of default values is questionable for me, but maybe you had a reason for that - in that case I would like to know about it.

We had to remove the defaults from the config otherwise we have no way of determining if a value for say `host` is the override or the default. The tests passed when you reverted because in the `dbal_allow_url_override` case - I was overriding everything. Which hide the fact that the `url`'s host value was `database`, and if i don't override the `host`, we are overriding the `host` with the default value (`localhost`)... Sounds a bit confusing but basically we dont want to override a legitimate value with a default value.. 

Anywho, im looking at it now to see if there is a way we can do this w/o removing the defaults from the `Configuration` but no luck so far.

ostrolucky:
I was testing exactly this case locally actually. I modified the test case by setting host to `localhost` and checked that output host was localhost. This still worked with my changes (adding defaults back and removing the mentioned block)

Tests/DependencyInjection/Fixtures/config/yml/dbal_allow_url_override.yml
jrushlow: done

DependencyInjection/Configuration.php
jrushlow: done

DependencyInjection/DoctrineExtension.php
jrushlow: using `trigger_error` in configuration

ConnectionFactory.php
jrushlow: The thought was to only allow specific overrides (`dbname`, `host`, `port`, `user` & `password`). https://github.com/doctrine/DoctrineBundle/pull/1290/commits/d1bbcb8f2ad708543269017af94194e4849bec2c streamlines the code by using `array_merge` - but in doing so will override / inject any connection param supplied in the `$connectionOverrideOptions` array. 

I'm good with reverting the aforementioned commit or keeping as is.

DependencyInjection/DoctrineExtension.php
jrushlow: Follow up to our slack conversation and to bring everyone else up to speed:

If we set default values in the `Configuration` - we have no way to determine if a value is an override or a default. https://github.com/doctrine/DoctrineBundle/pull/1290/files#diff-59ba6fdc91d16d6c28504e469b134f4bbf84e6b39d42c83d91652afe6015e1e7 asserts this behavior. If we set the defaults in the `Configuration`, this test will now fail(`host` would be `localhost` rather than the overridden value).

ostrolucky:
I've done some cleanup for you, let me know if you notice something suspicious in my changes.

jrushlow:
> I've done some cleanup for you, let me know if you notice something suspicious in my changes.

Everything looks good to me. Thanks @ostrolucky for the help on this!

nicolas-grekas:
I'm proposing another approach in #1328

nicolas-grekas:
Note that I don't agree with the deprecation and that I'm proposing to revert it in #1328

ostrolucky:
@jrushlow What do you think about that?

Issue #1293, opened 2021-02-13
ServiceEntityRepository shoud be generic like it parent 


Punk-UnDeaD:
EntityRepository are generic, but ServiceEntityRepository not

ostrolucky:
Already done https://github.com/doctrine/DoctrineBundle/blob/2.3.x/Repository/ServiceEntityRepository.php

PR #1294, opened 2021-02-15
Add generic/template annotations to ServiceEntityRepository

7ochem:
When creating a `FooRepository` for my `Foo` class I can use the `ServiceEntityRepository` to autowire the repository. However, I want to tell Vimeo Psalm that the `FooRepository` is operating on/with my `Foo` entity class.

The `\Doctrine\ORM\EntityRepository` has generic/template annotations on the class level doc block for that, but I cannot leverage those because my repository class is extending `ServiceEntityRepository` and that one doesn't have generic/template annotations.

This PR suggests to add those.

### DIFF
--- a/Repository/ServiceEntityRepository.php
+++ b/Repository/ServiceEntityRepository.php
@@ -19,6 +19,9 @@ use LogicException;
  *         parent::__construct($registry, YourEntity::class);
  *     }
  * }
+ *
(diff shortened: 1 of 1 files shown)


ostrolucky:
Duplicate of https://github.com/doctrine/DoctrineBundle/issues/1293

PR #1295, opened 2021-02-15
Add ServiceGeneratedValue annotation

nicolas-grekas:
Ths PR provides a `@ServiceGeneratedValue` annotation that one can use instead of the `@GeneratedValue` + `@CustomIdGenerator` combo.

See the [attached doc](https://github.com/doctrine/DoctrineBundle/pull/1295/files#diff-5edb9a6ef4c360d831febebdbdc972c4a6de5d673a57f2725339fecc09ac76ce) for more details and examples.

### DIFF
--- a/Tests/DependencyInjection/Fixtures/Bundles/AnnotationsBundle/Entity/TestServiceGeneratedValueEntity.php
+++ b/Tests/DependencyInjection/Fixtures/Bundles/AnnotationsBundle/Entity/TestServiceGeneratedValueEntity.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Fixtures\Bundles\AnnotationsBundle\Entity;
+
+use Doctrine\Bundle\DoctrineBundle\Mapping\ServiceGeneratedValue;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity
+ */
+class TestServiceGeneratedValueEntity
+{
+    /**
+     * @ORM\Id
+     * @ORM\Column(type="integer")
+     * @ServiceGeneratedValue(id="my_id_generator", method="theMethod", arguments={123})
+     *
+     * @var int
+     */
+    public $id;
+}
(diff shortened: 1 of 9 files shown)


nicolas-grekas:
About failures related to CS:

> Error: Type Attribute is not used in this file.
> Error: Perl-style comments are not allowed. Use "// Comment." or "/* comment */" instead.

This is related to PHP 8 attributes that aren't supported by the CS checker apparently.

> Error: Incorrect annotations group.

Dunno what this means. Any hints welcome.

The other failures are unrelated to this PR.

ostrolucky:
In my opinion such amount of code is not worth the syntax sugar it brings. If you think there should be shortcut for `@GeneratedValue` + `@CustomIdGenerator`, it would be better to contribute it to doctrine/orm instead.

Tests/DependencyInjection/Fixtures/Bundles/AnnotationsBundle/Entity/TestServiceGeneratedValueEntity.php
```diff
@@ -0,0 +1,21 @@
...
+use Doctrine\Bundle\DoctrineBundle\Mapping\ServiceGeneratedValue;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity
+ */
+class TestServiceGeneratedValueEntity
+{
+    /**
+     * @ORM\Id
```
greg0ire, line 14: > Dunno what this means. Any hints welcome.

@nicolas-grekas This one is about `@var` being separated from other annotations I think. It can usually be fixed by `vendor/bin/phpcbf`

ostrolucky:
I think we are in agreement and I don't expect further activity here, let's close.

juslintek:
What is the most recommended approach over here? Because this problem still exists and I cannot override UuidFactory defaultClass parameter in test environment for unknown reasons. Services configs are ignored. And doctrine 3.2 still has this problem. For time being using https://github.com/symfony/symfony/issues/41772#issuecomment-865255543 walk-around. Because nothing else helps, Uuidv6 is defaultClass always.

PR #1296, opened 2021-02-19
Autowire connections by argument type and name

HypeMC:
Similar to #1259 this will allow autowiring connections by argument type and name which is useful when you don't want to use the ORM and have multiple DBAL connections:

```php
use Doctrine\DBAL\Connection;

class Foo
{
    public function __construct(Connection $myNameConnection) {}
}
```

The last PR was a little controversial but I'd still like to hear peoples thoughts on this one.


nicolas-grekas (approved):
Sweet :)
Failures are unrelated.

DependencyInjection/DoctrineExtension.php
```diff
@@ -224,6 +227,10 @@ protected function loadDbalConnection($name, array $connection, ContainerBuilder
                 $connection['mapping_types'],
             ]);
 
+        $container
+            ->registerAliasForArgument($connectionId, Connection::class, sprintf('%sConnection', $name))
```
stof, line 231: is `registerAliasForArgument` available in all supported Symfony versions ? I don't remember when the shortcut was added

HypeMC: The minimum version for `symfony/dependency-injection` on the `2.3.x` branch is `4.3.3` [which does support this](https://github.com/symfony/symfony/blob/v4.3.3/src/Symfony/Component/DependencyInjection/ContainerBuilder.php#L1375).

dmaicher (approved):

PR #1297, opened 2021-02-25
Register remember me schema subscriber from Symfony Bridge

wouterj:
This PR registers the `RememberMeTokenProviderDoctrineSchemaSubscriber` schema listener that will be introduced in symfony/symfony#40145 . This way, Doctrine's schema system can help the user create and update their remember me table.


stof (changes requested):
This service definition should be conditional, to keep support for Symfony versions not having it (the bundle supports the 4.4 LTS too)

wouterj:
Thanks for the quick review @stof! The definition is optional now.

SenseException (approved):

wouterj:
The related Symfony PR has been merged and it will be part of the upcoming 5.3 release. So I think this PR is ready to be merged.

ostrolucky (approved):

PR #1298, opened 2021-03-03
Add template definitions to ServiceEntityRepository

iluuu1994:
Since `ServiceEntityRepository` doesn't define any `@template`s it's impossible to tell Psalm what `T` is. With this PR the following would become possible:

```php
/**
 * @template-extends ServiceEntityRepository<User>
 */
class UserRepository extends ServiceEntityRepository
{
    public function __construct(ManagerRegistry $registry)
    {
        parent::__construct($registry, User::class);
    }
}
```

whereas right now the only option is to avoid `ServiceEntityRepository` and extend `EntityRepository` directly.

### DIFF
--- a/Repository/ServiceEntityRepository.php
+++ b/Repository/ServiceEntityRepository.php
@@ -19,6 +19,9 @@ use LogicException;
  *         parent::__construct($registry, YourEntity::class);
  *     }
  * }
+ *
+ * @template T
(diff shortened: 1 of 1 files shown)


ostrolucky:
Another duplicate https://github.com/doctrine/DoctrineBundle/pull/1294

iluuu1994:
Just noticed, sorry.

PR #1309, opened 2021-03-23
Disable schema validation no matter the profiling configuration

bastnic:
On Symfony, there is a way with `APP_ENV=dev` and `APP_DEBUG=false` to have the profiling disable but still have the DataCollector running.

In this particular case, the schema validation will occured even if the configuration disable it.
The funny thing is that it's slower with debug disable...

Question: the correct fix shouldn't be to remove the collector when profiling is disabled?

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -156,6 +156,9 @@ class DoctrineExtension extends AbstractDoctrineExtension
 
         unset($connection['logging']);
 
+        $dataCollectorDefinition = $container->getDefinition('data_collector.doctrine');
+        $dataCollectorDefinition->replaceArgument(1, $connection['profiling_collect_schema_errors']);
+
         if ($connection['profiling']) {
             $profilingAbstractId = $connection['profiling_collect_backtrace'] ?
                 'doctrine.dbal.logger.backtrace' :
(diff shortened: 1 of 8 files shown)


greg0ire:
> Question: the correct fix shouldn't be to remove the collector when profiling is disabled?

In my opinion yes, not adding a service is the best way to make sure there is no profiling 👍

DependencyInjection/DoctrineExtension.php
```diff
@@ -164,9 +167,7 @@ protected function loadDbalConnection($name, array $connection, ContainerBuilder
             $profilingLoggerId = $profilingAbstractId . '.' . $name;
             $container->setDefinition($profilingLoggerId, new ChildDefinition($profilingAbstractId));
             $profilingLogger = new Reference($profilingLoggerId);
-            $container->getDefinition('data_collector.doctrine')
-                ->addMethodCall('addLogger', [$name, $profilingLogger])
-                ->replaceArgument(1, $connection['profiling_collect_schema_errors']);
```
greg0ire, line 169: Doing this conditionally caused the bug because the default value is `true`: https://github.com/doctrine/DoctrineBundle/blob/155a870ecbd035963e6b450e0f235e69c48cab30/DataCollector/DoctrineDataCollector.php#L34 Disabling profiling caused the schema validation to happen.

ostrolucky:
test needed

bastnic:
@ostrolucky tests added. 

I've an issue with my initial implemantation: the parameter is per connection but only the last connection will define the status in the Collector.

nicolas-grekas (approved):

nicolas-grekas:
> the parameter is per connection but only the last connection will define the status in the Collector.

The `profiling_collect_schema_errors` setting should not be per-connection, you're right.
But that's an issue that is unrelated to this very PR, so not a blocker here.

ostrolucky:
@bastnic Could you create issue for the bug you noticed?

Anyways thanks for this PR!

PR #1313, opened 2021-03-25
Add reference in container parameters to new attribute metadata driver

L41rx:
Hi! Recently https://github.com/doctrine/orm/pull/8266 was merged in. I had made a new symfony project and wanted to use the entity attributes to generate a schema, I was able to do so but needed to add this parameter entry

I am not sure if this is useful, sorry if it is not! (I am not sure which branch this should go to, because I guess this relies on doctrine/orm 2.9.x)


ostrolucky:
[These class parameters are deprecated and will be removed](https://github.com/doctrine/DoctrineBundle/blob/2.3.x/UPGRADE-2.1.md#parameters), so we don't want to add new ones.

PR #1316, opened 2021-03-29
Fix #1314, shard manager class option cannot be overridden

### LINKED ISSUE #1314: Unable to set custom shard manager
Good Day,

In https://github.com/doctrine/DoctrineBundle/blob/70be9d41da1c75c613d79e9d3c604d179cd705b0/DependencyInjection/DoctrineExtension.php#L405 the `$options['shardManagerClass']` will always be empty because it gets `unset()` a few lines prior as it is not part of `$nonRewrittenKeys` array.

So there is no way to provide a custom shard manager class via configs as it just gets unset.

Is this intentional or is it a bug?

cyberemissary:
Fixed a bug that does not allow to override the shard_manager_class

Closes #1314

### DIFF
--- a/Tests/DependencyInjection/DoctrineExtensionTest.php
+++ b/Tests/DependencyInjection/DoctrineExtensionTest.php
@@ -7,6 +7,8 @@ use Doctrine\Bundle\DoctrineBundle\DependencyInjection\DoctrineExtension;
 use Doctrine\Bundle\DoctrineBundle\Tests\Builder\BundleConfigurationBuilder;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Driver\Connection as DriverConnection;
+use Doctrine\DBAL\Sharding\PoolingShardManager;
+use Doctrine\DBAL\Sharding\SQLAzure\SQLAzureShardManager;
 use Doctrine\ORM\EntityManagerInterface;
 use InvalidArgumentException;
 use LogicException;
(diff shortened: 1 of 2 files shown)


Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -995,6 +1002,13 @@ public function testShardManager(): void
 
         $this->assertTrue($container->hasDefinition('doctrine.dbal.foo_shard_manager'));
         $this->assertFalse($container->hasDefinition('doctrine.dbal.bar_shard_manager'));
+        $this->assertTrue($container->hasDefinition('doctrine.dbal.baz_shard_manager'));
+
+        $fooManagerDef = $container->getDefinition('doctrine.dbal.foo_shard_manager');
+        $bazManagerDef = $container->getDefinition('doctrine.dbal.baz_shard_manager');
+
+        $this->assertEquals('Doctrine\\DBAL\\Sharding\\PoolingShardManager', $fooManagerDef->getClass());
```
ostrolucky: PoolingShardManager::class. Not sure why CS didn't complain.

ostrolucky (changes requested):

Tests/DependencyInjection/DoctrineExtensionTest.php
cyberemissary: Fixed

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -975,8 +976,9 @@ public static function cacheConfigurationProvider(): array
 
     public function testShardManager(): void
     {
-        $container = $this->getContainer();
-        $extension = new DoctrineExtension();
+        $container    = $this->getContainer();
+        $extension    = new DoctrineExtension();
+        $managerClass = 'Fake\\ShardManager';
```
ostrolucky: Sorry, one more. I missed this in my last review

cyberemissary: Not sure what's here, did you want me to use a real class and reference with `::class` ?

ostrolucky: yes

ostrolucky (approved):

PR #1319, opened 2021-04-02
Start using Psalm in CI

ostrolucky:
Let's start with least strict level


.github/workflows/static-analysis.properties.json
```diff
@@ -0,0 +1,12 @@
...
+  "description": "Perform static analysis on the source code and tests",
+  "iconName": "doctrine-logo",
+  "categories": [
+    "PHP"
+  ],
+  "filePatterns": [
+    "^phpstan\\.neon(?:\\.dist)$",
+    "^psalm\\.xml$"
+  ]
+}
```
greg0ire: That file should not be copied here, it's there just so you can see a nice icon while using the GUI to setup the action

.github/workflows/static-analysis.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          php-version: "${{ matrix.php-version }}"
+
+      - name: "Install dependencies with Composer"
+        uses: "ramsey/composer-install@v1"
+        with:
+          dependency-versions: "highest"
```
greg0ire: Maybe use stable stability for this job, to speed things up?

.github/workflows/static-analysis.yml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  static-analysis-psalm:
+    name: "Static Analysis with Psalm"
+    runs-on: "ubuntu-20.04"
+
+    strategy:
+      matrix:
+        php-version:
+          - "7.1"
```
greg0ire, line 20: I haven't given much thought at what the best version here for SA jobs, and maybe we need several, but why pick the lowest supported version here?

composer.json
```diff
@@ -42,14 +42,16 @@
         "doctrine/orm": "^2.6",
         "friendsofphp/proxy-manager-lts": "^1.0",
         "phpunit/phpunit": "^7.5 || ^8.0 || ^9.3",
+        "psalm/plugin-phpunit": "^0.15",
```
greg0ire: Do you observe fewer errors when you install this plugin? Nowadays phpunit has Psalm annotations, much like Doctrine, and the README does not make clear what benefit we get from this. Also there is a Symfony plugin, maybe that one should be installed, since Symfony does not accept Psalm annotations? It would help with the `Command` and `Controller` directories, which I see are ignored for now.

psalm.xml.dist
```diff
@@ -0,0 +1,32 @@
...
+        <directory name="DependencyInjection"/>
+        <directory name="EventSubscriber"/>
+        <directory name="Mapping"/>
+        <directory name="Repository"/>
+        <directory name="Tests"/>
+        <directory name="Twig"/>
+        <file name="ConnectionFactory.php"/>
+        <file name="DoctrineBundle.php"/>
+        <file name="ManagerConfigurator.php"/>
+        <file name="Registry.php"/>
```
greg0ire, line 30: Lost of directories in files ignored, but I suppose that's temporary, so that we can start having Psalm and remove them one by one?

CacheWarmer/DoctrineMetadataCacheWarmer.php
```diff
@@ -52,7 +54,7 @@ protected function doWarmUp($cacheDir, ArrayAdapter $arrayAdapter): bool
 
         if (method_exists($metadataFactory, 'setCache')) {
             $metadataFactory->setCache($arrayAdapter);
-        } else {
+        } elseif (class_exists(DoctrineProvider::class) && $metadataFactory instanceof AbstractClassMetadataFactory) {
```
greg0ire: Why is that condition necessary? are there 3 cases instead of just 2?

Command/DoctrineCommand.php
```diff
@@ -60,6 +61,10 @@ protected function getEntityManager($name, $shardId = null)
         $manager = $this->getDoctrine()->getManager($name);
 
         if ($shardId) {
+            if (! $manager instanceof EntityManagerInterface) {
+                throw new LogicException(sprintf('Sharding is supported only in EntityManager of instance "%s".', EntityManagerInterface::class));
```
greg0ire, line 65: This looks weird… I mean all `EntityManagers` implement `EntityManagerInterface`, right? Should that sentence mention `ObjectManager` or something like that instead?

DoctrineBundle.php
```diff
@@ -40,6 +40,7 @@ public function build(ContainerBuilder $container)
         $container->addCompilerPass(new RegisterEventListenersAndSubscribersPass('doctrine.connections', 'doctrine.dbal.%s_connection.event_manager', 'doctrine'), PassConfig::TYPE_BEFORE_OPTIMIZATION);
 
         if ($container->hasExtension('security')) {
+            /** @psalm-suppress MissingDependency We can safely assume here that if security extension exists, all seccurity deps are installed  */
```
greg0ire: Typo in the second "security"

greg0ire (commented):
This already catches a lot of things! Nice!

.github/workflows/static-analysis.yml
ostrolucky: Because 7.1 is most likely we will forget about and we need a kick from Psalm to remind us if we accidentally use some >7.1 functionality.

psalm.xml.dist
ostrolucky: These are inclusions, not exclusions actually. DoctrineBundle doesn't have central directory like `src` or `lib` at the moment.

Command/DoctrineCommand.php
ostrolucky: `getConnection` doesn't exist in ObjectManager, only EntityManagerInterface

composer.json
ostrolucky: I had some issues with that locally, but on second try looks like you are right and we don't need it

.github/workflows/static-analysis.yml
greg0ire: Okay, that make sense and since for now Psalm supports 7.1, it does not mean we will not get the latest updates 👍

psalm.xml.dist
greg0ire: Oh right, my bad!

CacheWarmer/DoctrineMetadataCacheWarmer.php
ostrolucky: I started with level 4 and you get this error there:

ERROR: UndefinedInterfaceMethod - CacheWarmer/DoctrineMetadataCacheWarmer.php:59:31 - Method Doctrine\Persistence\Mapping\ClassMetadataFactory::setCacheDriver does not exist (see https://psalm.dev/181)

But I have removed class_exists at least, it no longer seems necessary

greg0ire (approved):

PR #1322, opened 2021-04-10
Add support for ORM 2.9 AttributeDriver

beberlei:
This adds support for the `AttributeDriver` that is new in Doctrine ORM 2.9

Caveats:
- Auto Mapping is not possible for attributes, because the Entity folder is detected as annotation driver. Solution? We could add a global configuration that allows users to specifcy that attributes and not annotations should be used for auto mapping.
- Explicitly defining a bundle with attributes requires to specify the `type: attribute` in YML or XML.
- Methods `assertValidMappingConfiguration` and `registerMappingDrivers` had to be copied from `DoctrineBrdige/AbstractDoctrineExtension` to `DoctrineExtension`, because they need to mention `attribute` somewhere deep in the code. This can be removed once we move the code to DoctrineBridge.


beberlei:
@ostrolucky I assume this should better go against 2.4.x or?

ostrolucky:
yes I think so

ostrolucky (changes requested):
This looks like whole implementation consists of overriding 3 methods from doctrine-bridge... This patch needs to be implemented in doctrine-bridge.

beberlei:
@ostrolucky then this will not be released before November, so no

ostrolucky:
Well if we do this here, it will also mean any fixes in those methods in doctrine-bridge will not take effect for doctrine-bundle 2.4 users, which is also a no-no. This bundle is not for workarounding symfony release schedule...

Any opinions from symfony folks? Any chance this can be added as a patch release in doctrine-bridge perhaps? Without this, PHP 8 users cannot really use one of major PHP 8 features

cc @nicolas-grekas @weaverryan @stof @jderusse

beberlei:
The last change to this class that wasn't CS was in August 2019 and was not related to these methods b53d8ccfc1ee7680230c2089903bc37825460d7c. If you look more deeply at the git blame for the three methods, imho then nothing of value has changed in this code since Fabian and I wrote it in 2011.

I would be happy if this would be in a Symfony patch release, but I don't want to impose myself.

ostrolucky:
I think this can also land in Symfony 5.3 which will be released next month. Looking around in their PR list, it doesn't seem feature freeze happened yet. You really need to move this over there.

nicolas-grekas:
Which changes are we talking about? Maybe this could be considered as a bugfix, if no new method involved?

nicolas-grekas:
Please send a PR on Symfony :pray:

ostrolucky:
It's settled then. Will be moved over there.

beberlei:
@ostrolucky this still needs code changes here

ostrolucky:
Ok, which ones?

beberlei:
@ostrolucky all changes except these three methods I overwrote, so its still changes to 12 files.

ostrolucky (changes requested):
Work needed: 
- bump doctrine-bridge requirement (after it's released)
- drop changes in DoctrineExtension

stof:
the new requirement for the bridge will be `^4.4.22 || ^5.2.7`, as that change will be part of the next release of symfony (it has been merged)

beberlei:
@stof we don't need to raise the requirement, i would split out the AttributeBundle tests into their own methods and guard them with the package versions for the bridge and then the tests run only when the higher bridge is used.

ostrolucky:
At the same time, we don't need to support older versions. Raising the requirement would be simpler.

ostrolucky (approved):

Issue #1300, opened 2021-03-07
Integrate with doctrine/deprecations


beberlei:
As we begin to trigger deprecations through https://github.com/doctrine/deprecations/ the bundle should depend on the package and configure it to use Symfony's way of doing `@trigger_error` by calling the following in `DoctrineBundle::boot`.

```php
\Doctrine\Deprecations\Deprecation::enableWithSuppressedTriggerError();
```
Maybe there should be a configuration flag to disable individual things.

ostrolucky:
Where do you suggest to put this bootstrap code?

stof:
I think the bundle should keep using the Symfony way directly for things happening at build time in the bundle (in the DI extension, the Configuration class and compiler passes) and should call `\Doctrine\Deprecations\Deprecation::enableWithSuppressedTriggerError();` in the `boot` method.
This will be the easiest way (hooks running before the build time happens won't run if the cache is not regenerated).

ostrolucky:
That's not a good situation if we need to use mixed approach. I consider this a blocker for implementing this library. It has a global state other libraries will fight for. Solution should come from `doctrine/deprecations` @beberlei.

nicolas-grekas:
`doctrine/deprecations` looks like a terrible idea to me, @beberlei knows it already... Global state, options, which means choices to make, which means confusion and fighting with conflicting opinions...
I'd like its existence could be reconsidered in the first place. If it really needs to exist, most options should be removed to me, so that the use of global state can be really reduced.
E.g. there could be only one global state: $logger. When not `null`, all deprecations go there, with context to implement any custom logic (eg filtering by packages). And when `null`, a silenced `trigger_error()` is called.

beberlei:
How is this any different than what `@tirgger_error` does now? Error handling is a global problem in PHP, just have to live with it.

beberlei:
@nicolas-grekas nobody has to make choices, it is designed to adopt the way deprecations work in the framework it nests into. Doctrine is not used by Symfony alone, and Doctrine does not have control over the error handler as Symfony does. I can understand your approach, but it does not work for Doctrine.

Symfony uses suppressed trigger errors, so we are configuring it that way. 

A global logger is just the same global state with options to compete over.

nicolas-grekas:
> it is designed to adopt the way deprecations work in the framework it nests into

[...]

PR #1323, opened 2021-04-10
Add integration with doctrine/deprecations library.

beberlei:
A default off integration with doctrine/deprecations, enable with:

```
doctrine:
    deprecations: %kernel.debug%
```

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -55,10 +55,29 @@ class Configuration implements ConfigurationInterface
         $treeBuilder = new TreeBuilder('doctrine');
         $rootNode    = $treeBuilder->getRootNode();
 
+        $this->addDeprecationsSection($rootNode);
         $this->addDbalSection($rootNode);
         $this->addOrmSection($rootNode);
 
         return $treeBuilder;
+    }
+
+    /**
+     * Add Deprecations section
+     */
+    private function addDeprecationsSection(ArrayNodeDefinition $node)
+    {
+        $node
+            ->children()
+            ->arrayNode('deprecations')
+                ->treatFalseLike(['enabled' => false])
+                ->treatTrueLike(['enabled' => true])
+                ->treatNullLike(['enabled' => false])
+                ->addDefaultsIfNotSet()
+                ->children()
+                    ->scalarNode('enabled')->defaultFalse()->end()
+                ->end()
+            ->end();
     }
 
     /**
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -73,6 +73,10 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $configuration = $this->getConfiguration($configs, $container);
         $config        = $this->processConfiguration($configuration, $configs);
 
+        if (! empty($config['deprecations'])) {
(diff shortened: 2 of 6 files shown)


ostrolucky:
Shouldn't part of this integration also be to change current deprecations in bundle to utilize this library now?

greg0ire:
Wasn't there a plan at some point to split this bundle into orm bundle and dbal bundle? What's the status? And if it's still relevant, what will happen with this configuration node? It seems overkill to create a separate repository just to be able to configure this, so maybe we should have framework integrations directly in the repository `doctrine/deprecations`? See here an example of how it could look: https://github.com/sonata-project/exporter/tree/2.x/src/Bridge/Symfony

ostrolucky:
I'm not actually convinced we even need configuration for this. I would just enable it automatically, because that's what symfony does

beberlei:
@ostrolucky symfony does not trigger so many deprecations usually, the ORM alone can easily trigger DBAL deprecations over 1000 times at the moment per request. This has real overhead.

ostrolucky:
But symfony already has mechanisms to disable deprecations in production. It also uses FingersCrossed handler by default which doesn't log anything unless error happens. Not sure it makes sense to treat deprecations coming from libraries using doctrine/deprecations differently than symfony ones. Anyways I'm not comfortable on deciding this, any opinions from eg. symfony folks on this topic?

cc @nicolas-grekas @wouterj @stof @weaverryan

ostrolucky:
Alright no input from symfony folks, so I take it that they don't think it's going to be a problem that symfony users would be spammed with deprecations. Hence we will just turn it on by default, no need for config option for now. We can resurrect this PR later, if needed.

nicolas-grekas:
I agree to not make this configurable. If someone proves that this would provide an actual benefit, this could happen later as you mention.
But for now, this should always flow to the normal path of deprecations - aka trigger_error().

beberlei:
I want to revisit this, we should really make this configurable because our deprecations are a bit heavier due to the `Deprecation::triggerIfCalledFromOutside` API. 

When enabled it does a `debug_backtrace` call pretty early on every call even before deduplication. 

This will essentially make it highly recommended to be configurable based on `kernel.debug`, because this method is used from such deprecations as `Statement::fetch()` - which can happen 100-1000s of times per request.

In addition it is clear that there is no entry point in the bridge to offer this functionality. Users that don't use DoctrineBundle should just be advised to call `\Doctrine\Deprecations\Deprecation::enableWithTriggerError()` in their code.

nicolas-grekas:
See https://github.com/symfony/symfony/pull/41408 for a longer discussion. TLDR, many think this doesn't fit either the bridge nor the bundle. My preference would be for doctrine/deprecation to ship a small bundle for this purpose.

ostrolucky:
Yeah, also see https://github.com/doctrine/DoctrineBundle/issues/1300. 

We don't enable it by default, currently it's up to user. Since we don't enable it by default (and neither symfony looks interested in doing that), we don't need to consider performance. 

It makes sense to provide some kind of neat configuration for symfony users by default, but it's out of scope of this bundle, because doctrine/deprecations isn't for doctrine/orm and doctrine/dbal only, but doctrine/doctrine-bundle is. In other words, symfony users shouldn't have to install doctrine-bundle in order to configure this if they don't use orm/dbal, but do use some other doctrine libraries.

Let us know your opinion on that conclusion @beberlei.

PR #1326, opened 2021-04-18
Added missed attribute mapping driver class

a-menshchikov:
Fixing an error:
image


ostrolucky:
Work for that is being done here https://github.com/doctrine/DoctrineBundle/pull/1322. It will also not be merged until symfony releases new version

PR #1327, opened 2021-04-18
Make DoctrineBundle compatible with ORM 2.9

beberlei:
In ORM 2.9 similar to DBAL 2.11 we introduce a new way of injecting the `EntityManager` into commands via registries/providers. These register the option `em`, which would lead to conflicts in the proxy commands due to redefinition of the option.

The same change was already made for dbal in 86d2469d6be06d55ad7b9e2f076f6942476f2e87

Once ORM 2.9 is out we can think of increasing the `doctrine/orm` dependency from 2.6 to 2.9, since both support PHP 7.1. Then we can replace the whole proxy command code with proper dependency injection and shipping an entity manager provider.

This PR adds one test for one proxy command to test the variable entity manager option behavior. In addition since commands are services for a while already, the `ContainerTest` is also extended to fetch two out of the container

### DIFF
--- a/Tests/Command/Proxy/InfoDoctrineCommandTest.php
+++ b/Tests/Command/Proxy/InfoDoctrineCommandTest.php
@@ -7,12 +7,15 @@ use Doctrine\DBAL\Connection;
 use Doctrine\ORM\Configuration;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\Persistence\Mapping\Driver\MappingDriverChain;
+use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
 use Symfony\Bridge\Doctrine\ManagerRegistry;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
 use Symfony\Component\Console\Tester\CommandTester;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpKernel\Kernel;
+
+use array_merge;
 
 class InfoDoctrineCommandTest extends TestCase
 {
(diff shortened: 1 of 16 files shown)


Tests/Command/Proxy/InfoDoctrineCommandTest.php
```diff
@@ -0,0 +1,68 @@
...
+        $this->assertStringContainsString(
+            'You do not have any mapped Doctrine ORM entities',
+            $commandTester->getDisplay()
+        );
+    }
+
+    /**
+     * @return MockObject|Kernel
+     */
+    private function setupKernelMocks()
```
ostrolucky: return type should still be added for either Kernel or MockObject

ostrolucky (commented):
Just a nitpick. This PR LGTM otherwise, thanks for doing it 👍 Merge with squash.

simPod (approved):

simPod:
Also note the parent option has default value while this has none.

```php
->addOption('em', null, InputOption::VALUE_REQUIRED, 'Name of the entity manager to operate on', 'default')
```

I'm not entirely sure whats correct, though having default set as `'default'` and not `null` disables `AbstractManagerRegistry`'s functionality to use real default (`$this->defaultManager`) https://github.com/doctrine/persistence/blob/2.2.x/lib/Doctrine/Persistence/AbstractManagerRegistry.php#L143-L146 causing 

> Doctrine ORM Manager named "default" does not exist.

Issue for this https://github.com/doctrine/orm/issues/8643

beberlei:
@simPod hm thats annoying, i didn't want to make the manager name nullable in ORM if it could be avoided. :( Seems it is necessary just for backwards compatibility in Bundle

simPod:
Understand. I'm using only one EM for now (used to have more so it has custom name). Though it's not working with this change and it requires to specify `em` option which is also very annoying when there's only one EM 😄

ostrolucky:
Symfony release that we were waiting for is out

ostrolucky (approved):

Issue #1325, opened 2021-04-16
Charset passed via URL is overridden


dlegatt:
https://github.com/doctrine/DoctrineBundle/blob/e6a9c16d51a1f0399c7de0099a50cf06527a3ff4/ConnectionFactory.php#L87

Working with SQL Server and wanting to pass the charset through he URL configuration instead of modifying the doctrine.yaml file and I'm finding that this line is automatically overriding whatever charset option I pass.

Example:

In Symfony .env.local: `DATABASE_URL=sqlsrv://sa:password@127.0.0.1/my_db?charset=UTF-8`

I would expect "UTF-8" charset value to be passed to the SQL Server connection driver, but it gets overridden by this line. Can `$params['charset']` be checked before defaulting to `utf8`?

ostrolucky:
Do you have `override_url` option enabled? Can you check if behaviour is same with either value?

dlegatt:
I believe I did test changing that to true or false hd no effect. Debugging and watching the values of the Paramus array, any char set value that I pass via url is lost after this line.

ostrolucky:
Just asking because that line you passed is guarded behind this condition https://github.com/doctrine/DoctrineBundle/blob/e6a9c16d51a1f0399c7de0099a50cf06527a3ff4/ConnectionFactory.php#L60

dlegatt:
I don't believe that option affects this, `$overriddenOptions` is an empty array.  I just made a new Symfony skeleton and only installed the ORM pack. Here is my `$params` after setting `DATABASE_URL` to `sqlsrv://db_user:db_password@127.0.0.1/db_name?charset=UTF-8`. This was right before line 60:

```
array:8 [
  "url" => "sqlsrv://db_user:db_password@127.0.0.1/db_name?charset=UTF-8"
  "driver" => "pdo_mysql"
  "host" => "localhost"
  "port" => null
  "user" => "root"
  "password" => null
  "driverOptions" => []
  "defaultTableOptions" => []
]
```

The `$params` array is correctly defined when `DriverManager::getConnection()` is called on line 76

```
array:10 [
    "url" => "sqlsrv://db_user:db_password@127.0.0.1/db_name?charset=UTF-8"
    "driver" => "sqlsrv"
    "host" => "127.0.0.1"
    "port" => null
    "user" => "db_user"
    "password" => "db_password"
    "driverOptions" => []
    "defaultTableOptions" => []
    "dbname" => "db_name"
[...]

PR #1328, opened 2021-04-19
Add "dbname_suffix" config option to vary the dbname per env

nicolas-grekas:
This should provide better flexibility than #1290 when needing to vary the dbname per env (think tests of course)

### DIFF
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -57,7 +57,7 @@ class ConnectionFactory
         $overriddenOptions = $params['connection_override_options'] ?? [];
         unset($params['connection_override_options']);
 
-        if (! isset($params['pdo']) && (! isset($params['charset']) || $overriddenOptions)) {
+        if (! isset($params['pdo']) && (! isset($params['charset']) || $overriddenOptions || isset($params['dbname_suffix']))) {
             $wrapperClass = null;
 
             if (isset($params['wrapperClass'])) {
@@ -76,6 +76,10 @@ class ConnectionFactory
             $connection = DriverManager::getConnection($params, $config, $eventManager);
             $params     = array_merge($connection->getParams(), $overriddenOptions);
             $driver     = $connection->getDriver();
+
+            if (isset($params['dbname']) && isset($params['dbname_suffix'])) {
+                $params['dbname'] .= $params['dbname_suffix'];
+            }
 
             if ($driver instanceof AbstractMySQLDriver) {
                 $params['charset'] = 'utf8mb4';
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -239,6 +239,7 @@ class Configuration implements ConfigurationInterface
                 ->scalarNode('user')->info('Defaults to "root" at runtime.')->end()
                 ->scalarNode('password')->info('Defaults to null at runtime.')->end()
                 ->booleanNode('override_url')->defaultValue(false)->info('Allows overriding parts of the "url" parameter with dbname, host, port, user, and/or password parameters.')->end()
+                ->scalarNode('dbname_suffix')->end()
                 ->scalarNode('application_name')->end()
                 ->scalarNode('charset')->end()
                 ->scalarNode('path')->end()
--- a/Resources/config/schema/doctrine-1.0.xsd
+++ b/Resources/config/schema/doctrine-1.0.xsd
@@ -52,6 +52,7 @@
         <xsd:attribute name="user" type="xsd:string" />
         <xsd:attribute name="password" type="xsd:string" />
         <xsd:attribute name="override-url" type="xsd:boolean" />
+        <xsd:attribute name="dbname-suffix" type="xsd:string" />
         <xsd:attribute name="application-name" type="xsd:string" />
         <xsd:attribute name="path" type="xsd:string" />
         <xsd:attribute name="unix-socket" type="xsd:string" />
--- a/Resources/doc/configuration.rst
+++ b/Resources/doc/configuration.rst
@@ -31,6 +31,7 @@ Configuration Reference
                         user:                 root
                         password:             ~
                         charset:              "UTF8"
+                        dbname_suffix:        ~
 
                         # SQLite specific
                         path:                 ~
@@ -939,6 +940,7 @@ can configure. The following block shows all possible configuration keys:
                 port:                     1234
                 user:                     user
                 password:                 secret
+                dbname_suffix:            _test
                 driver:                   pdo_mysql
                 driver_class:             MyNamespace\MyDriverImpl
                 options:
--- a/Tests/ConnectionFactoryTest.php
+++ b/Tests/ConnectionFactoryTest.php
@@ -96,6 +96,16 @@ class ConnectionFactoryTest extends TestCase
 
         $this->assertEquals($params, array_intersect_key($connection->getParams(), $params));
     }
+
+    public function testDbnameSuffix(): void
+    {
+        $connection = (new ConnectionFactory([]))->createConnection([
+            'url' => 'mysql://root:password@database:3306/main?serverVersion=mariadb-10.5.8',
+            'dbname_suffix' => '_test',
+        ]);
+
+        $this->assertSame('main_test', $connection->getParams()['dbname']);
+    }
 }
 
 /**
--- a/Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
+++ b/Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
@@ -166,6 +166,15 @@ abstract class AbstractDoctrineExtensionTest extends TestCase
         $this->assertFalse(isset($config['override_url']));
     }
 
+    public function testDbalDbnameSuffix(): void
+    {
+        $container = $this->loadContainer('dbal_dbname_suffix');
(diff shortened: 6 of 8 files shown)


nicolas-grekas:
See corresponding recipe update at https://github.com/symfony/recipes/pull/939

weaverryan (approved):
This is great! A more useful and pragmatic solution to the original problem that we were trying to tackle.

javiereguiluz (approved):

ostrolucky:
How is this more flexible than https://github.com/doctrine/DoctrineBundle/pull/1290? When someone will ask for suffix instead of prefix we need to bake in then yet another option? What problem is this PR solving?

nicolas-grekas:
#1290 swaps the responsibility of configuring the database from the env to the config file. This is wrong: the env should always be in control, because *it* knows itself.
For testing, the app should vary the dbname, and nothing else (unlike #1290 which potentially overrides all params.) It should do so while preserving the dbname from the env. Not by ignoring it.
suffix is enough to cover all needs.
Can be turned into a sprintf template if you think that's worth it (I don't.)

ostrolucky:
> #1290 swaps the responsibility of configuring the database from the env to the config file. This is wrong: the env should always be in control, because it knows itself.

No it doesn't. Users are free to continue using DATABASE_URL as source of truth, but then they need to remove other options from config file, because they didn't take any effect. Having dead code around is very bad code smell, because it's misleading people that it's doing something.

> the env should always be in control, because it knows itself.

And yet, you introduced dbname_suffix option, which makes database name specified in DATABASE_URL not being true anymore. DATABASE_URL should either be true, so should override any other option and hence defining those options doesn't make sense, or it doesn't necessarily need to be true and then it's fine if options in yaml config override parts of it.

I don't agree with removing deprecation. Either it's ok to use combination of DATABASE_URL + options, or defining options doesn't make sense (because they are overridden). This is something we want users to realize and refactor their configs so it can't happen that they have options defined by mistake which are not actually in effect. If they want to configure everything via DATABASE_URL that's fine, but then remove those options.

nicolas-grekas:
All DSNs in Symfony work with this precedence (Cache, Lock, etc): the DSN wins over specific options. I would recommend against DoctrineBundle following with a different precedence, at least for consistency.

When one does define specific options, it can have a purpose: defining default values. If ppl use both options and DSN, they might have their reason (the default recipe doesn't encourage using these options.)

For the matter of the PR, ignoring the database name from the env is a lot less flexible. The DSN might say that the dbname is `foo` or `bar`, and suffix strategy adapts to it by creating a dbname of `foo_test` or `foo_bar`. This is where this approach is superior (and simpler).

nicolas-grekas:
Note that I appreciate that you're asking these questions @ostrolucky. They had to be answered!

ostrolucky:
> All DSNs in Symfony work with this precedence (Cache, Lock, etc): the DSN wins over specific options

This is misleading. Cache doesn't allow anything else than DSN. And Lock uses same config option key for either of them so it's not possible to define both env string and config options. These are very different cases from doctrine-bundle, where we allow both at the same time.

>  I would recommend against Doctrine following with a different precedence, at least for consistency.

Please watch more closely what happens in doctrine-bundle if you want to have a say what happens here. It's a tad too late complaining about this after release, we could have agreed on something better for both sides before we released the feature.

I still don't know what problem is this PR solving though. Before https://github.com/doctrine/DoctrineBundle/pull/1290, users had to have 

in doctrine_test.yaml
```yaml
doctrine:
    dbal:
        url: '%env(DATABASE_URL)%'
```
and in .env.test
```
DATABASE_URL=mysql://root:password@database:3306/main_test?serverVersion=mariadb-10.5.8
```

After https://github.com/doctrine/DoctrineBundle/pull/1290, **they can keep doing same thing**, they just need to add `override_url: true` to solve deprecation.

Functionality added in this PR is independent from #1290, so I would like to understand your motivation for doing this. Are you attempting to discourage usage of new `override_url`, or how is it related to #1290? Even if we accept that dbname_suffix is useful, deprecation removal needs to be extracted to different PR, because they are unrelated.

This functionality is very shortsighted IMHO. If you check original PR #1290, it's motivation was for overriding database name - without any mention of suffix. Database name can be completely different, but other parts of DATABASE_URL might be always same and you had to define whole string 10 times for 10 databases. I don't know why do you think database name in test environment is a problem here and coming with solution for that will solve all other use cases where user has one option different from DATABASE_URL so he has to have 10 different DATABASE_URL versions. But if you want to have solution which is still .env-first, I would suggest to redirect the focus and effort to https://github.com/symfony/symfony/issues/40717 If that issue was solved, we wouldn't have to introduce this thing here, it would solve it for every library and any part of env string. For example, in use case presented here, you would do

in .env
```
DATABASE_NAME=main
DATABASE_URL=mysql://root:password@database:3306/${DATABASE_NAME}?serverVersion=mariadb-10.5.8
```
in .env.test
```
DATABASE_NAME=${DATABASE_NAME}_test
```

That's the way to go, IMHO. Super flexible and we could stay with env being a source of truth. I guess it would not solve the problem with 10 different databases example, that's where override_options would still be useful I guess, though 🤔

nicolas-grekas:
> Cache doesn't allow anything else than DSN

Using configuration no, but using the constructor yes. The same logic applies when thinking about the desired precedence.

> we could have agreed on something better for both sides before we released the feature

Sorry about that, I missed that PR and it's quite unfortunate it's been merged and released yes. The PR is adding a mandatory deprecation to all projects, asking them all to add a new config option, which breaks consistency and creates risks for projects. That's a very high cost pushed to the community. I wish we can make a new release quickly to save this cost.

> in doctrine_test.yaml [...] and in .env.test

The point is: this doesn't work when using eg `docker-compose` + the `symfony` binary. In this setup, `.env` files are ignored and the DB is fully configured by real env vars. That's really nice for DX but most importantly, it means this has nothing to do with how `.env` files behave. symfony/symfony#40717 is totally unrelated.

This PR is related to #1290 because the only use case of it is adding a suffix to the DB (see how the PR has been introduced, and see also [the corresponding recipe](https://github.com/symfony/recipes/pull/900)).

But #1290 actually creates a risk, because when `override_url` is set, any param is overriding the DSN, not only the dbname. That's going to create hours of WTFs to many I'd bet.

I'm fine splitting this PR in two.

nicolas-grekas:
PR updated, split into https://github.com/doctrine/DoctrineBundle/pull/1331

nicolas-grekas:
Tests need #1331 to be green.

stof:
also, DBAL itself still makes the URL win over options. Only this bundle changes the precedence order when using the new `override_url`

ostrolucky:
> Using configuration no, but using the constructor yes. The same logic applies when thinking about the desired precedence.

It's still similar case I mentioned: It's not possible to instantiate adapter specifying both DSN and specific options at the same time, while DBAL allows that. Hence there is no precedence in Symfony.

> this doesn't work when using eg. docker-compose + the symfony binary. In this setup, .env files are ignored 

I can simply export `SYMFONY_DOTENV_VARS=DATABASE_URL` to solve that. But let's not pretend #1290 had something to do with this problem. Values in .env being ignored when real environment variables are set is a problem of a much, much broader scope than just focusing on dbname in test environment like here. How do I solve this problem for other, non-DATABASE_URL env vars? Well, just do very same thing here.

> also, DBAL itself still makes the URL win over options. Only this bundle changes the precedence order when using the new override_url

That ship has sailed thanks to our hacky ConnectionFactory, but is still conveniently used by people however they see fit. See: from an other side of argument - when introducing stuff like UTF-8 default charset. Now doctrine-bundle maintainers are stuck with bugs such as this one https://github.com/doctrine/DoctrineBundle/issues/1325. But I see it same way, it should have been added to DBAL instead. Hence I will most likely use this case as a precedence and start deprecating all of that crap + enforce such changes not land in bundle anymore - it should be in DBAL, not here.

nicolas-grekas:
> It's not possible to instantiate adapter specifying both DSN and specific options at the same time

Actually yes, all DSN-based factories allow that, see eg https://github.com/symfony/symfony/blob/8361713367e2735fd9b63cd2b6661411a0339811/src/Symfony/Component/Cache/Adapter/AbstractAdapter.php#L127

> export SYMFONY_DOTENV_VARS=DATABASE_URL to solve that

not really: in tests, I do want to use that DSN provided as real env by the `symfony` binary, because eg I don't know on which port Docker made the db container listen.

> How do I solve this problem for other, non-DATABASE_URL env vars?

that's a good question. I don't think there is a generic solution to that, because each network service has it's own concepts. Eg for a DB, we can change the name, but we could also prefix tables. For cache pools, Symfony already implements env-dependent namespaces, which means that there, we prefix cache keys, not dbnames. Etc, it's case by case.

derrabus (approved):

ostrolucky (approved):

jrushlow (approved):

ostrolucky:
can you rebase in order to see test failures are gone?

nicolas-grekas:
Can you please merge 2.3 into 2.4 before?

ostrolucky:
done

ostrolucky:
@wouterj our 2.4 branch is failing with latest symfony deps. Can you check that, since you contributed with https://github.com/doctrine/DoctrineBundle/pull/1297?

wouterj:
I'll have a look!

nicolas-grekas:
Thanks for merging!

PR #1331, opened 2021-04-21
Revert deprecation when "override_url" is not set

nicolas-grekas:
As explained in https://github.com/doctrine/DoctrineBundle/pull/1328

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -27,9 +27,6 @@ use function strlen;
 use function strpos;
 use function strtoupper;
 use function substr;
-use function trigger_error;
-
-use const E_USER_DEPRECATED;
 
 /**
  * This class contains the configuration information for the bundle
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
Note that we should consider deprecating `override_url` in 2.4. I'll submit the PR once this one is merged.

ostrolucky:
`override_url` was merged under the condition that we will introduce deprecation, otherwise I didn't want us to maintain both "modes" forever. If we remove `override_url` that would also solve this problem and I guess I would be fine with that, but I still believe having DATABASE_URL and other options defined together, while those options are completely ignored doesn't make sense and hence is bad UX. So what do you propose here to solve that? Having options not taking any effect is not something that should be acceptable.

nicolas-grekas:
The individual options can act as default values.
But in order to define default values, we can also provide a default `url`.
With this reasoning, I think that I would propose to deprecate all individual options in favor of `url` only.
I can submit a PR on 2.4 if you want once this one is merged.

ostrolucky (approved):

ostrolucky:
I would like that 👍 I'll let this sit here in case there is some feedback from other people till tomorrow, then merge it

jrushlow (approved):

stof:
Well, the issue with is that the `url` in DBAL allows overriding *any* DBAL option (through the query string).

However, supporting only the URL would make the bundle incompatible with hosting providing a standard URL without any DBAL-specific setting (for instance, on postgresql, the URL syntax is an official one). So we still need to be able to configure DBAL-specific options directly, otherwise we are incompatible with all these hosting (heroku for instance, but there might be others exposing a standard postgresql URL rather than a DBAL one)

stof:
but anyway, I agree that this `override_url` setting reversing the precedence of settings compared to DBAL (and causing lots of issues because the bundle config adds default values so this is not only project-provided options that override the URL) was a bad idea.
The dbname suffix suggested in #1328 seems like a better solution for the original need

nicolas-grekas:
See #1335 for the proposed deprecations.

PR #1332, opened 2021-04-23
Support setting PSR-6 caches with ORM 2.9+

alcaeus:
This adds support for doctrine/cache 2.0 as well as the new `setMetadataCache` method being introduced to ORM in https://github.com/doctrine/orm/pull/8651. This removes the need to wrap the configured cache pool in a doctrine/cache wrapper and injects this pool directly. If a service is configured as metadata cache, or ORM < 2.9 is installed, this will continue to use the then deprecated `setMetadataCacheImpl` to inject the cache. It will however wrap the cache pool in the compatibility layer provided by doctrine/cache to remove the need for Symfony to maintain their own adapter for doctrine/cache (see https://github.com/symfony/symfony/pull/40908).

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -724,7 +724,6 @@ class Configuration implements ConfigurationInterface
                 ->scalarNode('type')->defaultNull()->end()
                 ->scalarNode('id')->end()
                 ->scalarNode('pool')->end()
-                ->scalarNode('psr6')->defaultFalse()->end()
             ->end();
 
         if ($name === 'metadata_cache_driver') {
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -724,6 +724,7 @@ class Configuration implements ConfigurationInterface
                 ->scalarNode('type')->defaultNull()->end()
                 ->scalarNode('id')->end()
                 ->scalarNode('pool')->end()
+                ->scalarNode('psr6')->defaultFalse()->end()
             ->end();
 
         if ($name === 'metadata_cache_driver') {
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -21,7 +21,6 @@ use Doctrine\DBAL\Logging\LoggerChain;
 use Doctrine\DBAL\SQLParserUtils;
 use Doctrine\DBAL\Tools\Console\Command\ImportCommand;
 use Doctrine\DBAL\Tools\Console\ConnectionProvider;
-use Doctrine\ORM\Configuration as OrmConfiguration;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Id\AbstractIdGenerator;
 use Doctrine\ORM\Proxy\Autoloader;
@@ -558,16 +557,8 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $ormConfigDef = $container->setDefinition(sprintf('doctrine.orm.%s_configuration', $entityManager['name']), new ChildDefinition('doctrine.orm.configuration'));
         $ormConfigDef->addTag(IdGeneratorPass::CONFIGURATION_TAG);
 
-        $usePsr6             = false;
-        $metadataCacheSetter = 'setMetadataCacheImpl';
-
-        if (method_exists(OrmConfiguration::class, 'setMetadataCache')) {
-            $usePsr6             = true;
-            $metadataCacheSetter = 'setMetadataCache';
-        }
-
         $this->loadOrmEntityManagerMappingInformation($entityManager, $ormConfigDef, $container);
-        $this->loadOrmCacheDrivers($entityManager, $container, $usePsr6);
+        $this->loadOrmCacheDrivers($entityManager, $container);
 
         if (isset($entityManager['entity_listener_resolver']) && $entityManager['entity_listener_resolver']) {
             $container->setAlias(sprintf('doctrine.orm.%s_entity_listener_resolver', $entityManager['name']), $entityManager['entity_listener_resolver']);
@@ -578,7 +569,7 @@ class DoctrineExtension extends AbstractDoctrineExtension
         }
 
         $methods = [
-            $metadataCacheSetter => new Reference(sprintf('doctrine.orm.%s_metadata_cache', $entityManager['name'])),
+            'setMetadataCache' => new Reference(sprintf('doctrine.orm.%s_metadata_cache', $entityManager['name'])),
             'setQueryCacheImpl' => new Reference(sprintf('doctrine.orm.%s_query_cache', $entityManager['name'])),
             'setResultCacheImpl' => new Reference(sprintf('doctrine.orm.%s_result_cache', $entityManager['name'])),
             'setMetadataDriverImpl' => new Reference('doctrine.orm.' . $entityManager['name'] . '_metadata_driver'),
@@ -954,9 +945,9 @@ class DoctrineExtension extends AbstractDoctrineExtension
      *
      * @param array<string, mixed> $entityManager A configured ORM entity manager.
      */
-    protected function loadOrmCacheDrivers(array $entityManager, ContainerBuilder $container, bool $usePsr6)
+    protected function loadOrmCacheDrivers(array $entityManager, ContainerBuilder $container)
     {
-        $this->loadCacheDriver('metadata_cache', $entityManager['name'], $entityManager['metadata_cache_driver'], $container, $usePsr6);
+        $this->loadCacheDriver('metadata_cache', $entityManager['name'], $entityManager['metadata_cache_driver'], $container, true);
         $this->loadCacheDriver('result_cache', $entityManager['name'], $entityManager['result_cache_driver'], $container, false);
         $this->loadCacheDriver('query_cache', $entityManager['name'], $entityManager['query_cache_driver'], $container, false);
 
@@ -964,7 +955,7 @@ class DoctrineExtension extends AbstractDoctrineExtension
             return;
         }
 
-        $this->registerMetadataPhpArrayCaching($entityManager['name'], $container, $usePsr6);
+        $this->registerMetadataPhpArrayCaching($entityManager['name'], $container);
     }
 
     /**
@@ -1083,7 +1074,7 @@ class DoctrineExtension extends AbstractDoctrineExtension
         return $id;
     }
 
-    private function registerMetadataPhpArrayCaching(string $entityManagerName, ContainerBuilder $container, bool $usePsr6): void
+    private function registerMetadataPhpArrayCaching(string $entityManagerName, ContainerBuilder $container): void
     {
         $metadataCacheAlias              = $this->getObjectManagerElementName($entityManagerName . '_metadata_cache');
         $decoratedMetadataCacheServiceId = (string) $container->getAlias($metadataCacheAlias);
@@ -1097,22 +1088,8 @@ class DoctrineExtension extends AbstractDoctrineExtension
 
         $container->setAlias($metadataCacheAlias, $phpArrayCacheDecoratorServiceId);
 
-        if ($usePsr6) {
-            $container->register($phpArrayCacheDecoratorServiceId, PhpArrayAdapter::class)
-                ->addArgument($phpArrayFile)
-                ->addArgument(new Reference($decoratedMetadataCacheServiceId));
-        } else {
-            $container->register($phpArrayCacheDecoratorServiceId, DoctrineProvider::class)
-                ->setFactory([DoctrineProvider::class, 'wrap'])
-                ->addArgument(
-                    new Definition(PhpArrayAdapter::class, [
-                        $phpArrayFile,
-                        // Wrapping the DoctrineProvider created above in a CacheAdapter unwraps the pool from the provider
-                        (new Definition(CacheItemPoolInterface::class))
-                            ->setFactory([CacheAdapter::class, 'wrap'])
-                            ->addArgument(new Reference($decoratedMetadataCacheServiceId)),
-                    ])
-                );
-        }
+        $container->register($phpArrayCacheDecoratorServiceId, PhpArrayAdapter::class)
+            ->addArgument($phpArrayFile)
+            ->addArgument(new Reference($decoratedMetadataCacheServiceId));
     }
 }
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -57,6 +57,7 @@ use function array_intersect_key;
 use function array_keys;
 use function class_exists;
 use function interface_exists;
+use function is_a;
 use function method_exists;
 use function reset;
 use function sprintf;
@@ -557,17 +558,16 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $ormConfigDef = $container->setDefinition(sprintf('doctrine.orm.%s_configuration', $entityManager['name']), new ChildDefinition('doctrine.orm.configuration'));
         $ormConfigDef->addTag(IdGeneratorPass::CONFIGURATION_TAG);
 
-        $wrapLegacyCache     = true;
+        $usePsr6             = false;
         $metadataCacheSetter = 'setMetadataCacheImpl';
 
-        $metadataCacheType = $entityManager['metadata_cache']['type'] ?? 'pool';
-        if (method_exists(OrmConfiguration::class, 'setMetadataCache') && $metadataCacheType !== 'service') {
-            $wrapLegacyCache     = false;
+        if (method_exists(OrmConfiguration::class, 'setMetadataCache')) {
+            $usePsr6             = true;
             $metadataCacheSetter = 'setMetadataCache';
         }
 
         $this->loadOrmEntityManagerMappingInformation($entityManager, $ormConfigDef, $container);
-        $this->loadOrmCacheDrivers($entityManager, $container, $wrapLegacyCache);
+        $this->loadOrmCacheDrivers($entityManager, $container, $usePsr6);
 
         if (isset($entityManager['entity_listener_resolver']) && $entityManager['entity_listener_resolver']) {
             $container->setAlias(sprintf('doctrine.orm.%s_entity_listener_resolver', $entityManager['name']), $entityManager['entity_listener_resolver']);
@@ -892,9 +892,10 @@ class DoctrineExtension extends AbstractDoctrineExtension
     /**
      * {@inheritDoc}
      */
-    protected function loadCacheDriver($cacheName, $objectManagerName, array $cacheDriver, ContainerBuilder $container, bool $wrapLegacyCache = true): string
+    protected function loadCacheDriver($cacheName, $objectManagerName, array $cacheDriver, ContainerBuilder $container, bool $usePsr6 = false): string
     {
         $aliasId = $this->getObjectManagerElementName(sprintf('%s_%s', $objectManagerName, $cacheName));
+        $isPsr6  = null;
 
         switch ($cacheDriver['type'] ?? 'pool') {
             case 'service':
@@ -904,10 +905,8 @@ class DoctrineExtension extends AbstractDoctrineExtension
 
             case 'pool':
                 $pool      = $cacheDriver['pool'] ?? $this->createArrayAdapterCachePool($container, $objectManagerName, $cacheName);
-                $serviceId = $wrapLegacyCache
-                    ? $this->createPoolCacheDefinition($container, $pool)
-                    : $pool;
-
+                $serviceId = $pool;
+                $isPsr6    = true;
                 break;
 
             default:
@@ -917,6 +916,32 @@ class DoctrineExtension extends AbstractDoctrineExtension
                     $cacheName,
                     $objectManagerName
                 ));
+        }
+
+        if ($isPsr6 === null) {
+            $definition = $container->getDefinition($serviceId);
+            $isPsr6     = is_a($definition->getClass(), CacheItemPoolInterface::class, true);
+        }
+
+        $cacheName = str_replace('_cache', '', $cacheName);
+
+        // Create a wrapper as required
+        if ($isPsr6 && ! $usePsr6) {
+            $wrappedServiceId = sprintf('cache.doctrine.orm.provider.%s.%s', $objectManagerName, $cacheName);
+
+            $definition = $container->register($wrappedServiceId, CacheProvider::class);
+            $definition->setFactory([DoctrineProvider::class, 'wrap']);
+            $definition->addArgument(new Reference($serviceId));
+
+            $serviceId = $wrappedServiceId;
+        } elseif (! $isPsr6 && $usePsr6) {
+            $wrappedServiceId = sprintf('cache.doctrine.orm.adapter.%s.%s', $objectManagerName, $cacheName);
+
+            $definition = $container->register($wrappedServiceId, CacheItemPoolInterface::class);
+            $definition->setFactory([CacheAdapter::class, 'wrap']);
+            $definition->addArgument(new Reference($serviceId));
+
+            $serviceId = $wrappedServiceId;
         }
 
         $container->setAlias($aliasId, new Alias($serviceId, false));
@@ -929,17 +954,17 @@ class DoctrineExtension extends AbstractDoctrineExtension
      *
      * @param array<string, mixed> $entityManager A configured ORM entity manager.
      */
-    protected function loadOrmCacheDrivers(array $entityManager, ContainerBuilder $container, bool $wrapLegacyCache = true)
+    protected function loadOrmCacheDrivers(array $entityManager, ContainerBuilder $container, bool $usePsr6)
     {
-        $this->loadCacheDriver('metadata_cache', $entityManager['name'], $entityManager['metadata_cache_driver'], $container, $wrapLegacyCache);
-        $this->loadCacheDriver('result_cache', $entityManager['name'], $entityManager['result_cache_driver'], $container);
-        $this->loadCacheDriver('query_cache', $entityManager['name'], $entityManager['query_cache_driver'], $container);
+        $this->loadCacheDriver('metadata_cache', $entityManager['name'], $entityManager['metadata_cache_driver'], $container, $usePsr6);
+        $this->loadCacheDriver('result_cache', $entityManager['name'], $entityManager['result_cache_driver'], $container, false);
+        $this->loadCacheDriver('query_cache', $entityManager['name'], $entityManager['query_cache_driver'], $container, false);
 
         if ($container->getParameter('kernel.debug')) {
             return;
         }
 
-        $this->registerMetadataPhpArrayCaching($entityManager['name'], $container, $wrapLegacyCache);
+        $this->registerMetadataPhpArrayCaching($entityManager['name'], $container, $usePsr6);
     }
 
     /**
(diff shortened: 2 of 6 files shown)


DependencyInjection/DoctrineExtension.php
```diff
@@ -544,8 +545,17 @@ protected function loadOrmEntityManager(array $entityManager, ContainerBuilder $
         $ormConfigDef = $container->setDefinition(sprintf('doctrine.orm.%s_configuration', $entityManager['name']), new ChildDefinition('doctrine.orm.configuration'));
         $ormConfigDef->addTag(IdGeneratorPass::CONFIGURATION_TAG);
 
+        $wrapLegacyCache     = true;
+        $metadataCacheSetter = 'setMetadataCacheImpl';
+
+        $metadataCacheType = $entityManager['metadata_cache']['type'] ?? 'pool';
+        if (method_exists(OrmConfiguration::class, 'setMetadataCache') && $metadataCacheType !== 'service') {
```
alcaeus: @nicolas-grekas I believe this check is kosher. While users could theoretically change the class being used, ORM requires the configuration class to be of the instance we're checking here. Please let me know if this is not safe to do this way. Thanks!

stof: @alcaeus we might need to add a need configuration setting to support telling DoctrineBundle that the service is a PSR-6 cache pool rather than a doctrine/cache implementation

DependencyInjection/DoctrineExtension.php
```diff
@@ -556,7 +566,7 @@ protected function loadOrmEntityManager(array $entityManager, ContainerBuilder $
         }
 
         $methods = [
-            'setMetadataCacheImpl' => new Reference(sprintf('doctrine.orm.%s_metadata_cache', $entityManager['name'])),
+            $metadataCacheSetter => new Reference(sprintf('doctrine.orm.%s_metadata_cache', $entityManager['name'])),
             'setQueryCacheImpl' => new Reference(sprintf('doctrine.orm.%s_query_cache', $entityManager['name'])),
             'setResultCacheImpl' => new Reference(sprintf('doctrine.orm.%s_result_cache', $entityManager['name'])),
```
stof, line 574: what about the other kind of caches ?

alcaeus: These will not be changed in ORM 2.9.

DependencyInjection/DoctrineExtension.php
alcaeus: I was about to add this, then found out that https://github.com/doctrine/DoctrineBundle/pull/1260 deprecated setting a metadata cache entirely. @ostrolucky @dmaicher @ossinkine can you elaborate on this? From what I can tell, the PhpArrayCache adapter wraps the cache that was configured initially, which relies on setting a cache with the now deprecated key. I'm not sure I understand the whole idea here, so some clarification before adding new functionality would be nice.

dmaicher: @alcaeus so from what I can tell the config for `metadata_cache_driver` was deprecated as it does not change anything for no-debug cases when the `PhpArrayCache` is used.

In theory there cannot be any cache misses as the cache is properly warmed up fully for all metadata with a Symfony CacheWarmer. So the decorated adapter would never be used/accessed. This means even an in-memory cache (ArrayCache/Adpater) will do fine as a fallback.

See https://github.com/doctrine/DoctrineBundle/pull/1196#issuecomment-720124896 and the comment afterwards.

On my apps I removed the `metadata_cache_driver` config for prod completely and I don't have any issues. Things are actually a tiny bit faster now with the `PhpArrayCache`.

alcaeus: Sounds good. PSR-6 caches injected via service caches are now detected automatically and handled accordingly. The PhpArrayCache is used as it previously was. Deprecating the key sounds fine to me.

alcaeus:
@ostrolucky @dmaicher this is ready for another review.

DependencyInjection/DoctrineExtension.php
```diff
@@ -892,6 +906,32 @@ protected function loadCacheDriver($cacheName, $objectManagerName, array $cacheD
...
+        if ($isPsr6 === null) {
+            $definition = $container->getDefinition($serviceId);
+            $isPsr6     = is_a($definition->getClass(), CacheItemPoolInterface::class, true);
+        }
+
+        $cacheName = str_replace('_cache', '', $cacheName);
+
+        // Create a wrapper as required
+        if ($isPsr6 && ! $usePsr6) {
+            $wrappedServiceId = sprintf('cache.doctrine.orm.provider.%s.%s', $objectManagerName, $cacheName);
```
ostrolucky: Why not reusing `createPoolCacheDefinition` function? It's now unused. Either use it or remove it.

DependencyInjection/DoctrineExtension.php
```diff
@@ -902,17 +942,17 @@ protected function loadCacheDriver($cacheName, $objectManagerName, array $cacheD
      *
      * @param array<string, mixed> $entityManager A configured ORM entity manager.
      */
-    protected function loadOrmCacheDrivers(array $entityManager, ContainerBuilder $container)
+    protected function loadOrmCacheDrivers(array $entityManager, ContainerBuilder $container, bool $usePsr6)
```
ostrolucky: How come $usePsr6 was made optional in one protected function and non-optional here?

ostrolucky (changes requested):
BC layer is on too much places, me no bueno. I would rather drop doctrine/orm < 2.9 and doctrine/cache < 2.0

DependencyInjection/DoctrineExtension.php
alcaeus: I missed that. Will give that another pass.

alcaeus:
@ostrolucky can to elaborate what you mean with "BC layer is on too much places" to make the feedback actionable? The BC layer is where it needs to be, there's only so much we can do to cover this.

ostrolucky:
I know, that's why I proposed to bump requirements so we don't need such exhaustive bc layer. Layer would be much smaller if we drop support for those old versions.

alcaeus:
doctrine/cache can't be constrained to 2.0 only, as that makes no sense (why forbid someone from using an implementation when the API is compatible) and would most likely prevent everyone from installing this. As for bumping ORM dependency, this would delay this release since I can't tell when that will be released. This also includes figuring out what to do with https://github.com/doctrine/orm/pull/8672, as ORM 2.9 in its current form is not compatible with doctrine/cache 2.0 at all (since it needs a cache implementation for the setup tool). However, other Doctrine libraries will be released soon, so it's in our interest to not needlessly block other releases. In this case, the compatibility layer is also quite common in Symfony to allow for packages to be updated more independently of each other.

ostrolucky:
Ok well we don't actually need to bump doctrine/cache, since 1.11 is compatible with the changes, only ORM. As for ORM 2.9 release, doctrine-bundle 2.3 branch is compatible with it already so for doctrine-bundle 2.4, people can update ORM to 2.9 first and afterwards doctrine-bundle to 2.4, that's still within symfony goals to not having to update multiple packages at once.

You are right such change might delay release of doctrine-bundle 2.4 as we would have to wait for ORM 2.9 to be released first. But we are in control of that, aren't we? I think @beberlei will manage the next release and I think he is interested to ship it out of the door soon. WDYT @beberlei?

DependencyInjection/DoctrineExtension.php
```diff
@@ -544,8 +548,16 @@ protected function loadOrmEntityManager(array $entityManager, ContainerBuilder $
...
         $ormConfigDef->addTag(IdGeneratorPass::CONFIGURATION_TAG);
 
+        $usePsr6             = false;
+        $metadataCacheSetter = 'setMetadataCacheImpl';
+
+        if (method_exists(OrmConfiguration::class, 'setMetadataCache')) {
+            $usePsr6             = true;
+            $metadataCacheSetter = 'setMetadataCache';
+        }
+
```
alcaeus:
```suggestion
```

DependencyInjection/DoctrineExtension.php
```diff
@@ -544,8 +548,16 @@ protected function loadOrmEntityManager(array $entityManager, ContainerBuilder $
...
         $this->loadOrmEntityManagerMappingInformation($entityManager, $ormConfigDef, $container);
-        $this->loadOrmCacheDrivers($entityManager, $container);
+        $this->loadOrmCacheDrivers($entityManager, $container, $usePsr6);
```
alcaeus:
```suggestion
        $this->loadOrmCacheDrivers($entityManager, $container);
```

DependencyInjection/DoctrineExtension.php
```diff
@@ -556,7 +568,7 @@ protected function loadOrmEntityManager(array $entityManager, ContainerBuilder $
         }
 
         $methods = [
-            'setMetadataCacheImpl' => new Reference(sprintf('doctrine.orm.%s_metadata_cache', $entityManager['name'])),
+            $metadataCacheSetter => new Reference(sprintf('doctrine.orm.%s_metadata_cache', $entityManager['name'])),
```
alcaeus:
```suggestion
            'setMetadataCache' => new Reference(sprintf('doctrine.orm.%s_metadata_cache', $entityManager['name'])),
```

DependencyInjection/DoctrineExtension.php
alcaeus:
```suggestion
    protected function loadOrmCacheDrivers(array $entityManager, ContainerBuilder $container)
```

DependencyInjection/DoctrineExtension.php
```diff
@@ -902,17 +942,17 @@ protected function loadCacheDriver($cacheName, $objectManagerName, array $cacheD
      *
      * @param array<string, mixed> $entityManager A configured ORM entity manager.
      */
-    protected function loadOrmCacheDrivers(array $entityManager, ContainerBuilder $container)
+    protected function loadOrmCacheDrivers(array $entityManager, ContainerBuilder $container, bool $usePsr6)
     {
-        $this->loadCacheDriver('metadata_cache', $entityManager['name'], $entityManager['metadata_cache_driver'], $container);
-        $this->loadCacheDriver('result_cache', $entityManager['name'], $entityManager['result_cache_driver'], $container);
-        $this->loadCacheDriver('query_cache', $entityManager['name'], $entityManager['query_cache_driver'], $container);
+        $this->loadCacheDriver('metadata_cache', $entityManager['name'], $entityManager['metadata_cache_driver'], $container, $usePsr6);
```
alcaeus:
```suggestion
        $this->loadCacheDriver('metadata_cache', $entityManager['name'], $entityManager['metadata_cache_driver'], $container, true);
```

DependencyInjection/DoctrineExtension.php
```diff
@@ -902,17 +942,17 @@ protected function loadCacheDriver($cacheName, $objectManagerName, array $cacheD
...
+        $this->loadCacheDriver('result_cache', $entityManager['name'], $entityManager['result_cache_driver'], $container, false);
+        $this->loadCacheDriver('query_cache', $entityManager['name'], $entityManager['query_cache_driver'], $container, false);
 
         if ($container->getParameter('kernel.debug')) {
             return;
         }
 
-        $this->registerMetadataPhpArrayCaching($entityManager['name'], $container);
+        $this->registerMetadataPhpArrayCaching($entityManager['name'], $container, $usePsr6);
```
alcaeus:
```suggestion
        $this->registerMetadataPhpArrayCaching($entityManager['name'], $container);
```

DependencyInjection/DoctrineExtension.php
```diff
@@ -10,18 +10,23 @@
...
+use Doctrine\Common\Cache\Psr6\CacheAdapter;
+use Doctrine\Common\Cache\Psr6\DoctrineProvider;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Connections\MasterSlaveConnection;
 use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection;
 use Doctrine\DBAL\Logging\LoggerChain;
 use Doctrine\DBAL\SQLParserUtils;
 use Doctrine\DBAL\Tools\Console\Command\ImportCommand;
 use Doctrine\DBAL\Tools\Console\ConnectionProvider;
+use Doctrine\ORM\Configuration as OrmConfiguration;
```
alcaeus:
```suggestion
```

DependencyInjection/DoctrineExtension.php
```diff
@@ -1030,7 +1071,7 @@ private function createArrayAdapterCachePool(ContainerBuilder $container, string
         return $id;
     }
 
-    private function registerMetadataPhpArrayCaching(string $entityManagerName, ContainerBuilder $container): void
+    private function registerMetadataPhpArrayCaching(string $entityManagerName, ContainerBuilder $container, bool $usePsr6): void
```
alcaeus:
```suggestion
    private function registerMetadataPhpArrayCaching(string $entityManagerName, ContainerBuilder $container): void
```

DependencyInjection/DoctrineExtension.php
```diff
@@ -1043,12 +1084,23 @@ private function registerMetadataPhpArrayCaching(string $entityManagerName, Cont
...
+                ->addArgument(
+                    new Definition(PhpArrayAdapter::class, [
+                        $phpArrayFile,
+                        // Wrapping the DoctrineProvider created above in a CacheAdapter unwraps the pool from the provider
+                        (new Definition(CacheItemPoolInterface::class))
+                            ->setFactory([CacheAdapter::class, 'wrap'])
+                            ->addArgument(new Reference($decoratedMetadataCacheServiceId)),
+                    ])
+                );
+        }
```
alcaeus:
```suggestion
        $container->register($phpArrayCacheDecoratorServiceId, PhpArrayAdapter::class)
            ->addArgument($phpArrayFile)
            ->addArgument(new Reference($decoratedMetadataCacheServiceId));
```

Tests/ContainerTest.php
```diff
@@ -61,7 +63,13 @@ public function testContainer(): void
...
         $this->assertInstanceOf(Configuration::class, $container->get('doctrine.orm.default_configuration'));
         $this->assertInstanceOf(MappingDriverChain::class, $container->get('doctrine.orm.default_metadata_driver'));
-        $this->assertInstanceOf(DoctrineProvider::class, $container->get('doctrine.orm.default_metadata_cache'));
+
+        if (method_exists(Configuration::class, 'setMetadataCache')) {
+            $this->assertInstanceOf(PhpArrayAdapter::class, $container->get('doctrine.orm.default_metadata_cache'));
+        } else {
+            $this->assertInstanceOf(DoctrineProvider::class, $container->get('doctrine.orm.default_metadata_cache'));
+        }
+
```
alcaeus:
```suggestion
        $this->assertInstanceOf(PhpArrayAdapter::class, $container->get('doctrine.orm.default_metadata_cache'));
```

Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -426,16 +428,20 @@ public function testLoadMultipleConnections(): void
         $this->assertEquals('doctrine.orm.em2_configuration', (string) $arguments[1]);
 
         $definition = $container->getDefinition((string) $container->getAlias('doctrine.orm.em1_metadata_cache'));
-        $this->assertEquals(DoctrineProvider::class, $definition->getClass());
+        if (method_exists(OrmConfiguration::class, 'setMetadataCache')) {
+            $this->assertEquals(PhpArrayAdapter::class, $definition->getClass());
+        } else {
+            $this->assertEquals([DoctrineProvider::class, 'wrap'], $definition->getFactory());
+        }
```
alcaeus:
```suggestion
        $this->assertEquals(PhpArrayAdapter::class, $definition->getClass());
```

Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -470,10 +476,18 @@ public function testEntityManagerMetadataCacheDriverConfiguration(): void
         $container = $this->loadContainer('orm_service_multiple_entity_managers');
 
         $definition = $container->getDefinition((string) $container->getAlias('doctrine.orm.em1_metadata_cache'));
-        $this->assertDICDefinitionClass($definition, DoctrineProvider::class);
+        if (method_exists(OrmConfiguration::class, 'setMetadataCache')) {
+            $this->assertEquals(PhpArrayAdapter::class, $definition->getClass());
+        } else {
+            $this->assertEquals([DoctrineProvider::class, 'wrap'], $definition->getFactory());
+        }
```
alcaeus:
```suggestion
        $this->assertDICDefinitionClass($definition, PhpArrayAdapter::class);
```

Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -470,10 +476,18 @@ public function testEntityManagerMetadataCacheDriverConfiguration(): void
...
 
         $definition = $container->getDefinition((string) $container->getAlias('doctrine.orm.em2_metadata_cache'));
-        $this->assertDICDefinitionClass($definition, DoctrineProvider::class);
+        if (method_exists(OrmConfiguration::class, 'setMetadataCache')) {
+            $this->assertEquals(PhpArrayAdapter::class, $definition->getClass());
+        } else {
+            $this->assertEquals([DoctrineProvider::class, 'wrap'], $definition->getFactory());
+        }
```
alcaeus:
```suggestion
        $this->assertDICDefinitionClass($definition, PhpArrayAdapter::class);
```

Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -484,11 +498,17 @@ public function testDependencyInjectionImportsOverrideDefaults(): void
...
         $this->assertDICDefinitionMethodCallOnce($configDefinition, 'setAutoGenerateProxyClasses', ['%doctrine.orm.auto_generate_proxy_classes%']);
+
+        $cacheDefinition = $container->getDefinition((string) $container->getAlias('doctrine.orm.default_metadata_cache'));
+        if (method_exists(OrmConfiguration::class, 'setMetadataCache')) {
+            $this->assertEquals(PhpArrayAdapter::class, $cacheDefinition->getClass());
+            $this->assertDICDefinitionMethodCallOnce($configDefinition, 'setMetadataCache', [new Reference('doctrine.orm.default_metadata_cache')]);
+        } else {
+            $this->assertEquals([DoctrineProvider::class, 'wrap'], $cacheDefinition->getFactory());
+            $this->assertDICDefinitionMethodCallOnce($configDefinition, 'setMetadataCacheImpl', [new Reference('doctrine.orm.default_metadata_cache')]);
+        }
```
alcaeus:
```suggestion
        $this->assertEquals(PhpArrayAdapter::class, $cacheDefinition->getClass());
        $this->assertDICDefinitionMethodCallOnce($configDefinition, 'setMetadataCache', [new Reference('doctrine.orm.default_metadata_cache')]);
```

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -442,34 +445,52 @@ public function testDependencyInjectionConfigurationDefaults(): void
             $definition->getArguments()
         );
 
+        $wrapsLegacyCache = ! method_exists(Configuration::class, 'setMetadataCache');
+
```
alcaeus:
```suggestion
```

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -28,6 +30,7 @@
 use function array_values;
 use function class_exists;
 use function interface_exists;
+use function method_exists;
```
alcaeus:
```suggestion
```

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -442,34 +445,52 @@ public function testDependencyInjectionConfigurationDefaults(): void
...
+
         $definition = $container->getDefinition((string) $container->getAlias('doctrine.orm.default_metadata_cache'));
-        $this->assertEquals(DoctrineProvider::class, $definition->getClass());
+        if ($wrapsLegacyCache) {
+            $this->assertEquals(DoctrineProvider::class, $definition->getClass());
+            $arguments  = $definition->getArguments();
+            $definition = $arguments[0];
+            $this->assertInstanceOf(Definition::class, $definition);
+        }
+
```
alcaeus:
```suggestion
```

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -442,34 +445,52 @@ public function testDependencyInjectionConfigurationDefaults(): void
...
+        if ($wrapsLegacyCache) {
+            $this->assertInstanceOf(Definition::class, $arguments[1]);
+            $this->assertEquals([CacheAdapter::class, 'wrap'], $arguments[1]->getFactory());
+            $wrappedDefinition = $arguments[1]->getArguments()[0];
+            $this->assertInstanceOf(Reference::class, $wrappedDefinition);
+            $this->assertEquals('cache.doctrine.orm.provider.default.metadata', (string) $wrappedDefinition);
+        } else {
+            $this->assertInstanceOf(Reference::class, $arguments[1]);
+            $wrappedDefinition = (string) $arguments[1];
+        }
```
alcaeus:
```suggestion
        $this->assertInstanceOf(Reference::class, $arguments[1]);
        $wrappedDefinition = (string) $arguments[1];
```

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -442,34 +445,52 @@ public function testDependencyInjectionConfigurationDefaults(): void
...
+        $definition = $container->getDefinition((string) $wrappedDefinition);
+
+        if ($wrapsLegacyCache) {
+            $this->assertEquals([DoctrineProvider::class, 'wrap'], $definition->getFactory());
+            $arguments = $definition->getArguments();
+            $this->assertInstanceOf(Reference::class, $arguments[0]);
+            $this->assertEquals('cache.doctrine.orm.default.metadata', (string) $arguments[0]);
+            $definition = $container->getDefinition((string) $arguments[0]);
+        }
+
```
alcaeus:
```suggestion
```

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -928,25 +949,41 @@ public function testLegacyCacheConfiguration(string $expectedAliasName, string $
      */
     public static function legacyCacheConfigurationProvider(): array
     {
+        $usePsr6 = method_exists(Configuration::class, 'setMetadataCache');
+
```
alcaeus:
```suggestion
```

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -928,25 +949,41 @@ public function testLegacyCacheConfiguration(string $expectedAliasName, string $
...
         return [
             'metadata_cache_default' => [
                 'expectedAliasName' => 'doctrine.orm.default_metadata_cache',
-                'expectedAliasTarget' => 'doctrine.orm.cache.provider.cache.doctrine.orm.default.metadata.php_array',
+                'expectedAliasTarget' => $usePsr6
+                    ? 'cache.doctrine.orm.default.metadata.php_array'
+                    : 'cache.doctrine.orm.provider.default.metadata.php_array',
```
alcaeus:
```suggestion
                'expectedAliasTarget' => 'cache.doctrine.orm.default.metadata.php_array',
```

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -928,25 +949,41 @@ public function testLegacyCacheConfiguration(string $expectedAliasName, string $
...
                 'cacheName' => 'metadata_cache_driver',
                 'cacheConfig' => ['type' => null],
             ],
             'metadata_cache_pool' => [
                 'expectedAliasName' => 'doctrine.orm.default_metadata_cache',
-                'expectedAliasTarget' => 'doctrine.orm.cache.provider.metadata_cache_pool.php_array',
+                'expectedAliasTarget' => $usePsr6
+                    ? 'metadata_cache_pool.php_array'
+                    : 'cache.doctrine.orm.provider.default.metadata.php_array',
```
alcaeus:
```suggestion
                'expectedAliasTarget' => 'metadata_cache_pool.php_array',
```

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -928,25 +949,41 @@ public function testLegacyCacheConfiguration(string $expectedAliasName, string $
...
                 'cacheName' => 'metadata_cache_driver',
                 'cacheConfig' => ['type' => 'pool', 'pool' => 'metadata_cache_pool'],
             ],
             'metadata_cache_service' => [
                 'expectedAliasName' => 'doctrine.orm.default_metadata_cache',
-                'expectedAliasTarget' => 'service_target_metadata.php_array',
+                'expectedAliasTarget' => $usePsr6
+                    ? 'cache.doctrine.orm.adapter.default.metadata.php_array'
+                    : 'service_target_metadata.php_array',
```
alcaeus:
```suggestion
                'expectedAliasTarget' => 'cache.doctrine.orm.adapter.default.metadata.php_array',
```

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -928,25 +949,41 @@ public function testLegacyCacheConfiguration(string $expectedAliasName, string $
...
                 'cacheName' => 'metadata_cache_driver',
                 'cacheConfig' => ['type' => 'service', 'id' => 'service_target_metadata'],
             ],
+            'metadata_psr6_service' => [
+                'expectedAliasName' => 'doctrine.orm.default_metadata_cache',
+                'expectedAliasTarget' => $usePsr6
+                    ? 'service_target_metadata_psr6.php_array'
+                    : 'cache.doctrine.orm.provider.default.metadata.php_array',
```
alcaeus:
```suggestion
                'expectedAliasTarget' => 'service_target_metadata_psr6.php_array',
```

alcaeus (commented):
Went through the diff to see what we could save if we dropped ORM < 2.9. While it doesn't really reduce the amount of logic in the extension itself, the fact that the tests are no longer conditional makes this change worth it IMO, regardless of the release timeline for ORM 2.9.

alcaeus:
@ostrolucky dropped ORM < 2.8, please let me know what else we might need.

DependencyInjection/Configuration.php
```diff
@@ -724,6 +724,7 @@ private function getOrmCacheDriverNode(string $name): ArrayNodeDefinition
                 ->scalarNode('type')->defaultNull()->end()
                 ->scalarNode('id')->end()
                 ->scalarNode('pool')->end()
+                ->scalarNode('psr6')->defaultFalse()->end()
```
nicolas-grekas: Is there a way around adding the option?

DependencyInjection/DoctrineExtension.php
ostrolucky: I think this comment still applies. Also, are you sure with this naming? On most of the places we have doctrine.orm prefix

ostrolucky (commented):
Shouldn't we deprecate using non-psr6 caches?

DependencyInjection/Configuration.php
alcaeus: Apologies, this slipped in from during the refactoring. I've removed it.

DependencyInjection/DoctrineExtension.php
alcaeus: The unused `createPoolCacheDefinition` was removed and the cache name was changed for consistency.

ostrolucky (approved):

PR #1334, opened 2021-04-23
Fix tests compatibility with doctrine-bridge 5.x

wouterj:
With DoctrineBridge 5.x, there are 3 listeners instead of 2. The `attach_entity_listener` is always the last one.

cc @ostrolucky


ostrolucky:
thx

PR #1335, opened 2021-04-25
Deprecate mixing the `url` option with `host`, `dbname` and a few others

nicolas-grekas:
As proposed in #1331

For discussion. (I'll take care of tests once the discussion has settled.)

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -14,6 +14,7 @@ use Symfony\Component\DependencyInjection\Exception\LogicException;
 
 use function array_intersect_key;
 use function array_keys;
+use function array_pop;
 use function assert;
 use function class_exists;
 use function constant;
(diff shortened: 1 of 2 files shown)


ostrolucky:
You mean #1331, not #1334, right?

@stof I'm not quite sure what you meant with https://github.com/doctrine/DoctrineBundle/pull/1331#issuecomment-824912640, is that still a problem in this PR?

I like this PR though, IMO it's ok to finish it.

nicolas-grekas:
I extracted the part about `override_url` in #1342 because deprecating the other options requires updating a lot of tests, and also because the discussion should be split apart.

stof:
@ostrolucky this PR does not deprecate options like `server_version`, only the ones corresponding to the standard parts of the URL (the query string being used by DBAL for extra options is something totally specific to DBAL, relying on DBAL option names).
So it can be fine with me.

However, we should be careful when writing the upgrade guide (asking people to dynamically create a URL in the config file from separate username, host, password and db name is quite fragile, as it would probably miss the need for URL escaping, which will only trigger a bug after some time when rotating the password to one using special URL chars for instance)

ostrolucky:
Just a reminder that after https://github.com/doctrine/DoctrineBundle/pull/1331 has been merged, we expect this to be finished next.

nicolas-grekas:
I went with a more conservative approach that should provide the same benefit:

what is now deprecated is mixing the `url` option with `host`, `dbname` and a few others.

ostrolucky:
Yep that's good compromise I think

DependencyInjection/Configuration.php
```diff
@@ -245,12 +246,17 @@ private function configureDbalDriverNode(ArrayNodeDefinition $node): void
         $node
             ->validate()
             ->always(static function (array $values) {
-                $deprecatedOptions = ['override_url' => true];
-                $deprecatedValues  = array_intersect_key($values, $deprecatedOptions);
+                if (! isset($values['url'])) {
+                    return $values;
+                }
+
+                $urlConflictingOptions = ['host' => true, 'port' => true, 'user' => true, 'password' => true, 'path' => true, 'dbname' => true, 'unix_socket' => true, 'memory' => true];
```
ostrolucky, line 253: I think driver is missing, isn't it?

nicolas-grekas: `driver` is in another branch of the config tree.
I don't think that's an issue.

ostrolucky (approved):

ostrolucky:
thx @nicolas-grekas

PR #1342, opened 2021-05-10
Deprecate `override_url` configuration option

nicolas-grekas:
Extracted from #1335

### DIFF
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -16,7 +16,9 @@ use Doctrine\DBAL\Types\Type;
 use function array_merge;
 use function class_exists;
 use function is_subclass_of;
+use function trigger_error;
 
+use const E_USER_DEPRECATED;
 use const PHP_EOL;
 
 /** @psalm-import-type Params from DriverManager */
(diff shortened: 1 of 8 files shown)


DependencyInjection/Configuration.php
```diff
@@ -231,14 +239,27 @@ private function getDbalConnectionsNode(): ArrayNodeDefinition
...
     {
         $node
+            ->validate()
+            ->always(static function (array $values) {
+                $deprecatedOptions = ['override_url' => true];
+                $deprecatedValues  = array_intersect_key($values, $deprecatedOptions);
+
+                if ($deprecatedValues) {
+                    $message = count($deprecatedValues) > 1 ? 'options are' : 'option is';
+                    @trigger_error(sprintf('The "doctrine.dbal.%s" %s deprecated since DoctrineBundle 2.4, use the "doctrine.dbal.url" option instead.', implode('", "doctrine.dbal.', array_keys($deprecatedValues)), $message), E_USER_DEPRECATED);
```
stof, line 247: why not using `setDeprecated()` on the boolean node instead ?

stof (changes requested):

DependencyInjection/Configuration.php
nicolas-grekas: To reduce the diff with # #1335, since there, we'll want to group deprecations in one.

ConnectionFactory.php
```diff
@@ -54,8 +56,13 @@ public function createConnection(array $params, ?Configuration $config = null, ?
...
         }
 
-        $overriddenOptions = $params['connection_override_options'] ?? [];
-        unset($params['connection_override_options']);
+        if (isset($params['connection_override_options'])) {
+            @trigger_error('The "connection_override_options" connection parameter is deprecated since DoctrineBundle 2.4.', E_USER_DEPRECATED);
+            $overriddenOptions = $params['connection_override_options'];
+            unset($params['connection_override_options']);
+        } else {
+            $overriddenOptions = [];
```
ostrolucky: Why else? Make it default instead

ostrolucky (changes requested):
Can you please rebase, there were some cs changes and merging this PR would make target branch failing

ConnectionFactory.php
nicolas-grekas: updated

ostrolucky (approved):

Issue #1337, opened 2021-05-03
`default_entity_manager` config param is not inherited from parent config file


bravik:
This line in Configuratation:
https://github.com/doctrine/DoctrineBundle/blob/09339f7df6fc620ce805f52ca0c7424543d30515/DependencyInjection/Configuration.php#L376

makes `default_entity_manager` (=<some_none_default_value>) param defined in `config/packages/doctrine.yml`
been overriden by `config/packages/prod/doctrine.yml` without this option.
Instead of the value from `config/packages/doctirne.yml` we have `default` value.
Which is conter-intuitive and took a couple of time to investigate mapping issues.

Example:
```yaml
#c onfig/packages/doctrine.yml
doctrine:
  # ...
    orm:
        auto_generate_proxy_classes: true
        default_entity_manager: app
        entity_managers:
            app:
               # ...
            module:
               #...
```

```yaml
#config/packages/prod/doctrine.yml
# The default one, generated by flex

doctrine:
    orm:
        auto_generate_proxy_classes: false
        metadata_cache_driver:
            type: pool
            pool: doctrine.system_cache_pool
        query_cache_driver:
            type: pool
            pool: doctrine.system_cache_pool
        result_cache_driver:
            type: pool
            pool: doctrine.result_cache_pool
```

Expected: `app`
Given: `default`

ostrolucky:
Can be also tested via `bin/console debug:container --parameter=doctrine.default_entity_manager --env=prod` and `bin/console debug:container --parameter=doctrine.default_entity_manager --env=dev`, respectively. 

If anyone want to work on this, I think solution would be to extract this normalization https://github.com/doctrine/DoctrineBundle/blob/d6b3c37804539a24ba8a7d647a6144cab2f13242/DependencyInjection/Configuration.php#L376-L377 to DoctrineExtension

ostrolucky:
Patch has been merged into 2.4. It should prevent similar issues in future

PR #1347, opened 2021-05-15
Deprecate not defining entity_managers and connections due merge issues 

ostrolucky:
Context: https://github.com/doctrine/DoctrineBundle/issues/1337

### Description:
This reverts ab047c6f24420c44c7ad08cf2487ab47d4cba46d and ae0a3c649af899f95a2f45fa7dca79ba6a7b27c9

What wasn't accounted previously when deciding to implement these simplifications are issues when merging. Given following relevant parts of configs:

/doctrine.yaml
```yaml
doctrine:
    orm:
        default_entity_manager: app
        entity_managers:
            app:
            extra_manager:
```
/prod/doctrine.yaml
```yaml
doctrine:
    orm:
        metadata_cache_driver:
```

there are 2 issues:

* Issue no. 1:  When APP_ENV=prod, default_entity_manager is changed to `default`. And doctrine.orm.default.entity_manager is created, despite explicitly used configuration not using it
* Issue no. 2: When reading /prod/doctrine.yaml, instinctively, since there is no entity manager specified there, developers have impression options defined directly under doctrine.orm.* will be applied to all entity managers. But that's not what bundles does. Instead, it applies it to default entity manager only

These are the limitations of symfony/config component. When evaluating /prod/doctrine.yaml, in Configuration.php we don't have an access to values from /doctrine.yaml

### Status of this PR

This is not a finished PR, couple more deprecation fixes need to be done in our tests, as well as updates in .rst files. I will leave this here unfinished for couple days to give some room for possible discussion about alternatives, after that I'll finish it.

What I don't want to do is doing the job of Config component (merging and normalization) ourselves in DoctrineExtension, as that may be error prone. And even if we did, we can't fix issue no. 2 without breaking BC.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -89,6 +89,10 @@ class Configuration implements ConfigurationInterface
                             unset($v[$key]);
                         }
 
+                        if ($connection) {
+                            @trigger_error('Not defining "doctrine.dbal.connections" configuration option is deprecated since doctrine-bundle 2.4.', E_USER_DEPRECATED);
+                        }
+
                         $v['default_connection'] = isset($v['default_connection']) ? (string) $v['default_connection'] : 'default';
                         $v['connections']        = [$v['default_connection'] => $connection];
 
@@ -244,7 +248,7 @@ class Configuration implements ConfigurationInterface
 
                 if ($deprecatedValues) {
                     $message = count($deprecatedValues) > 1 ? 'options are' : 'option is';
-                    @trigger_error(sprintf('The "doctrine.dbal.%s" %s deprecated since DoctrineBundle 2.4, use the "doctrine.dbal.url" option instead.', implode('", "doctrine.dbal.', array_keys($deprecatedValues)), $message), E_USER_DEPRECATED);
+                    @trigger_error(sprintf('The "doctrine.dbal.%s" %s deprecated since doctrine-bundle 2.4, use the "doctrine.dbal.url" option instead.', implode('", "doctrine.dbal.', array_keys($deprecatedValues)), $message), E_USER_DEPRECATED);
                 }
 
                 return $values;
@@ -391,6 +395,10 @@ class Configuration implements ConfigurationInterface
 
                                 $entityManager[$key] = $v[$key];
                                 unset($v[$key]);
+                            }
+
+                            if ($entityManager) {
+                                @trigger_error('Not defining "doctrine.orm.entity_managers" configuration option is deprecated since doctrine-bundle 2.4.', E_USER_DEPRECATED);
                             }
 
                             $v['default_entity_manager'] = isset($v['default_entity_manager']) ? (string) $v['default_entity_manager'] : 'default';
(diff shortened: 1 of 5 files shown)


greg0ire:
> This reverts ab047c6 and ae0a3c6

It does not look like it does, but my guess is that you are going to do that in a subsequent PR targeting the next major branch, correct? What's weird is that both commits are very ancient (10 years old a few weeks ago lol :birthday: ), it's hard to believe that this bug is only spotted after 10 years :thinking: 

Also, I see the documentation for the config component mentions merging options that could be helpful here: https://symfony.com/doc/current/components/config/definition.html#merging-options

`cannotBeOverwritten()` does not seem to be in use in this bundle, but could be useful for `default_entity_manager` maybe, in addition to what you plan to do.

ostrolucky:
> you are going to do that in a subsequent PR targeting the next major branch, correct? 

Same approach like with all other deprecations here will be applied. Deprecate first and remove when we start working on next major release. Take revert keyword here as a beginning of undoing the adding of such true.

> it's hard to believe that this bug is only spotted after 10 years 🤔

I guess most people stick to having default entity manager called "default". So deprecating `default_entity_manager` option would be another "solution" for problem no. 1. But it still doesn't solve problem no. 2 and I think it would also be quite limitating to force people to use `default` name even if they don't want to.  People who do run into this issue probably just didn't report the issue because it's easy enough to workaround it in userspace by defining enity_managers key.

> cannotBeOverwritten() does not seem to be in use in this bundle, but could be useful for default_entity_manager maybe, in addition to what you plan to do.

That's a nice thought, but after thinking about this I can't really see why should we prevent users to overwrite some options. I don't actually see a use case for that symfony/config feature. If users want to have different default manager in prod, whatever, let them, not sure why should anyone restrict that. Anyways, this isn't only about default manager being overwritten. Configuration cannot detect which em is default, so it applies options in example prod file I gave to wrong em.

greg0ire:
> I will leave this here unfinished for couple days to give some room for possible discussion about alternatives

I see why you are doing that, deprecate this will require changes in many (most?) applications :grimacing: 

Maybe a suitable solution would be this to forbid mixing simplified and "complicated" configs:

```php
->ifTrue(function ($v) use (&$defaultEntityManager) {
    static $hasSimplifiedConfig = null;
    $isSimplified = is_array($v) && !array_key_exists('connections', $v) && !array_key_exists('connection', $v);
    if (isset($hasSimplifiedConfig && $hasSimplifiedConfig xor $isSimplified) {
        @trigger_error('Either use simplified syntax in all Doctrine configuration files, or in none of them', E_USER_DEPRECATED);
    }

    $hasSimplifiedConfig = $isSimplified;

    return $isSimplified; 
})
```

ostrolucky:
Cool idea, I like that over my solution, although I wish there was a way to do it without relying on global state.

greg0ire:
> Cool idea, I like that over my solution, although I wish there was a way to do it without relying on global state.

I think you might be able to access all the configuration arrays at once and perform that check in an implementation of `PrependExtensionInterface`: accessing all the arrays should be as simple as `$configs = $container->getExtensionConfig('doctrine');`.

https://symfony.com/doc/current/bundles/prepend_extension.html

That interface is meant for another use case (configuring one bundle with another), but I think it could be a solution for our use case as well.

ostrolucky:
Successor is https://github.com/doctrine/DoctrineBundle/pull/1356

PR #1353, opened 2021-05-25
Register `AttributeDriver` in configuration

romm:

Issue #1348, opened 2021-05-19
type: service caches are broken in 2.4


alexander-schranz:
I currently did test @sulu against current `dev` dependencies and we have the following configured in the `prod/doctrine.yaml`:

<details>

<summary>doctrine.yaml</summary>

```yaml
doctrine:
    orm:
        auto_generate_proxy_classes: false
        metadata_cache_driver:
            type: service
            id: doctrine.system_cache_provider
        query_cache_driver:
            type: service
            id: doctrine.system_cache_provider
        result_cache_driver:
            type: service
            id: doctrine.result_cache_provider

services:
    doctrine.result_cache_provider:
        class: Symfony\Component\Cache\DoctrineProvider
        public: false
        arguments:
            - '@doctrine.result_cache_pool'
    doctrine.system_cache_provider:
        class: Symfony\Component\Cache\DoctrineProvider
        public: false
        arguments:
            - '@doctrine.system_cache_pool'

framework:
    cache:
        pools:
            doctrine.result_cache_pool:
                adapter: cache.app
            doctrine.system_cache_pool:
                adapter: cache.system
```

</details>

> You have requested a non-existent service "doctrine.system_cache_provider".  

https://github.com/sulu/sulu/runs/2620361693

I think maybe there was a bc break introduced in the current current dev version, so I thought I will report this here. Updating the `doctrine.yaml` to something newer like this did fix it for us:

<details>

<summary>doctrine.yaml</summary>

```yaml
doctrine:
    orm:
        auto_generate_proxy_classes: false
        metadata_cache_driver:
            type: pool
            pool: doctrine.system_cache_pool
        query_cache_driver:
            type: pool
            pool: doctrine.system_cache_pool
        result_cache_driver:
            type: pool
            pool: doctrine.result_cache_pool

framework:
    cache:
        pools:
            doctrine.result_cache_pool:
                adapter: cache.app
            doctrine.system_cache_pool:
                adapter: cache.system
```

</details>

So its not an issue for us currently as we did update the skeleton some time ago just thought I would report it that there was unexpected change which could maybe crash older projects an older cache configuration.

stof:
@alcaeus we should re-add the `doctrine.system_cache_provider` service as a deprecated one to avoid breaking existing apps

alcaeus:
@alexander-schranz can you please test the current stable release with the same config? While we did make changes to metadata caching, I fail to see how those would be responsible for having a service disappear, let alone a service that you defined yourself.

[...]

PR #1354, opened 2021-05-25
Add "attribute" type to "mapping" for AttributeDriver

simivar:
In the new Doctrine ORM 2.9 there is the possibility to use Attributes to map entities but in this bundle, there's no `type` defined to use `AttributeDriver`


simivar:
I see that it's a duplicate of #1353 and works is already done in #1322. What's stopping the release of `2.4` and how? @ostrolucky maybe you would know as I saw you mentioned waiting for the new Symfony release (which one?).

ostrolucky:
https://github.com/doctrine/DoctrineBundle/issues/1348 is stopping it

PR #1356, opened 2021-05-25
Disallow mixed use of simplified and non-simplified configuration 

ostrolucky:
Continuation from #1347.

This is to resolve issues like reported in #1337 and https://github.com/doctrine/orm/discussions/8687

I did go with static variable, because with PrependExtension, I would get values pre-before-normalization. So eg. `entity-manager` instead of `entity_managers`. This logic is in [protected symfony/config method](https://github.com/symfony/symfony/blob/27b165434f32e71be01fcd8a491bc066ff059a1b/src/Symfony/Component/Config/Definition/ArrayNode.php#L49) and I didn't want to duplicate it. 😵

___

Simplified use was added in ab047c6 and ae0a3c6

What wasn't accounted previously when deciding to implement these simplifications are issues when merging. Given following relevant parts of configs:

/doctrine.yaml
```yaml
doctrine:
    orm:
        default_entity_manager: app
        entity_managers:
            app:
            extra_manager:
```
/prod/doctrine.yaml
```yaml
doctrine:
    orm:
        metadata_cache_driver:
```

there are 2 issues:

* Issue no. 1:  When APP_ENV=prod, default_entity_manager is changed to `default`. And `doctrine.orm.default.entity_manager` is created, despite explicitly used configuration not specifying any entity manager named `default`
* Issue no. 2: When reading /prod/doctrine.yaml, instinctively, since there is no entity manager specified there, developers have impression options defined directly under doctrine.orm.* will be applied to all entity managers. But that's not what bundles does. Instead, it applies it to default entity manager only

These are the limitations of symfony/config component. When evaluating /prod/doctrine.yaml, in Configuration.php we don't have an API to access values from /doctrine.yaml.

Originally, I inteded to deprecate simplified configuration completely, but due to wide BC impact let's at least disallowed mixed use, which is where the problem occurs.


greg0ire (approved):
Do you feel confident about using an exception and not a deprecation here? Approving in case it's not a concern.

ostrolucky:
I went with exception because I'm betting that people who use this mixed approach most likely already experience a bug they just didn't discover yet. For example, people don't tend to touch prod/doctrine.yaml, but default one counts with single entity manager only. So when people need second EM, they modify doctrine.yaml but keep prod/doctrine.yaml untouched. But if they didn't fix prod/doctrine.yaml, that means things like caches were not configured correctly for this second EM and they are unintentionally using default array caches there.

Webonaute:
I dont get this change. I have this config right now and getting that new exception... 

```
doctrine:
  dbal:
    logging: false
    driver: "%db_driver%"
    url: '%env(DATABASE_URL)%'
    replicas:
      lbslave:
        url: "%env(DATABASE_URL_SLAVE)%"
    charset: UTF8
    driver_class: 'App\Component\Doctrine\DBAL\Driver\Mysqli\Driver'
    wrapper_class: 'App\Component\Doctrine\DBAL\Connections\MasterSlaveConnection'
    platform_service: 'Webonaute\DoctrineMysqlExtra\Doctrine\DBAL\Platforms\MySql57Platform'
    server_version: '5.7'
    mapping_types:
      uuid_binary: binary
      enum: string
    keep_replica: true
    #options:
    #    1002: "SET NAMES 'UTF8' COLLATE 'utf8_general_ci'"
    types:
      #json: Sonata\Doctrine\Types\JsonType
      numeric: Webonaute\DoctrineMysqlExtra\Doctrine\DBAL\Types\NumericType
      cron_expression: Webonaute\DoctrineMysqlExtra\Doctrine\DBAL\Types\CronExpressionType
      uuid_binary: Ramsey\Uuid\Doctrine\UuidBinaryType
  orm:
    auto_generate_proxy_classes: '%doctrine_auto_generate_proxy_classes%' #with Swoole, setting this to true generete ErrorException("No such file or directory").
    default_entity_manager: default
    resolve_target_entities: { }
    entity_managers:
      default:
        metadata_cache_driver:
          id: cache.app
        query_cache_driver:
          id: cache.app
        result_cache_driver:
          type: pool
          pool: doctrine.redis_cache_pool
        auto_mapping: true
        dql:
          datetime_functions:
            now: Webonaute\DoctrineMysqlExtra\Doctrine\ORM\Query\Now
            utc_timestamp: Webonaute\DoctrineMysqlExtra\Doctrine\ORM\Query\UtcTimestamp
        mappings:
          app:
            type: annotation
            is_bundle: false
            prefix: App\Entity
            alias: App
        filters:
          softdeleteable:
            class: 'Gedmo\SoftDeleteable\Filter\SoftDeleteableFilter'
            enabled: true
```

what do I need to change to fix this?

ostrolucky:
First, you need to identify other configuration files where you configure doctrine.dbal

Webonaute:
There is no other file configuring doctrine.dbal. only one in config/packages/doctrine.yaml.

ostrolucky:
Then you must debug what's going on yourself. I don't see how can this be triggered otherwise.

Webonaute:
I actually found why this get mess up. I have this prependExtensionConfig in some dependencies injection. How can I go around to configure it properly? This code was to remove SonataMediaBundle mapping to add our own custom mapping.

```
        $container->prependExtensionConfig(
            'doctrine',
            [
                'orm' => [
                    'entity_managers' => [
                        'default' => [
                            'mappings' => [
                                'SonataMediaBundle' => null,
                            ],
                        ],
                    ],
                ],
            ]
        );

```

Webonaute:
I feel like am not the only one to use dependencies injection and blocking this might be a problem for lots of dev.

PR #1367, opened 2021-06-02
Make configuration error message better understandable

icanhazstring:
This simply changes the message of the newly introduces check of the configuration syntax.
The previous message imposed a lot of research to get it fixed.

Hope this message is will help users to spot the problem a bit more quickly.


ostrolucky:
thanks @icanhazstring

PR #1369, opened 2021-06-03
Allow to register AttributeDriver via DoctrineOrmMappingsPass

norkunas:


ostrolucky:
I need some description. What's the motivation of this PR, why don't you use constructor?

norkunas:
If there is a helper for other drivers why not add also for attribute driver for consistency?

ostrolucky:
I will answer your question after you answer mine. Since you reacted that way, I assume you don't actually have a need for this.

norkunas:
I have a need to adapt my bundles to use attribute driver so it will be easier to switch a method than constructing it myself

ostrolucky:
Ok I guess I accept that. Reason why I'm hesitant is because these utility methods have very limited benefit and we have to add new ones for each new driver. I could see benefit for the ones that have at least a little more meat inside, like xml and yaml, but staticphp and annotation could be one liners... This approach scales only up to certain point and now I feel we are reaching it. Any more and I fell will rather deprecate existing ones and tell everyone to use constructor. Constructor is generic and we don't have to keep updating class for each new driver.

ostrolucky (approved):

Chris53897:
Thanks for this PR.
It helped me by solving my problem.
I find it confusing to have functions for some drivers and not for others.
I was looking for the method "createAttributeMappingDriver" and did not find it. So i startet looking for "createPhpMappingDriver" and "createStaticPhpMappingDriver". ;(

Of course it is easy to write it manual. But the missing function makes is easier for consistency.
$driver = new Definition('Doctrine\ORM\Mapping\Driver\AttributeDriver', [$directories]);

PR #1378, opened 2021-06-29
workaround for issue 1310

### LINKED ISSUE #1310: DoctrineMetadataCacheWarmer must load metadata first, check priority  of you warmers when updated from 2.2.4 -> 2.3.0
Hello all.
I got this message when run `bin/console cache:clear` or `bin/console cache:warmup`.

Conditions:

I have symfony 5.2.4 project out box
Then i run `composer update` doctrine-bundle updated to 2.2.4 -> 2.3.0.

Then i got error.  

**DoctrineMetadataCacheWarmer must load metadata first, check priority  of you warmers**

dmaicher:
This is a first hacky idea to start a discussion on how to fix https://github.com/doctrine/DoctrineBundle/issues/1310 properly.

So to make sure we properly warm up complete metadata we need to ensure the metadataFactory does not have anything loaded yet when we replace the cache and load all available metadata. Hence the `LogicException` that is causing issues now.

Since the `AbstractClassMetadataFactory` does not provide a `clearLoadedMetadata` (or similar) method I don't see many options.

- hacky solution I implemented here 
  - but it relies on `isset` checks inside `AbstractClassMetadataFactory` to work and ignore `null` values
  - it violates the contract I think as `setMetadataFor` is not meant to be called with `null`
  
- use reflection to set `AbstractClassMetadataFactory::$loadedMetadata` to an empty array?

- add `AbstractClassMetadataFactory::clearLoadedMetadata` on doctrine/persistence?

- other ideas?

### DIFF
--- a/CacheWarmer/DoctrineMetadataCacheWarmer.php
+++ b/CacheWarmer/DoctrineMetadataCacheWarmer.php
@@ -45,8 +45,10 @@ class DoctrineMetadataCacheWarmer extends AbstractPhpFileCacheWarmer
         }
 
         $metadataFactory = $this->entityManager->getMetadataFactory();
-        if ($metadataFactory instanceof AbstractClassMetadataFactory && $metadataFactory->getLoadedMetadata()) {
-            throw new LogicException('DoctrineMetadataCacheWarmer must load metadata first, check priority of your warmers.');
+        if ($metadataFactory instanceof AbstractClassMetadataFactory) {
+            foreach (array_keys($metadataFactory->getLoadedMetadata()) as $class) {
+                $metadataFactory->setMetadataFor($class, null);
+            }
         }
 
         if (method_exists($metadataFactory, 'setCache')) {


ostrolucky:
We ship own implementation of AbstractClassMetadataFactory so I think we are already pretty tightly coupled to it. You said your workaround relies on isset call, so I assume it actually relies on output from hasMetadataFor method? If that's so, we could also override the function and implement custom logic in there, correct?

Another idea would be to somehow replace the service instance so that its state is cleared, similarly like doctrine-bridge does with EntityManager in ManagerRegistry::resetService. But at least in case of EntityManager it brings its own [can of worms](https://github.com/symfony/symfony/issues/35216), so not sure it would be cleaner here rather than using null or using reflection.

But in the end I don't quite understand the bug, so I can't be much of a help here I'm afraid.

dmaicher:
I will try to find some time in the next week to work on this

stof:
Replacing the service instance is hard to do, as it requires that no instantiated service in the container contain a reference to the old instance. This quickly becomes insane into what it implies. That's why I opened an issue a long time ago in doctrine/orm about implementing a native resetting feature (allowing to re-open a closed entity manager into a cleared state) to remove that need for the resetting proxy.

ostrolucky:
This PR is opened for a long time with no activity. Let's close until there is more activity, or feel free to open new PR later.

PR #1379, opened 2021-06-30
Loosen dependency on symfony/validator bundle

### LINKED ISSUE #1377: DoctrineLoader relies on symfony/validator as a peer dependency
**Description**
`DoctrineExtension#loadValidatorLoader` registers the `DoctrineLoader` as a service. This class uses the `Symfony\Component\Validator\Mapping\Loader\AutoMappingTrait` (from `symfony/validator`) which might not be available.

As soon as this service gets autloaded at container build time* this will produce a fatal error under PHP8.1:

```
PHP Fatal error:  During class fetch: Uncaught ReflectionException: Class "Symfony\Component\Validator\Mapping\Loader\AutoMappingTrait" not found while loading "Symfony\Bridge\Doctrine\Validator\DoctrineLoader`
```

*) one scenario is for instance reading class annotations/attributes and thereby calling `$container->getReflectionClass()` 

**How to reproduce**
1. Set up a Sf application with doctrine-bundle installed.
1. Add a compiler pass that does the following:
    ```php
    foreach ($container->getDefinitions() as $definition) {
      $container->getReflectionClass($definition->getClass(), false);
    }
    ```
1. Build the container.

----
This reminded me a bit of https://github.com/symfony/symfony/issues/32395. :slightly_smiling_face:  Maybe we could mark the service as synthetic if the `AutoMappingTrait` is not available or register a dummy instead?

(ref [symfony/symfony#41796](https://github.com/symfony/symfony/issues/41796))

m-vo:
Fixes #1377 

This PR skips registering a `doctrine.orm.<name>_entity_manager.validator_loader` service if the `Symfony\Component\Validator\Mapping\Loader\LoaderInterface` is not present. 

Not sure how to properly test this tbh.

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -47,6 +47,7 @@ use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Messenger\Bridge\Doctrine\Transport\DoctrineTransportFactory;
 use Symfony\Component\Messenger\MessageBusInterface;
 use Symfony\Component\PropertyInfo\PropertyInfoExtractorInterface;
+use Symfony\Component\Validator\Mapping\Loader\LoaderInterface;
 
 use function array_intersect_key;
 use function array_keys;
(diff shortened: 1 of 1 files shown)


m-vo:
@alcaeus Is this the right target branch?

alcaeus (approved):

alcaeus:
2.4 is correct, thank you for the PR. Build failures are unrelated and probably have to do with doctrine/annotations no longer being a required dependency in ORM but rather optional. I'll fix that in a separate PR.

PR #1383, opened 2021-08-06
Allow toggling MetadataPhpArrayCaching manually

ruudk:
After upgrading from 2.2 to 2.3 we noticed that the `metadata_cache_driver` option was deprecated.
It's now being toggled based on the value of `kernel.debug`.

This is a problem for us. We are running a very large (9 years old) Symfony 5.3 application with almost
10.000 PHP files. In `dev` environment we still like to have `debug` enabled, but we want caching
to be static. Cache invalidaton is managed by a background file watcher.

By introducing a `cache_metadata` configuration option we make it possible to manually toggle the
metadata caching while running in `kernel.debug`.

We profiled our application with PHP SPX and noticed that enabling the metadata cache reduces
request times with almost 100ms.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -576,6 +576,7 @@ class Configuration implements ConfigurationInterface
                     ->scalarNode('quote_strategy')->defaultValue('doctrine.orm.quote_strategy.default')->end()
                     ->scalarNode('entity_listener_resolver')->defaultNull()->end()
                     ->scalarNode('repository_factory')->defaultValue('doctrine.orm.container_repository_factory')->end()
+                    ->scalarNode('enable_metadata_cache')->defaultValue(! $this->debug)->end()
                 ->end()
                 ->children()
                     ->arrayNode('second_level_cache')
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -915,21 +915,21 @@ class DoctrineExtension extends AbstractDoctrineExtension
         if (isset($entityManager['metadata_cache_driver'])) {
             $this->loadCacheDriver('metadata_cache', $entityManager['name'], $entityManager['metadata_cache_driver'], $container);
         } else {
-            $this->createMetadataCache($entityManager['name'], $container);
+            $this->createMetadataCache($entityManager['name'], $entityManager['enable_metadata_cache'], $container);
         }
 
         $this->loadCacheDriver('result_cache', $entityManager['name'], $entityManager['result_cache_driver'], $container);
         $this->loadCacheDriver('query_cache', $entityManager['name'], $entityManager['query_cache_driver'], $container);
     }
 
-    private function createMetadataCache(string $objectManagerName, ContainerBuilder $container): void
+    private function createMetadataCache(string $objectManagerName, bool $enabled, ContainerBuilder $container): void
     {
         $aliasId = $this->getObjectManagerElementName(sprintf('%s_%s', $objectManagerName, 'metadata_cache'));
         $cacheId = sprintf('cache.doctrine.orm.%s.%s', $objectManagerName, 'metadata');
 
         $cache = new Definition(ArrayAdapter::class);
 
-        if (! $container->getParameter('kernel.debug')) {
+        if ($enabled) {
             $phpArrayFile         = '%kernel.cache_dir%' . sprintf('/doctrine/orm/%s_metadata.php', $objectManagerName);
             $cacheWarmerServiceId = $this->getObjectManagerElementName(sprintf('%s_%s', $objectManagerName, 'metadata_cache_warmer'));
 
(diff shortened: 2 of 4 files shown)


greg0ire (commented):
I think this will require a documentation update (see https://github.com/doctrine/DoctrineBundle/blob/2.4.x/Resources/doc/configuration.rst). Also, new features should target 2.5.x

ruudk:
@greg0ire Thanks for your quick reply. Updated the base branch to `2.5.x` and added the documentation line. Happy to make any more changes as requested 😊

DependencyInjection/DoctrineExtension.php
```diff
@@ -929,7 +929,7 @@ private function createMetadataCache(string $objectManagerName, ContainerBuilder
 
         $cache = new Definition(ArrayAdapter::class);
 
-        if (! $container->getParameter('kernel.debug')) {
+        if (! $entityManager['cache_metadata']) {
```
greg0ire: :thinking: shouldn't it be the opposite?
```suggestion
        if ($entityManager['enable_metadata_cache']) {
```

DependencyInjection/Configuration.php
```diff
@@ -576,6 +576,7 @@ private function getOrmEntityManagersNode(): ArrayNodeDefinition
                     ->scalarNode('quote_strategy')->defaultValue('doctrine.orm.quote_strategy.default')->end()
                     ->scalarNode('entity_listener_resolver')->defaultNull()->end()
                     ->scalarNode('repository_factory')->defaultValue('doctrine.orm.container_repository_factory')->end()
+                    ->scalarNode('cache_metadata')->defaultValue(!$this->debug)->end()
```
greg0ire:
```suggestion
                    ->booleanNode('enable_metadata_cache')->defaultValue(!$this->debug)->end()
```

Resources/doc/configuration.rst
```diff
@@ -311,6 +311,7 @@ Configuration Reference
                         quote_strategy:               doctrine.orm.quote_strategy.default
                         entity_listener_resolver:     ~
                         repository_factory:           ~
+                        cache_metadata:               false
```
greg0ire:
```suggestion
                        enable_metadata_cache:               false
```

DependencyInjection/DoctrineExtension.php
ruudk: Fixed it. I'll add some tests!

ruudk:
@greg0ire Added tests. Can you approve the GitHub Workflows so that we can see if they pass? Thanks!

DependencyInjection/Configuration.php
```diff
@@ -576,6 +576,7 @@ private function getOrmEntityManagersNode(): ArrayNodeDefinition
                     ->scalarNode('quote_strategy')->defaultValue('doctrine.orm.quote_strategy.default')->end()
                     ->scalarNode('entity_listener_resolver')->defaultNull()->end()
                     ->scalarNode('repository_factory')->defaultValue('doctrine.orm.container_repository_factory')->end()
+                    ->scalarNode('enable_metadata_cache')->defaultValue(!$this->debug)->end()
```
greg0ire: From reading the code, I see that by default, there still is a metadata cache, it's just that its duration is limited by the duration of the current execution of PHP.
```suggestion
                    ->scalarNode('enable_persistent_metadata_cache')->defaultValue(!$this->debug)->end()
```

ruudk: Technically you are right. 

But isn't that the case with all the things that are loaded? They are there during the lifecycle of the request / while the application runs.

For an end user, you want to cache (persist) it or not.

ostrolucky:
There is no point deprecating metadata_cache configuration if we are going to replace it with some (much less flexible) toggle. We may consider undeprecating the option if there is enough interest, but since you are the only one speaking out, that's not enough. Since you have such mature project and complex workflow involving file watchers, you may as well write a compiler pass adjusting metadata cache. My experience speak against having persistent metadata cache in dev environment, it was constant struggle with changes not having any effect and having to nuke whole cache manually. Create an issue if interested and let's see if we get more feedback.

ruudk:
It's fine that the driver is removed and deprecated.

My only problem is that now the cache is toggled based on `kernel.debug` and not a separate boolean.

I hoped that by taking initiative and providing a PR we could have a discussion if this was valuable or not. Instead of me creating yet another issue. Why can't the PR serve as the issue? It doesn't have to be merged immediately.

ostrolucky:
You need to see this from maintainers perspective. These toggles are shortsighted. Next thing you know, another person comes along complaining that in dev mode they want custom cache implementation and not php one. I wouldn't blame them TBH, it's rare to have opcache without timestamp validation enabled in dev environment and that's where is this cache designed to be home at. We deprecated it so we can decrease number of configuration options, not to have more of them (existing one + this new toggle). You can also see based on a diff in this PR that this change is actually more complex than undeprecating old option. Not sure why you feel this needs more discussion. I've provided alternative about your own compiler pass or undeprecating the option, but that's separate discussion, hence also warrants separate thread.

ruudk:
Created an issue to talk about un-deprecating the metadata driver configuration option. https://github.com/doctrine/DoctrineBundle/issues/1393

PR #1385, opened 2021-08-12
Adding PHP Attributes code sample

ThomasLandauer:
Since it's an array, it would be better to show how *multiple* listeners can be defined. It this the right syntax?:
```php
#[ORM\EntityListeners([Foo::class, Bar::class])]
```

### DIFF
--- a/Resources/doc/entity-listeners.rst
+++ b/Resources/doc/entity-listeners.rst
@@ -11,7 +11,7 @@ Full example:
 
 .. configuration-block::
 
-    .. code-block:: php-annotations
+    .. code-block:: annotation
 
         <?php
         // User.php
@@ -28,7 +28,7 @@ Full example:
             // ....
         }
     
-    .. code-block:: php-attributes
+    .. code-block:: attribute
 
         <?php
         // User.php
--- a/Resources/doc/entity-listeners.rst
+++ b/Resources/doc/entity-listeners.rst
@@ -9,22 +9,39 @@ which entity manager it should be registered with.
 
 Full example:
 
-.. code-block:: php
+.. configuration-block::
 
-    <?php
-    // User.php
+    .. code-block:: php-annotations
 
-    use Doctrine\ORM\Mapping as ORM;
+        <?php
+        // User.php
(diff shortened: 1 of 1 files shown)


ostrolucky:
We will update to PHP attributes version of documentation only after ORM does. Until then, annotations are more universal. We still support PHP 7.1.

ThomasLandauer:
@ostrolucky But ORM already supports attributes, doesn't it? (Just switching to them in my project :-)
What I'm suggesting would be a switcher between Annotation and Attributes, like here: https://symfony.com/doc/current/routing.html#creating-routes-as-attributes-or-annotations

ostrolucky:
Do you have some existing example where you saw `code-block:: php-annotations` being used? I don't see it in ORM docs and I'm worried this will break doctrine docs website build

ThomasLandauer:
Oh, I see what you mean :-) The page I want to edit is https://symfony.com/doc/current/bundles/DoctrineBundle/entity-listeners.html
And I was guessing it's built by Symfony, isn't it? So I copied the `code-block:: php-annotations` stuff from some other Symfony page (e.g. the one about routing linked above)
@javiereguiluz Could you take a look?

javiereguiluz:
This needs to be supported by the doctrine/rst-parser package used in Doctrine website.

I can find this: https://github.com/doctrine/rst-parser/blob/1.0.x/lib/Directives/CodeBlock.php but I don't know where to look for the actual supported languages.

In symfony.com website we use doctrine/rst-parser too ... but we have our own code to handle CodeBlocks (see https://github.com/symfony-tools/docs-builder/blob/4c36d3a48b8b6ee1b6928d095254984ef0a34209/src/Directive/ConfigurationBlockDirective.php) and that's where we added the needed `php-attributes`

SenseException:
Not really `php-annotations` but the PR https://github.com/doctrine/doctrine-website/pull/409 allows the addition of Annotations and Attributes as PHP highlighting.

ThomasLandauer:
OK, I changed it :-)

PR #1388, opened 2021-08-22
Add metadata for symfony.com/bundles

derrabus:
Symfony has a new website where we present selected bundles and render their documentation.

https://symfony.com/bundles

I'd like to add this configuration file to the repository. It allows us to include DoctrineBundle on that page.

### DIFF
--- a/.symfony.bundle.yaml
+++ b/.symfony.bundle.yaml
@@ -1,0 +1,9 @@
+branches:
+    - "1.12.x"
+    - "2.4.x"
+    - "2.5.x"
+maintained_branches:
+    - "2.4.x"
+    - "2.5.x"
+doc_dir: "Resources/doc/"
+dev_branch: "2.5.x"


.symfony.bundle.yaml
```diff
@@ -0,0 +1,24 @@
+branches:
```
fabpot, line 1: We should only list the versions for which we want the docs to be displayed on symfony.com. The idea is to have maintained branches and some "important" past versions.

fabpot (changes requested):

.symfony.bundle.yaml
derrabus: Understood. I've reduced the list to…
* `1.12.x` as the final 1.x version,
* `2.4.x` as the current stable version and
* `2.5.x` as the version currently under development

ostrolucky (approved):

fabpot (approved):

PR #1392, opened 2021-08-25
Add native return type to commands

nicolas-grekas:
Similar to https://github.com/doctrine/collections/pull/269 and related PRs.
Should help fix related deprecations.

### DIFF
--- a/Command/CreateDatabaseDoctrineCommand.php
+++ b/Command/CreateDatabaseDoctrineCommand.php
@@ -44,10 +44,7 @@ EOT
         );
     }
 
-    /**
-     * {@inheritDoc}
-     */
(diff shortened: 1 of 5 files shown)


greg0ire (approved):

ostrolucky:
Why not adding real type declarations?

nicolas-grekas:
> Why not adding real type declarations?

you mean we could because `@final`?

ostrolucky:
Well all of them are final or deprecated, so yeah it's fine to have native type declarations from my POV

nicolas-grekas:
Works for me, PR updated.

ostrolucky (approved):

PR #1394, opened 2021-08-30
[Doc] Fix minor RST syntax issue

javiereguiluz:
The single extra white space before `you're using an older ...` seems minor ... but RST considers it (and renders it as) a definition list. I found this while browsing https://symfony.com/bundles/DoctrineBundle/current/installation.html#step-2-enable-the-bundle and seeing the strange rendering of that paragraph.


ostrolucky:
thx!

PR #1399, opened 2021-09-11
Fix tests (DBAL 3, Symfony 6)

derrabus:
Doctrine ORM 2.10 will allow the installation of DBAL 3. This breaks our tests currently and this PR is my attempt to fix that.

### DIFF
--- a/Command/Proxy/DoctrineCommandHelper.php
+++ b/Command/Proxy/DoctrineCommandHelper.php
@@ -8,6 +8,7 @@ use Doctrine\ORM\Tools\Console\Helper\EntityManagerHelper;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
 
 use function assert;
+use function class_exists;
 
 /**
  * Provides some helper and convenience methods to configure doctrine commands in the context of bundles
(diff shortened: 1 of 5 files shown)


derrabus:
🟢

phpcs.xml.dist
```diff
@@ -20,6 +20,7 @@
         <exclude name="SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingNativeTypeHint"/> <!-- we can do it in doctrine-bundle 3.0 -->
         <exclude name="SlevomatCodingStandard.Classes.SuperfluousInterfaceNaming.SuperfluousSuffix"/>
         <exclude name="SlevomatCodingStandard.Classes.SuperfluousAbstractClassNaming.SuperfluousPrefix"/>
+        <exclude name="SlevomatCodingStandard.Classes.ModernClassNameReference.ClassNameReferencedViaFunctionCall" /> <!-- PHP 8.0 required -->
```
ostrolucky: this doesn't look right. We run phpcs under php 7.4

derrabus: I've reverted the line. However, that rule is a pitfall for everyone who uses PHP 8 locally. 🤔

ostrolucky: I see. In that case you could leave it in. But I think we also have similar issues with psalm, it will report different issues on different php versions, not much can be done about it.

ostrolucky (approved):

phpcs.xml.dist
derrabus: I created #1400 to update the PHPCS job to PHP 8. That job should fail without that line.

PR #1400, opened 2021-09-11
Run PHPCS on PHP 8

derrabus:


.github/workflows/coding-standards.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "7.4"
+          - "8.0"
```
ostrolucky, line 20: Perhaps we should run it with 7.1, that's the compatibility we need to maintain.

derrabus: I've added 7.1 as well.

greg0ire: :thinking: I think one run should be enough, if we add `<config name="php_version" value="70100"/>` to the configuration.

ostrolucky (approved):

.github/workflows/coding-standards.yml
derrabus: Let's check that. 🙂

SenseException (approved):

.github/workflows/coding-standards.yml
derrabus: Yes, that does the trick. I could even remove two more rules from the exclude list now. 🎉

Issue #1393, opened 2021-08-29
Undo deprecation of `metadata_cache_driver` configuration


ruudk:
After upgrading from 2.2 to 2.3 we noticed that the `metadata_cache_driver` option was deprecated. We removed the configuration key to get rid of the deprecation warning. We didn't think about it but after a while we noticed that our application started to slow down significantly.  After investigation we noticed it was related to this metadata cache.

We are running a very large (9 years old) Symfony 5.3 application with almost 10.000 PHP files. The metadata caching allowed us to reduce our load times. We don't make changes to your entities that often as compared to other code changes. By having the metadata cache enabled we shave 100ms off our request load times.

I created a PR to bring back some of the functionality that we need but it was rejected:
- https://github.com/doctrine/DoctrineBundle/pull/1383

Following up on @ostrolucky comment here:
> We may consider undeprecating the option if there is enough interest, but since you are the only one speaking out, that's not enough. Since you have such mature project and complex workflow involving file watchers, you may as well write a compiler pass adjusting metadata cache. My experience speak against having persistent metadata cache in dev environment, it was constant struggle with changes not having any effect and having to nuke whole cache manually. Create an issue if interested and let's see if we get more feedback.

So this issue is to see if more people are using the metadata caching driver and want to keep using it.

I'd also like to know more about the reasoning and urge to deprecate the metadata caching configuration. Is there an issue/PR that explains the reasoning?

dmaicher:
> I'd also like to know more about the reasoning and urge to deprecate the metadata caching configuration. Is there an issue/PR that explains the reasoning?

There is multiple PRs about this topic. Here the original PR that introduced the new caching: https://github.com/doctrine/DoctrineBundle/pull/1196#issuecomment-724655625

So the idea was that independent of the `metadata_cache_driver` configuration we would use the fastest cache available when in no-debug mode which would be the new `PhpArrayAdapter` cache.

Later on while providing compatibility with PSR-6 caches @alcaeus changed the approach slightly [here](https://github.com/doctrine/DoctrineBundle/pull/1352/files#diff-afe5f060789287a8ad428a4b5df1ad844184f496ece0297e421f6b1fc442b5a4R915).
So now we actually only register the fast `PhpArrayAdapter` if there is no other cache configured and we are in no-debug mode.

From my point of view there are not many use-cases where one would like to have a persistent cache while being in debug mode. 

[...]

PR #1405, opened 2021-10-05
Undo deprecation of `metadata_cache_driver`

ruudk:
This allows projects to still choose to opt-in on caching the metadata.

When the configuration key is not defined or used, the cache is created automatically.

See #1393

I targeted `2.5.x`. Is that correct?

### DIFF
--- a/Tests/DependencyInjection/Compiler/CacheCompatibilityPassTest.php
+++ b/Tests/DependencyInjection/Compiler/CacheCompatibilityPassTest.php
@@ -57,10 +57,8 @@ class CacheCompatibilityPassTest extends TestCase
         $this->addToAssertionCount(1);
     }
 
-    /** @group legacy */
     public function testMetadataCacheConfigUsingPsr6ServiceDefinedByApplication(): void
     {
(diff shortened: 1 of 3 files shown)


Tests/DependencyInjection/Compiler/CacheCompatibilityPassTest.php
```diff
@@ -60,7 +60,6 @@ public function registerContainerConfiguration(LoaderInterface $loader): void
     /** @group legacy */
```
ostrolucky: I think this can be removed

ruudk: Done!

dmaicher (approved):

ostrolucky (approved):

ruudk:
Thanks for merging this. 

I ran 2 Blackfire profiles (with 10 samples):

A: No metadata cache driver configured, and thus the smart cache warmer is enabled in debug mode
B: Metadata cache driver configured that caches on file system.

I get a 20% performance improvement (on every request) with version B 🎉 

Screenshot 2021-10-08 at 10 10 18@2x

PR #1408, opened 2021-10-19
added DatabaseDriverFactory to be injectable

oleg-andreyev:
Case

```
[Doctrine\ORM\Mapping\MappingException]                                             
  Property "d" in "Contract" was already declared, but it must be declared only once
```

- Need to reverse-engineer table with columns "D" and "_D" as result they are both are mapped to "D" field.
- Cannot call "setFieldNameForColumn" to add custom mapping

### DIFF
--- a/Command/ImportMappingDoctrineCommand.php
+++ b/Command/ImportMappingDoctrineCommand.php
@@ -2,7 +2,7 @@
 
 namespace Doctrine\Bundle\DoctrineBundle\Command;
 
-use Doctrine\ORM\Mapping\Driver\DatabaseDriver;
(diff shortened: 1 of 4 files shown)


stof:
Given that this command is deprecated, I don't think we should add new extension points in it.

oleg-andreyev:
Any alternative/replacement for this command?

stof:
No. Reverse-engineering a database table to generate the doctrine mapping has been discouraged by the Doctrine core team since the beginning of the Doctrine 2 era. There is no replacement doing reverse-engineering (at least no official one, maybe someone created a maintained third-party package for that)

oleg-andreyev:
Sad, but okay.
Thanks.

PR #1413, opened 2021-10-26
Add an Argument Resolver

### LINKED ISSUE #40333: #[CurrentUser] : Unable to guess how to get a Doctrine instance from the request information for parameter "user".
**Symfony version(s) affected**: 5.2.3
**sensio/framework-extra-bundle**: 6.1.1

**Description**
Hi,
I have an error when use #[CurrentUser]:
`Unable to guess how to get a Doctrine instance from the request information for parameter "user".
`

**How to reproduce**

```php
<?php

namespace App\Controller\User;

use App\Entity\User;
use App\Repository\NotificationRepository;
use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
use Symfony\Component\HttpFoundation\Request;
use Symfony\Component\HttpFoundation\Response;
use Symfony\Component\Routing\Annotation\Route;
use Symfony\Component\Security\Http\Attribute\CurrentUser;

class NotificationController extends AbstractController
{
    #[Route('/notifications', name: 'user_notification_index', methods: ['GET'])]
    public function index(#[CurrentUser] User $user): Response
    {
        ...
    }
}

```

jderusse:
This PR migrates the SensioFrameworkExtraBundle's Doctrine ParamsConverter to Argument Resolver.

Moving away from ParamConverter would fix https://github.com/symfony/symfony/issues/40333

usage:
enabled the behavior
```
doctrine:
    dbal:
        default_connection: default
        connections:
            default:
                dbname: db

    orm:
        argument_resolver: true
```

The default configuration works out of the box. The entity is fetch using `$repo->find()` using the route's attribute.
```
#[Route('/blog/{id}')]
public function show(Post $post)
{
}
```

Options can be configured with a new Attribute
```
use Doctrine\Bundle\DoctrineBundle\Attribute\Entity;

#[Route('/blog/{id}')]
public function show(
  #[Entity(entityManager: 'foo', expr: 'repository.findNotDeletedById(id)')]
  Post $post
)
{
}
```

note: the business logic and tests are copy/pasted from the [original repo](https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/src/Request/ParamConverter/DoctrineParamConverter.php). Tell me if you think some options are not needed anymore.

### DIFF
--- a/ArgumentResolver/EntityValueResolver.php
+++ b/ArgumentResolver/EntityValueResolver.php
@@ -1,0 +1,302 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\ArgumentResolver;
+
+use Doctrine\Bundle\DoctrineBundle\Attribute\Entity;
+use Doctrine\DBAL\Types\ConversionException;
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\NoResultException;
+use Doctrine\Persistence\ManagerRegistry;
+use Doctrine\Persistence\ObjectManager;
+use LogicException;
+use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
+use Symfony\Component\ExpressionLanguage\SyntaxError;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Controller\ArgumentValueResolverInterface;
+use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+use function array_combine;
+use function array_filter;
+use function array_merge;
+use function count;
+use function is_array;
+use function method_exists;
+use function sprintf;
+use function strstr;
+
+/**
+ * Yields the entity matching the criteria provided in the route
+ */
+final class EntityValueResolver implements ArgumentValueResolverInterface
+{
+    /** @var ManagerRegistry */
+    private $registry;
+    /** @var ExpressionLanguage|null */
+    private $language;
+    /** @var array<string, mixed> */
+    private $defaultOptions;
+
+    /** @param array<string, mixed> $defaultOptions */
+    public function __construct(ManagerRegistry $registry, ?ExpressionLanguage $expressionLanguage = null, array $defaultOptions = [])
+    {
+        $this->registry       = $registry;
+        $this->language       = $expressionLanguage;
+        $this->defaultOptions = array_merge([
+            'entity_manager' => null,
+            'expr' => null,
+            'auto_mapping' => true,
+            'mapping' => [],
+            'exclude' => [],
+            'strip_null' => false,
+            'id' => null,
+            'evict_cache' => false,
+        ], $defaultOptions);
+    }
+
+    public function supports(Request $request, ArgumentMetadata $argument): bool
+    {
+        if (count($this->registry->getManagerNames()) === 0) {
+            return false;
+        }
+
+        $options = $this->getOptions($argument);
+        if ($options['class'] === null) {
+            return false;
+        }
+
+        // Doctrine Entity?
+        $em = $this->getManager($options['entity_manager'], $options['class']);
+        if ($em === null) {
+            return false;
+        }
+
+        return ! $em->getMetadataFactory()->isTransient($options['class']);
+    }
+
+    /** @return object[] */
+    public function resolve(Request $request, ArgumentMetadata $argument): iterable
+    {
+        $options = $this->getOptions($argument);
+
+        $name  = $argument->getName();
+        $class = $options['class'];
+
+        $errorMessage = null;
+        if ($options['expr'] !== null) {
+            $object = $this->findViaExpression($class, $request, $options['expr'], $options);
+
+            if ($object === null) {
+                $errorMessage = sprintf('The expression "%s" returned null', $options['expr']);
+            }
+
+            // find by identifier?
+        } else {
+            $object = $this->find($class, $request, $options, $name);
+            if ($object === false) {
+                // find by criteria
+                $object = $this->findOneBy($class, $request, $options);
+                if ($object === false) {
+                    if (! $argument->isNullable()) {
+                        throw new LogicException(sprintf('Unable to guess how to get a Doctrine instance from the request information for parameter "%s".', $name));
+                    }
+
+                    $object = null;
+                }
+            }
+        }
+
+        if ($object === null && ! $argument->isNullable()) {
+            $message = sprintf('"%s" object not found by the "%s" Argument Resolver.', $class, self::class);
+            if ($errorMessage) {
+                $message .= ' ' . $errorMessage;
+            }
+
+            throw new NotFoundHttpException($message);
+        }
+
+        return [$object];
+    }
+
+    private function getManager(?string $name, string $class): ?ObjectManager
+    {
+        if ($name === null) {
+            return $this->registry->getManagerForClass($class);
+        }
+
+        return $this->registry->getManager($name);
+    }
+
+    /**
+     * @param array<string, string> $options
+     *
+     * @return false|object|null
+     */
+    private function find(string $class, Request $request, array $options, string $name)
+    {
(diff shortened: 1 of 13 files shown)


Resources/config/orm.xml
```diff
@@ -229,5 +229,15 @@
 
             <tag name="console.command" command="doctrine:mapping:import" />
         </service>
+
+        <!-- argument resolvers -->
+        <service id="doctrine.entity_value_resolver" class="Doctrine\Bundle\DoctrineBundle\ArgumentResolver\EntityValueResolver">
```
jderusse, line 234: shouldn't it be disabled by default :thinking: ?

ArgumentResolver/EntityValueResolver.php
```diff
@@ -0,0 +1,293 @@
...
+
+    public function __construct(?ManagerRegistry $registry = null, ?ExpressionLanguage $expressionLanguage = null)
+    {
+        $this->registry = $registry;
+        $this->language = $expressionLanguage;
+    }
+
+    public function supports(Request $request, ArgumentMetadata $argument): bool
+    {
+        // if there is no manager, this means that only Doctrine DBAL is configured
```
jvasseur: Maybe we could completely remove the service from the container in this case ?

ArgumentResolver/EntityValueResolver.php
```diff
@@ -0,0 +1,293 @@
...
+ * Yields the entity matching the criteria provided in the route
+ */
+final class EntityValueResolver implements ArgumentValueResolverInterface
+{
+    /** @var ManagerRegistry|null */
+    private $registry;
+    /** @var ExpressionLanguage|null */
+    private $language;
+
+    public function __construct(?ManagerRegistry $registry = null, ?ExpressionLanguage $expressionLanguage = null)
```
stof: why allowing the registry to be null ? It does not make sense to me

ArgumentResolver/EntityValueResolver.php
```diff
@@ -0,0 +1,293 @@
...
+    /**
+     * @param array<string, string> $options
+     *
+     * @return false|object|null
+     */
+    private function findOneBy(string $class, Request $request, array $options)
+    {
+        if (! $options['mapping']) {
+            $keys               = $request->attributes->keys();
+            $options['mapping'] = $keys ? array_combine($keys, $keys) : [];
```
stof, line 210: This default mapping that takes all request attributes to find the entity is something that I'd like to make configurable (through the config, not in each argument), as it is quite confusing. I had several bugs in my project where an optional argument was suddenly given a weird value instead of `null` just because the controller action had another argument with a name matching the name of an association (and that this default mapping would silently use it to build a criteria)

ArgumentResolver/EntityValueResolver.php
```diff
@@ -0,0 +1,293 @@
...
+                $message .= ' ' . $errorMessage;
+            }
+
+            throw new NotFoundHttpException($message);
+        }
+
+        return [$object];
+    }
+
+    private function getManager(?string $name, ?string $class): ?ObjectManager
```
stof: I don't think the class is ever null when calling this

ArgumentResolver/EntityValueResolver.php
jderusse: fixed

ArgumentResolver/EntityValueResolver.php
jderusse: good catch

ArgumentResolver/EntityValueResolver.php
jderusse: providing a semantic configuration that will change the default behavior is something that could/should be done in a dedicated PR IMHO.

Maybe related to https://github.com/doctrine/DoctrineBundle/pull/1413#discussion_r736867822

stof: I don't want to disable the whole argument resolver. I want a way to disable the "use any other request param as a lookup" behavior. In my project, I never found a case where I wanted it, and I got several bugs caused by it for optional parameters of controllers.
Cases that actually want to do a lookup by another argument than the current argument will often need to provide a mapping explicitly anyway, as they expect to use a given field, not "any field" either (maybe they are just lucky that the names of their other controller arguments don't match fields that exist in their entity today, and they won't add new fields or new arguments in the future that would break that lucky invariant)

Resources/config/orm.xml
stof: we probably need an opt-in (or an opt-out) for that (in the `orm` section of the semantic config) as not all projects want to use them (and the usage of param converters in SensioFrameworkExtraBundle is configurable.

ArgumentResolver/EntityValueResolver.php
jvasseur: Making this configurable globally would means bundles won't be able to rely on one the the behaviors. My take on this would be to completely remove the feature as something that is to much magic for an argument resolver.

stof: well, I'm also fine with removing the magic mapping entirely too (the `mapping` option should still exist for cases where it is configured explicitly, as it solves a use case)

jderusse: `id` and the attribute with the same name than the argument will both use the method `find` . here we are talking about the fallback that use attributes that match none of the 2 cases above.
Disabling this feature globally does not make sens. If the request does not contain `id` nor an attribute with the same, then there will be nothing to match, entity could not be found, and user would have to provide a mapping on every attributes anyway.

Sorry, I disagree with you

`/blogs/{slug}` or `/product/{uuid}` just work, and is often used. No need for custom mapping in such cases.

At the opposite, I never had issue with route containing several attributes. maybe I'm lucky, or because routes with multiple arguments are rare enough to provide mapping or expr and are not ambiguous

Resources/config/orm.xml
jderusse: Added a configuration section to enable the `argument_resolver` (disabled by default to avoid side effects)

ArgumentResolver/EntityValueResolver.php
jderusse: anyway, I added a configuration section and added an `auto_mapping` flag that allows users to disable this behavior. This will enforce people to either use `{id}` (or the same name as the argument) OR to explicitly provide a `mapping` option (or `id` or `expr`).

mbabker:
Maybe the resolver and attribute should go to the Doctrine Bridge instead of each bundle?  Based on https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/744 it seems like this could work OK for more than just the ORM.

jderusse:
> Maybe the resolver and attribute should go to the Doctrine Bridge instead of each bundle? Based on [sensiolabs/SensioFrameworkExtraBundle#744](https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/744) it seems like this could work OK for more than just the ORM.

The PR you are referring is about ParamConverter

Having the glue between Symfony and the ORM sounds

jderusse:
> Maybe the resolver and attribute should go to the Doctrine Bridge instead of each bundle? Based on [sensiolabs/SensioFrameworkExtraBundle#744](https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/744) it seems like this could work OK for more than just the ORM.

What do you mean by "each bundle". the PR you are referring is about ParamConverter and not Argument Resolver.

This PR is about making the glue between Symfony and the ORM. The bundle sounds an appropriate place for such code.

Edit sorry for multiple close/open currently typing on phone

mbabker:
Based on how that PR reads, it sounds like the original param converter code would work just fine for any object manager (be it the ORM or any of the ODM implementations).  So instead of this bundle, the MongoDB bundle, and the PHPCR bundle needing to re-implement basically the same class, they could all use one common class from the Doctrine Bridge and the bundles would just be left to handle the configuration bits.

ostrolucky:
I agree. I would rather this was in doctrine-bridge.

ostrolucky:
as mentioned, class should be in bridge and bundle will only provide wiring

PR #1430, opened 2021-11-22
Remove unknown branch name from doctrine-website config

SenseException:
The branch 2.6.x doesn't exist and breaks the website build.


ostrolucky:
Solution is create the branch, not remove it from config

PR #1434, opened 2021-11-30
Allow deprecation-contracts 3

derrabus:


stof (approved):

ostrolucky:
Would be nice to have more info. Why was this so urgent that it was qualified as a bug and release was created immediately after merge?

stof:
@ostrolucky this fixes an issue where DoctrineBundle cannot be added in a newly created Symfony 6.0 project by default, because the newly created project will contain deprecation-contracts v3. Users were reporting the fact that they could not install the ORM in the new project (using the `--update-with-dependencies` flag of composer might have helped by allowing composer to downgrade the installed contracts package but that's not as nice a DX).

derrabus:
Sorry for the emergency release, but I felt it was justified given the circumstances.

New Symfony 5.4 projects were created with a very old version of the bundle because Composer believed installing deprecation-contracts 3 was more important than pulling a recent version of DoctrineBundle, see symfony/symfony#44353.

This problem also screwed up MakerBundle's CI on PHP 8.

Issue #1442, opened 2021-12-05
Explain query doesn't work with Symfony 6


HypeMC:
The `Explain query` button in the profiler doesn't work with Symfony 6:
 
![image](https://user-images.githubusercontent.com/2445045/144752191-59da2053-6fee-4b0f-8f3b-0f2e78de0a01.png)

If I open the URL directly, e.g. `/_profiler/248883?panel=db&page=explain&connection=default&query=1` I get the following error:

> An exception has been thrown during the rendering of a template ("The "twig" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection instead.").

The problem is that the `ProfilerController` tries to fetch the `twig` service directly from the container, but in Symfony 6 [the service is private](https://github.com/symfony/symfony/blob/5.4/UPGRADE-6.0.md#twigbundle):

https://github.com/doctrine/DoctrineBundle/blob/97cea2188a6a2af8dd10e3208111b3d1b9699331/Controller/ProfilerController.php#L85

PR #1447, opened 2021-12-11
Fix ProfilerController compatibility with Symfony 6

ostrolucky:
https://github.com/doctrine/DoctrineBundle/issues/1442

### DIFF
--- a/Controller/ProfilerController.php
+++ b/Controller/ProfilerController.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\Bundle\DoctrineBundle\Controller;
 
+use Doctrine\Bundle\DoctrineBundle\Registry;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\ForwardCompatibility\Result;
 use Doctrine\DBAL\Platforms\OraclePlatform;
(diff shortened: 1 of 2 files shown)


Resources/config/orm.xml
```diff
@@ -233,5 +233,13 @@
 
             <tag name="console.command" command="doctrine:mapping:import" />
         </service>
+
+        <service id="Doctrine\Bundle\DoctrineBundle\Controller\ProfilerController">
+            <argument type="service" id="twig" on-invalid="ignore"/>
```
stof, line 238: `on-invalid="ignore"` is wrong, as the class cannot be instantiated without that argument.

ostrolucky: And that's fine, I don't want that controller instantiated without those arguments, as it would be useless. It cannot work without them.

Controller/ProfilerController.php
```diff
@@ -11,27 +12,29 @@
...
-     */
-    public function setContainer(?ContainerInterface $container = null)
+    /** @var Environment */
+    private $twig;
+    /** @var Registry */
+    private $registry;
+    /** @var Profiler */
+    private $profiler;
+
+    public function __construct(Environment $twig, Registry $registry, Profiler $profiler)
```
dmaicher, line 34: whats the BC policy for this class? :thinking: Maybe not likely that someone extended it? But possible...

Resources/config/orm.xml
stof: but why adding this `on-invalid="ignore"` then ?

ostrolucky: Because `twig` and `profiler` are registered only in certain configurations, as discovered by tests.

Controller/ProfilerController.php
stof: I would tag it as `@internal` (and document in the changelog that we removed that class from the public API because it was never meant to be part of it)

ostrolucky: Our BC policy is whatever we feel is right. We try to avoid this kind of things in patch releases, but if given issue is big enough like now, we can do this. We are purposefully vague to not limit ourselves. We don't have same policy as Symfony. Gymnastics of making this BC compatible is not worth it here for very unlikely scenario of someone extending this class.

dmaicher: ok then let's tag it as `@internal`? To make it more obvious that people should not extend this

ostrolucky: done

dmaicher (approved):

PR #1449, opened 2021-12-20
Add support for DoctrineOpenTransactionLoggerMiddleware

lyrixx:


lyrixx:
I'm not sure for the base branch

stof:
shouldn't the old abstract service also be kept for BC ?

stof:
to me, this is a new feature, so targetting 2.6.x makes sense.

lyrixx:
Oups, I messed'up the patch, I'll clean it! (fixed)

Resources/config/messenger.xml
```diff
@@ -28,6 +28,16 @@
             <argument type="service" id="doctrine" />
         </service>
 
+        <!--
+        The following service isn't prefixed by the "doctrine.orm" namespace in order for end-users to just use
+        the "doctrine_open_transaction_logger" shortcut in message buses middleware config
+        -->
+        <service id="messenger.middleware.doctrine_open_transaction_logger" class="Symfony\Bridge\Doctrine\Messenger\DoctrineOpenTransactionLoggerMiddleware" abstract="true" public="false">
+            <argument type="service" id="doctrine" />
+            <argument type="constant">null</argument>
```
stof: `type="constant"` with `null` as name looks suspicious to me

ostrolucky: Indeed. Even if it works, this is not conceptually correct. Pls reomve the type @lyrixx

lyrixx: I'm not sure to understand. What should I use?

ostrolucky: `<argument>null</argument>`

lyrixx: Thanks. fixed.

ostrolucky (approved):

PR #1450, opened 2021-12-24
fix: Doctrine ORM 3 support

dunglas:
`doctrine:mapping:convert`, `doctrine:ensure-production-settings` and `doctrine:mapping:import` have been removed from Doctrine 3. Registering the proxy classes prevent the container to build properly.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
     },
     "require-dev": {
         "doctrine/coding-standard": "^9.0",
-        "doctrine/orm": "^2.9",
+        "doctrine/orm": "^2.9 || ^3.0",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -59,7 +59,11 @@
         "vimeo/psalm": "^4.7"
     },
     "config": {
-        "sort-packages": true
+        "sort-packages": true,
+        "allow-plugins": {
```
ostrolucky, line 63: this is unrelated change, it needs comment from you

ostrolucky (changes requested):

composer.json
dunglas: It's a new behavior of Composer. This is added when running `composer update`.

Issue #1431, opened 2021-11-23
Implement a DBAL middleware for the profiler collector


stof:
Follows doctrine/dbal#4967

Maybe the Stopwatch integration should be a separate middleware than the collection of queries for the doctrine dbal panel.
Currently, we rely on a special SQLLogger to do that.

Refs #1429

ostrolucky:
Duplicate of https://github.com/doctrine/DoctrineBundle/issues/1429?

stof:
No. currently, our logger configuration covers 2 things:

- wiring the Psr3Logger to send queries to the `logger` service of Symfony
- wiring the special loggers related to the profiler (only when the profiler integration is enabled)

#1429 is about replacing the first part by wiring the logging middleware shipped in DBAL 3.2+.
this ticket is about *implementing* the necessary ProfilingMiddleware (or whatever its name) and using it instead of the special loggers, to replace the second part.

So they are related, but not duplicate (see how I even referenced it in the description)

stof:
The middleware is implemented in https://github.com/symfony/symfony/pull/45491

So it should be wired instead of the DbalLogger on supported Symfony and DBAL versions.

dmaicher:
done in https://github.com/doctrine/DoctrineBundle/pull/1517

PR #1456, opened 2021-12-29
Allow to use dbal logging middleware instead of deprecated SQLLogger

### LINKED ISSUE #1429: Wire a middleware for logging
Follows doctrine/dbal#4967

The way we currently use to wire a logger is deprecated in DBAL 3.2 and has been removed in DBAL 4. We should instead wire the new middleware.

l-vo:
Fixes #1429

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -13,6 +13,7 @@ use Doctrine\Bundle\DoctrineBundle\EventSubscriber\EventSubscriberInterface;
 use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepositoryInterface;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection;
+use Doctrine\DBAL\Driver\Middleware;
 use Doctrine\DBAL\Logging\LoggerChain;
 use Doctrine\DBAL\Tools\Console\Command\ImportCommand;
 use Doctrine\DBAL\Tools\Console\ConnectionProvider;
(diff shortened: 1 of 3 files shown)


stof:
To me, there is no reason for having a configuration for that. The core loggers should be wired as middles rather than loggers on newer DBAL versions without any config change.

ostrolucky:
like @stof said, this should work without change in configuration

l-vo:
@stof @ostrolucky thank you for your feedbacks, the point is the logged message formats are not the same; for a basic query, with SQLLogger:
```
[2022-01-02T12:47:18.370397+00:00] doctrine.DEBUG: SELECT t0.id AS id_1, t0.name AS name_2, t0.price AS price_3, t0.version AS version_4 FROM product t0 [] []
```
and with logging middleware:
```
[2022-01-01T13:45:40.827235+00:00] doctrine.DEBUG: Executing query: SELECT t0.id AS id_1, t0.name AS name_2, t0.price AS price_3, t0.version AS version_4 FROM product t0 {"sql":"SELECT t0.id AS id_1, t0.name AS name_2, t0.price AS price_3, t0.version AS version_4 FROM product t0"} []
```

It might be a problem if one do parsing of these logs for some reason. It's why I suggest an opt-in for this change.

ostrolucky:
I see. That change is fine, we don't provide BC promise for log format.

l-vo:
Ok, so I'm going to remove the opt-in.

ostrolucky:
Does this solve the deprecations we currently have? See https://github.com/doctrine/DoctrineBundle/runs/4681778064?check_suite_focus=true I am thinking if it wouldn't be better to remove definition of doctrine.dbal.logger completely, instead of only its arguments.

l-vo:
No, it doesn't, removing these deprecations needs two steps, the second (#1431) is about the profiler and involve to modify Doctrine Bridge (IMHO). I can work on it after this one. We can't remove `doctrine.dbal.logger` now because it also does time measurements (stopwatch) for the profiler.

dmaicher (approved):
Seems good to me. Tested it on my project and only noticed the different log message format. Apart from that all works well.

Next step would be to replace the data collection part for the profiler then? Like `DebugStack` and `BacktraceLogger`?

ostrolucky:
fair enough, let's merge it then

r-martins:
It does not really fix the deprecation warnings, right? It just fixes if we add a middleware class. Am I right?
If yes, how do we do that?

l-vo:
@r-martins Yes, this is a step to fix them since it involves changes on Symfony too. The changes on Symfony are now merged but a final step is still missing in DoctrineBundle. I'm working on it.

PR #1459, opened 2021-12-29
fix container linting for ProfilerController service in some cases

### LINKED ISSUE #1457: Invalid definition for service "Doctrine\Bundle\DoctrineBundle\Controller\ProfilerController"
When i run `bin/console lint:container` with version 2.5.3 of DoctrineBundle i get this error:
```
Invalid definition for service "Doctrine\Bundle\DoctrineBundle\Controller\ProfilerController": argument 3 of   
         "Doctrine\Bundle\DoctrineBundle\Controller\ProfilerController::__construct()" accepts                          
         "Symfony\Component\HttpKernel\Profiler\Profiler", "null" passed.
```
This seems to be introduced with version 2.5.3 as i upgraded stepwise from 2.5.0 and 2.5.2 worked for me.
Running symfony version 5.3.13.

dmaicher:
Closes https://github.com/doctrine/DoctrineBundle/issues/1457

### DIFF
--- a/Resources/config/orm.xml
+++ b/Resources/config/orm.xml
@@ -235,9 +235,9 @@
         </service>
 
         <service id="Doctrine\Bundle\DoctrineBundle\Controller\ProfilerController">
-            <argument type="service" id="twig" on-invalid="ignore"/>
+            <argument type="service" id="twig" on-invalid="ignore_uninitialized"/>
             <argument type="service" id="doctrine" />
-            <argument type="service" id="profiler" on-invalid="ignore" />
+            <argument type="service" id="profiler" on-invalid="ignore_uninitialized" />
 
             <tag name="controller.service_arguments" />
         </service>


stof (changes requested):
This needs to be done in a compiler pass, not in the DI extension.

DI extensions are working with a partial container. They don't see services defined by other bundles, so there won't ever be a `twig` service at that point.

dmaicher:
Thanks @stof. Indeed this was not working :man_facepalming: Now done within a compiler pass instead.

stof (approved):

dmaicher:
Interesting test fail with --prefer-lowest on Symfony 4.4.0 :thinking: checking

composer.json
```diff
@@ -34,7 +34,7 @@
         "symfony/cache": "^4.3.3|^5.0|^6.0",
         "symfony/config": "^4.4.3|^5.0|^6.0",
         "symfony/console": "^3.4.30|^4.3.3|^5.0|^6.0",
-        "symfony/dependency-injection": "^4.3.3|^5.0|^6.0",
+        "symfony/dependency-injection": "^4.4.18|^5.0|^6.0",
```
dmaicher, line 37: needs https://github.com/symfony/symfony/pull/39151

composer.json
```diff
@@ -34,7 +34,7 @@
         "symfony/cache": "^4.3.3|^5.0|^6.0",
         "symfony/config": "^4.4.3|^5.0|^6.0",
         "symfony/console": "^3.4.30|^4.3.3|^5.0|^6.0",
-        "symfony/dependency-injection": "^4.3.3|^5.0|^6.0",
+        "symfony/dependency-injection": "^4.4.18|^5.1.10|^6.0",
```
dmaicher: needs https://github.com/symfony/symfony/pull/39151

Resources/config/orm.xml
```diff
@@ -235,9 +235,9 @@
         </service>
 
         <service id="Doctrine\Bundle\DoctrineBundle\Controller\ProfilerController">
-            <argument type="service" id="twig" on-invalid="ignore"/>
```
ostrolucky, line 238: Instead of this whole PR, isn't proper fix to use `on-invalid="ignore_uninitialized"` instead?

dmaicher: This indeed makes the container linter happy :thinking: 

But technically the service definition is still useless/invalid if `twig` or `profiler` are not available? I don't mind :blush: Can also adjust it.

ostrolucky: It's symfony's job to remove useless services automatically. If it doesn't do that, that would be an issue for https://github.com/symfony/symfony. symfony/dependency-injection shouldn't shift such responsibility to all the bundles for such use cases instead.

dmaicher: Seems like this was discussed a while back: https://github.com/symfony/symfony/issues/36701

Anyway let's go with the simple solution for now. I will adjust the PR.

ostrolucky (approved):

PR #1465, opened 2022-01-20
New feature: "waiting for db" command

Nek-:
Hello,

Context: while writing test suite or deploying infrastructures with docker, we often need to wait for the database to be ready. Since `docker-compose` version 3 we cannot add a health check into docker, and the official way to go is [to use a tweak](https://docs.docker.com/compose/startup-order/). It never work as expected and finally the best option to make it simple is often to make a `sleep 120`. Even the command `pg_wait` provided in the docker image does not work properly for this use case (??!).

I'm getting bored of those tweaks and maintaining weak scripts even for a little project. So here is some code working for me. If you like the idea, do not hesitate to suggest improvements, and I will surely add a test and fix cs.

With the hope of helping people! :)

_(sorry no idea how to make it a draft)_

### DIFF
--- a/Command/WaitForDatabaseCommand.php
+++ b/Command/WaitForDatabaseCommand.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\Command;
+
+use Doctrine\DBAL\Connection;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class WaitForDatabaseCommand
+{
+    protected static $defaultName = 'doctrine:database:wait';
+
+    private $connection;
+
+    public function __construct(Connection $connection)
+    {
+        $this->connection = $connection;
+        parent::__construct();
+    }
+
+    protected function configure()
+    {
+        $this->addOption(
+            'timeout',
+            't',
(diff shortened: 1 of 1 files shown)


ostrolucky:
Sorry, but I don't believe this is something that should be in doctrine. There are tons of existing scripts for this all over the internet, you shouldn't need to maintain your own solution either.

dmaicher:
I personally use https://github.com/vishnubob/wait-for-it without any issues since years :thinking: 

I mean usually there are more services in the stack that one needs to wait for, or? Like ElasticSearch or Redis or Memcache etc

Nek-:
I didn't say that there are no solutions (I even post a link to the official documentation of docker about it). I'm saying standards tools should handle this problem.

I understand the doctrine bundle is not the best place to be, but it's a place where nobody suggested it to be so it has not been refused yet. 😄 

That said, please consider that wait-for-it is not as cool as it sounds:
1. It requires net tools to be installed
2. It requires actually modifying the image to add it inside
3. It is actually not enough! (for PostgreSQL at least) because it checks for networking, but it does not mean the db is fully ready, for example PostgreSQL will respond with an error saying it's not ready yet.

As I said, I think things should be more simple. I'm not fixing anything here but complexity. I hope my motivation is more clear like this.

ostrolucky:
There is dbal:query command, you can use that.

PR #1472, opened 2022-02-08
Allow userland middlewares

l-vo:
Related to the needs to replace SQLLogger by Middlewares, this implementations allows to easily inject middleware via a `doctrine.middleware` tag.

### DIFF
--- a/DependencyInjection/Compiler/MiddlewaresPass.php
+++ b/DependencyInjection/Compiler/MiddlewaresPass.php
@@ -1,0 +1,58 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
+
+use Doctrine\Bundle\DoctrineBundle\Middleware\ConnectionNameAwareInterface;
+use Symfony\Component\DependencyInjection\ChildDefinition;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+use function array_keys;
+use function in_array;
+use function is_subclass_of;
+use function sprintf;
+
+final class MiddlewaresPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
+    {
+        $middlewareAbstractDefs = [];
+        $middlewareConnections  = [];
+        foreach ($container->findTaggedServiceIds('doctrine.middleware') as $id => $tags) {
+            $middlewareAbstractDefs[$id] = $container->getDefinition($id);
+            // When a def has doctrine.middleware tags with connection attributes equal to connection names
+            // registration of this middleware is limited to the connections with these names
+            foreach ($tags as $tag) {
+                if (! isset($tag['connection'])) {
+                    continue;
+                }
+
+                $middlewareConnections[$id][] = $tag['connection'];
+            }
+        }
+
+        foreach (array_keys($container->getParameter('doctrine.connections')) as $name) {
+            $middlewareDefs = [];
+            foreach ($middlewareAbstractDefs as $id => $abstractDef) {
+                if (isset($middlewareConnections[$id]) && ! in_array($name, $middlewareConnections[$id], true)) {
+                    continue;
+                }
+
+                $middlewareDefs[] = $childDef = $container->setDefinition(
+                    sprintf('%s.%s', $id, $name),
+                    new ChildDefinition($id)
+                );
+
+                if (! is_subclass_of($abstractDef->getClass(), ConnectionNameAwareInterface::class)) {
+                    continue;
+                }
+
+                $childDef->addMethodCall('setConnectionName', [$name]);
+            }
+
+            $container
+                ->getDefinition(sprintf('doctrine.dbal.%s_connection.configuration', $name))
+                ->addMethodCall('setMiddlewares', [$middlewareDefs]);
+        }
+    }
+}
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -3,6 +3,7 @@
 namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection;
 
 use Doctrine\Bundle\DoctrineBundle\Attribute\AsEntityListener;
+use Doctrine\Bundle\DoctrineBundle\Attribute\AsMiddleware;
 use Doctrine\Bundle\DoctrineBundle\CacheWarmer\DoctrineMetadataCacheWarmer;
 use Doctrine\Bundle\DoctrineBundle\Command\Proxy\ImportDoctrineCommand;
 use Doctrine\Bundle\DoctrineBundle\Dbal\ManagerRegistryAwareConnectionProvider;
@@ -13,8 +14,9 @@ use Doctrine\Bundle\DoctrineBundle\EventSubscriber\EventSubscriberInterface;
 use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepositoryInterface;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection;
-use Doctrine\DBAL\Driver\Middleware;
+use Doctrine\DBAL\Driver\Middleware as MiddlewareInterface;
 use Doctrine\DBAL\Logging\LoggerChain;
+use Doctrine\DBAL\Logging\Middleware;
 use Doctrine\DBAL\Sharding\PoolingShardConnection;
 use Doctrine\DBAL\Sharding\PoolingShardManager;
 use Doctrine\DBAL\Tools\Console\Command\ImportCommand;
@@ -44,7 +46,6 @@ use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Alias;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
-use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
@@ -64,6 +65,8 @@ use function method_exists;
 use function reset;
 use function sprintf;
 use function str_replace;
+
+use const PHP_VERSION_ID;
 
 /**
  * DoctrineExtension is an extension for the Doctrine DBAL and ORM library.
@@ -143,9 +146,33 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $container->setParameter('doctrine.connections', $connections);
         $container->setParameter('doctrine.default_connection', $this->defaultConnection);
 
+        $connWithLogging = [];
         foreach ($config['connections'] as $name => $connection) {
+            if ($connection['logging']) {
+                $connWithLogging[] = $name;
+            }
+
             $this->loadDbalConnection($name, $connection, $container);
         }
+
+        /** @psalm-suppress UndefinedClass */
+        $container->registerForAutoconfiguration(MiddlewareInterface::class)->addTag('doctrine.middleware');
+
+        if (PHP_VERSION_ID >= 80000 && method_exists(ContainerBuilder::class, 'registerAttributeForAutoconfiguration')) {
+            $container->registerAttributeForAutoconfiguration(AsMiddleware::class, static function (ChildDefinition $definition, AsMiddleware $attribute) {
+                if ($attribute->connections === []) {
+                    $definition->addTag('doctrine.middleware');
+
+                    return;
+                }
+
+                foreach ($attribute->connections as $connName) {
+                    $definition->addTag('doctrine.middleware', ['connection' => $connName]);
+                }
+            });
+        }
+
+        $this->useMiddlewaresIfAvailable($container, $connWithLogging);
     }
 
     /**
@@ -160,7 +187,6 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $configuration = $container->setDefinition(sprintf('doctrine.dbal.%s_connection.configuration', $name), new ChildDefinition('doctrine.dbal.connection.configuration'));
         $logger        = null;
         if ($connection['logging']) {
-            $this->useMiddlewaresIfAvailable($connection, $container, $name, $configuration);
             $logger = new Reference('doctrine.dbal.logger');
         }
 
@@ -1073,11 +1099,11 @@ class DoctrineExtension extends AbstractDoctrineExtension
         return $id;
     }
 
-    /** @param array<string, mixed> $connection */
-    protected function useMiddlewaresIfAvailable(array $connection, ContainerBuilder $container, string $name, Definition $configuration): void
+    /** @param string[] $connWithLogging */
+    private function useMiddlewaresIfAvailable(ContainerBuilder $container, array $connWithLogging): void
     {
         /** @psalm-suppress UndefinedClass */
-        if (! interface_exists(Middleware::class)) {
+        if (! class_exists(Middleware::class)) {
             return;
         }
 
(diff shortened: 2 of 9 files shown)


greg0ire:
You can run `vendor/bin/phpcbf` to fix most CS issues :)

DependencyInjection/Compiler/MiddlewaresPass.php
```diff
@@ -0,0 +1,59 @@
...
+use Doctrine\Bundle\DoctrineBundle\Middleware\ConnectionNameAwareInterface;
+use Symfony\Component\DependencyInjection\ChildDefinition;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+use function array_keys;
+use function is_subclass_of;
+use function sprintf;
+
+class MiddlewaresPass implements CompilerPassInterface
```
greg0ire:
```suggestion
final class MiddlewaresPass implements CompilerPassInterface
```

DependencyInjection/Compiler/MiddlewaresPass.php
```diff
@@ -0,0 +1,59 @@
...
+{
+    /** @var string  */
+    private $connexionDefsParam;
```
greg0ire:
```suggestion
    private $connectionDefsParam;
```
:fr: spotted :slightly_smiling_face:

DependencyInjection/Compiler/MiddlewaresPass.php
```diff
@@ -0,0 +1,59 @@
...
+
+    public function __construct(
+        string $connexionDefsParam = 'doctrine.connections',
+        string $middlewareTag = 'doctrine.middleware'
+    ) {
+        $this->connexionDefsParam = $connexionDefsParam;
+        $this->middlewareTag      = $middlewareTag;
+    }
+
+    public function process(ContainerBuilder $container)
```
greg0ire:
```suggestion
    public function process(ContainerBuilder $container): void
```

DependencyInjection/Compiler/MiddlewaresPass.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        $middlewareAbstractDefs = [];
+        foreach (array_keys($container->findTaggedServiceIds($this->middlewareTag)) as $id) {
+            $def                         = $container->getDefinition($id);
+            $middlewareAbstractDefs[$id] = $def;
+        }
+
+        $connections = $container->getParameter($this->connexionDefsParam);
+        foreach ($connections as $name => $id) {
```
greg0ire:
```suggestion
        foreach ( $container->getParameter($this->connectionDefsParam) as $name => $id) {
```

DependencyInjection/Compiler/MiddlewaresPass.php
```diff
@@ -0,0 +1,59 @@
...
+        $this->connexionDefsParam = $connexionDefsParam;
+        $this->middlewareTag      = $middlewareTag;
+    }
+
+    public function process(ContainerBuilder $container)
+    {
+        $middlewareAbstractDefs = [];
+        foreach (array_keys($container->findTaggedServiceIds($this->middlewareTag)) as $id) {
+            $def                         = $container->getDefinition($id);
+            $middlewareAbstractDefs[$id] = $def;
```
greg0ire:
```suggestion
            $middlewareAbstractDefs[$id] = $container->getDefinition($id);
```

DependencyInjection/Compiler/MiddlewaresPass.php
```diff
@@ -0,0 +1,59 @@
...
+                $middlewareDefs[] = $childDef = $container->setDefinition(
+                    sprintf('%s.%s', $id, $name),
+                    new ChildDefinition($id)
+                );
+
+                if (! is_subclass_of($abstractDef->getClass(), ConnectionNameAwareInterface::class)) {
+                    continue;
+                }
+
+                $childDef->addMethodCall('setConnectionName', [$name]);
```
greg0ire, line 50: I get the need for early continue/return, etc. but here it seems needlessly convoluted as it requires a logical inversion and an extra statement.

```suggestion
                if (is_subclass_of($abstractDef->getClass(), ConnectionNameAwareInterface::class)) {
                    $childDef->addMethodCall('setConnectionName', [$name]);
                }
```

Middleware/ConnectionNameAwareInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\Middleware;
+
+interface ConnectionNameAwareInterface
+{
+    public function setConnectionName(string $name): void;
+}
```
greg0ire, line 8: What's the use case for this?

l-vo:
> You can run `vendor/bin/phpcbf` to fix most CS issues :)

Thank you, I missed that cs failed

DependencyInjection/Compiler/MiddlewaresPass.php
l-vo: 😅 thank you

Middleware/ConnectionNameAwareInterface.php
l-vo: I wrote this PR with the replacement of `DebugStack` by middlewares in mind. In the symfony profiler, queries are grouped by connection. I faced a choice:
- either I inject in the middleware a different service (kind of debug data holder) by connection
- either the middleware is aware of the connection and set queries in an unique debug data holder tied with their connection

It's not a big deal to duplicate an abstract middleware for each connection with a tag, but duplicating too the debug data holder seemed to me more complicated; it's why I chose the second possibility.

This is how I plan to create the debug data holder:
```php
class DebugDataHolder
{
    private array $data = [];

    public function addQuery(string $connexionName, Query $query): void
    {
        $this->data[$connexionName][] = [
            'sql' => $query->getSql(),
            'params' => $query->getParams(),
            'types' => $query->getTypes(),
            'executionMS' => static fn() => $query->getDuration(),  // stop() may not be called at this point
        ];
    }

    public function getData(string $connexionName): array
    {
        if (!isset($this->data[$connexionName])) {
            return [];
        }

        foreach ($this->data[$connexionName] as &$data) {
            if (is_callable($data['executionMS'])) {
                $data['executionMS'] = $data['executionMS']();
            }
        }

        return $this->data[$connexionName];
    }
}
```

DependencyInjection/Compiler/MiddlewaresPass.php
```diff
@@ -0,0 +1,57 @@
...
+{
+    /** @var string  */
+    private $connectionDefsParam;
+
+    /** @var string  */
+    private $middlewareTag;
+
+    public function __construct(
+        string $connectionDefsParam = 'doctrine.connections',
+        string $middlewareTag = 'doctrine.middleware'
```
ostrolucky: Why would this be configurable here if we hardcode it in DoctrineExtension?

DependencyInjection/DoctrineExtension.php
```diff
@@ -143,9 +144,18 @@ protected function dbalLoad(array $config, ContainerBuilder $container)
         $container->setParameter('doctrine.connections', $connections);
         $container->setParameter('doctrine.default_connection', $this->defaultConnection);
 
+        /** @psalm-suppress UndefinedClass */
+        if (interface_exists(Middleware::class)) {
+            $container
+                ->getDefinition('doctrine.dbal.logger')
+                ->replaceArgument(0, null);
+        }
+
```
ostrolucky: Why was this moved out of `loadMiddlewares`?

DependencyInjection/DoctrineExtension.php
```diff
@@ -143,9 +144,18 @@ protected function dbalLoad(array $config, ContainerBuilder $container)
         $container->setParameter('doctrine.connections', $connections);
         $container->setParameter('doctrine.default_connection', $this->defaultConnection);
 
+        /** @psalm-suppress UndefinedClass */
+        if (interface_exists(Middleware::class)) {
+            $container
+                ->getDefinition('doctrine.dbal.logger')
+                ->replaceArgument(0, null);
```
ostrolucky: This disables logging even when it's enabled, right? Basically it will block release of 2.6 once we merge this, until you finish follow-up PR

DependencyInjection/Compiler/MiddlewaresPass.php
l-vo: Not needed here indeed

DependencyInjection/DoctrineExtension.php
l-vo: It disables logging but the logging is now done by the logging middleware, see `Resources/config/middlewares.xml` in this PR.

DependencyInjection/DoctrineExtension.php
l-vo: I moved nothing out of `loadMiddlewares`, `loadMiddleware` has just been added in this PR. What am I missing ?

DependencyInjection/DoctrineExtension.php
ostrolucky: that's just abstract middleware

DependencyInjection/DoctrineExtension.php
ostrolucky: useMiddlewaresIfAvailable was renamed to loadMiddlewares. This condition was there.

l-vo: Ok sorry, it's a subjective choice, in my mind, `loadMiddleware` should only register middlewares, not disable what is replaced by a middleware. But I can move it back in `loadMiddleware` if you think it's better.

DependencyInjection/DoctrineExtension.php
l-vo: Yes but abstract middlewares are registered for each connection by `MiddlewarePass` 🙂

ostrolucky: I see. And which part disables middleware when $config['logging'] is false?

DependencyInjection/DoctrineExtension.php
ostrolucky: Point is to guarantee logger argument is replaced only when middlewares are registered. After this split, nothing guarantees it anymore and one can call these separately. Perhaps old name was better then, since `useMiddlewares` doesn't imply it only loads them.

l-vo: The tests guarantee that 🙂 But I understand the need from a code point of view, I moved it back into `useMiddlewaresIfAvailable` (I renamed the method too).

DependencyInjection/DoctrineExtension.php
l-vo: You're right, I missed this case, to allow to handle this case, I did some other changes:
- Middleware can now be registered only for chosen connections (with a tag attribute)
- An attribute `AsMiddleware` has been added

DependencyInjection/Compiler/MiddlewaresPass.php
```diff
@@ -0,0 +1,58 @@
...
+                }
+
+                $middlewareConnections[$id][] = $tag['connection'];
+            }
+        }
+
+        foreach (array_keys($container->getParameter('doctrine.connections')) as $name) {
+            $middlewareDefs = [];
+            foreach ($middlewareAbstractDefs as $id => $abstractDef) {
+                if (isset($middlewareConnections[$id]) && array_search($name, $middlewareConnections[$id]) === false) {
```
ostrolucky:
```suggestion
                if (isset($middlewareConnections[$id]) && ! in_array($name, $middlewareConnections[$id], true)) {
```

DependencyInjection/DoctrineExtension.php
```diff
@@ -1073,25 +1097,23 @@ private function createArrayAdapterCachePool(ContainerBuilder $container, string
...
-            new ChildDefinition('doctrine.dbal.logging_middleware')
-        );
-        $loggingMiddlewareDef->addArgument(new Reference('logger', ContainerInterface::NULL_ON_INVALID_REFERENCE));
-        $loggingMiddlewareDef->addTag('monolog.logger', ['channel' => 'doctrine']);
+        $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
+        $loader->load('middlewares.xml');
 
-        $configuration->addMethodCall('setMiddlewares', [[$loggingMiddlewareDef]]);
+        foreach ($connWithLogging as $connName) {
+            $container->getDefinition('doctrine.dbal.logging_middleware')->addTag('doctrine.middleware', ['connection' => $connName]);
```
ostrolucky: small optimization: `$container->getDefinition('doctrine.dbal.logging_middleware')` should be assigned to variable outside of loop so DI doesn't have to search for it on each iteration

Tests/DependencyInjection/Compiler/MiddlewarePassTest.php
```diff
@@ -0,0 +1,292 @@
...
+                ->setPublic(true); // Avoid removal and inlining
+        });
+
+        /** @psalm-suppress UndefinedClass */
+        $this->assertMiddlewareInjected($container, 'conn1', Middleware3::class);
+        /** @psalm-suppress UndefinedClass */
+        $this->assertMiddlewareInjected($container, 'conn2', Middleware3::class);
+    }
+
+    /** @return array<string, mixed[]> */
```
ostrolucky: in future instead please use syntax like such `@return array<string, array{0: class-string, 1: bool}>`

Tests/DependencyInjection/Compiler/MiddlewarePassTest.php
```diff
@@ -0,0 +1,292 @@
...
+class Middleware2 implements ConnectionNameAwareInterface
+{
+    public function setConnectionName(string $name): void
+    {
+    }
+}
+
+/** @psalm-suppress UndefinedClass */
+if (interface_exists(Middleware::class)) {
+    class Middleware3 implements Middleware
```
ostrolucky:
```suggestion
    class AutoconfiguredPHP7Middleware implements Middleware
```

Tests/DependencyInjection/Compiler/MiddlewarePassTest.php
```diff
@@ -0,0 +1,292 @@
...
+    {
+        public function wrap(Driver $driver): Driver
+        {
+            return $driver;
+        }
+    }
+
+    if (PHP_VERSION_ID >= 80000) {
+        #[AsMiddleware]
+        class Middleware4
```
ostrolucky:
```suggestion
        class AutoconfiguredMiddleware
```

Tests/DependencyInjection/Compiler/MiddlewarePassTest.php
```diff
@@ -0,0 +1,292 @@
...
+        {
+        }
+
+        #[AsMiddleware(connections: ['conn2'])]
+        class Middleware5
```
ostrolucky:
```suggestion
        class AutoconfiguredConnectionAwareMiddleware
```

Tests/DependencyInjection/Compiler/MiddlewarePassTest.php
```diff
@@ -0,0 +1,292 @@
...
+
+        return $middlewaresFound;
+    }
+}
+
+class Middleware1
+{
+}
+
+class Middleware2 implements ConnectionNameAwareInterface
```
ostrolucky:
```suggestion
class ConnectionAwarePHP7Middleware implements ConnectionNameAwareInterface
```

Tests/DependencyInjection/Compiler/MiddlewarePassTest.php
```diff
@@ -0,0 +1,292 @@
...
+
+                $middlewaresFound[] = $middlewareDef;
+            }
+        }
+
+        return $middlewaresFound;
+    }
+}
+
+class Middleware1
```
ostrolucky:
```suggestion
class PHP7Middleware
```

DependencyInjection/DoctrineExtension.php
```diff
@@ -143,9 +146,31 @@ protected function dbalLoad(array $config, ContainerBuilder $container)
...
         }
+
+        /** @psalm-suppress UndefinedClass */
+        $container->registerForAutoconfiguration(MiddlewareInterface::class)->addTag('doctrine.middleware');
+
+        if (PHP_VERSION_ID >= 80000 && method_exists(ContainerBuilder::class, 'registerAttributeForAutoconfiguration')) {
+            $container->registerAttributeForAutoconfiguration(AsMiddleware::class, static function (ChildDefinition $definition, AsMiddleware $attribute) {
+                if ($attribute->connections === []) {
+                    $definition->addTag('doctrine.middleware');
+                } else {
```
ostrolucky: early return would be better to avoid nesting

ostrolucky (changes requested):
Nice work, I think there are only minor comments

DependencyInjection/Compiler/MiddlewaresPass.php
l-vo: Indeed, I don't know why I used array_search 😅

Tests/DependencyInjection/Compiler/MiddlewarePassTest.php
l-vo: I took the freedom to rename it as `AutoconfiguredMiddlewareWithConnection` instead to avoid confusion with `ConnectionNameAwareInterface` (a middleware can target a specific connection without being aware of this connection name); is it ok for you ?

l-vo:
@ostrolucky thank you, all comments are addressed.

ostrolucky (approved):

ostrolucky:
LGTM, but would be nice if community properly tested this. I didn't try it in a project

greg0ire:
@l-vo sorry to comment about this more than 1 year later, is there documentation on this somewhere?

l-vo:
@greg0ire yes you're right, I forgot to document the configuration. I'm going to do that soon.

greg0ire:
Great! Thanks in advance!

l-vo:
@greg0ire done in #1653 :)

PR #1478, opened 2022-02-13
Deprecate "collate" default table option

greg0ire:
"collation" is what is expected now on the DBAL side.

### DIFF
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -11,6 +11,7 @@ use Doctrine\DBAL\Exception as DBALException;
 use Doctrine\DBAL\Exception\DriverException;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\Type;
+use Doctrine\Deprecations\Deprecation;
 
 use function array_merge;
 use function defined;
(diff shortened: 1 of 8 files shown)


ConnectionFactory.php
```diff
@@ -86,6 +87,11 @@ public function createConnection(array $params, ?Configuration $config = null, ?
                        after dropping support for doctrine/dbal 2, since this
                        package requires doctrine/dbal 3.3.2 or higher. */
                     if (isset($params['defaultTableOptions']['collate']) && defined('Doctrine\DBAL\Connection::PARAM_ASCII_STR_ARRAY')) {
+                        Deprecation::trigger(
+                            'doctrine/doctrine-bundle',
+                            'https://github.com/doctrine/DoctrineBundle/pull/1478',
```
ostrolucky: Wouldn't it be better if this was link to DBAL where they changed from collate to collation?

greg0ire: I think you are right, since here `defaultTableOptions` is only validated as an array, and there is no validation of the keys. Let me change that.

ostrolucky (approved):

PR #1481, opened 2022-03-02
Fix default MySQL charset with doctrine/dbal > 2

### LINKED ISSUE #1480: False collation on current 2.6.x-dev 89b7ed6 version
### Bug Report

I'm trying to find out why my table schema is not correclty utf8mb4 after using `@dev` dependencies.

|    Q        |   A
|------------ | ------
| Version     | 2.6.x-dev 89b7ed6
| MySQL       | 5.7.37 / 8.0.27
| PHP         | 8.1.2

#### Summary

On current stable version all is correctly in my database utf8mb4. I have no collation set manually in doctrine.yaml. But after ugprading to `@dev` version the collation is created not longer as utf8mb4 which was the default since I think dbal 3 for mysql.

#### Current behaviour

The schema has a strange mixed collation now:

Bildschirmfoto 2022-03-02 um 13 33 18

#### How to reproduce

```php
composer create-project sulu/skeleton
cd skeleton
composer config minimum-stability dev
composer update # installs doctrine-bundle 2.6.x-dev 89b7ed6
docker compose up
bin/console sulu:build dev
# check database e.g. `ac_activities` table
```

#### Expected behaviour

On 2.5.6 all works like expected:

Bildschirmfoto 2022-03-02 um 13 36 43

Maybe related to https://github.com/doctrine/DoctrineBundle/pull/1478 /cc @greg0ire

alexander-schranz:
The `$driver` variable can return `Doctrine\DBAL\Logging\Driver` which is not instance of `AbstractMySQLDriver`. So it need to be via the platform.

Sadly `AbstractMySQLDriver` is not available on prefer lowest so we need to check on the deprecated `getName` method on the platform.

fixes #1480

### DIFF
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -5,11 +5,9 @@ namespace Doctrine\Bundle\DoctrineBundle;
 use Doctrine\Common\EventManager;
 use Doctrine\DBAL\Configuration;
 use Doctrine\DBAL\Connection;
-use Doctrine\DBAL\Driver\AbstractMySQLDriver;
 use Doctrine\DBAL\DriverManager;
 use Doctrine\DBAL\Exception as DBALException;
 use Doctrine\DBAL\Exception\DriverException;
-use Doctrine\DBAL\Platforms\AbstractMySQLPlatform;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\Type;
 
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -8,6 +8,8 @@ use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DriverManager;
 use Doctrine\DBAL\Exception as DBALException;
 use Doctrine\DBAL\Exception\DriverException;
+use Doctrine\DBAL\Platforms\AbstractMySQLPlatform;
+use Doctrine\DBAL\Platforms\MySqlPlatform;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\Type;
 
@@ -72,7 +74,9 @@ class ConnectionFactory
             $driver     = $connection->getDriver();
 
             if (! isset($params['charset'])) {
-                if ($driver->getDatabasePlatform()->getName() === 'mysql') {
+                if ($driver->getDatabasePlatform() instanceof AbstractMySQLPlatform
+                    || $driver->getDatabasePlatform() instanceof MySqlPlatform
+                ) {
                     $params['charset'] = 'utf8mb4';
 
                     /* PARAM_ASCII_STR_ARRAY is defined since doctrine/dbal 3.3
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -41,6 +41,7 @@ class ConnectionFactory
      * @param mixed[]               $params
      * @param array<string, string> $mappingTypes
      * @psalm-param Params $params
+     * @psalm-suppress UndefinedClass
      *
      * @return Connection
      */
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -9,6 +9,7 @@ use Doctrine\DBAL\Driver\AbstractMySQLDriver;
 use Doctrine\DBAL\DriverManager;
 use Doctrine\DBAL\Exception as DBALException;
 use Doctrine\DBAL\Exception\DriverException;
+use Doctrine\DBAL\Platforms\AbstractMySQLPlatform;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\Type;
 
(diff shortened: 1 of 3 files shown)


Tests/ConnectionFactoryTest.php
```diff
@@ -71,7 +71,7 @@ public function testDefaultCharset(): void
         $connection    = $factory->createConnection($params);
 
         $this->assertInstanceof(FakeConnection::class, $connection);
-        $this->assertSame('utf8', $connection->getParams()['charset']);
+        $this->assertSame('utf8mb4', $connection->getParams()['charset']);
```
alexander-schranz: the FakeDriver is actually a MysqlPlatform that is why this test changees now.

ConnectionFactory.php
```diff
@@ -78,7 +77,7 @@ public function createConnection(array $params, ?Configuration $config = null, ?
             }
 
             if (! isset($params['charset'])) {
-                if ($driver instanceof AbstractMySQLDriver) {
+                if ($driver->getDatabasePlatform()->getName() === 'mysql') {
```
alexander-schranz: This is sadly calling a deprecated method but I could not find another method which works on this range of supported doctrine/dbal versions.

Tests/ConnectionFactoryTest.php
greg0ire: Are you sure this is OK? It does not look OK to change that default value like that (I know it's a bad default, but it might result in a lot of migrations for a lot of people).

alexander-schranz: The fake driver ist actually a mysqlplatform that is why it is changing : https://github.com/doctrine/DoctrineBundle/blob/6e578fd23ddb656af5c47c1014576490cebd4b0a/Tests/ConnectionFactoryTest.php#L156

alexander-schranz: Before the check was on the driver not on the platform.

greg0ire: :thinking: in fact, I think it might be fine, since people using MySQL ended up with `utf8mb4` before https://github.com/doctrine/DoctrineBundle/pull/1456, but maybe we should rename the test to `testDefaultCharsetForMySQL`, and duplicate it with `testDefaultCharsetForNonMySQL` to test the other codepath with `utf8`

ConnectionFactory.php
greg0ire: Maybe by performing several `instanceof` checks?

alexander-schranz:
@greg0ire any idea how we can get the `testContainer` working when we now access the `getDatabasePlatform` in the factory?

ConnectionFactory.php
```diff
@@ -78,7 +79,9 @@ public function createConnection(array $params, ?Configuration $config = null, ?
             }
 
             if (! isset($params['charset'])) {
-                if ($driver instanceof AbstractMySQLDriver) {
+                if ($driver->getDatabasePlatform() instanceof AbstractMySQLPlatform
```
greg0ire: The resulting Psalm error will have to be ignored through configuration I think.

greg0ire:
> @greg0ire any idea how we can get the testContainer working when we now access the getDatabasePlatform in the factory?

Let's improve the error message in the test first?  "Failed asserting that false is true." is not very helpful

alexander-schranz:
Improved the message by using `assertStringContainsString` instead of some strpos() > 0

ConnectionFactory.php
alexander-schranz: I'm not familiar with psalm do I need to ignore the whole file or is there another way?

alexander-schranz: okay seems `@psalm-suppress UndefinedClass` does the job

alexander-schranz:
Is it maybe to early to access the platform because the platform is validated later here: https://github.com/doctrine/DoctrineBundle/blob/eb7f3a25a483152238d63c377bba8a304ce3538c/ConnectionFactory.php#L110

Which is the method the `testContainer` test is testing. I'm a little bit stuck now.

dmaicher:
> The `$driver` variable can return `Doctrine\DBAL\Logging\Driver` which is not instance of `AbstractMySQLDriver`. So it need to be via the platform.

Ah nice catch. So this is caused by [wiring the logging middlewares](https://github.com/doctrine/DoctrineBundle/compare/2.5.6..2.6.x#diff-afe5f060789287a8ad428a4b5df1ad844184f496ece0297e421f6b1fc442b5a4R1103)

Tests/ConnectionFactoryTest.php
```diff
@@ -51,7 +54,7 @@ public function testContainer(): void
         try {
             $factory->createConnection($params, $config, $eventManager, $mappingTypes);
         } catch (Throwable $e) {
-            $this->assertTrue(strpos($e->getMessage(), 'can circumvent this by setting') > 0);
+            $this->assertStringContainsString($e->getMessage(), 'can circumvent this by setting');
```
greg0ire: It should be the other way around:

```suggestion
            $this->assertStringContainsString('can circumvent this by setting', $e->getMessage());
```

ostrolucky:
should target 2.5.x branch

greg0ire:
At that point I'm starting to think the way forward would be to make the middlewares opt-in, and throw if the charset is not specified when using them.
From my point of view, a lot of the code in  `ConnectionFactory` should go as it seems like hacks to fix issues that should either be fixed in DBAL, or defaults that should come from the framework (typically from Symfony recipes). Recipes are IMO the correct place to set defaults, because you can change the defaults there without that being a BC-break.

So I think we should also deprecate not specifying the charset.

Maybe @ostrolucky has an opinion to share on this as well?

ostrolucky:
I share same opinion

alexander-schranz:
Thank you all 👍

Issue #1485, opened 2022-03-30
composer cache:clear fails since upgrading DoctrineBundle 2.5.7->2.6.0


owsygavin:
I ran `composer upgrade`:

```
Lock file operations: 0 installs, 1 update, 0 removals
  - Upgrading doctrine/doctrine-bundle (2.5.7 => 2.6.0)
```

I got :

```
Executing script cache:clear [KO]
 [KO]
Script cache:clear returned with error code 255
!!  Symfony\Component\ErrorHandler\Error\UndefinedMethodError {#65817
!!    #message: "Attempted to call an undefined method named "setMiddlewares" of class "Doctrine\DBAL\Configuration"."
!!    #code: 0
!!    #file: "./var/cache/dev/ContainerMFCvxZk/App_KernelDevDebugContainer.php"
!!    #line: 1267
!!    trace: {
!!      ./var/cache/dev/ContainerMFCvxZk/App_KernelDevDebugContainer.php:1267 {
!!        ContainerMFCvxZk\App_KernelDevDebugContainer->getDoctrine_Dbal_DefaultConnectionService()
!!        › $a->setSchemaAssetsFilter(new \Doctrine\Bundle\DoctrineBundle\Dbal\SchemaAssetsFilterManager([0 => ($this->privates['doctrine.dbal.well_known_schema_asset_filter'] ?? ($this->privates['doctrine.dbal.well_known_schema_asset_filter'] = new \Doctrine\Bundle\DoctrineBundle\Dbal\BlacklistSchemaAssetFilter([0 => 'cache_items', 1 => 'sessions']))), 1 => new \Doctrine\Bundle\DoctrineBundle\Dbal\RegexSchemaAssetFilter('~^(?!(import\\.|reporting\\.|sqitch\\.|accounting\\.|history\\.|topology\\.|tiger\\.|tiger_data\\.))~')]));
!!        › $a->setMiddlewares([]);
!!        ›
!!      }
!!      ./var/cache/dev/ContainerMFCvxZk/App_KernelDevDebugContainer.php:1508 { …}
!!      ./var/cache/dev/ContainerMFCvxZk/App_KernelDevDebugContainer.php:1282 { …}
!!      ./var/cache/dev/ContainerMFCvxZk/EntityManager_9a5be93.php:246 { …}
!!      ./var/cache/dev/ContainerMFCvxZk/EntityManager_9a5be93.php:246 { …}
[...]

PR #1486, opened 2022-03-30
Do not configure middlewares in case of doctrine/dbal v2 installed

andrew-demb:
Should help for #1485

### DIFF
--- a/DependencyInjection/Compiler/MiddlewaresPass.php
+++ b/DependencyInjection/Compiler/MiddlewaresPass.php
@@ -3,6 +3,7 @@
 namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
 
 use Doctrine\Bundle\DoctrineBundle\Middleware\ConnectionNameAwareInterface;
+use Doctrine\DBAL\Driver\Middleware;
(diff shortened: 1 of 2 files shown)


DependencyInjection/Compiler/MiddlewaresPass.php
```diff
@@ -16,6 +17,11 @@ final class MiddlewaresPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container): void
     {
+        // Middlewares introduced in `doctrine/dbal:3.0.0`
+        if (! interface_exists(Middleware::class)) {
```
dmaicher, line 21: So as I see it this pass shouldn't even be registered using dbal 2?

https://github.com/doctrine/DoctrineBundle/blob/2.6.x/DoctrineBundle.php#L68

andrew-demb: Yes, it should not be registered.

Then I don't know the reason for the error from #1485.

andrew-demb: Yes, it should not be registered.

Then I don't know the reason for the error from #1485.
:(

PR #1488, opened 2022-03-30
Fix error for logging middleware when logger is missing

### LINKED ISSUE #1487: Error on middleware on 2.6 with logger
PHP Fatal error:  Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: Argument 1 passed to Doctrine\DBAL\Logging\Middleware::__construct() must implement interface Psr\Log\LoggerInterface, null given, called in /home/runner/work/SyliusResourceBundle/SyliusResourceBundle/src/Bundle/test/var/cache/test/ContainerRRyfIex/srcApp_KernelTestDebugContainer.php on line 3913 in /home/runner/work/SyliusResourceBundle/SyliusResourceBundle/vendor/doctrine/dbal/src/Logging/Middleware.php:16

versions
* doctrine/dbal 3.3.4

This is not the same error message as https://github.com/doctrine/DoctrineBundle/issues/1485 so, I added this one.

l-vo:
Fix #1487

### DIFF
--- a/Tests/DependencyInjection/Compiler/MiddlewarePassTest.php
+++ b/Tests/DependencyInjection/Compiler/MiddlewarePassTest.php
@@ -9,6 +9,7 @@ use Doctrine\Bundle\DoctrineBundle\Middleware\ConnectionNameAwareInterface;
 use Doctrine\DBAL\Driver;
 use Doctrine\DBAL\Driver\Middleware;
 use PHPUnit\Framework\TestCase;
+use Psr\Log\NullLogger;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\ParameterBag\ParameterBag;
@@ -169,6 +170,10 @@ class MiddlewarePassTest extends TestCase
     private function createContainer(callable $func): ContainerBuilder
     {
         $container = new ContainerBuilder(new ParameterBag(['kernel.debug' => false]));
+
+        $loggerDef = new Definition();
+        $loggerDef->setClass(NullLogger::class);
+        $container->setDefinition('logger', $loggerDef);
 
         $container->registerExtension(new DoctrineExtension());
         $container->loadFromExtension('doctrine', [
--- a/Tests/DependencyInjection/Compiler/RemoveLoggingMiddlewarePassTest.php
+++ b/Tests/DependencyInjection/Compiler/RemoveLoggingMiddlewarePassTest.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection\Compiler;
+
+use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\RemoveLoggingMiddlewarePass;
+use PHPUnit\Framework\TestCase;
+use Psr\Log\NullLogger;
+use Symfony\Component\Config\FileLocator;
+use Symfony\Component\DependencyInjection\ChildDefinition;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
+
+final class RemoveLoggingMiddlewarePassTest extends TestCase
+{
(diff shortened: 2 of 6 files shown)


Tests/DependencyInjection/Compiler/MiddlewarePassTest.php
```diff
@@ -290,3 +295,44 @@ class AutoconfiguredMiddlewareWithConnection
         }
     }
 }
+
+// phpcs:disable
+
+final class MyLogger implements LoggerInterface
```
dmaicher: instead we can probably use `Psr\Log\NullLogger`?

Tests/DependencyInjection/Compiler/RemoveLoggingMiddlewarePassTest.php
```diff
@@ -0,0 +1,102 @@
...
+            }
+        });
+
+        return $container;
+    }
+}
+
+// phpcs:disable
+
+final class Logger implements LoggerInterface
```
dmaicher: same

composer.json
```diff
@@ -47,6 +47,7 @@
         "phpunit/phpunit": "^7.5 || ^8.0 || ^9.3 || ^10.0",
         "psalm/plugin-phpunit": "^0.16.1",
         "psalm/plugin-symfony": "^3",
+        "psr/log": "^1.1.4",
```
dmaicher: I guess `^2.0` and `^3.0` are then also compatible, or?

dmaicher (commented):
thanks for looking into this :+1:

l-vo:
@dmaicher changes applied

dmaicher (approved):

dmaicher:
~@loic425 could you confirm that this fixes https://github.com/doctrine/DoctrineBundle/issues/1487 for you?~ tested it myself.

dmaicher:
Tested it myself: Works fine if no `logger` service is defined and fixes https://github.com/doctrine/DoctrineBundle/issues/1487

@ostrolucky do you want to have a look? Or shall I go ahead and merge it?

ostrolucky:
I probably won't review this today. You can go ahead if you feel this is urgent. I don't personally see the urgency: who the heck doesn't have logger in prod?

Tests/DependencyInjection/Compiler/RemoveLoggingMiddlewarePassTest.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        $container = $this->createContainer();
+
+        $logger = (new Definition())
+            ->setClass(NullLogger::class);
+        $container->setDefinition('logger', $logger);
+
+        $container->compile();
+
+        $this->assertTrue($container->hasDefinition('logging_middleware_child'));
```
ostrolucky, line 35: Why is this to be an expected case? This child definition will still carry reference to non-existent logger service, won't it? Wouldn't that cause issues?

l-vo: This test is about the case where the logger is present 🙂

ostrolucky (approved):

loic425:
Thank you! I'll try on monday. Have a good weekend

yura3d:
@l-vo Is it really need to always remove `doctrine.dbal.logging_middleware` definition in case no service named `logger` presents in container, even if the definition itself contains argument with its own logger?
https://github.com/doctrine/DoctrineBundle/blob/0ad7b24c6c6cfa492cd1815d676739f84193000a/DependencyInjection/Compiler/RemoveLoggingMiddlewarePass.php#L9-L19

I'm trying to set my custom `SQLLogger` service only for DBAL logs (if I name it `logger`, it starts to receive symfony-specific logs, that I don't need):
```yaml
App\Doctrine\SQLLogger:
    arguments:
        $logFilename: '%kernel.logs_dir%/sql.log'
doctrine.dbal.logging_middleware:
    class: 'Doctrine\DBAL\Logging\Middleware'
    arguments:
        - '@App\Doctrine\SQLLogger'
```
And this config doesn't work until I do something like this:
```yaml
logger:
    class: 'Psr\Log\NullLogger'
```

l-vo:
@yura3d What about using your own service id instead of reusing `doctrine.dbal.logging_middleware` ? You just need to add the `doctrine.middleware` tag or enable autoconfiguration for the service.

yura3d:
@l-vo Sorry, seems I didn't read the docs carefully. Thanks!

Issue #1491, opened 2022-03-31
Attempted to call an undefined method named "setMiddlewares" of class "Doctrine\DBAL\Configuration".


Ahummeling:
With version 2.6.0 of this bundle, I started getting the following error:
```
Attempted to call an undefined method named "setMiddlewares" of class "Doctrine\DBAL\Configuration".
```
I suppose this is due to some other Doctrine package not being updated as well on my end. In either way, I think the cause was introduced by #1472 .
I've made a PR that would fix the issue for me. 
PR: #1492 

Full stack trace:
```shell
bin/console lint:container
Symfony\Component\ErrorHandler\Error\UndefinedMethodError {#2659
  #message: "Attempted to call an undefined method named "setMiddlewares" of class "Doctrine\DBAL\Configuration"."
  #code: 0
  #file: "./Tests/App/var/cache/test/ContainerElFnfiZ/getDoctrine_Dbal_DefaultConnection_ConfigurationService.php"
  #line: 25
  trace: {
    ./Tests/App/var/cache/test/ContainerElFnfiZ/getDoctrine_Dbal_DefaultConnection_ConfigurationService.php:25 {
      ContainerElFnfiZ\getDoctrine_Dbal_DefaultConnection_ConfigurationService::do($container, $lazyLoad = true)
      › $instance->setSQLLogger(($container->privates['doctrine.dbal.logger.chain.default'] ?? $container->load('getDoctrine_Dbal_Logger_Chain_DefaultService')));
      › $instance->setMiddlewares([]);
      › 
    }
    ./Tests/App/var/cache/test/ContainerElFnfiZ/Adivare_AdminBundle_Tests_AppKernelTestDebugContainer.php:358 { …}
    ./Tests/App/var/cache/test/ContainerElFnfiZ/getDoctrine_Dbal_DefaultConnectionService.php:23 { …}
    ./Tests/App/var/cache/test/ContainerElFnfiZ/Adivare_AdminBundle_Tests_AppKernelTestDebugContainer.php:358 { …}
    ./Tests/App/var/cache/test/ContainerElFnfiZ/getDoctrine_Orm_DefaultEntityManagerService.php:39 { …}
    ./Tests/App/var/cache/test/ContainerElFnfiZ/getDoctrine_Orm_DefaultEntityManagerService.php:23 { …}
    ./Tests/App/var/cache/test/ContainerElFnfiZ/EntityManager_9a5be93.php:246 { …}
[...]

PR #1492, opened 2022-03-31
ensure middlewarespass checks for existence of method

Ahummeling:
this is the PR attempting to fix the issue described by #1491 

I am not very familiar with this code, but I wanted to propose my solution regardless.


dmaicher:
This is not an issue with DoctrineBundle but instead it was fixed in https://github.com/getsentry/sentry-symfony/pull/608

Ahummeling:
Oh the failing tests seem to indicate my "fix" prevents the `setMiddlewares` call from being set in every case.

PR #1495, opened 2022-04-01
#1494 - early return in MiddlewaresPass

### LINKED ISSUE #1494: MiddlewaresPass: You have requested a non-existent parameter "doctrine.connections"
Hi, there is an error in the `Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\MiddlewaresPass` in doctrine-bundle **2.6.0**, specifically on line 34:
```php
foreach (array_keys($container->getParameter('doctrine.connections')) as $name) {
```
The compiler pass expects the `doctrine.connections` parameter to be set at all times, but this doesn't have to be true, because in the bundle extension, there is this condition:
```php
if (! empty($config['dbal'])) {
```
The method `dbalLoad` populates the `doctrine.conntections` parameter.

Usually, this problem shouldn't be occurring, because in the applications, the dbal configuration has to be present when using Doctrine.
But there is a special case - when developing bundles where doctrine-bundle is just a soft dependency. Som test suites may have enabled doctrine globally, but not configured it for each of the tests. Such tests break because of the expectation of the MiddlewaresPass.

I still can workaround this in our test suite, but shouldn't this situation be considered as a bug?

Rastusik:
early return in MiddlewaresPass when doctrine.connections parameter is not present in the container

fixes https://github.com/doctrine/DoctrineBundle/issues/1494

### DIFF
--- a/DependencyInjection/Compiler/MiddlewaresPass.php
+++ b/DependencyInjection/Compiler/MiddlewaresPass.php
@@ -16,6 +16,10 @@ final class MiddlewaresPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container): void
     {
+        if (! $container->hasParameter('doctrine.connections')) {
+            return;
+        }
+
         $middlewareAbstractDefs = [];
         $middlewareConnections  = [];
         foreach ($container->findTaggedServiceIds('doctrine.middleware') as $id => $tags) {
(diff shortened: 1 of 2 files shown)


DependencyInjection/Compiler/MiddlewaresPass.php
```diff
@@ -31,6 +31,10 @@ public function process(ContainerBuilder $container): void
             }
         }
 
+        if (! $container->hasParameter('doctrine.connections')) {
```
dmaicher: we can even move this to the very top of the function, or? there is nothing useful to do in this case

dmaicher:
Covering this with a test would also be nice

Rastusik:
@dmaicher is tha last commit sufficient?

dmaicher (approved):

PR #1500, opened 2022-04-05
Logging: allow disabling logging for all connections

glaubinix:
Having one or multiple connections set up and all of them `logging: false` configured, activates logging via Middleware for all connection.

Sample configuration:
```
doctrine:
    dbal:
        connections:
            default:
                driver: 'pdo_pgsql'
                url: '%env(DATABASE_URL)%'
                logging: false
```

The `middlewares.xml` defines a `doctrine.middleware` tag by default for `Doctrine\DBAL\Logging\Middleware`, which the `MiddlewaresPass` then applies to all connections. Having one or more connections with `logging: true` will overwrite the tag and set `connections` to an array of connections, which then only enables logging for those configured connections. That second part is working as expected.

### DIFF
--- a/Resources/config/middlewares.xml
+++ b/Resources/config/middlewares.xml
@@ -8,7 +8,6 @@
         <service id="doctrine.dbal.logging_middleware" class="Doctrine\DBAL\Logging\Middleware" abstract="true">
             <argument type="service" id="logger" />
             <tag name="monolog.logger" channel="doctrine" />
-            <tag name="doctrine.middleware" />
         </service>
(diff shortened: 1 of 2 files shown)


stof (approved):

dmaicher:
@l-vo can you have a look at this too?

l-vo:
LGTM too, thank you :)

ostrolucky (approved):

dmaicher:
Thank you @glaubinix

Issue #1416, opened 2021-11-08
Driver options are not passed to the primary\replica configuration


GDXbsv:
When we use 
```
  'options'               => [
      \PDO::ATTR_STRINGIFY_FETCHES        => 1,
      \PDO::MYSQL_ATTR_USE_BUFFERED_QUERY => false
  ],
```
With one connection without replicas, this options passed in the driver configuration. (See \Doctrine\DBAL\Driver\PDO\MySQL\Driver::connect --> `$params['driverOptions']`)

When we add a replica, then the bundle do not include those driver options in the replica and primary configuration options.

ostrolucky:
We don't do that, because doing that causes other issues, like https://github.com/doctrine/DoctrineBundle/pull/1253. Can you try if you experience expected behaviour with doctrine-bundle 2.2.0? That's the version where we did pass driverOptions over to replicas.

GDXbsv:
I did try 2.2.0 it does not work.

1. It does not work with dbal3, so I had to test with dbal2
2. the bundle pass driver options in the new format for dbal3 which is not fit to dbal2, so it can not work at all.
3. But even in this case according to xdebug options exists only in primary but not in replica. In the last version, they were removed from primary as well.

I mostly need to pass to all the connections \PDO::MYSQL_ATTR_USE_BUFFERED_QUERY => false. Is it a way how I can do it differently?

ostrolucky:
I think you just have to pass options explicitly to each replica config instead of relying on them being passed from the primary config. How does your config look like?

GDXbsv:
But I can not pass. The bundle does not support it.
And in new version they even do not go in primary, the bundle do not pass them from the main config.

My config:
```
                    'default' => [
                        'server_version'        => '5.7',
                        'charset'               => 'utf8mb4',
                        'options'               => [
                            \PDO::ATTR_STRINGIFY_FETCHES        => 1,
                            \PDO::MYSQL_ATTR_USE_BUFFERED_QUERY => false
                        ],
                        'default_table_options' => [
                            'charset' => 'utf8mb4',
                            'collate' => 'utf8mb4_unicode_ci',
                        ],
                        'mapping_types'         => ['uuid_binary' => 'binary'],
                        'url'                   => '%env(resolve:DATABASE_URL)%',
                        'keep_replica'          => false,
                         'replicas'              => [
                             'replica' => [
                                 'url'            => '%env(DATABASE_REPLICA_URL)%',
                             ],
[...]

PR #1506, opened 2022-04-09
fix: propagate driverOptions to primary-replica connections

simPod:
_I kind of struggle to find out what's going on about propagating DriverOptions to primary-replica connections. Searching through issues it seems it was removed in https://github.com/doctrine/DoctrineBundle/pull/1253, then requested to be added in https://github.com/doctrine/DoctrineBundle/issues/1416 and that was closed as resolved even though there were no changes._

me confused

This PR does not break any tests so I consider it a fix that does not break anything 🤷🏾‍♀️.

[DBAL's PrimaryReadReplicaConnection::connectTo()](https://github.com/doctrine/dbal/blob/719663b15983278227669c8595151586a2ff3327/src/Connections/PrimaryReadReplicaConnection.php#L244) method calls `chooseConnectionConfiguration()` which creates params for PDO in my case.

The problem is that `primary` entry already does not have the `driverOptions` entry. `driverOptions` is in top level only. So `$connectionParams` does not contain it, it gets stripped.

<details>
<summary>Values</summary>

`$params`:

```php
array (
  'driver' => 'pdo_mysql',
  'driverOptions' => 
  array (
    17 => 1,
  ),
  'keepReplica' => true,
  'replica' => 
  array (
    'r1' => 
    array (
      'url' => 'mysql://root:root@127.0.0.1:3307/__tests_1649508676_a0a89602',
      'host' => '127.0.0.1',
      'port' => 3307,
      'user' => 'root',
      'password' => 'root',
      'driver' => 'pdo_mysql',
      'dbname' => '__tests_1649506773_2e8ed0d5',
    ),
  ),
  'primary' => 
  array (
    'charset' => 'utf8',
    'url' => 'mysql://root:root@127.0.0.1:3307/__tests_1649508676_a0a89602',
    'host' => '127.0.0.1',
    'port' => 3307,
    'user' => 'root',
    'password' => 'root',
    'driver' => 'pdo_mysql',
    'dbname' => '__tests_1649508676_a0a89602',
  ),
)
```

`$connectionParams`:
```php
array (
  'charset' => 'utf8',
  'url' => 'mysql://root:root@127.0.0.1:3307/__tests_1649508676_a0a89602',
  'host' => '127.0.0.1',
  'port' => 3307,
  'user' => 'root',
  'password' => 'root',
  'driver' => 'pdo_mysql',
  'dbname' => '__tests_1649508676_a0a89602',
)
```
</details>

Since those are passed into `$this->_driver->connect($connectionParams)` which bubbles that into [Driver, the new PDO instance is created without it so DriverOptions are effectively ignored](https://github.com/doctrine/dbal/blob/719663b15983278227669c8595151586a2ff3327/src/Driver/PDO/MySQL/Driver.php#L18).

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -360,10 +360,21 @@ class DoctrineExtension extends AbstractDoctrineExtension
             throw new InvalidArgumentException('Sharding and master-slave connection cannot be used together');
         }
 
+        foreach (['shards', 'replica', 'slaves'] as $connectionKey) {
+            foreach (array_keys($options[$connectionKey]) as $name) {
+                $driverOptions       = $options[$connectionKey][$name]['driverOptions'] ?? [];
+                $parentDriverOptions = $options['driverOptions'] ?? [];
+                if ($driverOptions === [] && $parentDriverOptions === []) {
+                    continue;
+                }
+
+                $options[$connectionKey][$name]['driverOptions'] = $driverOptions + $parentDriverOptions;
+            }
+        }
+
         if (! empty($options['slaves']) || ! empty($options['replica'])) {
             $nonRewrittenKeys = [
                 'driver' => true,
-                'driverOptions' => true,
                 'driverClass' => true,
                 'wrapperClass' => true,
                 'keepSlave' => true,
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -361,8 +361,8 @@ class DoctrineExtension extends AbstractDoctrineExtension
         }
 
         foreach (['shards', 'replica', 'slaves'] as $connectionKey) {
-            foreach (array_keys($options[$connectionKey]) as $name) {
-                $driverOptions       = $options[$connectionKey][$name]['driverOptions'] ?? [];
+            foreach ($options[$connectionKey] as $name => $value) {
+                $driverOptions       = $value['driverOptions'] ?? [];
                 $parentDriverOptions = $options['driverOptions'] ?? [];
                 if ($driverOptions === [] && $parentDriverOptions === []) {
                     continue;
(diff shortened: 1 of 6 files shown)


adamhoof (approved):

ostrolucky:
This is a can of worms and I no longer want to touch it without integration tests. Sorry, but no matter what we do here, every time somebody comes and complains. I don't use mysql replicas myself, so won't provide support on my own either.

simPod:
TBH when somebody comes and complains he can send PR with tests. This was not covered before so let's start with it.  
Bundle doesn't have integration tests but I don't think it's necessary:

We obviously want the connections to respect config provided by bundle which, as is visible from tests change, was ignored. I have checked https://github.com/doctrine/DoctrineBundle/pull/1253 again and the change was wrong as it fixes the issue only in commands and not systematically by providing proper config to DBAL. I have reverted that change since mine fixes it globally. Let's cover it with tests over time. When somebody comes and complains, assign me. We'll have use for master-slave for few years at least.

DependencyInjection/DoctrineExtension.php
```diff
@@ -360,10 +360,21 @@ protected function getConnectionOptions(array $connection): array
             throw new InvalidArgumentException('Sharding and master-slave connection cannot be used together');
         }
 
+        foreach (['shards', 'replica', 'slaves'] as $connectionKey) {
+            foreach (array_keys($options[$connectionKey]) as $name) {
```
ostrolucky: array_intersect_key? Also, I see no reason to do array_keys +$options[$connectionKey][$name]['driverOptions'] ?? [] instead of normal foreach + $connectionOptions[$name]['driverOptions'] ?? []

ostrolucky (commented):
alright, let's give it one more try

DependencyInjection/DoctrineExtension.php
simPod: I changed it to foreach without array_keys. Was not sure for what to use `array_intersect_key` tho.

dmaicher (approved):

PR #1507, opened 2022-04-13
refactor: drop master

simPod:
I decided to clean up the mess caused by dropping standard master-slave naming. Should I clean slaves as well?


stof:
This cannot be done in a minor version.

stof:
ah, actually, it can as this only cleans the parameters passed from the bundle to DBAL (and so it is fine if the bundle requires DBAL 2.11+). but only if this still works when users of the bundle have not updated their config yet (I don't remember whether the bundle normalizes the config between `master` and `primary`).

simPod:
I have removed all `master` appearances. So I don't think there's any normalization. DBAL requires `primary` https://github.com/doctrine/dbal/blob/2.11.3/lib/Doctrine/DBAL/Connections/PrimaryReadReplicaConnection.php#L106

ostrolucky (approved):

SenseException (approved):

Issue #1467, opened 2022-01-26
Deprecate `shard` configuration


dmaicher:
In DBAL 3 the sharding functionality was [removed](https://github.com/doctrine/dbal/pull/3602).

Since DBAL 2 will be [EOL soon](https://www.doctrine-project.org/2022/01/22/sunsetting-dbal-2.html) we should deprecate the sharding related configuration options and probably also throw some meaningful exception if someone tries to configure it using DBAL 3?

PR #1508, opened 2022-04-15
deprecate sharding functionality

dmaicher:
Fixes https://github.com/doctrine/DoctrineBundle/issues/1467

### DIFF
--- a/Command/CreateDatabaseDoctrineCommand.php
+++ b/Command/CreateDatabaseDoctrineCommand.php
@@ -13,6 +13,7 @@ use Throwable;
 use function array_merge;
 use function in_array;
 use function sprintf;
+use function trigger_deprecation;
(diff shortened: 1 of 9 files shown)


dmaicher:
What's our plan for dropping support for Doctrine DBAL v2? We could probably do it for 2.8 then? So people still get a chance to see the deprecations triggered here and can upgrade accordingly?

ostrolucky (approved):

ostrolucky:
If we are going to remove these options, we have to do it in 3.0

dmaicher:
> If we are going to remove these options, we have to do it in 3.0

yeah although this cannot work with DBAL 3 I guess it could still be considered a BC break to remove those options. Let's do it on 3.0 then.

PR #1511, opened 2022-04-22
Fix Psalm errors caused by ORM 2.12 and Persistence 2.5

derrabus:


ostrolucky (approved):

ostrolucky:
ah, also CS needs to be fixedbefore merge

Repository/ServiceEntityRepository.php
```diff
@@ -22,7 +22,7 @@
  *     }
  * }
  *
- * @template T
+ * @template T of object
```
wcoppens, line 25: @derrabus This change is triggering a `MoreSpecificReturnType` and `LessSpecificReturnStatement` issue for me while using ORM 2.12.2 and Persistance 2.5.3, while it was working properly before.

stof: triggerring them where ?

Issue #1509, opened 2022-04-20
Fix doctrine/orm deprecation: create + wire EntityManagerProvider for commands


dmaicher:
Since https://github.com/doctrine/orm/pull/8524 there is a new way of providing entity managers for the ORM commands. The currently used Helperset is deprecated.

So as a follow-up for https://github.com/doctrine/DoctrineBundle/pull/1327 we should create a new class that implements `Doctrine\ORM\Tools\Console\EntityManagerProvider` and uses a `Doctrine\Persistence\ManagerRegistry` to provide the default or named manager instance for the ORM commands.

We should probably also bump the minimum supported ORM version to make it easier.

Did I forget something @beberlei ? :blush:

ostrolucky:
We already have https://github.com/doctrine/DoctrineBundle/blob/2.6.x/Dbal/ManagerRegistryAwareConnectionProvider.php. What else is needed?

dmaicher:
> We already have https://github.com/doctrine/DoctrineBundle/blob/2.6.x/Dbal/ManagerRegistryAwareConnectionProvider.php. What else is needed?

Yeah we have that one for DBAL. We need something similar for ORM that implements [`Doctrine\ORM\Tools\Console\EntityManagerProvider`](https://github.com/doctrine/orm/blob/2.12.x/lib/Doctrine/ORM/Tools/Console/EntityManagerProvider.php#L9).

PR #1513, opened 2022-04-26
implement EntityManagerProvider and use for ORM proxy commands

dmaicher:
Fixes https://github.com/doctrine/DoctrineBundle/issues/1509

### DIFF
--- a/Command/Proxy/ClearMetadataCacheDoctrineCommand.php
+++ b/Command/Proxy/ClearMetadataCacheDoctrineCommand.php
@@ -3,15 +3,15 @@
 namespace Doctrine\Bundle\DoctrineBundle\Command\Proxy;
 
 use Doctrine\ORM\Tools\Console\Command\ClearCache\MetadataCommand;
-use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
-use Symfony\Component\Console\Output\OutputInterface;
 
 /**
  * Command to clear the metadata cache of the various cache drivers.
  */
 class ClearMetadataCacheDoctrineCommand extends MetadataCommand
 {
+    use OrmProxyCommand;
+
     protected function configure(): void
     {
         parent::configure();
@@ -25,12 +25,5 @@ class ClearMetadataCacheDoctrineCommand extends MetadataCommand
         }
 
         $this->addOption('em', null, InputOption::VALUE_OPTIONAL, 'The entity manager to use for this command');
-    }
-
-    protected function execute(InputInterface $input, OutputInterface $output): int
-    {
-        DoctrineCommandHelper::setApplicationEntityManager($this->getApplication(), $input->getOption('em'));
-
-        return parent::execute($input, $output);
     }
 }
--- a/Tests/Command/Proxy/InfoDoctrineCommandTest.php
+++ b/Tests/Command/Proxy/InfoDoctrineCommandTest.php
@@ -2,68 +2,27 @@
 
 namespace Command\Proxy;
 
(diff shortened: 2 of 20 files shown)


Command/Proxy/ClearMetadataCacheDoctrineCommand.php
```diff
@@ -3,15 +3,15 @@
...
 
 use Doctrine\ORM\Tools\Console\Command\ClearCache\MetadataCommand;
-use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
-use Symfony\Component\Console\Output\OutputInterface;
 
 /**
  * Command to clear the metadata cache of the various cache drivers.
  */
 class ClearMetadataCacheDoctrineCommand extends MetadataCommand
```
dmaicher, line 11: Pretty much all of those commands are only changing the name and description of the proxied ORM commands now.

Whats our plan with those? Should we remove them at some point?

stof: For descriptions, I think we should use the upstream description (or submit the bundle descriptions upstream if they are better).
For names, we need to make sure that upstream commands allow to inject a name in the constructor (and then pass it to the parent constructor of Command which accepts a name) instead of calling `->setName` with a fixed name. Once this is OK, we won't need the proxy commands anymore.

DependencyInjection/DoctrineExtension.php
```diff
@@ -602,6 +603,13 @@ protected function ormLoad(array $config, ContainerBuilder $container)
             ->addTag('container.preload', [
                 'class' => Autoloader::class,
             ]);
+
+        $container
```
ostrolucky: why is this registered here and not in orm.xml?

Command/Proxy/ClearMetadataCacheDoctrineCommand.php
ostrolucky: I think we could also change upstream command to use $defaultName property. Then we could override command name via console.command tag.

As for getting rid of these commands, definitely. They were needed at some point when upstream didn't support --em option, but that's no longer the case.

dmaicher: I will create a new issue then to tackle this

DependencyInjection/DoctrineExtension.php
dmaicher: yeah can be moved to xml indeed :+1:

Command/Proxy/ClearMetadataCacheDoctrineCommand.php
andrew-demb: I have a PR that declares $defaultName for doctrine/orm :)
https://github.com/doctrine/orm/pull/7975

You can reuse it to implement changes faster, if you prefer

Tests/Command/Proxy/InfoDoctrineCommandTest.php
```diff
@@ -2,68 +2,27 @@
...
-use Symfony\Component\HttpKernel\Kernel;
-
-use function array_merge;
 
 class InfoDoctrineCommandTest extends TestCase
 {
     public function testExecute(): void
     {
-        $kernel = $this->setupKernelMocks();
+        $kernel = new TestKernel();
```
dmaicher, line 14: made this a functional test instead

Command/Proxy/ClearMetadataCacheDoctrineCommand.php
dmaicher: See https://github.com/doctrine/DoctrineBundle/issues/1524

ostrolucky (approved):

PR #1517, opened 2022-05-02
Replace DbalLogger by DebugMiddleware

### LINKED ISSUE #1431: Implement a DBAL middleware for the profiler collector

l-vo:
Fixes #1431

### DIFF
--- a/DataCollector/DoctrineDataCollector.php
+++ b/DataCollector/DoctrineDataCollector.php
@@ -13,6 +13,7 @@ use Doctrine\ORM\Tools\SchemaValidator;
 use Doctrine\Persistence\ManagerRegistry;
 use Doctrine\Persistence\Mapping\AbstractClassMetadataFactory;
 use Symfony\Bridge\Doctrine\DataCollector\DoctrineDataCollector as BaseCollector;
+use Symfony\Bridge\Doctrine\Middleware\Debug\DebugDataHolder;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Throwable;
@@ -64,12 +65,18 @@ class DoctrineDataCollector extends BaseCollector
     /** @var bool */
     private $shouldValidateSchema;
 
-    public function __construct(ManagerRegistry $registry, bool $shouldValidateSchema = true)
+    /** @psalm-suppress UndefinedClass */
+    public function __construct(ManagerRegistry $registry, bool $shouldValidateSchema = true, ?DebugDataHolder $debugDataHolder = null)
     {
         $this->registry             = $registry;
         $this->shouldValidateSchema = $shouldValidateSchema;
 
-        parent::__construct($registry);
+        if ($debugDataHolder === null) {
+            parent::__construct($registry);
+        } else {
+            /** @psalm-suppress TooManyArguments */
+            parent::__construct($registry, $debugDataHolder);
+        }
     }
 
     /**
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -16,7 +16,6 @@ use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection;
 use Doctrine\DBAL\Driver\Middleware as MiddlewareInterface;
 use Doctrine\DBAL\Logging\LoggerChain;
-use Doctrine\DBAL\Logging\Middleware;
 use Doctrine\DBAL\Sharding\PoolingShardConnection;
 use Doctrine\DBAL\Sharding\PoolingShardManager;
 use Doctrine\DBAL\Tools\Console\Command\ImportCommand;
@@ -34,6 +33,7 @@ use Symfony\Bridge\Doctrine\IdGenerator\UlidGenerator;
 use Symfony\Bridge\Doctrine\IdGenerator\UuidGenerator;
 use Symfony\Bridge\Doctrine\Messenger\DoctrineClearEntityManagerWorkerSubscriber;
 use Symfony\Bridge\Doctrine\Messenger\DoctrineTransactionMiddleware;
+use Symfony\Bridge\Doctrine\Middleware\Debug\Middleware as SfDebugMiddleware;
 use Symfony\Bridge\Doctrine\PropertyInfo\DoctrineExtractor;
 use Symfony\Bridge\Doctrine\SchemaListener\DoctrineDbalCacheAdapterSchemaSubscriber;
 use Symfony\Bridge\Doctrine\SchemaListener\MessengerTransportDoctrineSchemaSubscriber;
@@ -115,9 +115,6 @@ class DoctrineExtension extends AbstractDoctrineExtension
     {
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('dbal.xml');
-        $chainLogger = $container->getDefinition('doctrine.dbal.logger.chain');
-        $logger      = new Reference('doctrine.dbal.logger');
-        $chainLogger->addArgument([$logger]);
 
         if (class_exists(ImportCommand::class)) {
             $container->register('doctrine.database_import_command', ImportDoctrineCommand::class)
@@ -146,10 +143,20 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $container->setParameter('doctrine.connections', $connections);
         $container->setParameter('doctrine.default_connection', $this->defaultConnection);
 
-        $connWithLogging = [];
+        $connWithLogging   = [];
+        $connWithProfiling = [];
+        $connWithBacktrace = [];
         foreach ($config['connections'] as $name => $connection) {
             if ($connection['logging']) {
                 $connWithLogging[] = $name;
+            }
+
+            if ($connection['profiling']) {
+                $connWithProfiling[] = $name;
+
+                if ($connection['profiling_collect_backtrace']) {
+                    $connWithBacktrace[] = $name;
+                }
             }
 
             $this->loadDbalConnection($name, $connection, $container);
@@ -172,7 +179,7 @@ class DoctrineExtension extends AbstractDoctrineExtension
             });
         }
 
-        $this->useMiddlewaresIfAvailable($container, $connWithLogging);
+        $this->useMiddlewaresIfAvailable($container, $connWithLogging, $connWithProfiling, $connWithBacktrace);
     }
 
     /**
@@ -186,7 +193,9 @@ class DoctrineExtension extends AbstractDoctrineExtension
     {
         $configuration = $container->setDefinition(sprintf('doctrine.dbal.%s_connection.configuration', $name), new ChildDefinition('doctrine.dbal.connection.configuration'));
         $logger        = null;
-        if ($connection['logging']) {
+
+        /** @psalm-suppress UndefinedClass */
+        if (! interface_exists(MiddlewareInterface::class) && $connection['logging']) {
             $logger = new Reference('doctrine.dbal.logger');
         }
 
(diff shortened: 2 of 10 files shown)


DataCollector/DoctrineDataCollector.php
```diff
@@ -2,311 +2,37 @@
...
-    public function getEntities()
-    {
-        return $this->data['entities'];
-    }
-
-    /** @return array<string, array<string, list<string>>> */
-    public function getMappingErrors()
+// phpcs:disable PSR1.Classes.ClassDeclaration.MultipleClasses
+if (class_exists(DebugDataHolder::class)) {
+    class DoctrineDataCollector extends BaseCollector
```
stof: There is no reason to have conditional class definitions here. Having an optional `?DebugDataHolder` argument in the constructor all the time is fine.

Middleware/BacktraceDebugDataHolder.php
```diff
@@ -0,0 +1,94 @@
...
+        parent::reset();
+
+        $this->backtraces = [];
+    }
+
+    public function addQuery(string $connectionName, Query $query): void
+    {
+        parent::addQuery($connectionName, $query);
+
+        // array_slice to skip middleware calls in the trace
```
stof: this comment should be on the line doing the array_slice instead

Middleware/BacktraceDebugDataHolder.php
```diff
@@ -0,0 +1,94 @@
...
+            $data[] = $this->addBacktracesIfAvailable($connectionName, $record, $idx);
+        }
+
+        return $data;
+    }
+
+    /**
+     * @param mixed[] $record
+     *
+     * @return mixed []
```
stof: extra space here

Middleware/BacktraceDebugDataHolder.php
```diff
@@ -0,0 +1,94 @@
...
+     */
+    private function addBacktracesIfAvailable(string $connectionName, array $record, int $idx): array
+    {
+        if (! isset($this->backtraces[$connectionName])) {
+            return $record;
+        }
+
+        return array_merge(
```
stof: no need to use `array_merge` to assign a single key

Middleware/BacktraceDebugDataHolder.php
```diff
@@ -0,0 +1,94 @@
...
+
+        // array_slice to skip middleware calls in the trace
+        if (! in_array($connectionName, $this->connWithBacktraces, true)) {
+            return;
+        }
+
+        $this->backtraces[$connectionName][] = array_slice(debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS), 2);
+    }
+
+    /** @return mixed[][][] */
```
stof: this is actually `array<string, mixed[][]>`

Middleware/BacktraceDebugDataHolder.php
```diff
@@ -0,0 +1,94 @@
...
+
+use const DEBUG_BACKTRACE_IGNORE_ARGS;
+
+/** @psalm-suppress UndefinedClass */
+class BacktraceDebugDataHolder extends DebugDataHolder
+{
+    /** @var string[] */
+    private $connWithBacktraces;
+
+    /** @var mixed[][][] */
```
stof: this is `array<string, list<mixed[]>`

Resources/config/dbal.xml
```diff
@@ -23,10 +22,6 @@
         <service id="Doctrine\DBAL\Driver\Connection" alias="database_connection" public="false" />
         <service id="Doctrine\DBAL\Connection" alias="database_connection" public="false" />
 
-        <service id="doctrine.dbal.logger.chain" class="%doctrine.dbal.logger.chain.class%" public="false" abstract="true" >
```
stof: Removing this one is a no-go. It is still needed if some projects add their own sql loggers. the removal must only happen in a new major version, with deprecations before

stof (changes requested):

DataCollector/DoctrineDataCollector.php
l-vo: Nice ! I didn't know class was not resolved in the case of a null value 👍

Resources/config/dbal.xml
l-vo: Indeed, thank you

Middleware/BacktraceDebugDataHolder.php
l-vo: Actually, it seems it's `array<string, array<string, mixed>[]>`

Middleware/BacktraceDebugDataHolder.php
l-vo: I think it's `array<string, array<string, mixed>[]>`; replaced

Resources/config/dbal.xml
l-vo: I simply reverted this change, from my POV, I can't deprecate this service while `doctrine/dbal` versions without middlewares and `symfony-bridge` without `DebugMiddleware` are still supported. Am I missing something ?

l-vo:
@stof PR updated according to your feedbacks 🙂

Resources/config/dbal.xml
stof: indeed. We cannot deprecated it yet

DependencyInjection/DoctrineExtension.php
```diff
@@ -186,7 +193,9 @@ protected function loadDbalConnection($name, array $connection, ContainerBuilder
     {
         $configuration = $container->setDefinition(sprintf('doctrine.dbal.%s_connection.configuration', $name), new ChildDefinition('doctrine.dbal.connection.configuration'));
         $logger        = null;
-        if ($connection['logging']) {
+
+        /** @psalm-suppress UndefinedClass */
```
stof, line 197: this looks useless here

DependencyInjection/DoctrineExtension.php
```diff
@@ -186,7 +193,9 @@ protected function loadDbalConnection($name, array $connection, ContainerBuilder
...
+        if (! $this->isSfDebugMiddlewareAvailable() && $connection['logging']) {
```
stof: this condition looks wrong to me. the SQL logging might use the Middleware system as soon as the DBAL logging middleware is available, even if the profiler integration still uses the old logger interface. DBAL supports using both APIs at the same time.

Middleware/DebugMiddleware.php
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Doctrine\Bundle\DoctrineBundle\Middleware;
+
+use Doctrine\DBAL\Driver as DriverInterface;
+use Doctrine\DBAL\Driver\Middleware;
+use Symfony\Bridge\Doctrine\Middleware\Debug\DebugDataHolder;
+use Symfony\Bridge\Doctrine\Middleware\Debug\Driver;
+use Symfony\Component\Stopwatch\Stopwatch;
+
+class DebugMiddleware implements Middleware, ConnectionNameAwareInterface
```
stof, line 11: why aren't you using the DebugMiddleware defined in Symfony ?

DependencyInjection/DoctrineExtension.php
l-vo: It's what is implemented, isn't it ? See https://github.com/doctrine/DoctrineBundle/blob/9cadf2c331b01a017482d8283b02b7edd6b82244/Tests/DependencyInjection/DoctrineExtensionTest.php#L1282, `isSfDebugMiddlewareAvailable` return true if Middleware interface is available but also the middleware for profiling provided by the Doctrine bridge.

Middleware/DebugMiddleware.php
l-vo: Because I need the  `ConnectionNameAwareInterface` in the bundle. Currently it's how the bundle is designed, one middleware (or previously a DebugStack) by connection. It was not a requirement in the bridge where the middleware sat, so by default the middleware doesn't implement `ConnectionNameAwareInterface`.

l-vo:
@stof Feedbacks answered :)

ostrolucky (approved):

DependencyInjection/DoctrineExtension.php
stof: And that's my whole point. `logging`  does not require the profiling middleware.

dmaicher:
I think this should go to 2.7.x, or? Not sure this qualifies as a patch?

ostrolucky:
indeed

l-vo:
I targeted 2.6 because on the Symfony side, the choice was to pass middlewares adds in a patch version to fix the deprecation about DebugStack. But maybe here the case is a bit different... It's up to you to choose 🙂

DependencyInjection/DoctrineExtension.php
ostrolucky: @l-vo can you address this concern?

l-vo: @stof ok, now I see what you mean :). Actually, here, if logging is enabled, a chain logger is used with `doctrine.dbal.logger` and `doctrine.dbal.logger.profiling`. If profiling middleware is available, logging middleware is also available and this chain has no reason to be created. Moreover, I must keep the condition on this line because below we have:
```
if ($logger) {
    $configuration->addMethodCall('setSQLLogger', [$logger]);
}
```
Is it ok for you ?

l-vo:
@dmaicher @ostrolucky So given my explanation https://github.com/doctrine/DoctrineBundle/pull/1517#issuecomment-1121016234 should I target 2.7 instead ?

DependencyInjection/DoctrineExtension.php
stof: Well, on this line, we should have a `if (!class_exists(Middleware::class)` condition only, so that the `doctrine.dbal.logger` is not added if logging is already handled by the middleware (and then, the `doctrine.dbal.logger.profiling` logger might be injected directly instead of wrapping it in a chain logger if it is alone, as is already the case if you enable profiling without logging)

l-vo: Indeed, my thinking was wrong. Fixed 👍

dmaicher:
@l-vo I just tested this on one of my apps and it looks like for me its not working as intended.

with DoctrineBundle 2.6.3:
Screenshot from 2022-06-07 20-26-45

With your changes on 2.7.x and the same page:
Screenshot from 2022-06-07 20-25-07

Did you test this on a real app?

l-vo:
@dmaicher Yes, I tested on a real app, which version of the doctrine bridge and dbal do you use ?

dmaicher:
> @dmaicher Yes, I tested on a real app, which version of the doctrine bridge and dbal do you use ?

I'm using those

```
doctrine/dbal                            3.3.6                    Powerful PHP database abstraction layer (DBAL) with many features for database schema introspection and management.
symfony/doctrine-bridge                  v5.4.9                   Provides integration for Doctrine with various Symfony components

**Comment by l-vo** [comment]
It works on b7002752b555cb61cc9ca8d4c89a5b6e8200f765 (after #1488) for me, could you try this commit on your project ? Maybe a bad conflict resolution has broken the system ?

**Comment by l-vo** [comment]
> It works on [b700275](https://github.com/doctrine/DoctrineBundle/commit/b7002752b555cb61cc9ca8d4c89a5b6e8200f765) (after #1488) for me, could you try this commit on your project ? Maybe a bad conflict resolution has broken the system ?

My bad, I didn't look the good commit; indeed; there is a problem, I'll take a look.

**Comment by l-vo** [comment]
@dmaicher I don't understand how it occurred, but it seems the dbal.xml file hasn't been pushed. Fix in #1535.

**Comment by mat-twg** [comment]
```
doctrine/doctrine-bundle: 2.7.0
```

```
/srv/app # php bin/console doctrine:migrations:migrate 0

 WARNING! You are about to execute a migration in database "db" that could result in schema changes and data loss. Are you sure you wish to continue? (yes/no) [yes]:
 > yes

[notice] Migrating down to 0
[notice] finished in 4224.6ms, used 40.5M memory, 1 migrations executed, 3 sql queries

2022-06-21T09:43:20+00:00 [info] User Deprecated: The "Symfony\Bridge\Doctrine\Logger\DbalLogger" class implements "Doctrine\DBAL\Logging\SQLLogger" that is deprecated Use {@see \Doctrine\DBAL\Logging\Middleware} or implement {@see \Doctrine\DBAL\Driver\Middleware} instead.

```

ostrolucky:
?

PR #1522, opened 2022-05-09
Update .symfony.bundle.yaml

dmaicher:
Wondering how we can ensure that this files stay up-to-date :thinking: We will forget it and it will be outdated again soon I guess.

### DIFF
--- a/.symfony.bundle.yaml
+++ b/.symfony.bundle.yaml
@@ -1,9 +1,9 @@
 branches:
-    - "1.12.x"
-    - "2.4.x"
     - "2.5.x"
+    - "2.6.x"
+    - "2.7.x"
 maintained_branches:
-    - "2.4.x"
-    - "2.5.x"
+    - "2.6.x"
+    - "2.7.x"
 doc_dir: "Resources/doc/"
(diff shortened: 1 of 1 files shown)


.symfony.bundle.yaml
```diff
@@ -1,9 +1,9 @@
...
+    - "2.6.x"
+    - "2.7.x"
 maintained_branches:
-    - "2.4.x"
-    - "2.5.x"
+    - "2.6.x"
+    - "2.7.x"
 doc_dir: "Resources/doc/"
-dev_branch: "2.5.x"
+dev_branch: "2.7.x"
```
dmaicher, line 9: not sure which one to take here. Technically 2.6.x is also a dev branch :thinking:

ostrolucky: can this reference non-existing branch?

dmaicher: not sure I follow. 2.7.x [exists](https://github.com/doctrine/DoctrineBundle/tree/2.7.x)?

ostrolucky (approved):

PR #1523, opened 2022-05-09
Added naming_strategy option to config reference defaults

stollr:
The `namong_strategy` option of the configuration reference is currenty missing from the ORM defaults section. So I have added it together with its default value.

### DIFF
--- a/Resources/doc/configuration.rst
+++ b/Resources/doc/configuration.rst
@@ -758,6 +758,7 @@ the ORM resolves to:
             metadata_cache_driver: ~
             query_cache_driver: ~
             result_cache_driver: ~
+            naming_strategy: doctrine.orm.naming_strategy.default
 
 There are lots of other configuration options that you can use to overwrite
 certain classes, but those are for very advanced use-cases only.


Resources/doc/configuration.rst
```diff
@@ -758,6 +758,7 @@ the ORM resolves to:
             metadata_cache_driver: ~
             query_cache_driver: ~
             result_cache_driver: ~
+            naming_strategy: doctrine.orm.naming_strategy.default
 
 There are lots of other configuration options that you can use to overwrite
```
ostrolucky, line 763: This explains why is this missing from there. I don't think anything needs to be done here. naming_strategy option itself is explained in this file already on different location

ostrolucky (changes requested):

Resources/doc/configuration.rst
stollr: First, the sentence above that code paragraph says "This following configuration example shows **all** the configuration defaults". As far as I understand this means that all options with any defaults should be listed.

Secondly, I cannot find any explanation about this Option in this file (https://www.doctrine-project.org/projects/doctrine-bundle/en/2.2/configuration.html). Maybe I am overseeing anything. Please correct me if I am wrong.

ostrolucky: > Secondly, I cannot find any explanation about this Option in this file ([doctrine-project.org/projects/doctrine-bundle/en/2.2/configuration.html](https://www.doctrine-project.org/projects/doctrine-bundle/en/2.2/configuration.html)). Maybe I am overseeing anything. Please correct me if I am wrong.

Alright. explained -> mentioned

> First, the sentence above that code paragraph says "This following configuration example shows all the configuration defaults". As far as I understand this means that all options with any defaults should be listed.

Then that text is wrong. There are lot of other configuration options not specified there, but they are in huge code block before. Surely purpose of second code block wasn't to specify all of them again.

stollr: Then I do not understand what the second code block is even for :-/

Anyway, maybe a bit off-topic. I realized that there is no place where the options of the bundle configuration is explained in detail. Searching in the web for the term "doctrine naming strategy" the only page in the doctrine documentation related to that is [Implementing a NamingStrategy](https://www.doctrine-project.org/projects/doctrine-orm/en/2.10/reference/namingstrategy.html)

Would it make sense to add it anywhere in the reference? In fact this was the reason I came here - I didn't find any details on the naming strategy in the doc and had to look it up in the code.

ostrolucky: I think it's not mentioned because defaults are fine for most people. And also because this is a doctrine bundle, not doctrine. Bundle is just a bridge for ORM. Options should be primarily explained in ORM docs, we shouldn't have to repeat explanations here. So I would suggest to improve docs of ORM primarily.

stollr: Okay, this makes sense. Thanks for your patience ;-)

PR #1528, opened 2022-05-24
Call clear() when trying to reset ghost object proxies

nicolas-grekas:
Similar to https://github.com/symfony/symfony/pull/46443
Builds on https://github.com/doctrine/DoctrineBundle/pull/1118


stof:
Last time I checked, clearing a closed entity manager was not reopening it, which would break the feature.

nicolas-grekas:
At least using ghost objects explode right now so this makes them follow the existing "not-lazy" code path.
I agree it's not the end of the story, just a small guard step maybe.

nicolas-grekas:
Closing in favor of https://github.com/symfony/symfony/pull/46443, aka let Symfony throw if this happens (it shouldn't.)

PR #1529, opened 2022-05-24
Ensure a lazy *value holder* proxy is always generated with proxy-manager-bridge

nicolas-grekas:
While working on https://github.com/symfony/symfony/issues/35345, I figured out that we need to ensure that the entity manager will continue to use a value-holder proxy (and not a ghost-object proxy, at least not until https://github.com/doctrine/orm/issues/5933 is solved.)

This is the way to do so.

See also https://github.com/symfony/symfony/issues/35345#issuecomment-1136017356

### DIFF
--- a/Resources/config/orm.xml
+++ b/Resources/config/orm.xml
@@ -100,6 +100,7 @@
 
         <service id="doctrine.orm.entity_manager.abstract" class="%doctrine.orm.entity_manager.class%" abstract="true" lazy="true">
             <factory class="%doctrine.orm.entity_manager.class%" method="create" />
+            <tag name="proxy" interface="%doctrine.orm.entity_manager.class%" />
         </service>
 
         <service id="doctrine.orm.container_repository_factory" class="Doctrine\Bundle\DoctrineBundle\Repository\ContainerRepositoryFactory" public="false">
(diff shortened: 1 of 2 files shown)


stof (approved):

ostrolucky:
I don't quite follow. Is default proxy strategy going to be changed in bridge? Otherwise I don't see why this insurance needs to be put in place. Ghost-object proxy is not used at the moment. And if it is going to be changed in bridge, shouldn't the tag be renamed there first?

nicolas-grekas:
> Is default proxy strategy going to be changed in bridge

I'm working on it yes

ostrolucky:
Isn't using `<tag name="proxy"` _only_ for Holder Value proxies non-intuitive? I mean, nobody would expect ghost proxies are using some different tag without digging through docs. Or what's the plan there. If both are going to be supported by symfony, I would either assume `proxy` tag would get deprecated (and 2 new tags created for each strategy), or `proxy` tag would get new tag attribute where you can specify which proxy type is desired.

nicolas-grekas:
The "proxy" tag allows decorating an instance of Foo and exposing only it as FooInterface - aka the template for the proxy generation is the interface. Because the class of the proxy cannot be an `instanceof` the object it decorates (as desired), only value-holder is possible when "proxy" is used.

Which means: no, the tag won't be deprecated, it has a clear purpose. Other kind of lazy proxies can use ghost-objects, unless they are created by a factory and the type declared by this factory is abstract.

Aka I don't think we need new tags.

ostrolucky (approved):

nicolas-grekas:
Looks like I managed to make ghost object proxified entity managers resettable, so this is not needed anymore (I'll reopen if I realize I was wrong of course.)
See https://github.com/symfony/symfony/pull/46458

PR #1532, opened 2022-05-30
feat add support for ORM SchemaIgnoreClasses

### LINKED ISSUE #1454: Add support for `SchemaIgnoreClasses`
I was thinking if we could add `SchemaIgnoreClasses` config (https://github.com/doctrine/orm/pull/9202)

_I checked if I could do it myself but I'm not very familiar with symfony bundles and it seemed too difficult when considered the time I'd have to invest in it_

simPod:
Resolves #1454


simPod:
Does anyone has any idea how to fix xml support?

DependencyInjection/Configuration.php
```diff
@@ -466,6 +467,9 @@ private function addOrmSection(ArrayNodeDefinition $node): void
                         ->end()
                         ->scalarNode('proxy_dir')->defaultValue('%kernel.cache_dir%/doctrine/orm/Proxies')->end()
                         ->scalarNode('proxy_namespace')->defaultValue('Proxies')->end()
+                        ->arrayNode('schema_ignore_classes')
```
stof: this needs a `->fixXmlConfig('schema_ignore_class', 'schema_ignore_classes')` on the parent node (outside the `->children()` block)

DependencyInjection/DoctrineExtension.php
```diff
@@ -643,6 +643,7 @@ protected function loadOrmEntityManager(array $entityManager, ContainerBuilder $
             'setProxyDir' => '%doctrine.orm.proxy_dir%',
             'setProxyNamespace' => '%doctrine.orm.proxy_namespace%',
             'setAutoGenerateProxyClasses' => '%doctrine.orm.auto_generate_proxy_classes%',
+            'setSchemaIgnoreClasses' => '%doctrine.orm.schema_ignore_classes%',
```
stof: Should we really have a single list of ignored classes common for all entity managers ? Shouldn't we have separate list per managers instead ? This is not a configuration mapping to a global state

stof (changes requested):

DependencyInjection/DoctrineExtension.php
simPod: I'm not sure it's possible since the `doctrine.orm.configuration` is global and not per EM 🤔

stof: really ? This looks weird to me...

simPod: Is it this? https://github.com/doctrine/DoctrineBundle/blob/4f4387cbbd71a3a5037e44cefa466734499e1e57/DependencyInjection/DoctrineExtension.php#L608

ostrolucky:
There are still some test failures

ostrolucky (approved):

dmaicher (approved):

dmaicher:
I tested this on one of my apps and it works as expected :+1: 

Thanks @simPod

PR #1535, opened 2022-06-08
Missing file in replacement of dballogger by debugmiddleware

l-vo:
Missing file in #1517

### DIFF
--- a/Resources/config/dbal.xml
+++ b/Resources/config/dbal.xml
@@ -40,6 +40,7 @@
             <tag name="data_collector" template="@Doctrine/Collector/db.html.twig" id="db" priority="250" />
             <argument type="service" id="doctrine" />
             <argument>true</argument>
+            <argument type="service" id="doctrine.debug_data_holder" on-invalid="null"/>
         </service>
 
         <service id="doctrine.dbal.connection_factory" class="%doctrine.dbal.connection_factory.class%">


ostrolucky:
@nicolas-grekas can you pls check the failure, looks like some issue with DI phpdumper

l-vo:
@ostrolucky why do you think it's a problem with the phpdumper ?

dmaicher:
> @ostrolucky why do you think it's a problem with the phpdumper ?

I guess @ostrolucky is referring to the [build fail ](https://github.com/doctrine/DoctrineBundle/runs/6801941885?check_suite_focus=true) which indeed looks like some compiled container syntax issues

dmaicher (approved):
I can confirm this fixes [the issue](https://github.com/doctrine/DoctrineBundle/pull/1517#issuecomment-1149023230) and the profiler is working again :+1:

nicolas-grekas:
See https://github.com/symfony/symfony/pull/46628 for the fix for the dumped container.

l-vo:
> > @ostrolucky why do you think it's a problem with the phpdumper ?
> 
> I guess @ostrolucky is referring to the [build fail ](https://github.com/doctrine/DoctrineBundle/runs/6801941885?check_suite_focus=true) which indeed looks like some compiled container syntax issues

Oh of course! I thought it was about my modification. Thank you for the clarification 🙂

PR #1538, opened 2022-06-20
add conflict with doctrine/orm < 2.11

### LINKED ISSUE #1536: DoctrineBundle 2.7.0 calls undefined method setSchemaIgnoreClasses
In orm 2.11 the method setSchemaIgnoreClasses was introduced. DoctrineBundle's composer.json has a conflict doctrine/orm <2.10 but that should be <2.11, or the version should be checked before setSchemaIgnoreClasses is called.

```
16:05:38 CRITICAL  [php] Uncaught Error: Call to undefined method Doctrine\ORM\Configuration::setSchemaIgnoreClasses() ["exception" => Error { …}]

In App_KernelDevDebugContainer.php line 1283:

  Attempted to call an undefined method named "setSchemaIgnoreClasses" of class "Doctrine\ORM\Configuration".
```

dmaicher:
Fixes https://github.com/doctrine/DoctrineBundle/issues/1536

Also thought about conditionally checking for existence of `Doctrine\ORM\Configuration::setSchemaIgnoreClasses()` but bumping the orm requirement seems better here.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -67,7 +67,7 @@
         }
     },
     "conflict": {
-        "doctrine/orm": "<2.10|>=3.0",
+        "doctrine/orm": "<2.11|>=3.0",
         "twig/twig": "<1.34|>=2.0,<2.4"
     },
     "suggest": {


stof:
Given that require-dev already requires 2.11+, this is indeed the right fix.

stof (approved):

PR #1539, opened 2022-06-20
drop support for unmaintained Symfony versions

dmaicher:
Some of those versions are long dead. So I think we can clean up a bit?


dmaicher:
Whats our policy for bumping requirements? fine for a patch release? Or should be a minor?

ostrolucky:
Well I generally avoid bumping if I can't have benefit from it, eg. in form of getting rid of some BC code. I like when requirements truthfully say which version are _actually_ required and are not just periodically bumped for political reasons. But since multiple people asked about this already, I don't think it's such a big deal. I think it was even pointed out before that test suite is not even testing these, because even --prefer-lowest doesn't install lowest constraints,since they are blocked by some other dependencies.

As for patch/minor, I don't think we have a policy for this. As for me, I don't care in this instance.

ostrolucky (approved):

ostrolucky:
Feel free to rebase, we can merge afterwards

PR #1540, opened 2022-06-21
Add "Composer Lint" job to CI pipeline

phansys:
Applies `composer-lint` workflow from [doctrine/.github](https://github.com/doctrine/.github/blob/87f8373f3668edfe31b6aa8d3049b3fd35f90362/.github/workflows/composer-lint.yml#L2).

### DIFF
--- a/.github/workflows/composer-lint.yml
+++ b/.github/workflows/composer-lint.yml
@@ -1,0 +1,18 @@
+name: "Composer Lint"
+
+on:
+  pull_request:
+    branches:
+      - "*.x"
+    paths:
+      - "composer.json"
+  push:
+    branches:
+      - "*.x"
+    paths:
+      - "composer.json"
+
+jobs:
+  composer-lint:
+    name: "Composer Lint"
+    uses: "doctrine/.github/.github/workflows/composer-lint.yml@1.5.0"
--- a/composer.json
+++ b/composer.json
@@ -1,11 +1,14 @@
 {
     "name": "doctrine/doctrine-bundle",
-    "type": "symfony-bundle",
     "description": "Symfony DoctrineBundle",
-    "keywords": ["DBAL", "ORM", "Database", "Persistence"],
-    "homepage": "https://www.doctrine-project.org",
     "license": "MIT",
-    "minimum-stability": "dev",
+    "type": "symfony-bundle",
+    "keywords": [
+        "DBAL",
+        "ORM",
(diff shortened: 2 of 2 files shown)


dmaicher (approved):

ostrolucky (changes requested):
Can you rather embed this in https://github.com/doctrine/DoctrineBundle/blob/2.7.x/.github/workflows/coding-standards.yml?

ostrolucky (approved):

ostrolucky:
Hmm maybe we should have this in https://github.com/doctrine/.github/blob/main/.github/workflows/coding-standards.yml so it applies for all projects, cc @greg0ire

ostrolucky (approved):

greg0ire:
I think it could indeed be contributed to that repo. If we add it to the CS workflow. If we do that, the next release should probably be 2.0.0 to make it clear that an action must be taken to make the build green (fixing `composer.json`). It would be great to make this workflow kick in only when appropriate, using [`paths`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths), otherwise it will mean one more job per project, and a job that is unlikely to fail when `composer.json` is not touched.

.github/workflows/coding-standards.yml
```diff
@@ -15,3 +15,7 @@ jobs:
     uses: "doctrine/.github/.github/workflows/coding-standards.yml@1.4.1"
     with:
       composer-options: "--prefer-dist --prefer-stable"
+
+  lint-composer:
+    name: "Lint Composer"
+    uses: "phansys/doctrine.github/.github/workflows/composer-lint.yml@composer_lint"
```
greg0ire: This needs to be in a separate workflow so that you can use `paths`

stof: This should not reference your fork

stof (dismissed):

.github/workflows/coding-standards.yml
greg0ire: It's temporary, for test purposes, in order to proove that the upstream PR works before merging it.

greg0ire: Let's convert this to a draft in the meantime.

greg0ire (approved):
:ok_hand:

ostrolucky (approved):

PR #1545, opened 2022-07-12
Allow resetting lazy managers on Symfony 6.2

nicolas-grekas:
Sidekick of https://github.com/symfony/symfony/pull/46752

### DIFF
--- a/Registry.php
+++ b/Registry.php
@@ -8,6 +8,7 @@ use Doctrine\ORM\Proxy\Proxy;
 use ProxyManager\Proxy\LazyLoadingInterface;
 use Psr\Container\ContainerInterface;
 use Symfony\Bridge\Doctrine\ManagerRegistry;
+use Symfony\Component\VarExporter\LazyGhostObjectInterface;
 use Symfony\Contracts\Service\ResetInterface;
 
 use function array_keys;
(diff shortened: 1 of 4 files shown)


ostrolucky:
I've fixed flex issue in 2.7 branch, can you rebase please (some CI matrices will still fail due to unrelated issue with DBAL, but at least some should succeed).

Registry.php
```diff
@@ -75,7 +76,7 @@ private function resetOrClearManager(string $managerName, string $serviceId): vo
 
         assert($manager instanceof EntityManagerInterface);
 
-        if (! $manager instanceof LazyLoadingInterface || $manager->isOpen()) {
+        if ((! $manager instanceof LazyLoadingInterface && ! $manager instanceof LazyGhostObjectInterface) || $manager->isOpen()) {
```
ostrolucky, line 79: feel free to suppress this by adding the LazyGhostObjectInterface here https://github.com/doctrine/DoctrineBundle/blob/2d8779dbf8035bc13ae88b7c8537cfd662a4d689/psalm.xml.dist#L43-L43

nicolas-grekas: updated, thanks for the hint

nicolas-grekas:
Looks like I need to have a closer look at failures.

nicolas-grekas:
Green after https://github.com/symfony/symfony/pull/46929
Help wanted for the psalm failure.
I'm on vacation starting today, feel free to push on my fork if needed.
You might need to borrow (part of) the patch to fix the builds on lower maintained branches.

PR #1551, opened 2022-07-28
Don't use a factory to create the EM when possible

nicolas-grekas:
Follows https://github.com/doctrine/orm/pull/9915

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -20,6 +20,7 @@ use Doctrine\DBAL\Sharding\PoolingShardConnection;
 use Doctrine\DBAL\Sharding\PoolingShardManager;
 use Doctrine\DBAL\Tools\Console\Command\ImportCommand;
 use Doctrine\DBAL\Tools\Console\ConnectionProvider;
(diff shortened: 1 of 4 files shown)


DependencyInjection/DoctrineExtension.php
```diff
@@ -493,6 +495,11 @@ protected function ormLoad(array $config, ContainerBuilder $container)
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('orm.xml');
 
+        if ((new ReflectionMethod(EntityManager::class, '__construct'))->isPublic()) {
```
greg0ire: Should there be a negation here? This seems to do the opposite of what your commit message says.

```suggestion
        if (! (new ReflectionMethod(EntityManager::class, '__construct'))->isPublic()) {
```

nicolas-grekas: Indeed, fixed!

nicolas-grekas:
Remaining deprecations will be fixed once https://github.com/symfony/symfony/pull/47097 is propagated to the subtree split.

greg0ire (approved):

PR #1554, opened 2022-08-18
Wire EntityArgumentResolver when available

nicolas-grekas:
Wires https://github.com/symfony/symfony/pull/43854 by adding a new configuration node.

The default is:
```yaml
doctrine:
  orm:
    controller_resolver:
      enabled: true
      auto_mapping: true
      evict_cache: false
```


jderusse (approved):

Resources/config/orm.xml
```diff
@@ -172,6 +172,14 @@
             <argument type="service" id="doctrine" />
         </service>
 
+        <service id="doctrine.orm.entity_value_resolver" class="Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver">
+            <argument type="service" id="doctrine" />
+            <argument type="service" id="doctrine.orm.entity_value_resolver.expression_language" on-invalid="ignore" />
+            <tag name="controller.argument_value_resolver" priority="-40" />
```
HypeMC: Shouldn't this have a higher priority than the `RequestAttributeValueResolver`, I've tested it with the following code & got an error:

```php
    #[Route('/foo/{foo<\d+>}', name: 'app_foo')]
    public function index(Foo $foo): Response
    {
        dd($foo);
    }
```

```
App\Controller\FooController::index(): Argument #1 ($foo) must be of type App\Entity\Foo, string given, called in /vendor/symfony/http-kernel/HttpKernel.php on line 161
```

Since both the route parameter & the method argument have the same name, the `RequestAttributeValueResolver` resolves it to a string which is then passed to the controller.

nicolas-grekas: Good catch. I'd rather fix `RequestAttributeValueResolver` to make it skip arguments that are incompatible with the value of the attribute!

DependencyInjection/Configuration.php
```diff
@@ -466,6 +466,13 @@ private function addOrmSection(ArrayNodeDefinition $node): void
                         ->end()
                         ->scalarNode('proxy_dir')->defaultValue('%kernel.cache_dir%/doctrine/orm/Proxies')->end()
                         ->scalarNode('proxy_namespace')->defaultValue('Proxies')->end()
+                        ->arrayNode('controller_resolver')
+                            ->canBeDisabled()
+                            ->children()
+                                ->booleanNode('auto_mapping')->defaultTrue()->end()
+                                ->booleanNode('evict_cache')->defaultFalse()->end()
```
lyrixx: Can you add some info here ?

nicolas-grekas: Done

Resources/config/orm.xml
nicolas-grekas: Actually that won't work because it's legal for a resolver to give incompatible values when there is a controller argument listener that fixes them.
I've registered the entity resolver with priority 110 (the request one is at 100.)

DependencyInjection/Configuration.php
```diff
@@ -466,6 +466,19 @@ private function addOrmSection(ArrayNodeDefinition $node): void
                         ->end()
                         ->scalarNode('proxy_dir')->defaultValue('%kernel.cache_dir%/doctrine/orm/Proxies')->end()
                         ->scalarNode('proxy_namespace')->defaultValue('Proxies')->end()
+                        ->arrayNode('controller_resolver')
+                            ->canBeDisabled()
+                            ->children()
+                                ->booleanNode('auto_mapping')
```
stof, line 472: I'm not sure about this naming (or maybe just the description). What this boolean will do is disabling the automatic mapping of _all_ request attributes as a criteria. But the first lookup by primary key (attempted before the criteria lookup) is not disabled by that flag. So setting this one to `false` will still not require to add an explicit `MapEntity` on each argument explicitly (that behavior is triggered when setting `enabled: false`, not `auto_mapping: false`)

nicolas-grekas: Right! How would you formulate that in a single sentence?

nicolas-grekas: I updated to `Set to false to disable using route placeholders as lookup criteria when the primary key doesn't match the argument name`, WDYT?

ostrolucky (approved):

nicolas-grekas:
PR rebased, green, thus ready.

PR #1558, opened 2022-09-23
(WIP) Update the profiler panel for Symfony 6.2+

javiereguiluz:
This is still WIP. When it's ready for review I'll post some screenshots.

### DIFF
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -8,7 +8,11 @@
         {% set icon %}
             {% set status = collector.invalidEntityCount > 0 ? 'red' : collector.querycount > 50 ? 'yellow' %}
 
-            {{ include('@Doctrine/Collector/icon.svg') }}
+            {% if profiler_markup_version < 3 %}
+                <span class="icon">{{ include('@Doctrine/Collector/icon.svg') }}</span>
+            {% else %}
+                <span class="icon">{{ include('@Doctrine/Collector/database.svg') }}</span>
+            {% endif %}
 
             {% if collector.querycount == 0 and collector.invalidEntityCount > 0 %}
                 <span class="sf-toolbar-value">{{ collector.invalidEntityCount }}</span>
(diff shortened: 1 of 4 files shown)


Resources/views/Collector/db.html.twig
```diff
@@ -98,289 +107,332 @@
...
-        <div class="metric">
-            <span class="value">{{ collector.querycount }}</span>
-            <span class="label">Database Queries</span>
-        </div>
-
-        <div class="metric">
-            <span class="value">{{ collector.groupedQueryCount }}</span>
-            <span class="label">Different statements</span>
-        </div>
+        <div class="metric-group">
```
stof, line 117: Will this markup work in Symfony 6.1 too ?

Resources/views/Collector/db.html.twig
```diff
@@ -98,289 +107,332 @@
...
+                <span class="label">Queries</span>
+            </div>
 
-        <div class="metric">
-            <span class="value">{{ '%0.2f'|format(collector.time * 1000) }} ms</span>
-            <span class="label">Query time</span>
-        </div>
+            <div class="metric">
+                <span class="value">{{ collector.groupedQueryCount }}</span>
+                <span class="label">Statements</span>
```
stof, line 125: The previous label was `Different statements`, and the key thing to distinguish it from the previous number is the word `Different` that you removed.

Resources/views/Collector/db.html.twig
```diff
@@ -98,289 +107,332 @@
...
-                    <span class="label">Misses</span>
-                </div>
+        </div>
 
-                <div class="metric">
-                    <span class="value">{{ collector.cacheCounts.puts }}</span>
-                    <span class="label">Puts</span>
-                </div>
+        <div class="tab">
+            <h3 class="tab-title">Connections & EM</h3>
```
stof, line 411: should use `&amp;`

Resources/views/Collector/db.html.twig
```diff
@@ -98,289 +107,332 @@
...
+                    <div class="empty">
+                        <p>Second Level Cache is not enabled.</p>
+                    </div>
+                {% else %}
+                    {% if not collector.cacheCounts %}
+                        <div class="empty">
+                            <p>Second level cache information is not available.</p>
+                        </div>
+                    {% else %}
+                        <div class="metric-gro">
```
HeahDude, line 374:
```suggestion
                        <div class="metric-group">
```
?

ostrolucky:
Sorry @javiereguiluz but this is open for too long with no traction. Let's reopen when you are ready to continue working on it, alright? :)

nickygerritsen:
From the default profiler panes Doctrine is currently the only one that is not in the new style, which is most obvious looking at the sidebar:
image

What needs to happen to actually make this update? Maybe we can resurrect/recreate the PR? Or should we at least create an issue to track that this is something that needs to be done?

PR #1560, opened 2022-10-11
Enable attaching an entity listener without specifying an event

HypeMC:
Requires https://github.com/doctrine/orm/pull/10122

Currently, when registering an entity listener the event name needs to be specified, eg:

```php
use Doctrine\Bundle\DoctrineBundle\Attribute\AsEntityListener;
use Doctrine\ORM\Events;

#[AsEntityListener(event: Events::prePersist, entity: MyEntity::class)]
final class MyEntityListener
{}
```

If the event is not specified, it's not registered. This is because of the following check:
https://github.com/doctrine/DoctrineBundle/blob/23c23c79a0a0ddeb695854ee57a79198a8bca56a/DependencyInjection/Compiler/EntityListenerPass.php#L56-L58

However, when the `EntityListeners` attribute is used the events get read from the listener class by matching the method names to the names of possible events, eg:

```php

#[ORM\Entity]
#[ORM\EntityListeners([MyEntityListener::class])]
class MyEntity
{}
```

The following code is triggers this behavior:

https://github.com/doctrine/orm/blob/06c77cebb5aa7ee3cb57f8cc0d2224538de154d7/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php#L527-L529

The problem with this approach is that it's not very intuitive since the configuration is split across multiple files making it hard to use.

With this PR, the events will attempt to be auto-mapped in the `AttachEntityListenersListener` if `null` is passed as the event name (see https://github.com/doctrine/orm/pull/10122), eg:

```php
use Doctrine\Bundle\DoctrineBundle\Attribute\AsEntityListener;

#[AsEntityListener(entity: MyEntity::class)]
final class MyEntityListener
{}
```

### DIFF
--- a/DependencyInjection/Compiler/EntityListenerPass.php
+++ b/DependencyInjection/Compiler/EntityListenerPass.php
@@ -53,7 +53,7 @@ class EntityListenerPass implements CompilerPassInterface
                 $resolver = $container->findDefinition($resolverId);
                 $resolver->setPublic(true);
 
-                if (isset($attributes['entity']) && isset($attributes['event'])) {
+                if (isset($attributes['entity'])) {
                     $this->attachToListener($container, $name, $this->getConcreteDefinitionClass($container->findDefinition($id), $container, $id), $attributes);
                 }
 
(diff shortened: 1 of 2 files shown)


ostrolucky (approved):

ostrolucky:
thx!

PR #1566, opened 2022-10-27
fix some DBAL 3.5 deprecations in commands

dmaicher:
Fixes https://github.com/doctrine/DoctrineBundle/issues/1565

PR #1567, opened 2022-10-27
fix remaining getSchemaManager deprecation

dmaicher:
follow-up: https://github.com/doctrine/DoctrineBundle/pull/1566#discussion_r1006816262

### DIFF
--- a/Command/CreateDatabaseDoctrineCommand.php
+++ b/Command/CreateDatabaseDoctrineCommand.php
@@ -106,7 +106,10 @@ EOT
             );
         }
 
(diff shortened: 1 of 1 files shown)


dmaicher:
@stof @ostrolucky there is also one more usage in `ImportMappingDoctrineCommand`. But the command itself is deprecated and also using quite a few other deprecated things. So probably not worth fixing?

stof:
As the command is deprecated, I think it is indeed fine to leave it as is.

stof (approved):

PR #1568, opened 2022-10-30
Add option "doctrine.orm.enable_lazy_ghost_objects"

nicolas-grekas:
Sidekick of https://github.com/doctrine/orm/pull/10187

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -15,6 +15,7 @@ use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepositoryInterface;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection;
 use Doctrine\DBAL\Driver\Middleware as MiddlewareInterface;
+use Doctrine\ORM\Configuration as OrmConfiguration;
 use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Id\AbstractIdGenerator;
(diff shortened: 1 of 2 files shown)


ostrolucky:
LGTM, but upstream PR should be merged first

stof:
The upstream PR is merged

stof:
We might want to wait for a release though

ostrolucky:
what about documenting this, at least by ading option description?

DependencyInjection/DoctrineExtension.php
```diff
@@ -489,8 +490,15 @@ protected function ormLoad(array $config, ContainerBuilder $container)
 
         $container->setParameter('doctrine.default_entity_manager', $config['default_entity_manager']);
 
-        $options = ['auto_generate_proxy_classes', 'proxy_dir', 'proxy_namespace'];
+        $options = ['auto_generate_proxy_classes', 'enable_lazy_ghost_objects', 'proxy_dir', 'proxy_namespace'];
         foreach ($options as $key) {
+            if ($key === 'enable_lazy_ghost_objects' && $config[$key] && ! method_exists(OrmConfiguration::class, 'setLazyGhostObjectEnabled')) {
```
derrabus: We should also check if VarExporter and Persistence have been installed in the correct versions.

nicolas-grekas: this is not needed: the checks are already done in setLazyGhostObjectEnabled

nicolas-grekas:
> what about documenting this, at least by ading option description?

I was about to add this but I think it's quite difficult to convey something useful for now.
I'd suggest waiting for doctrine/common to deprecate its proxies before.
And then to deprecate *not* setting the new setting to true later on.

DependencyInjection/DoctrineExtension.php
derrabus: Yes, but we would turn a runtime error into a compile-time one.

nicolas-grekas: Good call, PR updated.

derrabus (approved):

derrabus:
Upstream PR has been released with ORM 2.14.0.

Issue #1550, opened 2022-07-22
Drop doctrine/dbal 2.* support


amberovsky:
Hi all,

Is there a reason to support unmaintained doctrine/dbal 2.* ?

Among many other things (in the 3.* version) MySql was renamed to MySQL [see there](https://github.com/doctrine/dbal/commit/885bf615a5820c56ddee60a8fbd3d7ce973587ed), it is still MySql in the current codebase , [e.g. here](https://github.com/doctrine/DoctrineBundle/blob/74d9ae81989b263cf7906de6557cad2e870a2eae/ConnectionFactory.php#L13) - that fails some code analyzers

stof:
According to https://www.doctrine-project.org/2022/01/22/sunsetting-dbal-2.html, the EOL date for DBAL 2.x is 6 months after the release date of ORM 2.10 (supporting DBAL 3), which was on April 3 2022. this means that DBAL 2 is not EOL yet. It will reach EOL only in October.

amberovsky:
Right, so they plan to address security issues after that for at most a year, does that mean you will keep supporting it as well?

Is it worth creating a new branch without 2.* support? As at the moment it is not possible to even backport MySQL renaming to the 2.* branch

greg0ire:
https://github.com/doctrine/orm/releases/tag/2.10.0 was actually released on October 3rd 2021, the phrasing has mislead you @stof, I'm afraid :wink: 

We should IMO drop DBAL 2 support from the 2.8.x branch (which does not exist yet but I can fix that)

ostrolucky:
Yes we can drop 2.x if it means we can support some 3.x feature more easily. But doesn't mean to just increase composer.json constraint, but also drop all the BC layers.

amberovsky:
I'm happy to rename MySql to MySQL once 2.* is dropped. Or it can be done now - the only difference is when the errors are generated - with dbal 3 or dbal 2 installed. 
At the moment it happens with dbal 3, which I would expect is the majority of the cases.

dmaicher:
@ostrolucky do we need a new major to properly drop every BC layer for DBAL 2? I guess we would have to remove some classes. I mean they are useless/broken anyway with DBAL 3 but not sure if we can remove them in a minor release?

Like

https://github.com/doctrine/DoctrineBundle/blob/2.7.x/Command/Proxy/ImportDoctrineCommand.php
[...]

PR #1570, opened 2022-11-01
Drop support for DBAL 2

dmaicher:
Closes https://github.com/doctrine/DoctrineBundle/issues/1550


Controller/ProfilerController.php
```diff
@@ -128,17 +125,6 @@ private function explainSQLServerPlatform(Connection $connection, array $query):
 
         $stmt = $connection->executeQuery($sql, $params, $query['types']);
 
-        // DBAL 2.13 "forward compatibility" BC break handling
-        if ($stmt instanceof Result) {
```
dmaicher, line 132: so I guess with DBAL 3 this never worked :thinking:

Controller/ProfilerController.php
```diff
@@ -128,17 +125,6 @@ private function explainSQLServerPlatform(Connection $connection, array $query):
...
-            $stmt = $stmt->getIterator();
-        }
-
-        if (! $stmt instanceof PDOStatement) {
-            throw new LogicException('We need nextRowSet() functionality feature, which is not available with current DBAL driver');
```
dmaicher, line 137: anyone knows why we need to call `nextRowSet` here?

Tests/ConnectionFactoryTest.php
```diff
@@ -7,59 +7,19 @@
...
-use function defined;
-
-// Compatibility with DBAL < 3
-class_exists(\Doctrine\DBAL\Platforms\MySqlPlatform::class);
 
 class ConnectionFactoryTest extends TestCase
 {
     use VerifyDeprecations;
 
-    public function testContainer(): void
```
dmaicher, line 32: I don't understand what this test actually tests. That exception throwing logic all seems DBAL related.

reading https://github.com/doctrine/DoctrineBundle/issues/673 I don't think its relevant anymore.

composer.json
```diff
@@ -29,10 +29,10 @@
     ],
     "homepage": "https://www.doctrine-project.org",
     "require": {
-        "php": "^7.1 || ^8.0",
+        "php": "^7.3 || ^8.0",
```
derrabus: PHP 7.3 is supported by unmaintained DBAL 3 branches only. We should bump to PHP 7.4 directly.

See also the [statistics on Packagist](https://packagist.org/packages/doctrine/dbal/php-stats#3): PHP 7.3 accounts for only 1.8% of all DBAL 3 installations.

Command/DoctrineCommand.php
```diff
@@ -52,36 +46,13 @@ protected function getEntityGenerator()
...
      * Get a doctrine entity manager by symfony name.
      *
-     * @param string   $name
-     * @param int|null $shardId
+     * @param string $name
      *
      * @return EntityManager
      */
-    protected function getEntityManager($name, $shardId = null)
+    protected function getEntityManager($name)
```
derrabus: Passing a shard ID to this method should raise an exception as before.

DependencyInjection/DoctrineExtension.php
```diff
@@ -643,7 +536,7 @@ protected function ormLoad(array $config, ContainerBuilder $container)
         $container->registerForAutoconfiguration(AbstractIdGenerator::class)
             ->addTag(IdGeneratorPass::ID_GENERATOR_TAG);
 
-        if (method_exists($container, 'registerAttributeForAutoconfiguration')) {
+        if (PHP_VERSION_ID >= 80000 && method_exists(ContainerBuilder::class, 'registerAttributeForAutoconfiguration')) {
```
derrabus: Why is this change necessary?

dmaicher: Psalm was complaining with `RedundantCondition` :man_shrugging: I aligned this with https://github.com/doctrine/DoctrineBundle/blob/4bbe5880469d3cfab660f4dd887959b22bcf9dc4/DependencyInjection/DoctrineExtension.php#L162

derrabus: That's a false positive. Psalm analyzes the code with Symfony 6 deps installed and this condition makes sure we don't run into errors with Symfony 4. We should rather suppress the error here.

dmaicher: Ok I don't mind. Changed it back.

Controller/ProfilerController.php
```diff
@@ -128,17 +125,6 @@ private function explainSQLServerPlatform(Connection $connection, array $query):
...
-        }
-
-        $stmt->nextRowset();
-
         return $stmt->fetchAll(PDO::FETCH_ASSOC);
```
derrabus: This doesn't work with DBAL 3 either, does it. I'm tempted to replace this method with an exception and a call for contributors, tbh. 🙈

dmaicher: yeah I have no easy way of testing this :cry: :see_no_evil: seems completely broken indeed

dmaicher: Although maybe it works?

That method should still work: https://github.com/doctrine/dbal/blob/3.5.x/src/Result.php#L300

derrabus: Yes, but it does not accept PDO constants anymore.

dmaicher: Well... `FetchMode::ASSOCIATIVE = PDO::FETCH_ASSOC = 2` :see_no_evil:

derrabus: What a lucky coincidence. 😁 

Shall we just switch to `fetchAllAssociative()` and cross fingers then?

.github/workflows/continuous-integration.yml
```diff
@@ -21,9 +21,6 @@ jobs:
       fail-fast: false
       matrix:
         php-version:
-          - "7.1"
-          - "7.2"
-          - "7.3"
           - "7.4"
```
stof, line 24: we should add jobs on PHP 8

.github/workflows/continuous-integration.yml
```diff
@@ -81,7 +78,7 @@ jobs:
 
       - name: "Install PHP with PCOV"
         uses: "shivammathur/setup-php@v2"
-        if: "${{ matrix.php-version != '7.1' }}"
+        if: "${{ matrix.php-version != '7.4' }}"
```
stof: This change looks wrong to me. We should keep using PCov on PHP 7.4 (removing the PHP 7.1 logic entirely)

Controller/ProfilerController.php
stof: I think it is related to the way SQL server returns the statistics

Resources/doc/configuration.rst
```diff
@@ -562,32 +494,6 @@ Configuration Reference
                         <!-- sslcrl: The name of a file containing the SSL certificate revocation list (CRL) -->
                         <!-- pooled: True to use a pooled server with the oci8/pdo_oracle driver -->
                         <!-- MultipleActiveResultSets: Configuring MultipleActiveResultSets for the pdo_sqlsrv driver -->
```
stof: You should remove these comments too

.github/workflows/continuous-integration.yml
```diff
@@ -81,7 +80,6 @@ jobs:
 
       - name: "Install PHP with PCOV"
         uses: "shivammathur/setup-php@v2"
-        if: "${{ matrix.php-version != '7.1' }}"
```
derrabus, line 84: If we do that, the `"Install PHP with XDebug"` step does not make sense anymore. Let's remove it.

dmaicher: done

Command/CreateDatabaseDoctrineCommand.php
```diff
@@ -60,51 +56,17 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
-        }
-
         $hasPath = isset($params['path']);
         $name    = $hasPath ? $params['path'] : ($params['dbname'] ?? false);
         if (! $name) {
             throw new InvalidArgumentException("Connection does not contain a 'path' or 'dbname' parameter and cannot be created.");
         }
 
         // Need to get rid of _every_ occurrence of dbname from connection configuration and we have already extracted all relevant info from url
+        /** @psalm-suppress InvalidArrayOffset */
```
ostrolucky, line 66: What was the error here? Does it complain about `url` key? Because that looks correct to me,  `$connection->getParams()` cannot really return `url` key, can it? Same in DropDatabaseDoctrineCommand

Command/DoctrineCommand.php
```diff
@@ -61,24 +56,8 @@ protected function getEntityManager($name, $shardId = null)
...
-                throw new LogicException(sprintf('Sharding is supported only in EntityManager of instance "%s".', EntityManagerInterface::class));
-            }
-
-            $connection = $manager->getConnection();
-            if (! $connection instanceof PoolingShardConnection) {
-                throw new LogicException(sprintf("Connection of EntityManager '%s' must implement shards configuration.", $name));
-            }
-
-            $connection->connect($shardId);
+        if ($shardId !== null) {
```
ostrolucky, line 59: I would just remove argument

Controller/ProfilerController.php
ostrolucky: Then let's just throw exception in this function to make it clear this is not supported for SQLSRV? Because I'm pretty sure it ccannot work without nextRowset()

Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -789,6 +785,9 @@ public function testAnnotationsBundleMappingDetection(): void
         ]);
     }
 
+    // Disabled to prevent changing the comment below to a single-line annotation
+    // phpcs:disable SlevomatCodingStandard.Commenting.RequireOneLineDocComment.MultiLineDocComment
```
ostrolucky: better to put exclude rule in phpcs.xml.dist, or to use if PHP_VERSION_ID + $this->skip, wouldn't it?

ostrolucky (commented):
nice work, few minor stuff

Tests/DependencyInjection/DoctrineExtensionTest.php
derrabus: PHPUnit 9.5 can handle single-line annotations. And if we bump to PHP 7.4, we don't need to care about earlier PHPUnit versions.

derrabus:
In composer.json, we can drop older PHPUnit versions:

```JSON
"phpunit/phpunit": "^9.5.26 || ^10.0",
```

Command/DoctrineCommand.php
dmaicher: At first I removed it but then I added it back. See https://github.com/doctrine/DoctrineBundle/pull/1570#discussion_r1010801285

Command/CreateDatabaseDoctrineCommand.php
dmaicher: Yes exactly. The error was about the `url` key.

> $connection->getParams() cannot really return url key, can it?

Actually it can :cry: I tested it. If I configure a DBAL connection using `url` in the bundle config then that key is also present in connection params.

Controller/ProfilerController.php
dmaicher: removed + added an exception

Command/DoctrineCommand.php
ostrolucky: alright then

derrabus (approved):

ostrolucky (approved):

derrabus:
Thank you, @dmaicher!

PR #1574, opened 2022-11-07
Lowering the EntityValueResolver to be lower than the #[CurrentUser]

weaverryan:
Hi!

Currently, this, unfortunately, doesn't work:

```php
#[Route('/products/edit/{id}', name: 'product_edit')]
public function foo(Product $product, #[CurrentUser] User $user)
```

The `EntityValueResolver` runs before the `#[CurrentUser]` system and so it will try to query for `User`. This changes the priority to be lower than that system so that `#[CurrentUser]` runs first.

The trade-off is that this situation won't work: https://github.com/doctrine/DoctrineBundle/pull/1554#discussion_r949478887

However, because this is a new feature, we can choose which use-cases we want to support and not support. Querying from a `Foo` entity whose property is called `foo` looks weird to me.

ALSO, Symfony could implement some sort of two-pass system to allow for "conflicting" entity resolvers to be handled more intelligently. So, for right now, the decision, I think, is to decide which use-case is more important to support today.

Cheers!

### DIFF
--- a/Resources/config/orm.xml
+++ b/Resources/config/orm.xml
@@ -175,7 +175,7 @@
         <service id="doctrine.orm.entity_value_resolver" class="Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver">
             <argument type="service" id="doctrine" />
             <argument type="service" id="doctrine.orm.entity_value_resolver.expression_language" on-invalid="ignore" />
-            <tag name="controller.argument_value_resolver" priority="110" />
+            <tag name="controller.argument_value_resolver" priority="30" />
         </service>
 
         <service id="doctrine.orm.entity_value_resolver.expression_language" class="Symfony\Component\ExpressionLanguage\ExpressionLanguage" />


kbond:
@HypeMC, WDYT?

HypeMC:
Hi guys.

> @HypeMC, WDYT?

Since neither solution is ideal, I think it boils down to which one has the better workaround for its corresponding drawback. If you decide to keep the current priority, an easy way to get around the issue would be to use the `#[MapEntity]` attribute, eg:

```php
#[Route('/products/edit/{id}', name: 'product_edit')]
public function foo(Product $product, #[CurrentUser, MapEntity(disabled: true)] User $user)
```

Even though it requires using an extra attribute, that's really all it requires, everything else remains the same.

On the other hand, if you decide to lower the resolver's priority, you'd be reintroducing the problem with the `RequestAttributeValueResolver` mentioned in https://github.com/doctrine/DoctrineBundle/pull/1554#discussion_r949478887 . Of course, the easy workaround here would be to rename the variable, but IMO this is a bad solution, as it would mean that the name of the variable is reliant on the name of the route param. I don't think the route param should in any way influence how you name your variables (unless you actually want the `RequestAttributeValueResolver` to resolve your argument). Personally, I'd rather go with the extra attribute option.

> However, because this is a new feature, we can choose which use-cases we want to support and not support. Querying from a `Foo` entity whose property is called `foo` looks weird to me.

I think you misunderstood the problem here, it had nothing to do with a `Foo` entity whose property is called `foo`, but with a variable whose name (`$foo`) is the same as that of a route param (`{foo}`), eg:

```php
    #[Route('/{foo}', name: 'app_test')]
    public function index(SomeEntity $foo): Response
    {
        dd($foo);
    }
```
Since the `RequestAttributeValueResolver` had a higher priority, it would resolve `$foo` to the value of the route param.

HypeMC:
Then again, it just occurred to me that this is not a problem specific to the `EntityValueResolver`, other resolvers such as the `UserValueResolver` have it as well, eg:

```php
#[Route('/{user}', name: 'app_user')]
public function index(#[CurrentUser] User $user): Response {}
```

Since the `RequestAttributeValueResolver` has a higher priority than the `UserValueResolver`, `$user` is resolved to the value of the route param:
```
Argument #1 ($user) must be of type App\Entity\User, string given
```

So I guess it would make sense to lower the priority of the `EntityValueResolver`, at least for the time being. :thinking:

nicolas-grekas (approved):
:+1: on my side. The conflict with `RequestAttributeValueResolver` is not a real one to me: one should use different names for different things (route param / controller arg).

stof:
Using the same name for the route parameter and the controller argument transformed to an entity is actually a very important use case to allow seamless migration from SensioFrameworkExtraBundle's ParamConverter, where this was the way to make it work without custom mapping.
And it is also the way this EntityValueResolver works when mapping ids without a custom attribute.

stof:
maybe both the EntityValueResolver and the UserValueResolver should have higher priority than RequestAttributeValueResolver

weaverryan:
> Using the same name for the route parameter and the controller argument transformed to an entity is actually a very important use case to allow seamless migration from SensioFrameworkExtraBundle's ParamConverter, where this was the way to make it work without custom mapping.

> And it is also the way this EntityValueResolver works when mapping ids without a custom attribute.

Can you elaborate with an example for both of these?

> maybe both the EntityValueResolver and the UserValueResolver should have higher priority than RequestAttributeValueResolver

It does seem like that would make more sense - i.e. "only grab from request attributes if these other value resolvers have not been able to resolve".

nicolas-grekas:
We don't need to map `user` route param to `$user` argument to make ParamConverter / EntityValueResolver work AFAIK. Even the doc says the route param is typically `id` when the argument is `$user`. Having "user" at both positions would mean that the user has an identifier named also user. Looks the uncommon case to me.

If I'm not missing something (which I could), then I'd prefer this change over reordering the precedence of other things.

stof:
@weaverryan You can have routes like those:

```yaml
post_show:
  path: /posts/{space}

event_show:
  path: /events/{event}

talk_show:
  path: /events/{event}/talk/{talk}
```

This pattern naming the request parameter (taking an id) with a similar name than the controller argument (taking the entity instance). This works out of the box in SensioFrameworkExtraBundle, and was working out of the box (i.e. without having to tag arguments with `MapEntity` with EntityValueResolver before this priority change), with or without auto_mapping enabled. After this change, I'm not even sure it would work even when using `MapEntity` explicitly (as RequestAttributeValueResolver might still win).

The alternatives are:
1. use `id` as the parameter name in routes. This has several drawbacks:
    - it does not work when a route has several arguments needing to be resolved (see `talk_show` in my example)
    - it does not make it clear in the route definition which kind of object id is expected when generating the URL
2. find different names for each parameter (maybe adding a `_id` suffix in route parameters) and add `MapEntity` for all arguments. This is a huge pain for DX.

And both alternatives are adding a huge cost to the migration path from ParamConverter to the EntityValueResolver

stof:
> Even the doc says the route param is typically `id` when the argument is `$user`

the code supports both `id` and param name. So what is the typical usage would require looking at actual usage. And using `id` works only for the basic case. So I'm not sure it is actually the typical usage in bigger projects.

stof:
And even if it were the typical usage, you still break the migration path from ParamConverter, making it a lot harder to deprecate ParamConverter.

If I have to change the parameter names for all my routes in the Incenteev project to stop using the same name than the controller for y controller arguments (which is the case for 95% of my route parameters at least), migrating from ParamConverter to the EntityValueResolver would cost be days or weeks of work (having to go over all route generation to use the renamed argument and over all controllers to add `MapEntity`, without a way to do the work incrementally as ParamConverter would not read those `MapEntity` before changing the resolver). That's not a good migration path.

weaverryan:
If I understand it correctly, with SensioFrameworkExtraBundle, this currently works:

```php
#[Route('/posts/{post}')
public function edit(Post $post)
{

}
```

Where `{post}` is actually the `id`. Is that correct? If so, I actually had no idea this worked. I've never had a problem using `{id}`. The situations where want to auto-query for multiple entities on a single route (`/events/{event}/talk/{talk}`) are not something I have run into. But you're right that we're guessing at what's the common usage.

weaverryan:
> If I have to change the parameter names for all my routes in the Incenteev project to stop using the same name than the controller for y controller arguments (which is the case for 95% of my route parameters at least), migrating from ParamConverter to the EntityValueResolver would cost be days or weeks of work 

That's a fair point. Is that all fixed if we did this?

> maybe both the EntityValueResolver and the UserValueResolver should have higher priority than RequestAttributeValueResolver

stof:
Well, given I'm not using the UserValueResolver myself, I don't know what would be the impact of moving it before the RequestAttributeValueResolver (but I think it is fine given that it requires the type to be explicitly the UserInterface or to use the `CurrentUser` attribute).

However, having EntityValueResolver before RequestAttributeValueResolver is indeed what is needed to solve that use case (making the resolved entity win over the raw request attribute)

nicolas-grekas:
Got it, let's change the priorities then.

Maybe we should also consider requiring a more explicit mappings at some point, eg user_id => $user->id, user_slug => $user->slug, etc.

stof:
Well, the current implementation has these modes:

1. `MapEntity` with a mapping provided => use this mapping and nothing else
2. Use `$request->attributes->get($paramName)` or `$request->attributes->get('id')` and try to find by primary key
3. if auto_mapping is turned on, get all attributes and keep the ones that correspond to mapped properties in the entity. Then find by this combined criteria

Adding some kind of prefix support would add even more cases. While supporting `$paramName . '_id'` as part of the id retrieval might make sense, I'm not convinced by adding support for the prefix in auto-mapping (but I'm also not convinced by auto-mapping at all, as it caused more WTF bugs than it helped in my own experience, as soon as multiple entities are involved in the controller signature, so I'd rather encourage using `MapEntity` for cases wanting to search by something else than id)

nicolas-grekas:
We'd then deprecate the other ways if we go with prefixed mapping.

stof:
That would be a very painful deprecation, as it would require renaming route parameters, which then requires updating every place using the url generator (as well as any code accessing the route attributes directly for some reason)

To me, the case of using the parameter name works quite fine without needing an `_id` suffix in the route parameter.

Note however that I would totally be in favor of deprecating the auto-mapping feature as I think an explicit mapping is a lot more robust for those cases. But this is something to be discussed on the symfony/symfony repo to have more insight of the community (at least I managed to make the auto_mapping feature configurable in the EntityValueResolver while the DoctrineParamConverter had no way to disable it, but it is still on by default for now).

nicolas-grekas:
Please check #1578

Issue #1524, opened 2022-05-09
Deprecate ORM proxy commands


dmaicher:
As mentioned in https://github.com/doctrine/DoctrineBundle/pull/1513#discussion_r859096142 the [proxy commands](https://github.com/doctrine/DoctrineBundle/tree/2.6.x/Command/Proxy) would just change the command name and description once we remove the deprecated helper approach of providing the entity manager in 3.0.

So we should deprecate them and prepare a path of removing them. 

As suggested in the linked discussion we could make the name and description configurable on the ORM commands or change them there directly.

stof:
The need for the ORM is actually to keep the command name configurable (the base Command constructor accepts a name).

Then, the bundle could add the name in the tag (for lazyness) and in the constructor (so that the command has the right name).
but the current code setting the name in `configure` in the ORM means that it overrides any name set in the constructor
For the description, I don't think we need to override it in the bundle.

ostrolucky:
As far as I can see, DBAL also defines name in `configure` and we don't care about it, we just specify name in tag. Like here https://github.com/doctrine/DoctrineBundle/blob/b7002752b555cb61cc9ca8d4c89a5b6e8200f765/Resources/config/dbal.xml#L104-L104 Why wouldn't this work for ORM?

dmaicher:
@ostrolucky indeed I think this works. I just tried this diff on one of my apps:

```diff
-        <service id="doctrine.cache_clear_query_cache_command" class="Doctrine\Bundle\DoctrineBundle\Command\Proxy\ClearQueryCacheDoctrineCommand">
+        <service id="doctrine.cache_clear_query_cache_command" class="Doctrine\ORM\Tools\Console\Command\ClearCache\QueryCommand">
             <argument type="service" id="doctrine.orm.command.entity_manager_provider" />
             <tag name="console.command" command="doctrine:cache:clear-query" />
         </service>
```

and the configured name from the tag works:

```
$ bin/console doctrine:cache:clear-query --help
Description:
  Clear all query cache of the various cache drivers

Usage:
  doctrine:cache:clear-query [options]
``` 

Just the description is now of course the default one provided by the ORM command.

stof:
In case the description in the proxy command is better than the one in the ORM command, we should update the description upstream.

PR #1581, opened 2022-11-14
deprecate all ORM proxy commands

dmaicher:
Closes https://github.com/doctrine/DoctrineBundle/issues/1524

List of all doctrine commands from this bundle as tested on one of my apps using this branch:

```
      doctrine:database:create                  Creates the configured database                                                                                    
      doctrine:database:drop                    Drops the configured database                                                                                      
      doctrine:query:sql                        Executes arbitrary SQL directly from the command line.                                                             
      doctrine:cache:clear-metadata             Clear all metadata cache of the various cache drivers                                                              
      doctrine:cache:clear-query                Clear all query cache of the various cache drivers                                                                 
      doctrine:cache:clear-result               Clear all result cache of the various cache drivers                                                                
      doctrine:cache:clear-collection-region    Clear a second-level cache collection region                                                                       
      doctrine:mapping:convert                  Convert mapping information between supported formats                                                              
      doctrine:schema:create                    Processes the schema and either create it directly on EntityManager Storage Connection or generate the SQL output  
      doctrine:schema:drop                      Drop the complete database schema of EntityManager Storage Connection or generate the corresponding SQL output     
      doctrine:ensure-production-settings       Verify that Doctrine is properly configured for a production environment                                           
      doctrine:cache:clear-entity-region        Clear a second-level cache entity region                                                                           
      doctrine:mapping:info                     Show basic information about all mapped entities                                                                   
      doctrine:cache:clear-query-region         Clear a second-level cache query region                                                                            
      doctrine:query:dql                        Executes arbitrary DQL directly from the command line                                                              
      doctrine:schema:update                    Executes (or dumps) the SQL needed to update the database schema to match the current mapping metadata             
      doctrine:schema:validate                  Validate the mapping files                                                                                         
      doctrine:mapping:import                   Imports mapping information from an existing database     
```

### DIFF
--- a/Command/Proxy/ClearMetadataCacheDoctrineCommand.php
+++ b/Command/Proxy/ClearMetadataCacheDoctrineCommand.php
@@ -7,6 +7,8 @@ use Symfony\Component\Console\Input\InputOption;
 
 /**
  * Command to clear the metadata cache of the various cache drivers.
+ *
+ * @deprecated use Doctrine\ORM\Tools\Console\Command\ClearCache\MetadataCommand instead
  */
 class ClearMetadataCacheDoctrineCommand extends MetadataCommand
 {
--- a/Command/Proxy/ClearQueryCacheDoctrineCommand.php
+++ b/Command/Proxy/ClearQueryCacheDoctrineCommand.php
@@ -7,6 +7,8 @@ use Symfony\Component\Console\Input\InputOption;
 
 /**
  * Command to clear the query cache of the various cache drivers.
+ *
+ * @deprecated use Doctrine\ORM\Tools\Console\Command\ClearCache\QueryCommand instead
  */
 class ClearQueryCacheDoctrineCommand extends QueryCommand
 {
--- a/Command/Proxy/ClearResultCacheDoctrineCommand.php
+++ b/Command/Proxy/ClearResultCacheDoctrineCommand.php
@@ -7,6 +7,8 @@ use Symfony\Component\Console\Input\InputOption;
 
 /**
  * Command to clear the result cache of the various cache drivers.
+ *
+ * @deprecated use Doctrine\ORM\Tools\Console\Command\ClearCache\ResultCommand instead
  */
 class ClearResultCacheDoctrineCommand extends ResultCommand
 {
--- a/Command/Proxy/CollectionRegionDoctrineCommand.php
+++ b/Command/Proxy/CollectionRegionDoctrineCommand.php
@@ -7,6 +7,8 @@ use Symfony\Component\Console\Input\InputOption;
 
 /**
  * Command to clear a collection cache region.
+ *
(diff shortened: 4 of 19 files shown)


ostrolucky (approved):

stof:
The `doctrine:query:dql` command is redefining the help text instead of inheriting it from the parent. Can you check why this was done ? Is the help text of the bundle better than the help text of the ORM ?

stof:
This should probably also deprecate the proxy commands for DBAL commands. For now, you handle only ORM ones.

dmaicher:
> The `doctrine:query:dql` command is redefining the help text instead of inheriting it from the parent. Can you check why this was done ? Is the help text of the bundle better than the help text of the ORM ?

ORM:
```
Executes arbitrary DQL directly from the command line.
``` 

Bundle:
```
The <info>%command.name%</info> command executes the given DQL query and
outputs the results:

<info>php %command.full_name% "SELECT u FROM UserBundle:User u"</info>

You can also optional specify some additional options like what type of
hydration to use when executing the query:

<info>php %command.full_name% "SELECT u FROM UserBundle:User u" --hydrate=array</info>

Additionally you can specify the first result and maximum amount of results to
show:

<info>php %command.full_name% "SELECT u FROM UserBundle:User u" --first-result=0 --max-result=30</info>
``` 

So yeah seems better :smile:

stof:
I suggest improving the help in the ORM (without using the entity namespace feature that has been removed in doctrine/persistence 3.x)

dmaicher:
@stof see https://github.com/doctrine/orm/pull/10233

dmaicher:
@stof regarding

> This should probably also deprecate the proxy commands for DBAL commands. For now, you handle only ORM ones.

Which DBAL command are you referring to exactly?

From what I can tell all commands inside the `Command\Proxy` namespace are now deprecated.


Screenshot from 2022-11-15 17-03-21

stof:
Actually, the deprecation of DBAL proxy commands was half-done already. For instance, `Doctrine\Bundle\DoctrineBundle\Command\Proxy\RunSqlDoctrineCommand` is marked as deprecated, but it is still used when running the command (unlike your change where you use the ORM commands directly for service definitions)

dmaicher:
@stof I will do this for the DBAL command(s) separately. Ok?

dmaicher:
@stof @ostrolucky actually this is done already for the DBAL command, right?

Checking 

https://github.com/doctrine/DoctrineBundle/commit/95c9daff56c0eba9a8df1838c46a63a52735dde0#diff-11eb636a4aa9dc6c6b58251b233250d85cb1b8c16ec5845574e7a5b4fd90ff42R105

I see that because of the changed command name `doctrine:query:sql` vs `dbal:run-sql` basically we now have both classes registered as services.

ostrolucky:
yeah

PR #1584, opened 2022-11-28
Clear second level cache stats after data collected

tourze:
I am writing a Symfony application which integrated with Swoole.
I found that when request finished, the profiler page will not clear the counter in profiler page.
After some research, I think the problem is that DoctrineBundle froget to reset the stat data after profiler data collected.
This PR also fix a small OOM problem in long-run runtime.

### DIFF
--- a/DataCollector/DoctrineDataCollector.php
+++ b/DataCollector/DoctrineDataCollector.php
@@ -180,6 +180,8 @@ class DoctrineDataCollector extends BaseCollector
 
                 $caches['regions']['misses'][$key] += $value;
             }
+
+            $cacheLoggerStats->clearStats();
         }
 
         $this->data['entities'] = $entities;


ostrolucky:
This problem should be fixed by adding kernel.reset tag to https://github.com/doctrine/DoctrineBundle/blob/2f094c91fa3107bf03934b1f30ffcc975abba216/DependencyInjection/DoctrineExtension.php#L857-L857

Not reset it when collecting

tourze:
> kernel.reset

I try to add this line `$loggerStatsDef->addTag('kernel.reset', ['method' => 'clearStats']);`, and the service is tagged as kernel.reset successfully, but however the second level cache is not clear.

image
image

Is there anything i am missing?

ostrolucky:
How do you have swoole integrated with Symfony? Are you using Swoole bridge? This looks like issue there. Your runtime should support kernel.reset semantics.

tourze:
> How do you have swoole integrated with Symfony? Are you using Swoole bridge? This looks like issue there. Your runtime should support kernel.reset semantics.

Using this component https://symfony.com/doc/current/components/runtime.html
I think the problem will exists in any other long-run runtime, like roadrunner/swoole/reactphp

ostrolucky:
I can debug further if demo could be provided. symfony/runtime also needs to be set up in certain way with swoole runner.

Issue #1585, opened 2022-11-29
Issues with DbalLogger on version 2.7


jbnemanja:
I'm getting a deprecation warning:

```
User Deprecated: The "Symfony\Bridge\Doctrine\Logger\DbalLogger" class implements "Doctrine\DBAL\Logging\SQLLogger" that is deprecated Use {@see \Doctrine\DBAL\Logging\Middleware} or implement {@see \Doctrine\DBAL\Driver\Middleware} instead.
```

I saw the issue was brought up in #1431 and resolved in #1517 for v2.7. But I'm still getting the warning, these are my dependencies

```
"require": {
    "php": ">=8.0",
    "ext-ctype": "*",
    "ext-iconv": "*",
    "doctrine/annotations": "^1.13",
    "doctrine/doctrine-bundle": "^2.7",
    "doctrine/doctrine-migrations-bundle": "^3.2",
    "doctrine/orm": "^2.13",
    "gedmo/doctrine-extensions": "^3.10",
    "gesdinet/jwt-refresh-token-bundle": "^1.1",
    "lexik/jwt-authentication-bundle": "^2.16",
    "sensio/framework-extra-bundle": "^6.2",
    "stof/doctrine-extensions-bundle": "^1.7",
    "symfony/console": "5.4.*",
    "symfony/dotenv": "5.4.*",
    "symfony/flex": "^1.17|^2",
    "symfony/framework-bundle": "5.4.*",
    "symfony/http-client": "5.4.*",
    "symfony/http-foundation": "5.4.*",
    "symfony/property-access": "5.4.*",
    "symfony/proxy-manager-bridge": "5.4.*",
    "symfony/runtime": "5.4.*",
    "symfony/security-bundle": "5.4.*",
    "symfony/serializer": "5.4.*",
    "symfony/uid": "5.4.*",
    "symfony/validator": "5.4.*",
    "symfony/yaml": "5.4.*"
},
"require-dev": {
    "friendsofphp/php-cs-fixer": "^3.12",
    "phpstan/phpstan": "^1.8",
    "symfony/stopwatch": "5.4.*",
    "symfony/web-profiler-bundle": "5.4.*",
    "vimeo/psalm": "^4.29"
},
```

I'm using PHP 8.1 and Symfony 5.4

dmaicher:
[...]

PR #1586, opened 2022-11-29
Drop support for Symfony 4

derrabus:
Active support for Symfony 4 ends this month. This PR drops compat code we have maintained for Symfony 4.


.github/workflows/continuous-integration.yml
```diff
@@ -82,7 +76,7 @@ jobs:
 
       - name: "Require symfony/messenger"
         run: "composer require --dev symfony/messenger --no-update"
-        if: "${{ startsWith(matrix.symfony-require, '4.') }}"
+        if: "${{ startsWith(matrix.symfony-require, '5.') }}"
```
stof, line 79: Why having a single job installing it ?

ostrolucky: Yeah this is something I wanted to investigate too. IIRC we had to install symfony/messenger for 4.x because it wasn't otherwise installed, while with 5.x it was. But I'm not 100% sure. But if it's true we can just remove this.

dmaicher:
I think we should then also remove all of those?

https://github.com/doctrine/DoctrineBundle/blob/2.8.x/Resources/config/dbal.xml#L26-L37

This would then also fix https://github.com/doctrine/DoctrineBundle/issues/1585

.github/workflows/continuous-integration.yml
derrabus: … which would be strange because Symfony Messenger did exist in Symfony 4. I mean, I could just add the messenger to `require-dev` and drop this block. This would also allow to simplify a couple of tests and fix some Psalm errors.

Right now, we test that the bundle still works if Messenger is not installed. But then again, we don't do this for other Symfony components, only for Messenger. 🤷🏻‍♂️ 

I have no strong opinion here. Tell me which way to go and I'll change the code accordingly.

derrabus:
> I think we should then also remove all of those?

There are tests in place that assert that the `doctrine.dbal.logger` service exists even if we use the logger middleware. But of course, that service serves no purpose anymore. It's just there. If removing it is not considered a BC break, I can remove it and adjust the tests.

ostrolucky:
People will be upset about removing this service for sure, but better to throw it in their face that they need to change something, rather than their code not doing anything. So I agree with removal.

.github/workflows/continuous-integration.yml
ostrolucky: I think it's good that we test both without and with messenger to ensure code works with both, but I don't have strong opinion. Btw this aproach was introduced in https://github.com/doctrine/DoctrineBundle/pull/817

stof:
Symfony supports marking a service as deprecated. That's what we should do for the logger service when we use the middleware

dmaicher:
> Symfony supports marking a service as deprecated. That's what we should do for the logger service when we use the middleware

But with this PR we will always use the middlewares. So the legacy logging services are not doing anything anymore. You would prefer to still keep them and deprecate them?

EDIT: I personally would be fine with removing them. I see those as internal details and not as an extension point that one should rely on.

.github/workflows/continuous-integration.yml
derrabus: Okay, let's keep this as-is for now. We could revisit that in a follow-up. Maybe we could create a CI job that runs tests with minimal deps, so that we cover other optional deps as well.

ostrolucky (approved):

PR #1587, opened 2022-12-04
remove legacy DBAL logging services + parameters

dmaicher:
WDYT about cleaning this up entirely?

### DIFF
--- a/Resources/config/dbal.xml
+++ b/Resources/config/dbal.xml
@@ -5,9 +5,6 @@
     xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
     <parameters>
-        <parameter key="doctrine.dbal.logger.chain.class">Doctrine\DBAL\Logging\LoggerChain</parameter>
-        <parameter key="doctrine.dbal.logger.profiling.class">Doctrine\DBAL\Logging\DebugStack</parameter>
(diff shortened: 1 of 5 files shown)


ostrolucky (approved):

stof (approved):

Resources/config/dbal.xml
```diff
@@ -23,19 +20,6 @@
         <service id="Doctrine\DBAL\Driver\Connection" alias="database_connection" public="false" />
         <service id="Doctrine\DBAL\Connection" alias="database_connection" public="false" />
 
-        <service id="doctrine.dbal.logger.chain" class="%doctrine.dbal.logger.chain.class%" public="false" abstract="true" >
```
theofidry, line 26: AFAICS this is a BC break isn't it? Although the service has been deprecated (not sure it actually has been marked as such), this was supposedly done in 2.3.0

ostrolucky: We favored BC break over userspace hooking into code that does nothing, which would cause functionality to silently stop working. BTW we don't guarantee BC promise on service definitions, similarly as [Symfony doesn't ](https://symfony.com/doc/current/contributing/code/bc.html).

theofidry: Thanks for the note, indeed a BC makes more sense in that context.

>BTW we don't guarantee BC promise on service definitions, similarly as Symfony doesn't

I was not aware it did not.

PR #1596, opened 2022-12-20
Allow Doctrine Annotations 2

derrabus:
The bundle itself should be compatible with Annotations 2 right away.

However, I wonder if the Annotations package should remain a mandatory dependency. If we use the bundle to configure the DBAL without the ORM, requiring Annotations feels pointless. Moreover, in a PHP 8 project with native attributes, we don't necessarily need Annotations for the ORM either.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
     "homepage": "https://www.doctrine-project.org",
     "require": {
         "php": "^7.4 || ^8.0",
-        "doctrine/annotations": "^1",
+        "doctrine/annotations": "^1|^2",
         "doctrine/cache": "^1.11 || ^2.0",
         "doctrine/dbal": "^3.4.0",
         "doctrine/persistence": "^2.2 || ^3",


derrabus:
Regarding the Psalm failure: psalm/psalm-plugin-symfony#296

derrabus:
The stable deps jobs will remain red until Symfony makes another round of bugfix releases, see symfony/symfony#48718

greg0ire:
> However, I wonder if the Annotations package should remain a mandatory dependency. If we use the bundle to configure the DBAL without the ORM, requiring Annotations feels pointless. Moreover, in a PHP 8 projects with native attributes, we don't necessarily need Annotations for the ORM either.

Looks like the dependency was added one year ago as a quick fix for the CI: https://github.com/doctrine/DoctrineBundle/commit/d7afbaca3274078f5d44e3c8c38acd0e6dc66cc7

I don't think it was reconsidered.

It seems to be only used here: https://github.com/doctrine/DoctrineBundle/blob/d2088fc50494e4e7441fecca54732245a613eeb6/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php#L131-L137

I have no idea what uses that method though; looked in Symfony, couldn't find a reference.

mbabker:
> Looks like the dependency was added one year ago as a quick fix for the CI: [d7afbac](https://github.com/doctrine/DoctrineBundle/commit/d7afbaca3274078f5d44e3c8c38acd0e6dc66cc7)
> 
> I don't think it was reconsidered.

The quickest fix I could come up with in a PR to another bundle yesterday was to run `composer require --no-update "doctrine/annotations:<constraint>"` before installing all the dependencies.  The FrameworkBundle uses `ContainerBuilder::willBeAvailable()` to decide if annotations support is enabled or disabled by default, so unless something has the package as a direct production dependency, it ends up not enabling annotations.

> I have no idea what uses that method though; looked in Symfony, couldn't find a reference.

It's for apps/bundles that don't register their mapping files in an auto-mapped folder, you're not going to find too many uses of it TBH.

nicolas-grekas:
I would also vote for removing the dep (or move it to req-dev if needed for tests)

stof:
I agree with @nicolas-grekas

ostrolucky:
I agree in principle, but if we simply remove it and install it in CI build, we wouldn't test for a case when it's not installed

mbabker:
https://github.com/doctrine/DoctrineBundle/compare/2.8.x...mbabker:DoctrineBundle:28.x is pretty ugly but it gets the test suite to run when the annotations package isn't installed (all of the skips are needed because the test fixtures use an annotation based bundle).

Tobion:
Agree that the dependency on annotations should be removed.

mbabker:
See #1598 for a full PR

derrabus:
All right, let's continue there. Thanks for taking over.

PR #1598, opened 2022-12-21
Allow Annotations 2, add attribute support to testing fixtures and test with attributes when able, allow the Annotations package to be an optional dependency

mbabker:
This can supersede #1596

This PR will:

- Allow `doctrine/annotations` 2.x to be installed
- Makes the Annotations package optional by moving it to `require-dev` and unloading the `doctrine.orm.metadata.annotation_reader` service alias if not installed
- Explicitly configures whether annotation support should be enabled in test fixtures instead of relying on feature detection in the FrameworkBundle (which relies on the package existing in someone's `require` list)
- Adds attributes to the `RepositoryServiceBundle` test fixture bundle that is loaded by `Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection\Fixtures\TestKernel`
- Configures resources using the `RepositoryServiceBundle` test fixture bundle to use attributes when run on PHP 8 and annotations on PHP 7
- Skips tests if run on PHP 7 and the Annotations package isn't installed

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -12,6 +12,7 @@ use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\IdGeneratorPass;
 use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\ServiceRepositoryCompilerPass;
 use Doctrine\Bundle\DoctrineBundle\EventSubscriber\EventSubscriberInterface;
 use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepositoryInterface;
+use Doctrine\Common\Annotations\Annotation;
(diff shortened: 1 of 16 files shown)


maxhelias:
The CI will be fixed only when https://github.com/symfony/symfony/pull/48725 is released

mbabker:
No, it's https://github.com/symfony/symfony/pull/48718 that most of the CI's waiting for (there's also https://github.com/psalm/psalm-plugin-symfony/pull/296 which was needed to unblock SA but that looks to be released now); that particular change removes some code from Symfony 6.3 only that was only needed to support old Annotations versions.

maxhelias:
Yeah i missed that one 👍

derrabus (approved):

derrabus:
Thank you @mbabker!

PR #1602, opened 2022-12-29
Remove messenger transport service conditionally

### LINKED ISSUE #1601: [2.8.0] Container compilation fails if symfony/doctrine-messenger is not used
After upgrading from `2.7.2` to `2.8.0` (with symfony 6.2), container compilation fails:

```
In AbstractRecursivePass.php line 170:
                                                                               
  Invalid service "messenger.transport.doctrine.factory": class "Symfony\Comp  
  onent\Messenger\Bridge\Doctrine\Transport\DoctrineTransportFactory" does no  
  t exist.         
```

This is caused by #1586 removing the checks [here](https://github.com/doctrine/DoctrineBundle/pull/1586/files#diff-afe5f060789287a8ad428a4b5df1ad844184f496ece0297e421f6b1fc442b5a4L1043-L1064).

We use Symfony Messenger, but not the Doctrine transport. This use case should be allowed, as the Doctrine transport [became optional in Symfony 6.0](https://github.com/symfony/symfony/blob/6.3/UPGRADE-6.0.md#messenger).

Just because `symfony/messenger` is used, does not mean that `symfony/doctrine-messenger` is also used.

gndk:
Fixes #1601

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -48,6 +48,7 @@ use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
 use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Messenger\Bridge\Doctrine\Transport\DoctrineTransportFactory;
 use Symfony\Component\Messenger\MessageBusInterface;
 use Symfony\Component\PropertyInfo\PropertyInfoExtractorInterface;
 use Symfony\Component\Validator\Mapping\Loader\LoaderInterface;
@@ -1030,7 +1031,8 @@ class DoctrineExtension extends AbstractDoctrineExtension
 
     private function loadMessengerServices(ContainerBuilder $container): void
     {
-        // If the Messenger component is installed and the doctrine transaction middleware is available, wire it:
+        // If the Messenger component is installed, wire it:
+
         /** @psalm-suppress UndefinedClass Optional dependency */
         if (! interface_exists(MessageBusInterface::class)) {
             return;
@@ -1038,6 +1040,18 @@ class DoctrineExtension extends AbstractDoctrineExtension
 
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('messenger.xml');
+
+        /**
+         * The Doctrine transport component (symfony/doctrine-messenger) is optional.
+         * Remove service definition, if it is not available.
+         */
+
+        /** @psalm-suppress UndefinedClass Optional dependency */
+        if (class_exists(DoctrineTransportFactory::class)) {
+            return;
+        }
+
+        $container->removeDefinition('messenger.transport.doctrine.factory');
     }
 
     private function createArrayAdapterCachePool(ContainerBuilder $container, string $objectManagerName, string $cacheName): string
--- a/Resources/config/messenger.xml
+++ b/Resources/config/messenger.xml
@@ -39,8 +39,7 @@
         </service>
 
         <!--
-        The following service isn't tagged as transport factory because the class may not exist.
-        The tag is added conditionally in DoctrineExtension.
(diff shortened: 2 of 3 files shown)


gndk:
Not sure about the best way to test based on optional dependencies, so I manually composer required/removed `symfony/messenger` and `symfony/doctrine-messenger` to verify the new tests locally. Should they be added to `require-dev` like `doctrine/orm` for some of the other tests in `DoctrineExtensionTest`?

Resources/config/messenger.xml
```diff
@@ -44,7 +44,6 @@
         -->
         <service id="messenger.transport.doctrine.factory" class="Symfony\Component\Messenger\Bridge\Doctrine\Transport\DoctrineTransportFactory" public="false">
             <argument type="service" id="doctrine" />
-            <tag name="messenger.transport_factory" />
```
ostrolucky: Wouldn't it be better to define as much things as we can here and remove the whole definition if criteria don't match? As far as I can see, at the moment you are leaving this service registered, but is unused because it won't have a tag.

DependencyInjection/DoctrineExtension.php
```diff
@@ -1030,14 +1031,27 @@ protected function getMetadataDriverClass(string $driverType): string
...
+        // The Doctrine transport component is optional. Only tag if it is available.
+
+        /** @psalm-suppress UndefinedClass Optional dependency */
+        if (! class_exists(DoctrineTransportFactory::class) || ! $container->hasDefinition('messenger.transport.doctrine.factory')) {
+            return;
+        }
+
+        $container->getDefinition('messenger.transport.doctrine.factory')
+            /** @psalm-suppress UndefinedClass Optional dependency */
+            ->setClass(DoctrineTransportFactory::class)
```
ostrolucky: class is defined already in messenger.xml

ostrolucky (commented):
> Not sure about the best way to test based on optional dependencies, so I manually composer required/removed symfony/messenger and symfony/doctrine-messenger to verify the new tests locally. Should they be added to require-dev like doctrine/orm for some of the other tests in DoctrineExtensionTest?

The way we do it is that we manually require symfony/messenger in CI already (for some jobs) https://github.com/doctrine/DoctrineBundle/blob/0421ebc069519a0f19b9c39e5dc18c359be0feab/.github/workflows/continuous-integration.yml#L77 so I think the way you did it is sufficient

Resources/config/messenger.xml
gndk: Probably not much of a difference between both approaches, since, from my understanding, the unused service will be removed during compilation. So I just went with the previous approach. Or does it stay in despite not having a tag? I'm not very familiar with this. But sure, I could reverse it if you want?

gndk:
> > Not sure about the best way to test based on optional dependencies, so I manually composer required/removed symfony/messenger and symfony/doctrine-messenger to verify the new tests locally. Should they be added to require-dev like doctrine/orm for some of the other tests in DoctrineExtensionTest?
> 
> The way we do it is that we manually require symfony/messenger in CI already (for some jobs)
> 
> https://github.com/doctrine/DoctrineBundle/blob/0421ebc069519a0f19b9c39e5dc18c359be0feab/.github/workflows/continuous-integration.yml#L77
> 
> so I think the way you did it is sufficient

Maybe it would make sense to add the new combinations (`symfony/doctrine-messenger` require on/off) to the matrix then?

Resources/config/messenger.xml
ostrolucky: Sure, it doesn't make huge difference, but it will be less work for symfony optimizer if we remove it ourselves. Plus it's following existing approach we use. We try to define all the tags in xml files and if it cannot be used, we remove whole service definition. I think it's also more clean approach, because you expect definitions we define in xml files to be as complete as possible, otherwise it's required to crosscheck them with with what's defined in Extension, which will get messy fast.

ostrolucky:
I don't think adding another combination to matrix has enough ROI, so no need to do something around that.

Resources/config/messenger.xml
gndk: I started working on it before you replied and I agree, its cleaner now 👍

ostrolucky (approved):

dmaicher (approved):

ostrolucky:
Thx, merged manually

PR #1607, opened 2023-01-05
Update Profiler UI for Symfony 6.2+.

nickygerritsen:
This is based on the closed PR #1558, but only does two changes:

* It updates the icon in 6.2+.
* It adds the metric groups.

I will do a separate PR to add the tabs.

Screenshot on 6.1:
image

And on 6.2:
image

### DIFF
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -8,7 +8,7 @@
         {% set icon %}
             {% set status = collector.invalidEntityCount > 0 ? 'red' : collector.querycount > 50 ? 'yellow' %}
 
-            {{ include('@Doctrine/Collector/icon.svg') }}
+            <span class="icon">{{ include('@Doctrine/Collector/' ~ (profiler_markup_version < 3 ? 'icon' : 'database') ~ '.svg') }}</span>
 
             {% if collector.querycount == 0 and collector.invalidEntityCount > 0 %}
                 <span class="sf-toolbar-value">{{ collector.invalidEntityCount }}</span>
@@ -68,7 +68,7 @@
 
 {% block menu %}
     <span class="label {{ collector.invalidEntityCount > 0 ? 'label-status-error' }} {{ collector.querycount == 0 ? 'disabled' }}">
-        <span class="icon">{{ include('@Doctrine/Collector/icon.svg') }}</span>
+        <span class="icon">{{ include('@Doctrine/Collector/' ~ (profiler_markup_version < 3 ? 'icon' : 'database') ~ '.svg') }}</span>
         <strong>Doctrine</strong>
         {% if collector.invalidEntityCount %}
             <span class="count">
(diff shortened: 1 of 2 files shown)


nickygerritsen:
Not sure why the test fails btw. If I add `$twigLoaderFilesystem->addPath(__DIR__ . '/../Resources/views', 'Doctrine');` to the `setUp` method it works, but we already load other resources in the twig...?

Resources/views/Collector/db.html.twig
```diff
@@ -426,7 +486,9 @@
             }
 
             header.setAttribute('data-sort-direction', (-1*direction).toString());
-            header.children[0].innerHTML = direction > 0 ? '<span class="text-muted">&#9650;</span>' : '<span class="text-muted">&#9660;</span>';
+            header.children[0].innerHTML = direction > 0
+                ? `<span class="sort-icon">{{ include('@Doctrine/Collector/arrow-narrow-up.svg') }}</span>`
```
ostrolucky: Would be nice to know why are we replacing UTF-8 character with SVG, which is larger and harder to change.

Resources/views/Collector/db.html.twig
```diff
@@ -98,289 +107,340 @@
...
-            {% if collector.managers|length > 1 %}
-                <h3>{{ manager }} <small>entity manager</small></h3>
-            {% endif %}
-
-            {% if classes is empty %}
-                <div class="empty">
-                    <p>No loaded entities.</p>
-                </div>
-            {% else %}
-                <table>
```
ostrolucky: Not sure why are we changing position of this section, this makes the PR quite hard to review

Resources/views/Collector/db.html.twig
```diff
@@ -98,289 +107,340 @@
...
-                <span class="label">Cache puts</span>
+                <span class="value">{{ collector.invalidEntityCount }}</span>
+                <span class="label">Invalid entities</span>
             </div>
         {% endif %}
     </div>
 
-    {% set group_queries = request.query.getBoolean('group') %}
-    {% if group_queries %}
-        <h2>Grouped Statements</h2>
```
ostrolucky: why are these headings removed?

Resources/views/Collector/db.html.twig
```diff
@@ -98,289 +107,340 @@
...
-        <div class="metric">
-            <span class="value">{{ '%0.2f'|format(collector.time * 1000) }} ms</span>
-            <span class="label">Query time</span>
-        </div>
+            <div class="metric">
+                <span class="value">{{ collector.groupedQueryCount }}</span>
+                <span class="label">Different statements</span>
+            </div>
 
-        <div class="metric">
```
ostrolucky, line 121: Why was this moved under? Seems quite random change to me.

Resources/views/Collector/db.html.twig
```diff
@@ -98,289 +107,340 @@
...
+                    <span class="label">Cache misses</span>
+                </div>
+                <div class="metric">
+                    <span class="value">{{ collector.cachePutsCount }}</span>
+                    <span class="label">Cache puts</span>
+                </div>
             </div>
+        {% endif %}
+
+        {% if collector.invalidEntityCount %}
```
ostrolucky: This also seem quite arbitrary decision to me to hide the section when count is 0

Resources/views/Collector/db.html.twig
```diff
@@ -8,7 +8,11 @@
         {% set icon %}
             {% set status = collector.invalidEntityCount > 0 ? 'red' : collector.querycount > 50 ? 'yellow' %}
 
-            {{ include('@Doctrine/Collector/icon.svg') }}
+            {% if profiler_markup_version < 3 %}
```
ostrolucky: these can be simplified like so
```twig
<span class="icon">{{ include('@Doctrine/Collector/' ~ (profiler_markup_version < 3 ? 'icon' : 'database') ~ '.svg') }}</span>
```

nickygerritsen:
I changed the PR quite a bit and only updated the icon and added the metric groups. Once this is merged we can work from there to add the tabs for the different things.

I also didn't drop any metrics or changed any order anymore.

Resources/views/Collector/db.html.twig
nickygerritsen: Agreed, got rid of this.

Resources/views/Collector/db.html.twig
nickygerritsen: Agreed, reverted.

Resources/views/Collector/db.html.twig
nickygerritsen: I think because the UTF-8 characters don't look nice. But I also didn't really like the arrows, so I leave this to someone with actual UX/UI skills.

Resources/views/Collector/db.html.twig
```diff
@@ -103,38 +103,42 @@
...
-        <div class="metric">
-            <span class="value">{{ '%0.2f'|format(collector.time * 1000) }} ms</span>
-            <span class="label">Query time</span>
-        </div>
-
-        <div class="metric">
-            <span class="value">{{ collector.invalidEntityCount }}</span>
-            <span class="label">Invalid entities</span>
-        </div>
+        <div class="metric-group">
```
ostrolucky, line 106: What are these for? Is it so that it stretches to full width?

nickygerritsen: It “groups” the metrics. Look at the screenshots: in 6.1 it is 4 blocks, in 6.2 it is 4 parts of a bigger block. Without it, there would be whitespace between the blocks

ostrolucky (approved):

PR #1608, opened 2023-01-05
Profiler: Linkify classes in entities mapping section

nickygerritsen:
While working on #1607 I noticed entity classes are not clickable to your IDE unlike most other things in the profiler, so I added support for it. It's based on how the RequestDataCollector from Symfony's HttpKernel does it.

Screenshot:
image

### DIFF
--- a/DataCollector/DoctrineDataCollector.php
+++ b/DataCollector/DoctrineDataCollector.php
@@ -12,8 +12,6 @@ use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\ORM\Tools\SchemaValidator;
 use Doctrine\Persistence\ManagerRegistry;
 use Doctrine\Persistence\Mapping\AbstractClassMetadataFactory;
-use ReflectionClass;
-use ReflectionException;
 use Symfony\Bridge\Doctrine\DataCollector\DoctrineDataCollector as BaseCollector;
 use Symfony\Bridge\Doctrine\Middleware\Debug\DebugDataHolder;
 use Symfony\Component\HttpFoundation\Request;
@@ -117,21 +115,13 @@ class DoctrineDataCollector extends BaseCollector
                         continue;
                     }
 
-                    $classErrors = $validator->validateClass($class);
-                    try {
-                        $r                                  = new ReflectionClass($class->getName());
-                        $entities[$name][$class->getName()] = [
-                            'class' => $class->getName(),
-                            'file' => $r->getFileName(),
-                            'line' => $r->getStartLine(),
-                        ];
-                    } catch (ReflectionException $e) {
-                        $entities[$name][$class->getName()] = [
-                            'class' => $class->getName(),
-                            'file' => 'n/a',
-                            'line' => 'n/a',
-                        ];
-                    }
+                    $classErrors                        = $validator->validateClass($class);
+                    $r                                  = $class->getReflectionClass();
+                    $entities[$name][$class->getName()] = [
+                        'class' => $class->getName(),
+                        'file' => $r->getFileName(),
+                        'line' => $r->getStartLine(),
+                    ];
 
                     if (empty($classErrors)) {
                         continue;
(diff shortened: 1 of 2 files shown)


Resources/views/Collector/db.html.twig
```diff
@@ -359,10 +359,15 @@
                     </tr>
                     </thead>
                     <tbody>
-                    {% for class in classes %}
+                    {% for class in classes|keys %}
```
ostrolucky: since class is part of array, I think it would be better to change `class` instances below with `class.class` instead of iterating over the keys and then getting that same array through `classes[class]`

DataCollector/DoctrineDataCollector.php
```diff
@@ -115,8 +117,21 @@ public function collect(Request $request, Response $response, ?Throwable $except
                         continue;
                     }
 
-                    $classErrors                        = $validator->validateClass($class);
-                    $entities[$name][$class->getName()] = $class->getName();
+                    $classErrors = $validator->validateClass($class);
+                    try {
+                        $r                                  = new ReflectionClass($class->getName());
```
ostrolucky: have you tried using $class->getReflectionClass() instead?

DataCollector/DoctrineDataCollector.php
```diff
@@ -115,8 +117,21 @@ public function collect(Request $request, Response $response, ?Throwable $except
...
+                        $entities[$name][$class->getName()] = [
+                            'class' => $class->getName(),
+                            'file' => $r->getFileName(),
+                            'line' => $r->getStartLine(),
+                        ];
+                    } catch (ReflectionException $e) {
+                        $entities[$name][$class->getName()] = [
+                            'class' => $class->getName(),
+                            'file' => 'n/a',
```
ostrolucky: this will go into URL and PhpStorm will probably fail to open it, am I wrong? We could add some checking in template, but I would rather we remove whole try catching and see in real world if we actually ever need it.

Resources/views/Collector/db.html.twig
```diff
@@ -359,10 +359,15 @@
...
                         {% set contains_errors = collector.mappingErrors[manager] is defined and collector.mappingErrors[manager][class] is defined %}
                         <tr class="{{ contains_errors ? 'status-error' }}">
-                            <td>{{ class }}</td>
+                            <td>
+                                {%- set link = classes[class].file|file_link(classes[class].line) %}
+                                {%- if link %}<a href="{{ link }}" title="{{ class }}">{% else %}<span title="{{ class }}">{% endif %}
```
ostrolucky: From brief look, if you use `format_file` instead of `file_link`, looks like we wouldn't have to do this if checking ourselves

ostrolucky (changes requested):

DataCollector/DoctrineDataCollector.php
nickygerritsen: One of the tests fails since it has a non existing class. Not on pc now so can’t check which one, but how should we handle this?

ostrolucky: It fails because entity fqcs we configured in DoctrineDataCollectorTest don't exist, let's fix that.

nickygerritsen: So you suggest to just create an empty class for it?

ostrolucky: It can even be any existing class, like stdClass, doesn't have to be valid entity class

nickygerritsen: Ok let me try this tomorrow. I like the idea

Resources/views/Collector/db.html.twig
nickygerritsen: That will show `App\Entity\Hello at line 9`. Or we drop the line but then clicking the link will not send you to the start of the class. Which do you prefer?

ostrolucky: Yeah let's remove all conditionals and leave in just 
```twig
<a href="{{ class.file|file_link(class.line) }}">{{class.class}}</a>
```

(I also couldn't find a purpose for that `<span>`)

We do exactly this on other place and nobody has complained yet.

DataCollector/DoctrineDataCollector.php
nickygerritsen: ->getReflectionClass fixes this actually 😄

Resources/views/Collector/db.html.twig
nickygerritsen: Yup the class will always be there anyway. Done.

ostrolucky (approved):

PR #1599, opened 2022-12-27
Make ServiceEntityRepository lazy on Symfony 6.2+

nicolas-grekas:
Best reviewed [ignoring whitespaces](https://github.com/doctrine/DoctrineBundle/pull/1599/files?w=1).
See https://github.com/symfony/symfony/issues/48666 for background.

PR #1611, opened 2023-01-06
Add generics types on ServiceEntityRepository using the LazyServiceEntityRepository

jdecool:
https://github.com/doctrine/DoctrineBundle/pull/1599 has introduced a new `ServiceEntityRepository` definition using the `LazyServiceEntityRepository`.

This new definition doesn't declare the generic type that cause problems for static analysis when this service implementation is used.

### DIFF
--- a/Repository/ServiceEntityRepository.php
+++ b/Repository/ServiceEntityRepository.php
@@ -11,6 +11,10 @@ use function sprintf;
 use function trait_exists;
 
 if (trait_exists(LazyGhostTrait::class)) {
+    /**
+     * @template T of object
+     * @template-extends LazyServiceEntityRepository<T>
+     */
     class ServiceEntityRepository extends LazyServiceEntityRepository
     {
     }


ostrolucky (approved):

stof:
@ostrolucky what do you think about releasing 2.8.2 with that fix, as it is a regression for anyone using SA tools ?

ostrolucky:
As it is I don't believe fix for SA alone warrants a release. 2.8.1 also isn't super important so users can hold off from updating to it. But let's see if this will get further feedback.

stof:
Well, the fix in https://github.com/doctrine/DoctrineBundle/pull/1599 actually fixes issues that happens depending on how you write the entity manager in your project. So it might actually be important depending on whether the project is impacted or no.

> As it is I don't believe fix for SA alone warrants a release

For projects that require SA to pass in their CI (just like Doctrine does for instance), a _regression_ in the types (the class not being generic anymore) can be a blocker.
As we have no other issues pending for 2.8.x, I don't see a benefit of delaying the release (getting duplicate reports for an already solved issue is a risk on the other hand)

ro0NL:
@jdecool, by any chance did you manage to make psalm recognize ServiceEntityRepository

all our repositories got wiped from the baseline :(

jdecool:
@ro0NL I don't use psalm (only phpstan).

This PR solve all issues with phpstan.

ro0NL:
@jdecool i figured :} cool :+1:

ro0NL:
actually i think nobody uses psalm in symfony/doctrine projects :joy:

MihailSerov:
this pr made issues https://github.com/psalm/psalm-plugin-doctrine/issues/145#issuecomment-1608140134

PR #1618, opened 2023-01-16
[DoctrineBundle] register  LockStoreSchemaSubscriber

alli83:
linked to PR symfony/symfony#48999

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -35,6 +35,7 @@ use Symfony\Bridge\Doctrine\DependencyInjection\AbstractDoctrineExtension;
 use Symfony\Bridge\Doctrine\IdGenerator\UlidGenerator;
 use Symfony\Bridge\Doctrine\IdGenerator\UuidGenerator;
 use Symfony\Bridge\Doctrine\PropertyInfo\DoctrineExtractor;
+use Symfony\Bridge\Doctrine\SchemaListener\LockStoreSchemaSubscriber;
 use Symfony\Bridge\Doctrine\SchemaListener\PdoCacheAdapterDoctrineSchemaSubscriber;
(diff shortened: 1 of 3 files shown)


DependencyInjection/DoctrineExtension.php
```diff
@@ -392,6 +392,10 @@ protected function ormLoad(array $config, ContainerBuilder $container)
             $container->removeDefinition('doctrine.orm.listeners.pdo_session_handler_schema_subscriber');
         }
 
+        if (! class_exists(LockStoreSchemaSubscriber::class)) {
```
ostrolucky, line 397: You are missing the import. Add a comment since which Symfony version is this available as well please.

DoctrineBundle.php
```diff
@@ -46,6 +46,10 @@ public function build(ContainerBuilder $container)
         $container->addCompilerPass(new class () implements CompilerPassInterface {
             public function process(ContainerBuilder $container): void
             {
+                if (! $container->findTaggedServiceIds('lock.store')) {
```
ostrolucky: I would rather not have this, the only call in that subscriber outside foreach is rather cheap.

ostrolucky (commented):
Can we have a test please?

DoctrineBundle.php
alli83: So if it doesn't cause an error, as it does for session.handler service, we should rather let it be loaded knowing that the component is not installed ? 

Thanks for you feedback !

nicolas-grekas: Part of the overhead in Symfony is listeners registered that don't do much so I'd prefer core bundles like this one do their cleanups. But I'm fine discussing this in a dedicated PR for all listeners at once, so let's remove this from the current PR.

alli83: ok !

PR #1621, opened 2023-01-18
Fix checking for lazy ghost entities

### LINKED ISSUE #1620: Wrong proxy classes check in AbstractManagerRegistry if lazy ghost is enabled
doctrine/orm creates proxy class implementations of various interfaces depending on whether the lazy ghost setting is enabled or not.
[ProxyFactory line 116](https://github.com/doctrine/orm/blob/2.14.x/lib/Doctrine/ORM/Proxy/ProxyFactory.php#L116)

doctrine/persistence checks a proxy class by its interface.
[AbstractManagerRegistry line 176](https://github.com/doctrine/persistence/blob/3.1.x/src/Persistence/AbstractManagerRegistry.php#L176)

doctrine/DoctrineBundle doesn't care and intializes doctrine/persistence with the old Proxy interface.
[Registry line 30](https://github.com/doctrine/DoctrineBundle/blob/2.8.x/Registry.php#L30)

Doctrine\Persistence\AbstractManagerRegistry->getManagerForClass($proxyClassName) returns null when lazy ghost is enabled.

nicolas-grekas:
Fix #1620

### DIFF
--- a/Registry.php
+++ b/Registry.php
@@ -4,7 +4,7 @@ namespace Doctrine\Bundle\DoctrineBundle;
 
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\ORMException;
-use Doctrine\ORM\Proxy\Proxy;
+use Doctrine\Persistence\Proxy;
 use ProxyManager\Proxy\LazyLoadingInterface;
 use Psr\Container\ContainerInterface;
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Failures unrelated.

stof (approved):

PR #1569, opened 2022-10-31
Register PdoSessionHandlerSchemaSubscriber

alli83:
link to  PR  https://github.com/symfony/symfony/pull/48059

PR #1623, opened 2023-01-24
Fix registering PdoSessionHandlerSchemaSubscriber

alli83:
linked to https://github.com/symfony/symfony/pull/49097

### DIFF
--- a/DependencyInjection/Compiler/CacheSchemaSubscriberPass.php
+++ b/DependencyInjection/Compiler/CacheSchemaSubscriberPass.php
@@ -7,7 +7,6 @@ use Symfony\Component\Cache\Adapter\PdoAdapter;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Reference;
-use Symfony\Component\HttpFoundation\Session\Storage\Handler\PdoSessionHandler;
 
 /**
  * Injects Doctrine DBAL and legacy PDO adapters into their schema subscribers.
@@ -25,10 +24,6 @@ class CacheSchemaSubscriberPass implements CompilerPassInterface
 
         // available in Symfony 5.1 and up to Symfony 5.4 (deprecated)
         $this->injectAdapters($container, 'doctrine.orm.listeners.pdo_cache_adapter_doctrine_schema_subscriber', PdoAdapter::class);
-
-        // available in Symfony 6.3 and higher
-        /** @psalm-suppress UndefinedClass */
-        $this->injectAdapters($container, 'doctrine.orm.listeners.pdo_session_handler_schema_subscriber', PdoSessionHandler::class);
     }
 
     private function injectAdapters(ContainerBuilder $container, string $subscriberId, string $class)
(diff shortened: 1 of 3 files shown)


Resources/config/orm.xml
```diff
@@ -145,7 +145,7 @@
             <tag name="doctrine.event_subscriber" />
         </service>
         <service id="doctrine.orm.listeners.pdo_session_handler_schema_subscriber" class="Symfony\Bridge\Doctrine\SchemaListener\PdoSessionHandlerSchemaSubscriber">
-            <argument type="collection" /> <!-- PdoSessionHandler instances -->
+            <argument type="tagged" tag="session.handler_abstract" /> <!-- AbstractSessionHandler instances -->
```
nicolas-grekas: `<argument type="service" id="session.handler" />`? but we should also ensure that the service exists, and remove the listener if not
maybe we can do this in DependencyInjection/Compiler/CacheSchemaSubscriberPass.php?

alli83: I don't think my approach is right in the CacheSchemaSubscriberPass to avoid the serviceNotFoundException:  when I remove the listener if the service doesn't exist I end up with 'doctrine.dbal.default_connection.event_manager' which always has a dependency on the listener and I receive the serviceNotFound exception because of the RegisterEventListenersAndSubscribersPass called before
Thanks for your feedback 🙏

DependencyInjection/Compiler/CacheSchemaSubscriberPass.php
```diff
@@ -26,9 +30,35 @@ public function process(ContainerBuilder $container)
...
+
+            $eventManager = $container->getDefinition('doctrine.dbal.default_connection.event_manager');
+
+            if (! isset($eventManager->getArguments()[1])) {
+                return;
+            }
+
+            $subscriberIds = $eventManager->getArguments()[1];
+
+            $key = array_search('doctrine.orm.listeners.pdo_session_handler_schema_subscriber', $subscriberIds);
```
nicolas-grekas: I think we reached the point where this deserves a dedicated compiler pass
here, we should remove the service before it is registered as a service, so that we don't have to unwire it as the next lines try to do for now

alli83: yep, i was thinking about a new compiler pass called before the RegisterEventListenersAndSubscribersPass ?

nicolas-grekas: RegisterPdoSessionHandlerSchemaSubscriberPass?

alli83: what I mean, if I'm not wrong, is that we should remove the service before the Symfony\Bridge\Doctrine\DependencyInjection\CompilerPass\RegisterEventListenersAndSubscriberPass replace the service container argument of the event manager ?

nicolas-grekas (approved):
Ready on my side (I updated the PR)

ostrolucky:
If this is a fix, shouldn't it go to 2.8.x?

alli83:
It's a fix for https://github.com/doctrine/DoctrineBundle/pull/1569. if I'm not wrong it has been merged on 2.9. I let @nicolas-grekas confirm it  :)

PR #1626, opened 2023-02-01
Fix: Link to issue

localheinz:
This pull request

- [x] adds a link to #673

### DIFF
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -120,7 +120,8 @@ class ConnectionFactory
      *
      * This could fail if types should be registered to an predefined/unused connection
      * and the platform version is unknown.
-     * For details have a look at DoctrineBundle issue #673.
+     *
+     * @link https://github.com/doctrine/DoctrineBundle/issues/673
      *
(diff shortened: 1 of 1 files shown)


ostrolucky:
I think it's too late to add reference to 5 year old BC break. Or what's the justification you see for adding this there?

ConnectionFactory.php
```diff
@@ -132,6 +132,8 @@ public function createConnection(array $params, ?Configuration $config = null, ?
      * and the platform version is unknown.
      * For details have a look at DoctrineBundle issue #673.
```
localheinz: @ostrolucky 

The reference is already here, just not clickable.

ostrolucky: Hmm I would say perhaps better then to change it to a link rather than adding a link additionally to ID

ostrolucky (approved):

dmaicher (approved):

localheinz:
Thank you, @dmaicher and @ostrolucky!

PR #1628, opened 2023-02-06
add precision for doctrine 2.6 and symfony 5.4

EmmanuelleOmni:
Configuration with service.yaml dones not works with doctrine/bundle 2.6, 2.7, 2.8 and symfony 5.4

### DIFF
--- a/Resources/doc/entity-listeners.rst
+++ b/Resources/doc/entity-listeners.rst
@@ -2,7 +2,7 @@ Entity Listeners
 ================
 
 Entity listeners that are services must be registered with the entity listener
-resolver. On top of the annotation in the entity class, you have to tag the
+resolver. Starting with doctrine/orm 2.6, on top of the annotation in the entity class, you have to tag the
 service with ``doctrine.orm.entity_listener`` for it to be automatically added
 to the resolver. Use the (optional) ``entity_manager`` attribute to specify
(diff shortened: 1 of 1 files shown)


stof:
Registering an EntityListener service does not remove the need to declare the EntityListener in the ORM mapping (through annotations or attributes) as that's required by Doctrine ORM.

However, defining a tagged service allows you to inject some dependencies in it instead of requiring a constructor without any arguments.

franckranaivo:
Indeed @stof , you're right we forgot this. 

In this case, telling in the doc 

> "On top of the annotation in the entity class, you have to tag the service with doctrine.orm.entity_listener" 

seems a bit confusing to me.

It should be (as you said) :

> "On top of the annotation in the entity class, prior to doctrine 2.5 you have to declare the Entity Listener" 

rather than mentionning the tag set in the yaml file which is useful if you add dependency to your listener.

Also, we should remove:

> "instead of registering the entity listener on the entity"

As it suggest that registering the entity Listener in the ORM mapping is optional. What do you think ?

Sorry, for the long comment as I don't know how to include code inside comments on github

ostrolucky:
Sorry but I don't think it makes sense to start mentioning doctrine/orm 2.6.x at this point, as such old version is unsupported for a long time

PR #1629, opened 2023-02-08
add priority argument to AsEntityListener attribute

Jontsa:
The `doctrine.orm.entity_listener` tag has optional `priority` attribute which is missing from `AsEntityListener`.

### DIFF
--- a/Attribute/AsEntityListener.php
+++ b/Attribute/AsEntityListener.php
@@ -12,6 +12,7 @@ class AsEntityListener
 {
     public function __construct(
         public ?string $event = null,
+        public ?int $priority = null,
         public ?string $method = null,
         public ?bool $lazy = null,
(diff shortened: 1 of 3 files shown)


Attribute/AsEntityListener.php
```diff
@@ -12,6 +12,7 @@ class AsEntityListener
 {
     public function __construct(
         public ?string $event = null,
+        public ?int $priority = null,
```
ostrolucky: needs to be moved as a last parameter, otherwise it's a BC break

Jontsa: Thank you. Moved as last parameter in constructor.

ostrolucky (approved):

stof (approved):

PR #1630, opened 2023-02-08
show doctrine deprecations with phpunit bridge

dmaicher:
WDYT about showing the doctrine deprecations?

```
Remaining indirect deprecation notices (7)

  1x: Not configuring a schema manager factory is deprecated. Use Doctrine\DBAL\Schema\DefaultSchemaManagerFactory which is going to be the default in DBAL 4. (Connection.php:225 called by DriverManager.php:193, https://github.com/doctrine/dbal/issues/5812, package doctrine/dbal)
    1x in CreateDatabaseDoctrineTest::testExecute from Doctrine\Bundle\DoctrineBundle\Tests\Command

  1x: SqliteSchemaManager::createDatabase() is deprecated. The engine will create the database file automatically. (SqliteSchemaManager.php:140 called by CreateDatabaseDoctrineCommand.php:85, https://github.com/doctrine/dbal/issues/4963, package doctrine/dbal)
    1x in CreateDatabaseDoctrineTest::testExecute from Doctrine\Bundle\DoctrineBundle\Tests\Command

  1x: The "platform" connection parameter is deprecated. Use a driver middleware that would instantiate the platform instead. (Connection.php:207 called by DriverManager.php:193, https://github.com/doctrine/dbal/pull/5699, package doctrine/dbal)
    1x in ContainerTest::testContainer from Doctrine\Bundle\DoctrineBundle\Tests

  1x: Using XML mapping driver with XSD validation disabled is deprecated and will not be supported in Doctrine ORM 3.0. (XmlDriver.php:65 called by SimplifiedXmlDriver.php:23, https://github.com/doctrine/orm/pull/6728, package doctrine/orm)
    1x in ContainerTest::testContainer from Doctrine\Bundle\DoctrineBundle\Tests

  1x: Passing an instance of Doctrine\ORM\Mapping\ClassMetadataInfo to Doctrine\ORM\Tools\SchemaValidator::validateClass is deprecated, please pass a ClassMetadata instance instead. (SchemaValidator.php:82 called by DoctrineDataCollector.php:116, https://github.com/doctrine/orm/pull/249, package doctrine/orm)
    1x in DoctrineDataCollectorTest::testCollectEntities from Doctrine\Bundle\DoctrineBundle\Tests\DataCollector

  1x: DebugStack is deprecated. (DebugStack.php:41 called by Generator.php:715, https://github.com/doctrine/dbal/pull/4967, package doctrine/dbal)
    1x in DoctrineDataCollectorTest::testGetGroupedQueries from Doctrine\Bundle\DoctrineBundle\Tests\DataCollector

  1x: Passing an instance of Doctrine\Common\Cache\Psr6\DoctrineProvider to Doctrine\ORM\Cache\DefaultCacheFactory::__construct is deprecated, pass a Psr\Cache\CacheItemPoolInterface instead. (DefaultCacheFactory.php:62 called by getDoctrine_Orm_DefaultEntityManagerService.php:36, https://github.com/doctrine/orm/pull/9322, package doctrine/orm)
    1x in CacheCompatibilityPassTest::testCacheConfigUsingServiceDefinedByApplication from Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection\Compiler

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: "ubuntu-20.04"
     env:
       SYMFONY_REQUIRE: ${{matrix.symfony-require}}
-      SYMFONY_DEPRECATIONS_HELPER: ${{matrix.symfony-deprecations-helper}}
+      #SYMFONY_DEPRECATIONS_HELPER: ${{matrix.symfony-deprecations-helper}}
 
     strategy:
       fail-fast: false
@@ -37,25 +37,25 @@ jobs:
           - dependencies: "lowest"
             stability: "stable"
             php-version: "7.4"
-            symfony-deprecations-helper: "weak"
+            #symfony-deprecations-helper: "weak"
 
           # Test LTS
           - symfony-require: "5.4.*"
             dependencies: "highest"
             php-version: "8.0"
-            symfony-deprecations-helper: "weak"
+            #symfony-deprecations-helper: "weak"
 
           # Test last supported minor version
           - symfony-require: "6.*"
             dependencies: "highest"
             php-version: "8.1"
-            symfony-deprecations-helper: "weak"
+            #symfony-deprecations-helper: "weak"
 
           # Bleeding edge
           - php-version: "8.1"
             dependencies: "highest"
             stability: "dev"
-            symfony-deprecations-helper: "weak"
+            #symfony-deprecations-helper: "weak"
 
     steps:
       - name: "Checkout"
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -15,7 +15,6 @@ jobs:
     runs-on: "ubuntu-20.04"
     env:
       SYMFONY_REQUIRE: ${{matrix.symfony-require}}
-      #SYMFONY_DEPRECATIONS_HELPER: ${{matrix.symfony-deprecations-helper}}
 
     strategy:
       fail-fast: false
@@ -37,25 +36,21 @@ jobs:
           - dependencies: "lowest"
             stability: "stable"
             php-version: "7.4"
-            #symfony-deprecations-helper: "weak"
 
           # Test LTS
           - symfony-require: "5.4.*"
             dependencies: "highest"
             php-version: "8.0"
-            #symfony-deprecations-helper: "weak"
 
           # Test last supported minor version
           - symfony-require: "6.*"
             dependencies: "highest"
             php-version: "8.1"
-            #symfony-deprecations-helper: "weak"
 
           # Bleeding edge
           - php-version: "8.1"
             dependencies: "highest"
             stability: "dev"
-            #symfony-deprecations-helper: "weak"
 
     steps:
       - name: "Checkout"
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -15,6 +15,7 @@ jobs:
     runs-on: "ubuntu-20.04"
     env:
       SYMFONY_REQUIRE: ${{matrix.symfony-require}}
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by stof** [comment]
This looks like a good idea.

File: .github/workflows/test-dev-stability.yml [hunk 1]
```diff
@@ -52,4 +52,4 @@ jobs:
           composer-options: "--prefer-dist"
 
       - name: "Run PHPUnit"
-        run: "SYMFONY_DEPRECATIONS_HELPER=${{env.GITHUB_WORKSPACE}}/tests/baseline-ignore vendor/bin/phpunit"
```
ostrolucky: Purpose of this workflow is to alert us on upcoming deprecations by failing the build, instead of failing the main build and blocking our pipeline (often with deprecations that are not possible to fix by us).  Adding deprecations coming from doctrine/deprecations in this PR is fine, but changing the main workflow so that it fails the main build when there are deprecations will bring back the old problem. Rationale of all this is coming from https://github.com/orgs/doctrine/teams/doctrinecore/discussions/13?from_comment=5 People opening a PR shouldn't have their builds failed just because day before DBAL or some other package released a new version - that's completely unrelated to their changes they want to do in PR. Also, this cronjob is not limited to self and direct deprecations.

dmaicher: makes sense :+1: Will adjust it

dmaicher: Should this workflow then fail due to those indirect doctrine deprecations?

ostrolucky: Yeah I think it will fail if we start counting doctrine/deprecations and not fix them. But it's ok temporarily until somebody tries to fix those.

phpunit.xml.dist
```diff
@@ -1,14 +1,14 @@
...
+<phpunit colors="true" bootstrap="Tests/bootstrap.php">
     <testsuites>
         <testsuite name="DoctrineBundle for the Symfony Framework">
             <directory>./Tests</directory>
         </testsuite>
     </testsuites>
 
     <php>
-        <env name="SYMFONY_DEPRECATIONS_HELPER" value="ignoreFile=./Tests/baseline-ignore"/>
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[indirect]=9999&amp;ignoreFile=./Tests/baseline-ignore"/>
```
dmaicher: @ostrolucky is this fine? So the main CI build will never fail due to deprecations now

dmaicher: Or I can restore the behavior with setting it to `weak`? don't mind

ostrolucky: I am not sure what's the difference between these, but if you believe there is none then yeah keep it. `weak` is no longer mentioned in SF docs, it's something legacy I guess.

.github/workflows/continuous-integration.yml
```diff
@@ -15,7 +15,6 @@ jobs:
     runs-on: "ubuntu-20.04"
     env:
       SYMFONY_REQUIRE: ${{matrix.symfony-require}}
-      SYMFONY_DEPRECATIONS_HELPER: ${{matrix.symfony-deprecations-helper}}
```
ostrolucky, line 18: I think reason why we didn't simply put it to phpunit.xml but here only was that even if we don't want CI build to fail, we want tests to fail locally at least, so that people writing the PR are aware they might be introducing a new deprecation. I would say we should stick to that.

ostrolucky (approved):

ostrolucky (changes requested):
hmm looks like it fails the build now

dmaicher:
yeah pushed again after you approved. Should be good now

ostrolucky (approved):

ostrolucky:
thx :)

PR #1632, opened 2023-02-11
allow configuring schema manager factory

dmaicher:
Will get rid of this deprecation

```
  1x: Not configuring a schema manager factory is deprecated. Use Doctrine\DBAL\Schema\DefaultSchemaManagerFactory which is going to be the default in DBAL 4. (Connection.php:225 called by DriverManager.php:193, https://github.com/doctrine/dbal/issues/5812, package doctrine/dbal)
```

Also see https://github.com/doctrine/dbal/pull/5876

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -195,6 +195,9 @@ class Configuration implements ConfigurationInterface
                     ->useAttributeAsKey('name')
                     ->prototype('scalar')->end()
                 ->end()
+                ->scalarNode('schema_manager_factory')
+                    ->defaultValue('doctrine.dbal.default_schema_manager_factory')
+                ->end()
             ->end();
 
         // dbal < 2.11
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -196,8 +196,7 @@ class Configuration implements ConfigurationInterface
                     ->prototype('scalar')->end()
                 ->end()
                 ->scalarNode('schema_manager_factory')
-                    ->defaultValue('doctrine.dbal.legacy_schema_manager_factory')
-                    ->cannotBeEmpty()
+                    ->defaultNull()
                     ->end()
                 ->end()
             ->end();
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -196,7 +196,9 @@ class Configuration implements ConfigurationInterface
                     ->prototype('scalar')->end()
                 ->end()
                 ->scalarNode('schema_manager_factory')
-                    ->defaultValue('doctrine.dbal.default_schema_manager_factory')
+                    ->defaultValue('doctrine.dbal.legacy_schema_manager_factory')
+                    ->cannotBeEmpty()
+                    ->end()
                 ->end()
             ->end();
 
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -3,6 +3,7 @@
 namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection;
 
 use Doctrine\Common\Proxy\AbstractProxyFactory;
+use Doctrine\DBAL\Schema\LegacySchemaManagerFactory;
 use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\EntityRepository;
 use Doctrine\ORM\Mapping\ClassMetadataFactory;
@@ -196,7 +197,8 @@ class Configuration implements ConfigurationInterface
                     ->prototype('scalar')->end()
                 ->end()
                 ->scalarNode('schema_manager_factory')
-                    ->defaultNull()
+                    ->cannotBeEmpty()
+                    ->defaultValue($this->getDefaultSchemaManagerFactory())
                 ->end()
             ->end();
 
@@ -789,5 +791,14 @@ class Configuration implements ConfigurationInterface
             'names' => $namesArray,
             'values' => $valuesArray,
         ];
+    }
+
+    private function getDefaultSchemaManagerFactory(): string
+    {
+        if (class_exists(LegacySchemaManagerFactory::class)) {
+            return 'doctrine.dbal.legacy_schema_manager_factory';
+        }
+
+        return 'doctrine.dbal.default_schema_manager_factory';
     }
 }
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -197,7 +197,6 @@ class Configuration implements ConfigurationInterface
                 ->end()
                 ->scalarNode('schema_manager_factory')
                     ->defaultNull()
-                    ->end()
                 ->end()
             ->end();
 
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -242,6 +242,8 @@ class DoctrineExtension extends AbstractDoctrineExtension
             ManagerRegistryAwareConnectionProvider::class,
             new Definition(ManagerRegistryAwareConnectionProvider::class, [$container->getDefinition('doctrine')])
         );
+
+        $configuration->addMethodCall('setSchemaManagerFactory', [new Reference($connection['schema_manager_factory'])]);
     }
 
     /**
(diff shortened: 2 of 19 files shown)


DependencyInjection/Configuration.php
```diff
@@ -195,6 +195,9 @@ private function getDbalConnectionsNode(): ArrayNodeDefinition
                     ->useAttributeAsKey('name')
                     ->prototype('scalar')->end()
                 ->end()
+                ->scalarNode('schema_manager_factory')
+                    ->defaultValue('doctrine.dbal.default_schema_manager_factory')
```
dmaicher: @derrabus do you think its a good idea to use `Doctrine\DBAL\Schema\DefaultSchemaManagerFactory` by default?

Or should we make this opt-in?

ostrolucky (approved):

DependencyInjection/Configuration.php
derrabus: This will break apps that:

* Use a custom DBAL driver.
* Override the schema manager via a custom platform class.

I'll let you decide whether this is acceptable. 🙂

stof: Given that the DefaultSchemaManagerFactory is a BC-breaking solution, it should not be the default in DoctrineBundle for now (unless DBAL 4.x is used).

I suggest having a simple boolean flag to opt-in for the new default, as done in several places in Symfony for such migrations (see how SecurityBundle 5.3 introduced  a boolean flag to opt-in for the new security system).

The behavior of that flag should be like that:

- when using DBAL 3.6+, trigger a deprecation warning if the flag is not set or is set to using the legacy behavior
- when using DBAL 4+, throw an exception if the flag is set to `false`

and then, make a PR on the symfony/recipes repo configuring the flag to use the new behavior, so that new projects start with the new behavior.

ostrolucky: I don't see why it should be a boolean. We can make it use legacy factory by default if it exist, otherwise default factory.

dmaicher: @ostrolucky yeah I was also thinking to keep the setting but using legacy as default + set the new default factory then via recipe

stof: well, if we want to support custom factories instead of only the legacy one and the new one, this can indeed be done with a non-boolean config

derrabus: The possibility of custom factories is kinda the whole point of this new setting: it enables the app to override the schema manager.

Resources/config/dbal.xml
```diff
@@ -97,5 +97,8 @@
             <tag name="controller.service_arguments" />
         </service>
 
+        <service id="doctrine.dbal.default_schema_manager_factory" class="Doctrine\DBAL\Schema\DefaultSchemaManagerFactory" />
+        <service id="doctrine.dbal.legacy_schema_manager_factory" class="Doctrine\DBAL\Schema\LegacySchemaManagerFactory" />
```
dmaicher, line 101: todo: deprecate this service

DependencyInjection/DoctrineExtension.php
```diff
@@ -242,6 +243,19 @@ protected function loadDbalConnection($name, array $connection, ContainerBuilder
             ManagerRegistryAwareConnectionProvider::class,
             new Definition(ManagerRegistryAwareConnectionProvider::class, [$container->getDefinition('doctrine')])
         );
+
+        $schemaManagerFactoryId = $connection['schema_manager_factory'];
+        if ($schemaManagerFactoryId === 'doctrine.dbal.legacy_schema_manager_factory') {
+            trigger_deprecation(
```
dmaicher: any idea how to easily/centrally fix or silence this deprecation for the testsuite?

```
Remaining self deprecation notices (170)

  170x: Since doctrine/doctrine-bundle 2.9: Not configuring schema_manager_factory is deprecated. Use "doctrine.dbal.default_schema_manager_factory" or a custom factory instead. See UPGRADE-2.9.md for details.
    6x in DoctrineExtensionTest::testAutomapping from Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection
    6x in XmlDoctrineExtensionTest::testDbalLoadFromXmlMultipleConnections from Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection
    6x in YamlDoctrineExtensionTest::testDbalLoadFromXmlMultipleConnections from Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection
    4x in MiddlewarePassTest::testAddMiddlewareWithExplicitTag from Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection\Compiler
    ...

```

I don't really want to change all the configuration fixtures and set `schema_manager_factory = doctrine.dbal.default_schema_manager_factory` :see_no_evil:

dmaicher: Or is deprecating the service enough actually? :thinking:

derrabus: > I don't really want to change all the configuration fixtures and set `schema_manager_factory = doctrine.dbal.default_schema_manager_factory` 🙈

As soon as this bundle starts to support DBAL 4, this has to be done anyway. So why not do it now? (sorry)

ostrolucky: I don't see a need to trigger deprecation there actually. I don't see why users must specify schema manager explicitly.

dmaicher: Ok so currently there is no default for this config. So it would be null. Which is deprecated in DBAL 3.6+. Also using the legacy factory is deprecated and not working with the upcoming DBAL 4.

So we should encourage users to use the default factory, right?

ostrolucky: Why not class_exists(LegacyFactory::class) ? LegacyFactory::class : DefaultFactory like I suggested? This way nothing breaks neither with DBAL 3 nor 4, deprecation will still be there but will come from DBAL and not doctrine bundle and users can get rid of it.

dmaicher: Ok let me have a look at implementing it that way.

stof: @ostrolucky migrating from the legacy factory to the default factory is a BC break. And the DefaultFactory exists in DBAL 3.6 (while not being the default yet). We cannot do a class_exists check to use the default factory as that would break things for DBAL 3 users. We need to have an opt-in.

However, if DBAL 4 is used, the bundle should indeed default to using the DefaultFactory.

ostrolucky: class_exists(LegacyFactory::class) ? LegacyFactory::class : DefaultFactory is an opt in. It still defaults to legacyfactory so nothing changes.

stof: this is not an opt-in. It forces users of DBAL 3.6 to stay on the legacy behavior (which will trigger a deprecation warning) instead of being able to intentionally choose to upgrade to the new behavior (without waiting for the DBAL 4 upgrade itself).

ostrolucky: I was just talking about default value for `schema_manager_factory` bundle option. Option will still be there and users can change it to default factory. When I said I don't see the need why users have to specify schema_manager_factory, I didn't mean we should drop the option, only that it's fine if they don't specify which factory explicitly.

DependencyInjection/Configuration.php
```diff
@@ -787,4 +792,13 @@ private function getAutoGenerateModes(): array
             'values' => $valuesArray,
         ];
     }
+
+    private function getDefaultSchemaManagerFactory(): string
+    {
+        if (class_exists(LegacySchemaManagerFactory::class)) {
+            return 'doctrine.dbal.legacy_schema_manager_factory';
```
dmaicher, line 799: is this what you had in mind @ostrolucky ?

UPGRADE-2.9.md
```diff
@@ -0,0 +1,32 @@
...
+=======================
+
+Configuration
+--------
+
+### Schema manager factory
+
+Due to changes and deprecations on `doctrine/dbal` 3.6+ we introduced a new option to configure a Schema manager factory.
+
+Not setting this option is deprecated and won't work with the upcoming DBAL 4 anymore.
```
dmaicher: todo: needs update

DependencyInjection/Configuration.php
ostrolucky: yes

ostrolucky (approved):

PR #1633, opened 2023-02-27
improve DoctrineDataCollectorTest and fix some deprecations

dmaicher:
Fixes

```
Remaining self deprecation notices (1)

  1x: Method "Symfony\Bridge\Doctrine\DataCollector\DoctrineDataCollector::collect()" might add "void" as a native return type declaration in the future. Do the same in child class "Doctrine\Bundle\DoctrineBundle\DataCollector\DoctrineDataCollector" now to avoid errors or add an explicit @return annotation to suppress this message.
    1x in DoctrineDataCollectorTest::testCollectEntities from Doctrine\Bundle\DoctrineBundle\Tests\DataCollector

Remaining indirect deprecation notices (2)

  1x: Passing an instance of Doctrine\ORM\Mapping\ClassMetadataInfo to Doctrine\ORM\Tools\SchemaValidator::validateClass is deprecated, please pass a ClassMetadata instance instead. (SchemaValidator.php:82 called by DoctrineDataCollector.php:116, https://github.com/doctrine/orm/pull/249, package doctrine/orm)
    1x in DoctrineDataCollectorTest::testCollectEntities from Doctrine\Bundle\DoctrineBundle\Tests\DataCollector

  1x: DebugStack is deprecated. (DebugStack.php:41 called by Generator.php:715, https://github.com/doctrine/dbal/pull/4967, package doctrine/dbal)
    1x in DoctrineDataCollectorTest::testGetGroupedQueries from Doctrine\Bundle\DoctrineBundle\Tests\DataCollector
```

### DIFF
--- a/DataCollector/DoctrineDataCollector.php
+++ b/DataCollector/DoctrineDataCollector.php
@@ -73,10 +73,7 @@ class DoctrineDataCollector extends BaseCollector
         }
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function collect(Request $request, Response $response, ?Throwable $exception = null)
+    public function collect(Request $request, Response $response, ?Throwable $exception = null): void
     {
         parent::collect($request, $response, $exception);
 
--- a/Tests/DataCollector/DoctrineDataCollectorTest.php
+++ b/Tests/DataCollector/DoctrineDataCollectorTest.php
@@ -87,7 +87,16 @@ class DoctrineDataCollectorTest extends TestCase
     {
         $debugDataHolder = new DebugDataHolder();
 
+        /**
+         * @psalm-suppress InternalClass
+         * @psalm-suppress InternalMethod
+         */
         $debugDataHolder->addQuery('default', new Query('SELECT * FROM foo WHERE bar = :bar'));
+
+        /**
+         * @psalm-suppress InternalClass
+         * @psalm-suppress InternalMethod
+         */
         $debugDataHolder->addQuery('default', new Query('SELECT * FROM foo WHERE bar = :bar'));
 
         $collector = $this->createCollector([], true, $debugDataHolder);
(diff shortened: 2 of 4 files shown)


psalm.xml.dist
```diff
@@ -75,5 +75,15 @@
                 <file name="Repository/ServiceEntityRepository.php"/>
             </errorLevel>
         </DuplicateClass>
+        <InternalClass>
```
dmaicher: needed for using `Symfony\Bridge\Doctrine\Middleware\Debug\Query` in the new test case I added. Or I can remove the test again? :man_shrugging:

ostrolucky: Add a psalm-suppress in an offending code instead please

Tests/DataCollector/DoctrineDataCollectorTest.php
```diff
@@ -82,6 +84,42 @@ public function testDoesNotCollectEntities(): void
...
 
     public function testGetGroupedQueries(): void
+    {
+        $debugDataHolder = new DebugDataHolder();
+
+        /**
+         * @psalm-suppress InternalClass
+         * @psalm-suppress InternalMethod
+         */
+        $debugDataHolder->addQuery('default', new Query('SELECT * FROM foo WHERE bar = :bar'));
```
ostrolucky: Sorry I was writing as I was thinking, you are way too fast :D Now I got the idea that we could avoid using Query class here by mocking `DebugDataHolder::getData`, right?

dmaicher: Ah yeah we could also mock it... true! I thought it was final but actually its not :smile:

DataCollector/DoctrineDataCollector.php
```diff
@@ -73,10 +73,7 @@ public function __construct(ManagerRegistry $registry, bool $shouldValidateSchem
         }
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function collect(Request $request, Response $response, ?Throwable $exception = null)
+    public function collect(Request $request, Response $response, ?Throwable $exception = null): void
```
dmaicher, line 76: is this actually a BC break if someone extended this class (for whatever reason)? :thinking:

ostrolucky: Yes, why did you decide to do it? I mean I am not opposed, we don't have same BC promise as Symfony, but it's strange that this return type was adding seemingly arbitrarily while it wasn't done in other methods.

dmaicher: Yeah did it just to fix the reported deprecation. I can also revert this change then

ostrolucky (approved):

DataCollector/DoctrineDataCollector.php
dmaicher: Ok then let's keep it as I think the chances are low that someone extended this class

PR #1634, opened 2023-02-28
Fix all deprecations

dmaicher:
All deprecations are fixed now :blush:


dmaicher:
@greg0ire I might need some help to figure out how to resolve this one

```
  1x: Using XML mapping driver with XSD validation disabled is deprecated and will not be supported in Doctrine ORM 3.0. (XmlDriver.php:65 called by SimplifiedXmlDriver.php:23, https://github.com/doctrine/orm/pull/6728, package doctrine/orm)
    1x in ContainerTest::testContainer from Doctrine\Bundle\DoctrineBundle\Tests
```

I had a look at https://github.com/doctrine/orm/pull/6728 and figured out that we seem to be using https://github.com/doctrine/orm/blob/2.14.x/lib/Doctrine/ORM/Mapping/Driver/SimplifiedXmlDriver.php#L23

So I cannot really get rid of this deprecation easily, or?

stof:
Well, looks like the ORM forgot to update the constructor of the SimplifiedXmlDriver to expose the boolean argument added in the parent

DoctrineBundle.php
```diff
@@ -36,9 +36,7 @@ class DoctrineBundle extends Bundle
 {
     private ?Closure $autoloader = null;
 
-    /**
-     * {@inheritDoc}
-     */
+    /** @return void */
```
stof, line 40: I suggest also tagging the class as `@final since 2.9`. If we were already trating our bundle class as final, we would add actual return types (and there is no valid use case to extend the bundle class IMO)

stof: btw, we should also do it for the DI extension and all compiler passes

greg0ire:
@Kern046 can you please take a look at this simplified xml driver stuff? I'm AFK now and for a few more days I think.

DependencyInjection/Compiler/CacheCompatibilityPass.php
```diff
@@ -59,7 +59,7 @@ private function updateSecondLevelCache(ContainerBuilder $container, Definition
             $factoryDefinition = $methodCall[1][0];
             assert($factoryDefinition instanceof Definition);
             $aliasId = (string) $factoryDefinition->getArgument(1);
-            $this->wrapIfNecessary($container, $aliasId, (string) $container->getAlias($aliasId), false);
+            $this->wrapIfNecessary($container, $aliasId, (string) $container->getAlias($aliasId), true);
```
dmaicher: second level also supports PSR6 caches since ORM 2.11 (which we require). See https://github.com/doctrine/orm/pull/9322

Tests/TestCase.php
```diff
@@ -46,9 +44,8 @@ public function createXmlBundleTestContainer(): ContainerBuilder
                 'dbal' => [
                     'connections' => [
                         'default' => [
-                            'driver' => 'pdo_mysql',
+                            'driver' => 'pdo_sqlite',
                             'charset' => 'UTF8',
-                            'platform-service' => 'my.platform',
```
dmaicher, line 51: is it important to cover this case with a test?

See

```
  1x: The "platform" connection parameter is deprecated. Use a driver middleware that would instantiate the platform instead. (Connection.php:207 called by DriverManager.php:193, https://github.com/doctrine/dbal/pull/5699, package doctrine/dbal)
    1x in ContainerTest::testContainer from Doctrine\Bundle\DoctrineBundle\Tests
```
    
This is completely deprecated

dmaicher: I guess we should also deprecate the `platform_service` option?

Edit: done

DependencyInjection/Compiler/CacheCompatibilityPass.php
```diff
@@ -82,48 +80,40 @@ private function updateSecondLevelCache(ContainerBuilder $container, Definition
...
+    private function createCompatibilityLayerDefinition(ContainerBuilder $container, string $definitionId): ?Definition
     {
         $definition = $container->getDefinition($definitionId);
 
         while (! $definition->getClass() && $definition instanceof ChildDefinition) {
             $definition = $container->findDefinition($definition->getParent());
         }
 
-        if ($shouldBePsr6 === is_a($definition->getClass(), CacheItemPoolInterface::class, true)) {
+        if (is_a($definition->getClass(), CacheItemPoolInterface::class, true) === true) {
```
stof: ` === true` should be removed. `is_a` already returns a boolean

dmaicher: yeah that makes no sense :+1:

DependencyInjection/Compiler/XmlMappingDriverPass.php
```diff
@@ -0,0 +1,53 @@
...
+
+use Doctrine\ORM\Mapping\Driver\SimplifiedXmlDriver;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+use function array_keys;
+use function count;
+use function sprintf;
+
+final class XmlMappingDriverPass implements CompilerPassInterface
```
dmaicher: todo: check if we can do this within our extension instead. We are extending Symfonys `AbstractDoctrineExtension` so it should be doable

ostrolucky: And in any case we should put in a TODO comment to drop this once we bump ORM version

stof: and we should tag the class as `@internal` to be able to drop it without BC break.

DependencyInjection/DoctrineExtension.php
```diff
@@ -1123,4 +1132,32 @@ private function useMiddlewaresIfAvailable(
                 ->addTag('doctrine.middleware', ['connection' => $connName]);
         }
     }
+
+    /**
+     * TODO: remove this once we dropped support for ORM < 3.0
```
ostrolucky, line 1137:
```suggestion
     * TODO: remove this once we drop support for ORM < 2.14
```

Because 3rd argument is supported since 2.14. So once we start requiring it, then we don't need conditional logic and we can hardcode last argument.

Btw do we need to put it here instead of to `\Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\DoctrineOrmMappingsPass::createXmlMappingDriver`?

dmaicher: But with ORM 3 we don't need to hardcode anything, or? Then the validation is always enabled I believe. I mean then we can drop the whole method.

DependencyInjection/DoctrineExtension.php
```diff
@@ -1123,4 +1132,32 @@ private function useMiddlewaresIfAvailable(
...
+        }
+
+        if ($numberOfArgs < 2) {
+            $args[] = SimplifiedXmlDriver::DEFAULT_FILE_EXTENSION;
+        }
+
+        // enable validation
+        $args[] = true;
+
+        $xmlDriverDef->setArguments($args);
```
ostrolucky, line 1161: @stof can we move this to symfony/doctrine-bridge instead? I mean that bridge is the one that specifies constructor args there. I am just not sure if symfony folks are ok to enable strict xml validation with symfony 6.3. If not, we will do it here but end result will be same.

ostrolucky (approved):
Let's be pragmatic, bite a bullet and merge this as is instead of moving code around

DependencyInjection/Configuration.php
```diff
@@ -156,7 +158,13 @@ private function getDbalConnectionsNode(): ArrayNodeDefinition
             ->fixXmlConfig('default_table_option')
             ->children()
                 ->scalarNode('driver')->defaultValue('pdo_mysql')->end()
-                ->scalarNode('platform_service')->end()
+                ->scalarNode('platform_service')
+                    ->setDeprecated(
+                        'doctrine/doctrine-bundle',
+                        '2.9',
+                        'The "platform_service" configuration key is deprecated since doctrine-bundle 2.9. DBAL 4 will not support setting a custom platform via connection params anymore.'
```
lyrixx, line 165: Hello, what is the new way to change the platform ?

stof: drivers are responsible for providing the platform. So using a driver middleware (the main extension point of DBAL) is probably the way to go for your case.

dmaicher: I think you can achieve this with a custom DBAL driver middleware

PR #1635, opened 2023-03-10
[DoctrineBundle] feat: doctrine schema subscribers as listeners

alli83:
linked to PR https://github.com/symfony/symfony/pull/49610

Doctrine schema listeners are deprecated and changed as listeners

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -36,9 +36,12 @@ use Symfony\Bridge\Doctrine\DependencyInjection\AbstractDoctrineExtension;
 use Symfony\Bridge\Doctrine\IdGenerator\UlidGenerator;
 use Symfony\Bridge\Doctrine\IdGenerator\UuidGenerator;
 use Symfony\Bridge\Doctrine\PropertyInfo\DoctrineExtractor;
-use Symfony\Bridge\Doctrine\SchemaListener\LockStoreSchemaSubscriber;
+use Symfony\Bridge\Doctrine\SchemaListener\DoctrineDbalCacheAdapterSchemaListener;
+use Symfony\Bridge\Doctrine\SchemaListener\LockStoreSchemaListener;
+use Symfony\Bridge\Doctrine\SchemaListener\MessengerTransportDoctrineSchemaListener;
 use Symfony\Bridge\Doctrine\SchemaListener\PdoCacheAdapterDoctrineSchemaSubscriber;
-use Symfony\Bridge\Doctrine\SchemaListener\PdoSessionHandlerSchemaSubscriber;
+use Symfony\Bridge\Doctrine\SchemaListener\PdoSessionHandlerSchemaListener;
+use Symfony\Bridge\Doctrine\SchemaListener\RememberMeTokenProviderDoctrineSchemaListener;
 use Symfony\Bridge\Doctrine\Validator\DoctrineLoader;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
 use Symfony\Component\Cache\Adapter\PhpArrayAdapter;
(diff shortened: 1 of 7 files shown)


alli83:
It seems the dependencies installed for psalm check are  Symfony 5.4. It could explained why the class `DoctrineDbalCacheAdapterSchemaListener`  is not found ? 

psalm.xml.dist updated in order to suppress error

nicolas-grekas (approved):

DependencyInjection/DoctrineExtension.php
```diff
@@ -394,18 +397,36 @@ protected function ormLoad(array $config, ContainerBuilder $container)
             $container->removeAlias('doctrine.orm.metadata.annotation_reader');
         }
 
+        // available in Symfony 6.3
+        if (! class_exists(DoctrineDbalCacheAdapterSchemaListener::class)) {
+            $container->removeDefinition('doctrine.orm.listeners.doctrine_dbal_cache_adapter_schema_listener');
+        }
+
+        if (class_exists(DoctrineDbalCacheAdapterSchemaListener::class)) {
```
ostrolucky: In all of the new 3 conditionals you added in this file you are removing the definition both when class does not exist and when class exists, check condition above and please add tests that actually cover this mistake.

psalm.xml.dist
```diff
@@ -53,6 +53,7 @@
                 <referencedClass name="Symfony\Bridge\Doctrine\Attribute\MapEntity"/>
                 <referencedClass name="Symfony\Component\VarExporter\LazyGhostTrait"/>
                 <referencedClass name="Symfony\Component\VarExporter\LazyObjectInterface"/>
+                <referencedClass name="Symfony\Bridge\Doctrine\SchemaListener\DoctrineDbalCacheAdapterSchemaListener"/>
```
ostrolucky: Replace with `@psalm-suppress`. We add exclusion rules to global config only when inline suppression does not work.

ostrolucky (changes requested):

DependencyInjection/DoctrineExtension.php
nicolas-grekas: it's not the same definition: see suffix, listener in one case, subscriber in the other

ostrolucky: I'm confused, I don't see suffix difference

image

nicolas-grekas: ah, it's in the service name. The double "if" could be replaced by an if/else if that's more clear. It looks correct to me: if the class exists, remove A, else remove B, so that only one of (listener, subscriber) is registered.

ostrolucky: Since we are always removing a definition, let's replace with inline condition to determine service ID

psalm.xml.dist
alli83: ok !

alli83:
updated

DependencyInjection/DoctrineExtension.php
```diff
@@ -394,18 +397,28 @@ protected function ormLoad(array $config, ContainerBuilder $container)
             $container->removeAlias('doctrine.orm.metadata.annotation_reader');
         }
 
+        // available in Symfony 6.3
+        class_exists(DoctrineDbalCacheAdapterSchemaListener::class) ?
```
nicolas-grekas: `$container->removeDefinition('doctrine.orm.listeners.doctrine_dbal_cache_adapter_schema_' . (class_exists(DoctrineDbalCacheAdapterSchemaListener::class) ? 'listener' : 'subscriber));`
and let's do this also for pre-existing lines below

alli83: ok ! if I'm not mistaken, there is just the case for the 3 new listeners. I'll double check

nicolas-grekas (approved):

ostrolucky (approved):

PR #1646, opened 2023-04-14
update deprecation baseline

dmaicher:
The baseline was outdated and used the wrong syntax on the scheduled job


stof:
On CI, the syntax was wrong. It misses the `ignoreFile=` prefix (which is there in the config file defining a fallback)

ostrolucky:
Indeed. Let's remove it from there. And instead of removing baseline, we should update it for this deprecation, until we fix it. Otherwise our cronjob will keep failing and we won't notice new issues.

ostrolucky (approved):

PR #1648, opened 2023-04-14
Fix configuration inheritance to not override default connection/EM values

### LINKED ISSUE #1337: `default_entity_manager` config param is not inherited from parent config file

tucksaun:
Hi there!

I recently unveiled a weird behavior when doing some cleanup on a project I work on.
Long story short, we have several connections, use a different name than `default` for the main one, and because we have config files defining only `doctrine.dbal.types`, the `default_connection` ends up being reset to `default`. 

While trying to fix this I realized this is similar to #1337 (just another section of the config really), so I try to fix #1337 at the same time (therefore being an alternative to #1356).

The true nature of the fix is mostly not to run the custom pre-normalization process but then this makes the default empty configuration not work anymore.
So we move the side effect of this process (creating an empty 'default' connection configuration) out in the Extension and we rerun the normalization process again so that the default values for connections and entity manager get resolved properly.

This should be considered a bugfix, but I'm not really sure which branch to target in that case, let me know if you want me to rebase and target another branch instead

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -14,6 +14,7 @@ use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
 use Symfony\Component\DependencyInjection\Exception\LogicException;
 
+use function array_diff_key;
 use function array_intersect_key;
 use function array_key_exists;
 use function array_keys;
@@ -72,17 +73,27 @@ class Configuration implements ConfigurationInterface
      */
     private function addDbalSection(ArrayNodeDefinition $node): void
     {
+        // Key that should not be rewritten to the connection config
+        $excludedKeys = ['default_connection' => true, 'types' => true, 'type' => true];
+
         $node
             ->children()
             ->arrayNode('dbal')
                 ->beforeNormalization()
-                    ->ifTrue(static function ($v) {
-                        return is_array($v) && ! array_key_exists('connections', $v) && ! array_key_exists('connection', $v);
+                    ->ifTrue(static function ($v) use ($excludedKeys) {
+                        if (! is_array($v)) {
+                            return false;
+                        }
+
+                        if (array_key_exists('connections', $v) || array_key_exists('connection', $v)) {
+                            return false;
+                        }
+
+                        // Is there actually anything to use once excluded keys are considered?
+                        return (bool) array_diff_key($v, $excludedKeys);
                     })
-                    ->then(static function ($v) {
-                        // Key that should not be rewritten to the connection config
-                        $excludedKeys = ['default_connection' => true, 'types' => true, 'type' => true];
-                        $connection   = [];
+                    ->then(static function ($v) use ($excludedKeys) {
+                        $connection = [];
                         foreach ($v as $key => $value) {
                             if (isset($excludedKeys[$key])) {
                                 continue;
@@ -92,8 +103,7 @@ class Configuration implements ConfigurationInterface
                             unset($v[$key]);
                         }
 
-                        $v['default_connection'] = isset($v['default_connection']) ? (string) $v['default_connection'] : 'default';
-                        $v['connections']        = [$v['default_connection'] => $connection];
+                        $v['connections'] = [($v['default_connection'] ?? 'default') => $connection];
 
                         return $v;
                     })
@@ -380,30 +390,39 @@ class Configuration implements ConfigurationInterface
      */
     private function addOrmSection(ArrayNodeDefinition $node): void
     {
+        // Key that should not be rewritten to the entity-manager config
+        $excludedKeys = [
+            'default_entity_manager' => true,
+            'auto_generate_proxy_classes' => true,
+            'enable_lazy_ghost_objects' => true,
+            'proxy_dir' => true,
+            'proxy_namespace' => true,
+            'resolve_target_entities' => true,
+            'resolve_target_entity' => true,
+            'controller_resolver' => true,
+        ];
+
         $node
             ->children()
                 ->arrayNode('orm')
                     ->beforeNormalization()
-                        ->ifTrue(static function ($v) {
+                        ->ifTrue(static function ($v) use ($excludedKeys) {
                             if (! empty($v) && ! class_exists(EntityManager::class)) {
                                 throw new LogicException('The doctrine/orm package is required when the doctrine.orm config is set.');
                             }
 
-                            return $v === null || (is_array($v) && ! array_key_exists('entity_managers', $v) && ! array_key_exists('entity_manager', $v));
+                            if (! is_array($v)) {
+                                return false;
+                            }
+
+                            if (array_key_exists('entity_managers', $v) || array_key_exists('entity_manager', $v)) {
+                                return false;
+                            }
+
+                            // Is there actually anything to use once excluded keys are considered?
+                            return (bool) array_diff_key($v, $excludedKeys);
                         })
-                        ->then(static function ($v) {
-                            $v = (array) $v;
-                            // Key that should not be rewritten to the entity-manager config
-                            $excludedKeys  = [
-                                'default_entity_manager' => true,
-                                'auto_generate_proxy_classes' => true,
-                                'enable_lazy_ghost_objects' => true,
-                                'proxy_dir' => true,
-                                'proxy_namespace' => true,
-                                'resolve_target_entities' => true,
-                                'resolve_target_entity' => true,
-                                'controller_resolver' => true,
-                            ];
+                        ->then(static function ($v) use ($excludedKeys) {
                             $entityManager = [];
                             foreach ($v as $key => $value) {
                                 if (isset($excludedKeys[$key])) {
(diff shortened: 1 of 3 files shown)


Tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -695,6 +709,27 @@ public function testOverwriteEntityAliases(): void
...
+            self::markTestSkipped('This test requires ORM');
+        }
+
+        $container = $this->getContainer();
+        $extension = new DoctrineExtension();
+
+        $config        = BundleConfigurationBuilder::createBuilder()
+            ->addBaseConnection()
+            ->build();
+        $config['orm'] = ['default_entity_manager' => 'app', 'entity_managers' => ['app' => ['mappings' => ['YamlBundle' => ['alias' => 'yml']]]]];
```
ostrolucky: Instead of using both config builder and plain array, let's stick to one approach. If you can't use config  builder, switch to plain array fully.

DependencyInjection/Configuration.php
```diff
@@ -72,17 +75,29 @@ public function getConfigTreeBuilder(): TreeBuilder
...
+                        if (! is_array($v)) {
+                            return false;
+                        }
+
+                        if (array_key_exists('connections', $v) || array_key_exists('connection', $v)) {
+                            return false;
+                        }
+
+                        // Is there actually anything to use once excluded keys considered?
+                        return ! empty(array_filter($v, static function ($key) use ($excludedKeys) {
```
ostrolucky: This seems like you reimplemented array_diff_key. Let's also get rid of !empty(), it can be replaced by (bool).

DependencyInjection/Configuration.php
```diff
@@ -380,30 +394,47 @@ private function configureDbalDriverNode(ArrayNodeDefinition $node): void
...
                     ->beforeNormalization()
-                        ->ifTrue(static function ($v) {
+                        ->ifTrue(static function ($v) use ($excludedKeys) {
                             if (! empty($v) && ! class_exists(EntityManager::class)) {
                                 throw new LogicException('The doctrine/orm package is required when the doctrine.orm config is set.');
                             }
 
-                            return $v === null || (is_array($v) && ! array_key_exists('entity_managers', $v) && ! array_key_exists('entity_manager', $v));
+                            if ($v === null) {
+                                return false;
```
ostrolucky: This seems like you reversed the condition. Can you please ensure line is covered by tests?

DependencyInjection/DoctrineExtension.php
```diff
@@ -104,6 +120,20 @@ public function load(array $configs, ContainerBuilder $container)
...
+                    'orm' => [
+                        'entity_managers' => [
+                            ($config['orm']['default_entity_manager'] ?? 'default') => [],
+                        ],
+                    ],
+                ],
+            ], $configs));
+        }
+
         $this->ormLoad($config['orm'], $container);
```
ostrolucky: Let's move above to ormLoad method

DependencyInjection/DoctrineExtension.php
```diff
@@ -91,6 +92,21 @@ public function load(array $configs, ContainerBuilder $container)
...
+                        'dbal' => [
+                            'connections' => [
+                                ($config['dbal']['default_connection'] ?? 'default') => [],
+                            ],
+                        ],
+                    ],
+                ], $configs));
+            }
+
             $this->dbalLoad($config['dbal'], $container);
```
ostrolucky, line 96: Let's move above to dbalLoad method

DependencyInjection/DoctrineExtension.php
```diff
@@ -91,6 +92,21 @@ public function load(array $configs, ContainerBuilder $container)
         $config        = $this->processConfiguration($configuration, $configs);
 
         if (! empty($config['dbal'])) {
+            // if no DB connection defined, prepend an empty one for the default
```
ostrolucky: Does it actually need to prepend instead of append?

DependencyInjection/DoctrineExtension.php
```diff
@@ -91,6 +92,21 @@ public function load(array $configs, ContainerBuilder $container)
...
+            if (empty($config['dbal']['connections'])) {
+                $config = $this->processConfiguration($configuration, array_merge([
+                    [
+                        'dbal' => [
+                            'connections' => [
+                                ($config['dbal']['default_connection'] ?? 'default') => [],
+                            ],
+                        ],
+                    ],
+                ], $configs));
```
ostrolucky: I don't like how verbose these are, let's do something like
```suggestion
                $configs[] = ['dbal' => ['connections' => [($config['dbal']['default_connection'] ?? 'default') => []]]];
                $config    = $this->processConfiguration($configuration, $configs);
```
instead

ostrolucky (changes requested):
This was a can of worms last time, so I think I will need review from someone else as well.

Tests/DependencyInjection/DoctrineExtensionTest.php
tucksaun: I'm fine with a plain array, I just mimicked the logic from above which is done this way I guess to not rely on knowing the default dbal configuration in an EM related test 🤷

DependencyInjection/Configuration.php
tucksaun: 🤦
let's keep that between you and me 😅

DependencyInjection/Configuration.php
tucksaun: Yep, I reversed it because a null value here was interesting to get into the normalization callback that would create the empty default entity manager configuration which we are now doing in the Extension.
This what not covered before but I added a test case.
And because now we are now sure to enter the normalization callback with an array I removed cast to an array there

DependencyInjection/DoctrineExtension.php
tucksaun: not really possible atm because it needs access to the original `$configs`.

DependencyInjection/DoctrineExtension.php
tucksaun: not really possible atm because it needs access to the original `$configs`.

DependencyInjection/DoctrineExtension.php
tucksaun: good question...
I would say that it does not matter given it only adds an empty connection array and that there's no change it would get overridden because otherwise, we would not be in that case.
but my gut tells me pre-pending is better because it's closer to the concept of `processConfiguration`

tucksaun: but now that I read your comment bellow your question makes more sense to me...
test passes so I guess we are good to go

DependencyInjection/DoctrineExtension.php
tucksaun: done 👍

tucksaun:
> This was a can of worms last time, so I think I will need review from someone else as well.

Yeah, I've read that 😞 
I guess @nicolas-grekas and @stof would be great ones for reviewing this PR

tucksaun:
@ostrolucky thank you for the review btw!
all comments (except dbalLoad/ormLoad) should have been addressed 💪

stof:
could we do the checks on default connection and default entity manager together at the beginning, to process the configuration only twice instead of 3 times ? Otherwise, we might report deprecations 3 times when a deprecated setting is used.

tucksaun:
Good call @stof !
We could then wrap this logic into processConfiguration. Wdyt?

ostrolucky:
Good idea I think

DependencyInjection/DoctrineExtension.php
```diff
@@ -107,6 +109,40 @@ public function load(array $configs, ContainerBuilder $container)
...
+     * default values for each.
+     *
+     * @param array<array<mixed>> $configs
+     *
+     * @return array<mixed>
+     */
+    private function processConfigurationPrependingDefaults(ConfigurationInterface $configuration, array $configs): array
+    {
+        $config      = $this->processConfiguration($configuration, $configs);
+        $configToAdd = [];
```
ostrolucky, line 124: Why this instead of doing `$configs[] =` ?

tucksaun: if we want to call `processConfiguration` a second time only when required we need to track if some configuration has been added or not.
if we directly append to `configs` we can't really tell 🤷

ostrolucky (approved):

DependencyInjection/DoctrineExtension.php
tucksaun: But thinking about it, we can use a Boolean value to track this if you prefer

ostrolucky: no it's ok

dmaicher (approved):
LGTM and also tested it quickly on one of my apps: no issues :+1:

PR #1661, opened 2023-05-23
Add a new `report_fields_where_declared` config setting

mpdude:
This adds a new config setting `report_fields_where_declared` that makes it easy to opt-in to the new mapping driver mode that was added in Doctrine ORM 2.16 and will be mandatory in ORM 3.0. For details, see https://github.com/doctrine/orm/pull/10455.

I think that since this bundle allows to specify the mapping configuration per entity manager, it would make sense to have this new config setting at the entity manager level as well.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -628,6 +628,7 @@ class Configuration implements ConfigurationInterface
                     ->arrayNode('schema_ignore_classes')
                         ->prototype('scalar')->end()
                     ->end()
+                    ->scalarNode('report_fields_where_declared')->defaultFalse()->info('Set to "true" to opt-in to the new mapping driver mode that was added in Doctrine ORM 2.16 and will be mandatory in ORM 3.0. See https://github.com/doctrine/orm/pull/10455.')->end()
                 ->end()
                 ->children()
                     ->arrayNode('second_level_cache')
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -823,6 +823,22 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $this->loadMappingInformation($entityManager, $container);
         $this->registerMappingDrivers($entityManager, $container);
 
+        $chainDriverDef = $container->getDefinition($this->getObjectManagerElementName($entityManager['name'] . '_metadata_driver'));
+        foreach (array_keys($this->drivers) as $driverType) {
+            $mappingService   = $this->getObjectManagerElementName($entityManager['name'] . '_' . $driverType . '_metadata_driver');
+            $mappingDriverDef = $container->getDefinition($mappingService);
+            $args             = $mappingDriverDef->getArguments();
+            if ($driverType === 'annotation') {
+                $args[2] = $entityManager['report_fields_where_declared'];
+            } elseif ($driverType === 'attribute') {
+                $args[1] = $entityManager['report_fields_where_declared'];
+            } else {
+                continue;
+            }
+
+            $mappingDriverDef->setArguments($args);
+        }
+
         $ormConfigDef->addMethodCall('setEntityNamespaces', [$this->aliasMap]);
     }
 
--- a/Resources/doc/configuration.rst
+++ b/Resources/doc/configuration.rst
@@ -246,6 +246,8 @@ Configuration Reference
                         class_metadata_factory_name:  Doctrine\ORM\Mapping\ClassMetadataFactory
                         default_repository_class:     Doctrine\ORM\EntityRepository
                         auto_mapping:                 false
+                        # Opt-in to new mapping driver mode as of Doctrine ORM 2.16, https://github.com/doctrine/orm/pull/10455
+                        report_fields_where_declared: false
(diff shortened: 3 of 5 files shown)


greg0ire:
> how do I make sure that the config setting is used only for ORM versions that already support it?

You may use feature detection with `trait_exists(ReflectionBasedDriver::class)`

> And once we're at Doctrine 3.0, how to notify users that the setting is a no-op and should be removed?

Likewise, use feature detection based on something hopefully related to this feature.

DependencyInjection/Configuration.php
```diff
@@ -628,6 +628,7 @@ private function getOrmEntityManagersNode(): ArrayNodeDefinition
                     ->arrayNode('schema_ignore_classes')
                         ->prototype('scalar')->end()
                     ->end()
+                    ->scalarNode('report_fields_where_declared')->defaultFalse()->end()
```
greg0ire: I think a call to `info()` will be quite important for this setting, because it's hard to understand.

ostrolucky (changes requested):
Docs should be added too (configuration.rst and UPGRADE-2.10.md)

DependencyInjection/Configuration.php
mpdude: added

mpdude:
@ostrolucky Do I need the UPGRADE entry when there is nothing users have to do during the upgrade? They can opt in at any time as long as ORM ^2.16 is available.

Don't get me wrong, I'll update `configuration.rst` for sure.

mpdude:
> > how do I make sure that the config setting is used only for ORM versions that already support it?
> 
> You may use feature detection with `trait_exists(ReflectionBasedDriver::class)`
> 
> > And once we're at Doctrine 3.0, how to notify users that the setting is a no-op and should be removed?
> 
> Likewise, use feature detection based on something hopefully related to this feature.

Not sure this is going to work, since I cannot tell whether I am before ORM 2.16 (config setting makes no sense, need not be available) or at ORM 3.0 (config setting should be removed).

Can I use https://getcomposer.org/doc/07-runtime.md#knowing-whether-package-x-is-installed-in-version-y – that requires Composer 2.x?

greg0ire:
>  I cannot tell whether I am before ORM 2.16 (config setting makes no sense, need not be available) or at ORM 3.0 (config setting should be removed).

Well you could check if `ClassMetadataInfo` exists, for instance.

> Can I use https://getcomposer.org/doc/07-runtime.md#knowing-whether-package-x-is-installed-in-version-y – that requires Composer 2.x?

Might be fine as well.

ostrolucky:
> @ostrolucky Do I need the UPGRADE entry when there is nothing users have to do during the upgrade?

I'm basing this on what we have in UPGRADE-2.9.md but IIRC we don't document things that users have to change only, but also things that _should be_ changed. Especially things that help them avoid deprecations. I mean, let's say ORM 3.x or 4.x decides to throw exception if you don't set the value to true. In which UPGRADE-* file would you suggest then to do add this docs once such ORM is released and all users code starts breaking because they didn't change that config?

mpdude:
The thing is that users should set the config option once they upgrade doctrine/orm to >= 2.16, and remove it when they upgrade doctrine/orm to >= 3.0. Both times, they probably don't look at DoctrineBundle upgrade notes, but rather at the doctrine/orm ones.

It may happen that when making the DoctrineBundle 2.10 upgrade, ORM 2.16 will not even be released in the first place.

ostrolucky:
Again, check how UPGRADE-2.9.md is written. It explains everything, similar like you explained in your last comment, that's how it should be. We are not just going to tell user to set it, but to set it if they use ORM 2.16 and to do it if they want to solve deprecation and prepare for ORM 3.

> they probably don't look at DoctrineBundle upgrade notes, but rather at the doctrine/orm ones.

Opposite for me. I don't look at orm notes because they are often meaningless in symfony context. For example orm notes will tell you to set some argument of some class/function, that's meaningless in context of symfony user. Symfony user needs to know which config value they need to flip in doctrine.yaml, since bundle is doing the passing of the argument and not the user.

dmaicher:
Maybe we can just bump the [conflict](https://github.com/doctrine/DoctrineBundle/blob/2.9.x/composer.json#L69) to `<2.16` for the ORM (once released?) to address the challenge you mentioned in the PR description?

mpdude:
I have added an explanatory text in a new `UPGRADE-2.10.md` file, explaining the changes and possible impact.

Regarding feature detection, what if we simply go without it?

Turning the new flag on for pre-ORM 2.16 does no harm: It adds a new constructor parameter to mapping drivers that will simply be ignored.

For ORM ^2.16, the ORM will trigger a deprecation notice unless the flag is turned on.

In ORM ^3.1, a deprecation notice will be triggered stating that the flag is a no-op and the configuration should be removed.

... so maybe nothing we need to take care of here at all?

ostrolucky:
Yeah I think it's good enough like this. Can you rebase? Then we can merge.

mpdude:
@ostrolucky Rebased, thank you

ostrolucky (approved):

stof:
This misses the update of the XSD file (so the documented configuration for the XML format will break)

Issue #1662, opened 2023-05-25
Logging Doctrine deprecations with Symfony?


mpdude:
What is the easiest way to log deprecations emitted by the Doctrine codebase in the same way, same place, using the same mechanism that is used to deal with Symfony deprecations?

Is this something this bundle can help with?

greg0ire:
The easiest way would be to do this I suppose: https://github.com/doctrine/DoctrineBundle/blob/e6da248067559c88792aa12320f9b11323583937/Tests/bootstrap.php#L7

I think maybe that's one of the things the bundle should do… when booting? Or maybe later, conditionally, based on a config setting? In case this might be unwanted/considered a breaking change? Or maybe it should be possible to fully configure the library with https://github.com/doctrine/deprecations#usage-from-consumer-perspective

IIRC there was at some point a discussion about splitting this repo into a DBAL bundle and an ORM bundle, so I'm not 100% sure if this is the right place, or if the doctrine bridge would preferred for this.

ostrolucky:
See discussion at https://github.com/symfony/symfony/pull/41408 for reasons why this wasn't and won't be done in a bundle.

stof:
See https://github.com/doctrine/DoctrineBundle/issues/1300#issuecomment-1563899734 for the discussion.

The dedicated bundle should do it on boot time IMO.
I don't think we should expose the full config of the library. In the context of Symfony, it makes sense to always bring them with other deprecations (Symfony already allows to customize how they are logged as it uses a dedicated channel of the MonologBundle configuration)

nicolas-grekas:
I'm glad this discussion finally moves forward.
Instead of a bundle, what about a package that'd ship a file required by composer directly? This file would run the line that configures the deprecation lib so that it uses trigger_error.
People that'd get this as a transitive dep could "replace" the lib if they'd prefer another configuration.
This strategy would remove the overhead of a bundle.

greg0ire:
I think that would be acceptable, since it's about a one-liner that does not need any configuration. 

>  People that'd get this as a transitive dep could "replace" the lib if they'd prefer another configuration.

Definitely something we should document if we go that way.

@stof said it could be required by the bundle, I suppose it would be suggested in 2.x, and required in 3.x then, right?
We would only be left with the hard problem of naming… `doctrine/deprecations-symfony`?

stof:
[...]

PR #1663, opened 2023-05-25
Add a new `use_trigger_error_for_deprecations` config setting

mpdude:
This adds a new `use_trigger_error_for_deprecations` config setting that can be used to make `doctrine/deprecations` report deprecations through `trigger_error`.

That way, those deprecations should show up in the same place/way as other Symfony deprecations.

See #1662 for the preceding discussion.

#### TODO

- [ ] Document the new setting
- [ ] Add UPGRADING notice?

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -65,6 +65,12 @@ class Configuration implements ConfigurationInterface
         $this->addDbalSection($rootNode);
         $this->addOrmSection($rootNode);
 
+        $rootNode
+            ->children()
+            ->booleanNode('use_trigger_error_for_deprecations')
+            ->defaultFalse()
+            ->info('Setting this to "true" will make Doctrine use trigger_error() to report deprecations, just like Symfony does');
+
         return $treeBuilder;
     }
 
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -92,6 +92,8 @@ class DoctrineExtension extends AbstractDoctrineExtension
         $configuration = $this->getConfiguration($configs, $container);
         $config        = $this->processConfigurationPrependingDefaults($configuration, $configs);
 
+        $container->setParameter('doctrine.use_trigger_error_for_deprecations', $config['use_trigger_error_for_deprecations']);
+
         if (! empty($config['dbal'])) {
             $this->dbalLoad($config['dbal'], $container);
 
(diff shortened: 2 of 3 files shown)


DependencyInjection/Configuration.php
```diff
@@ -65,6 +65,12 @@ public function getConfigTreeBuilder(): TreeBuilder
         $this->addDbalSection($rootNode);
         $this->addOrmSection($rootNode);
 
+        $rootNode
+            ->children()
+            ->booleanNode('use_trigger_error_for_deprecations')
+            ->defaultFalse()
```
greg0ire, line 71: Will this default change in 3.0.0? If yes, should we deprecate not configuring this node explicitly?
As a follow-up, I think a contribution should be made to the Symfony recipe, setting it to `true`.

mpdude: I don’t see why it should change?

greg0ire (commented):
If we merge like this, we close the door on having the possibility to rather use a PSR logger, or even something custom that will use `getTriggeredDeprecations`.

Another possibility would be to have something more complex:

```
deprecations:
    strategy: trigger_error|psr_logger|just_track # defaults to trigger_error
    strategy_options:
        logger: my_psr_logger # only valid when the strategy is psr_logger
```

mpdude:
That's true.

My point is that, from a Symfony user's perspective, I do not care all too much about all the options the doctrine/deprecations package offers. I understand that it needs to provide this flexibility since it is a library underlying Symfony and other frameworks.

However, in Symfony, there is a way to capture, see and log deprecation messages – and I do not see why DoctrineBundle users would want to use any other (additional) way.

So, this flag is available to say "I want to see and notice Doctrine deprecations just like I see Symfony deprecations".

If you need any other way of specialized Doctrine deprecation handling, you can still configure it yourself.

ostrolucky:
See discussion at symfony/symfony#41408 for reasons why this wasn't and won't be done in a bundle.

mpdude:
I find that discussion difficult to understand. If I get it right, the conclusion is that since `doctrine/deprecations` would be enabled (or should be possible to enable it) for *all* Doctrine project libraries, not only ORM or ODM, the only place suitable for such code would be a `DoctrineDeprecationsBundle`.

Speaking for Doctrine ORM users, those that actively go out and search for ORM deprecations will end up adding a single line to their bootstrapping code. Those who only know about (and rely on) the “Symfony way” of reporting deprecations will be surprised when they try the next major upgrade.

mpdude:
https://github.com/doctrine/DoctrineBundle/issues/1300 was discussed two years ago but did not implement anything in the end.

ostrolucky:
From what I understood it doesn't need to be a new package but bundle code could be part of doctrine/deprecations. So effort to improve this situation should go there. And yeah I guess reason why it wasn't implemented was that nobody cares enough to implement it, like usual. Btw this should be non-confiruable `true` IMHO, nothing else makes sense in symfony framework world.

PR #1650, opened 2023-04-21
[DoctrineBundle] remove doctrine.event_subscriber tag and replace by doctrine.event_listener tag

alli83:
Remove doctrine.event_subscribers tag in bundle and replace them with doctrine.event_listener tag.

following https://github.com/symfony/symfony/pull/49918

PR #1664, opened 2023-05-26
[DoctrineBundle] deprecate EventSubscriberInterface

alli83:
Doctrine subscribers are flawed because they force eager loading by design so the goal since https://github.com/symfony/symfony/pull/49918 and https://github.com/doctrine/DoctrineBundle/pull/1650 is to push projects to use attributes to register their listeners through autoconfiguration. In order to achieve this goal, this PR propose to deprecate the EventSubscriberInterface available in doctrine/doctrine-bundle.

### DIFF
--- a/EventSubscriber/EventSubscriberInterface.php
+++ b/EventSubscriber/EventSubscriberInterface.php
@@ -4,6 +4,9 @@ namespace Doctrine\Bundle\DoctrineBundle\EventSubscriber;
 
 use Doctrine\Common\EventSubscriber;
 
+/**
+ * @deprecated use the {@see AsDoctrineListener} attribute instead
+ */
 interface EventSubscriberInterface extends EventSubscriber
 {
 }
--- a/UPGRADE-2.10.md
+++ b/UPGRADE-2.10.md
@@ -1,0 +1,6 @@
+UPGRADE FROM 2.9 to 2.10
+========================
+
+### Deprecations
+
+- `Doctrine\Bundle\DoctrineBundle\EventSubscriber\EventSubscriberInterface` has been deprecated. Use the `#[AsDoctrineListener]` attribute instead.


EventSubscriber/EventSubscriberInterface.php
```diff
@@ -2,8 +2,16 @@
 
 namespace Doctrine\Bundle\DoctrineBundle\EventSubscriber;
 
+use Doctrine\Bundle\DoctrineBundle\Attribute\AsDoctrineListener;
 use Doctrine\Common\EventSubscriber;
 
+use function trigger_deprecation;
+
+trigger_deprecation('doctrine/doctrine-bundle', '2.10', 'The "%s" class is deprecated, use "%s" attribute instead.', EventSubscriberInterface::class, AsDoctrineListener::class);
```
stof: no need for this deprecation triggered at the time of autoloading the class. The proper deprecation will be triggered by DebugClassLoader when loading a class that implements the interface based on the `@deprecated` phpdoc tag.

EventSubscriber/EventSubscriberInterface.php
```diff
@@ -2,8 +2,16 @@
...
+
+/**
+ * @deprecated use {@link AsDoctrineListener} instead
```
stof:
```suggestion
 * @deprecated use the {@link AsDoctrineListener} attribute instead
```

stof (changes requested):

UPGRADE-2.10.md
```diff
@@ -0,0 +1,6 @@
+UPGRADE FROM 2.9 to 2.10
+=======================
```
nicolas-grekas:
```suggestion
========================
```

UPGRADE-2.10.md
```diff
@@ -0,0 +1,6 @@
...
+
+### Deprecations
+
+- `Doctrine\Bundle\DoctrineBundle\EventSubscriber\EventSubscriberInterface` have been deprecated. Use AsDoctrineListener attribute instead.
```
nicolas-grekas:
```suggestion
- `Doctrine\Bundle\DoctrineBundle\EventSubscriber\EventSubscriberInterface` has been deprecated. Use the `#[AsDoctrineListener]` attribute instead.
```

EventSubscriber/EventSubscriberInterface.php
alli83: ah ok thanks 🙏

UPGRADE-2.10.md
alli83: oups thanks😅

EventSubscriber/EventSubscriberInterface.php
```diff
@@ -2,8 +2,12 @@
 
 namespace Doctrine\Bundle\DoctrineBundle\EventSubscriber;
 
+use Doctrine\Bundle\DoctrineBundle\Attribute\AsDoctrineListener;
 use Doctrine\Common\EventSubscriber;
 
+/**
+ * @deprecated use the {@link AsDoctrineListener} attribute instead
```
stof: I think this should be `@see` rather than `@link` here (`@link` is about linking to a URL)

stof (approved):

qdequippe:
Hi,

I have the following use case that I needs in one subscriber to listen to multiple events at once

```php

class MySubscriber implements EventSubscriberInterface
{
    ...

    public function getSubscribedEvents(): array
    {
        return array(
            Events::postUpdate,
            Events::preUpdate,
            Events::postLoad,
            Events::onFlush,
            Events::preFlush,
            Events::postFlush,
        );
    }
}

```

Do I need to declare 6 attributes `#[AsDoctrineListener(event: Events::postPersist)]`, `#[AsDoctrineListener(event: Events::preFlush)]` ... to remove the deprecation and resolve my use case?

suggestion: what about create attribute `#[AsDoctrineSubscriber]`?

Regards

ostrolucky:
Yes, define 6 attributes, I don't see the issue.

qdequippe:
> Yes, define 6 attributes, I don't see the issue.

No issue just a question (and suggestion for lazy ones like me 😄)

PR #1668, opened 2023-06-02
Fix XSD for the new `report-fields-where-declared` setting from #1661

mpdude:


mpdude:
@stof re. https://github.com/doctrine/DoctrineBundle/pull/1661#issuecomment-1573930908 – is this the right way to fix it?

Resources/config/schema/doctrine-1.0.xsd
```diff
@@ -218,6 +218,7 @@
         <xsd:attribute name="quote-strategy" type="xsd:string" />
         <xsd:attribute name="entity-listener-resolver" type="xsd:string" />
         <xsd:attribute name="repository-factory" type="xsd:string" />
+        <xsd:attribute name="report-fields-where-declared" type="xsd:string" />
```
stof:
```suggestion
        <xsd:attribute name="report-fields-where-declared" type="xsd:boolean" />
```

stof (changes requested):

mpdude:
😑

dmaicher (approved):

PR #1670, opened 2023-06-05
Expose a setting to opt-in for the XSD validation for mapping

### LINKED ISSUE #1647: Fix XSD valdiation deprecation from ORM
See https://github.com/doctrine/DoctrineBundle/issues/1643

We need to agree on an approach so that users can get rid of the deprecation reported by the ORM

> Using XML mapping driver with XSD validation disabled is deprecated and will not be supported in Doctrine ORM 3.0. (XmlDriver.php:60 called by SimplifiedXmlDriver.php:23, https://github.com/doctrine/orm/pull/6728, package doctrine/orm)

stof:
Closes #1647

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -659,6 +659,7 @@ class Configuration implements ConfigurationInterface
                         ->prototype('scalar')->end()
                     ->end()
                     ->scalarNode('report_fields_where_declared')->defaultFalse()->info('Set to "true" to opt-in to the new mapping driver mode that was added in Doctrine ORM 2.16 and will be mandatory in ORM 3.0. See https://github.com/doctrine/orm/pull/10455.')->end()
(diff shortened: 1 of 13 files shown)


ostrolucky:
This means we can remove issue from https://github.com/doctrine/DoctrineBundle/blob/2.9.x/Tests/baseline-ignore now, right?

stof:
@ostrolucky indeed. I removed it now (removing the baseline file entirely as it is empty) by fixing the remaining test triggering the error.

stof:
@ostrolucky @dmaicher I suggest that once this PR is merged, we trigger the 2.10.0 release. This would allow shipping this code to projects upgrading to Symfony 6.3 to avoid getting new reports about deprecations. I see no reason to delay the release (we have nothing else in-progress and that's the last deprecation was can solve in the bundle itself as the remaining one is in the ORM code)

dmaicher (approved):

ostrolucky (approved):

PR #1671, opened 2023-06-05
Avoid indirect deprecations for the AttributeDriver in the testsuite

stof:
this makes the contributor experience nicer when running tests if they don't get reports about them. Opting in for the new behavior in those tests does not change what the tests are intended to cover and avoids the deprecation.


dmaicher (approved):
Ah I thought I fixed all of them :yum:

PR #1672, opened 2023-06-05
updates for .doctrine-project.json

dmaicher:


stof (approved):

.doctrine-project.json
```diff
@@ -4,16 +4,22 @@
   "slug": "doctrine-bundle",
   "versions": [
     {
-      "name": "2.10",
-      "branchName": "2.10.x",
+      "name": "2.11",
+      "branchName": "2.11.x",
       "slug": "latest",
```
dmaicher, line 9: should it really be `latest` :thinking: Maybe something like `next` or `upcoming` would be better?

ostrolucky: I guess it would be better, but not sure if this doesn't break something and also not sure we want to make doctrine projects inconsistent. dbal and other also use latest

ostrolucky (approved):

.doctrine-project.json
dmaicher: yeah ok I see

PR #1673, opened 2023-06-06
Allow using connection `url` with `driver_class`

kbond:
2.10 (#1669) removed passing the `url` param to custom drivers. Instead, it parses it with `DsnParser` from dbal. It's not currently possible to use a custom driver and have the url parsed. You end up with an unknown driver exception because the custom `driver_class` is not used.

I propose we allow `driver_class` to take precedence over any driver parsed by `DsnParser`.

### DIFF
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -242,12 +242,6 @@ class ConnectionFactory
             throw new Exception('Malformed parameter "url".', 0, $e);
         }
 
-        if (isset($parsedParams['driver'])) {
-            // The requested driver from the URL scheme takes precedence
-            // over the default custom driver from the connection parameters (if any).
-            unset($params['driverClass']);
-        }
-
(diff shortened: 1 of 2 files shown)


stof:
I'm in favor of it, but I'm wondering whether we should treat it as a bug (so merged in 2.10.x) or a feature (so merged with 2.11.x).

The legacy handling of `url` in DBAL was also removing the driverClass (which is why DoctrineBundle 2.10.x does it as I copied the code from there), so this looks like a new feature.

kbond:
I'm fine having this as a new feature.

ostrolucky:
We are not going to do this because DSN has a precedence over other params. We had and have this problem with other params and we are not going to make it worse. See https://github.com/doctrine/DoctrineBundle/pull/1328#issuecomment-824112731

PR #1676, opened 2023-06-14
Add a `priority` attribute on `doctrine.middleware` tag and `#[AsMiddleware]` attribute

okhoshi:
Currently, the ordering of the middlewares is only driven by the order in which they were registered in the container, which is not very predictable and sometimes impossible to modify (if one middleware is injected by another bundle for instance). By setting the priority on the `doctrine.middleware` tag, this ordering can be modified more easily.

### DIFF
--- a/DependencyInjection/Compiler/MiddlewaresPass.php
+++ b/DependencyInjection/Compiler/MiddlewaresPass.php
@@ -7,10 +7,13 @@ use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 
+use function array_key_exists;
 use function array_keys;
-use function in_array;
+use function array_map;
+use function array_values;
 use function is_subclass_of;
 use function sprintf;
+use function uasort;
 
 final class MiddlewaresPass implements CompilerPassInterface
 {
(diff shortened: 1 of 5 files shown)


DependencyInjection/Compiler/MiddlewaresPass.php
```diff
@@ -54,6 +66,18 @@ public function process(ContainerBuilder $container): void
...
+            $middlewareDefs = array_map(
+                static fn ($id, $def) => [
+                    $middlewareConnections[$id][$name] ?? $middlewarePriorities[$id] ?? 0,
+                    $def[1],
+                    $def[0],
+                ],
+                array_keys($middlewareDefs),
+                array_values($middlewareDefs),
+            );
+            uasort($middlewareDefs, static fn ($a, $b) => $b[0] <=> $a[0] ?: $a[1] <=> $b[1]);
```
dmaicher, line 78: maybe we can use https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/DependencyInjection/Compiler/PriorityTaggedServiceTrait.php#L40

okhoshi: It was my first idea, but it doesn't work :(. On [line 56](https://github.com/symfony/symfony/blob/52a92926f7fed15cdff399c6921100a10e0d6f61/src/Symfony/Component/DependencyInjection/Compiler/PriorityTaggedServiceTrait.php#L56), `PriorityTaggedServiceTrait` calls `$container->findTaggedServiceIds` with `throwOnAbstract: true`, which is incompatible with Doctrine Middleware that can be abstract definitions (as it is the case in the tests).

dmaicher: Ok I see :confused: 

But from what I can see you made sure we are doing a stable sort, right? So the order for null priority tagged services remains the same as before this change?

okhoshi: Yes, exactly. In case of identical priority (or no priority set), the order from registration in the container is preserved.

dmaicher:
From my side :+1: for that feature. I will also let @ostrolucky have a look.

For finalizing it I think we should also document it in https://github.com/doctrine/DoctrineBundle/blob/2.11.x/Resources/doc/middlewares.rst

dmaicher:
I might also be able to leverage that here :smile: 

https://github.com/dmaicher/doctrine-test-bundle/blob/master/src/DAMA/DoctrineTestBundle/DependencyInjection/DoctrineTestCompilerPass.php#L81

okhoshi:
> I might also be able to leverage that here 😄
> 
> https://github.com/dmaicher/doctrine-test-bundle/blob/master/src/DAMA/DoctrineTestBundle/DependencyInjection/DoctrineTestCompilerPass.php#L81

Happy to see it might be useful immediately 🎉 

> For finalizing it I think we should also document it in https://github.com/doctrine/DoctrineBundle/blob/2.11.x/Resources/doc/middlewares.rst

Adding it now 👍

ostrolucky (changes requested):
Looks ok, but please also document it

dmaicher (approved):

PR #1679, opened 2023-06-27
Expose the XSD validation in the DoctrineOrmMappingsPass

### LINKED ISSUE #1647: Fix XSD valdiation deprecation from ORM

stof:
Closes https://github.com/doctrine/DoctrineBundle/issues/1647

This allows bundles registering a custom mapping through the DoctrineOrmMappingsPass to opt-in for the XSD validation, which was forgotten when adding support for this opt-in in other places in the bundle.

### DIFF
--- a/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
+++ b/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
@@ -66,10 +66,10 @@ class DoctrineOrmMappingsPass extends RegisterMappingsPass
      *
      * @return self
      */
-    public static function createXmlMappingDriver(array $namespaces, array $managerParameters = [], $enabledParameter = false, array $aliasMap = [])
+    public static function createXmlMappingDriver(array $namespaces, array $managerParameters = [], $enabledParameter = false, array $aliasMap = [], bool $enableXsdValidation = false)
     {
         $locator = new Definition(SymfonyFileLocator::class, [$namespaces, '.orm.xml']);
-        $driver  = new Definition(XmlDriver::class, [$locator]);
+        $driver  = new Definition(XmlDriver::class, [$locator, null, $enableXsdValidation]);
 
         return new DoctrineOrmMappingsPass($driver, $namespaces, $managerParameters, $enabledParameter, $aliasMap);
     }


DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
```diff
@@ -66,10 +66,10 @@ public function __construct($driver, array $namespaces, array $managerParameters
      *
      * @return self
      */
-    public static function createXmlMappingDriver(array $namespaces, array $managerParameters = [], $enabledParameter = false, array $aliasMap = [])
+    public static function createXmlMappingDriver(array $namespaces, array $managerParameters = [], $enabledParameter = false, array $aliasMap = [], bool $enableXsdValidation = false)
     {
         $locator = new Definition(SymfonyFileLocator::class, [$namespaces, '.orm.xml']);
-        $driver  = new Definition(XmlDriver::class, [$locator]);
+        $driver  = new Definition(XmlDriver::class, [$locator, null, $enableXsdValidation]);
```
greg0ire, line 72: In the previous situation, we were omitting the second argument. I think as a result, `XmlDriver::DEFAULT_FILE_EXTENSION` was used: https://github.com/doctrine/orm/blob/55d477dc507a52675c5a61267a00d389c71182f4/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php#L50


```suggestion
        $driver  = new Definition(XmlDriver::class, [$locator, XmlDriver::DEFAULT_FILE_EXTENSION, $enableXsdValidation]);
```

If I don't sound very sure about this, that's because its value is `.dcm.xml`, and I don't recall ever seeing that extension in Symfony projects. Also, on line 71, `.orm.xml` is mentioned so maybe it takes over somehow?

greg0ire (approved):
Approving in case my comment is irrelevant.

DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
stof: The extension argument of the driver is ignored when passing a locator anyway. It is used only when passing a path.

And the locator we use uses the default extension of the _simplified_ driver, which is not the same, so it would be confusing.

dmaicher (approved):

PR #1680, opened 2023-06-28
Fix SQL syntax highlight on Symfony 6.3

HypeMC:
The SQL syntax highlighting is broken on Symfony 6.3:

image

The CSS was removed in https://github.com/symfony/symfony/pull/48757 (https://github.com/symfony/symfony/commit/b5a10cebca5f5eb90150c613ca2de08b8f5e2c18).

Since the idea of the PR was to move panel specific styles to their respective template files, it'd make sense to move this CSS to the Doctrine bundle. The only problem with this approach is that on Symfony 6.2 the CSS would be displayed twice, once by Symfony's WebProfilerBundle and once by the DoctrineBundle. I don't think that's a big issue however.

Maybe @javiereguiluz has some input?

### DIFF
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -102,6 +102,16 @@
         .time-bar { display: block; position: absolute; top: 0; left: 0; bottom: 0; background: #e0e0e0; }
         .sql-runnable.sf-toggle-content.sf-toggle-visible { display: flex; flex-direction: column; }
         .sql-runnable button { align-self: end; }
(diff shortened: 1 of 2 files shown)


javiereguiluz (approved):
Thanks @HypeMC!

I agree with you that it's not a big deal to includes these CSS styles twice in 6.2.

ostrolucky (approved):

dmaicher (approved):

dmaicher:
Thanks @HypeMC

PR #1686, opened 2023-07-21
Ensure connection is closed before deleting sqlite file

### LINKED ISSUE #1658: Cannot drop Sqlite database on windows using doctrine:database:drop
Hi everyone,

Since version 2.9.0 we have issue using the `doctrine:database:drop` command with sqlite database on windows env.

Since this version, the command does not drop the database (as `dropDatabase()` is deprecated for Sqlite) but directly delete the _.db_ file (using `unlink()`).

Unfortunately, a connection is open before droppping the database but never closed.
So, on windows, because of this non-closed connection, the _.db_ file can't be deleted.
That throws an error, which is catched, resulting in this error : 

> Warning: unlink(...): Resource temporarily unavailable

The right thing to do seems to close the connection before unlink. Something like this :

```php
// Command/DropDatabaseDoctrineCommand.php:106

                if ($schemaManager instanceof SqliteSchemaManager) {
                    // dropDatabase() is deprecated for Sqlite
                    $connection->close(); // Add this to fix
                    if (file_exists($name)) {
                        unlink($name);
                    }
                } else {
                    $schemaManager->dropDatabase($name);
                }
```
But I let you fix it in the way you think is the best.

Thanks a lot in advance

ostrolucky:
This should fix inability to drop sqlite DB on Windows OS

Fixes https://github.com/doctrine/DoctrineBundle/issues/1658

### DIFF
--- a/Command/DropDatabaseDoctrineCommand.php
+++ b/Command/DropDatabaseDoctrineCommand.php
@@ -104,6 +104,7 @@ EOT
             if ($shouldDropDatabase) {
                 if ($schemaManager instanceof SqliteSchemaManager) {
                     // dropDatabase() is deprecated for Sqlite
+                    $connection->close();
                     if (file_exists($name)) {
                         unlink($name);
                     }


dmaicher (approved):
hope it fixes the issue :blush: theoretically it should

olberger:
I confirm a fix in 2.10.2. Many thx

PR #1688, opened 2023-07-29
Add psalm and phpcs files to .gitattributes

VincentLanglet:
Last one I have in mind @greg0ire

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,7 +1,6 @@
 /.github         export-ignore
 .gitattributes   export-ignore
 .gitignore       export-ignore
-.travis.yml      export-ignore
 phpunit.xml.dist export-ignore
 phpcs.xml.dist   export-ignore
 psalm.xml.dist   export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,5 +3,7 @@
 .gitignore       export-ignore
 .travis.yml      export-ignore
 phpunit.xml.dist export-ignore
+phpcs.xml.dist   export-ignore
+psalm.xml.dist   export-ignore
 /Resources/doc   export-ignore
 /Tests           export-ignore


.gitattributes
```diff
@@ -3,5 +3,7 @@
 .gitignore       export-ignore
 .travis.yml      export-ignore
```
greg0ire: While you're at it, I think you can remove this line.

greg0ire (approved):

.gitattributes
VincentLanglet: Done,

Also, this bundle download a lot of UPGRADE-*.md file.
Seems like all doctrine lib are not excluded which can be understood ; but shouldn't at least old files be removed ?

What about a PR which remove the UPGRADE-2.x or UPGRADE-1.x files where x is a non-anymore-supported version ?

ostrolucky (approved):

.gitattributes
greg0ire: > all doctrine lib are not excluded which can be understood

I don't quite understand that bit…

Ignoring old UPGRADE files while generating the archive sounds good to me 👍

VincentLanglet: > > all doctrine lib are not excluded which can be understood
> 
> I don't quite understand that bit…

My sentence is awful
I meant:
All doctrine lib are not excluding the `Upgrade-*` files, which "may" be understood like some developer will read them directly in the vendor (in the same way docs folders are not excluded too)

> Ignoring old UPGRADE files while generating the archive sounds good to me 👍

Do you mean adding `export-ignore` for all the old UPGRADE file in the .gitattributes ?

When I said "What about a PR which remove the UPGRADE-2.x or UPGRADE-1.x files where x is a non-anymore-supported version ?", I thought about deleting those files from the doctrine repository. Because on the `2.10.x` branch I don't need the UPGRADE-2.9 or UPGRADE-2.8 (etc) files ; juste the UPGRADE-2.10. And if someone need the `UPGRADE-2.9` file it can still be found on the 2.9 branch. (Like we do for https://github.com/sonata-project/SonataAdminBundle)

PR #1689, opened 2023-07-30
Remove outdated Upgrade file

VincentLanglet:
This what I had in mind here https://github.com/doctrine/DoctrineBundle/pull/1688#discussion_r1278341158 @greg0ire 

On the 2.10.x, outdated Upgrade files are unnecessary doctrine should only keep the "Upgrade from 2.9 to 2.10" notice.
This will
- avoid to download those files when installing the library 
- avoid to add/remove those files in the gitattributes and update the entries all the time

If someone still need the "Upgrade from 2.8 to 2.9" notice he can still find it on the branch 2.9.x
https://github.com/doctrine/DoctrineBundle/tree/2.9.x
and so on


greg0ire:
I'll defer to @ostrolucky on this.

PR #1692, opened 2023-08-02
Add Symfony 7 support

ostrolucky:
I was curious what needs to be done so far. Not much so far.

@nicolas-grekas Is that added Container type declaration (instead of ContainerInterface) in `\Symfony\Bridge\Doctrine\ManagerRegistry` legit?


Tests/RegistryTest.php
```diff
@@ -24,54 +25,57 @@
 
 class RegistryTest extends TestCase
 {
+    /** @var MockObject&Container  */
+    private MockObject $container;
+
+    protected function setUp(): void
+    {
+        $this->container = $this->createMock(Container::class);
```
stof: If you want to reduce the diff, you could simply replace `$container = $this->getMockBuilder(ContainerInterface::class)->getMock();` with `$container = $this->createMock(Container::class);` without also moving it to an instance property (which does not have much benefit as there is no shared setup logic

nicolas-grekas (approved):
> Is that added Container type declaration (instead of ContainerInterface) in \Symfony\Bridge\Doctrine\ManagerRegistry legit?

yes it is, because we need that kind of container in `resetService()`

Tests/RegistryTest.php
ostrolucky: Thanks but I know. My goal wasn't to reduce the diff.

derrabus (approved):

stof (approved):

nicolas-grekas:
Could help fix the failing job:
```diff
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -39,7 +39,7 @@ jobs:
           # Test LTS
           - symfony-require: "5.4.*"
             dependencies: "highest"
-            php-version: "8.0"
+            php-version: "8.1"
 
```

stof:
@nicolas-grekas to me, this is a bug in Symfony if installing deps on PHP 8.0 with Symfony 5.4 brings incompatible dependencies between Symfony packages.

nicolas-grekas:
I agree, but I'd wait for someone to be affected to have a closer look :)

ostrolucky:
I think it's just because https://github.com/symfony/translation-contracts/commit/9fca65891b3305a38959d766b3bb13ef35ff0600 was still not tagged

DependencyInjection/DoctrineExtension.php
```diff
@@ -1095,19 +1095,6 @@ private function loadValidatorLoader(string $entityManagerName, ContainerBuilder
...
 
-    /**
-     * @param array<string, mixed> $objectManager
-     * @param string               $cacheName
-     *
-     * @return void
-     *
-     * @psalm-suppress MoreSpecificImplementedParamType
-     */
-    public function loadObjectManagerCacheDriver(array $objectManager, ContainerBuilder $container, $cacheName)
```
xabbuh, line 1106: What is the reason for this change?

ostrolucky: Same method is defined in parent class, but with changed signature

xabbuh (approved):

SenseException (approved):

derrabus (approved):

nicolas-grekas:
Thank you @ostrolucky

PR #1694, opened 2023-08-04
Add missing `reportFieldsWhereDeclared` field parameter to DoctrineOrmMappingsPass methods

jdecool:
A new parameter has been introduced in https://github.com/doctrine/orm/pull/10455 to make Annotations/Attribute mapping drivers report fields for the classes where they are declared.

The bundle has been updated with https://github.com/doctrine/DoctrineBundle/pull/1661 to add a new configuration variable to handle this behavior, but the `DoctrineOrmMappingsPass` has been forgotten.


jdecool:
There are two errors but it doesn't related to my code.

PR #1701, opened 2023-09-04
Don't mock the service container

derrabus:
Spotted while reviewing #1692.

Mocking the service container for tests requires us to write some pretty complicated mocking logic for little to no gain. I think, our tests are easier to comprehend and maintain, if we use Symfony's `Conatiner` class instead.


ostrolucky:
Mocking allows us to verify that tested class is trying to pull the service out of container. Don't we need that? But yeah I am not opposed, although benefit is not that high. And what's up with gh actions?

derrabus:
> Mocking allows us to verify that tested class is trying to pull the service out of container. Don't we need that?

I don't think we do. Our tests verify that a service returned by the registry this the very same that we've put into the container. I think that is enough.

> benefit is not that high.

Looking at the `withConsecutive()`/`willReturnOnConsecutiveCalls()` gymnastics that I could remove, I'd say it's worth it. 🙂 

> And what's up with gh actions?

Dunno. I restarted the jobs and they passed. 🤷🏻‍♂️

dmaicher (approved):

PR #1707, opened 2023-10-17
DBAL 4 compatibility

dmaicher:
This adds compatibility with DBAL 4 (ignoring ORM 3 for now).

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -54,6 +54,12 @@ jobs:
             stability: "stable"
             remove-orm: true
 
+          # DBAL 4
+          - php-version: "8.2"
+            dependencies: "highest"
+            stability: "dev"
+            remove-orm: true
+
           # Bleeding edge
           - php-version: "8.2"
             dependencies: "highest"
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -6,17 +6,19 @@ use Doctrine\Common\EventManager;
 use Doctrine\DBAL\Configuration;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DriverManager;
-use Doctrine\DBAL\Exception;
 use Doctrine\DBAL\Exception as DBALException;
 use Doctrine\DBAL\Exception\DriverException;
-use Doctrine\DBAL\Exception\MalformedDsnException;
+use Doctrine\DBAL\Exception\DriverRequired;
+use Doctrine\DBAL\Exception\InvalidWrapperClass;
 use Doctrine\DBAL\Platforms\AbstractMySQLPlatform;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Tools\DsnParser;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\Deprecations\Deprecation;
+use RuntimeException;
 
 use function array_merge;
+use function class_exists;
 use function is_subclass_of;
 use function trigger_deprecation;
 
@@ -92,6 +94,10 @@ class ConnectionFactory
 
             if (isset($params['wrapperClass'])) {
                 if (! is_subclass_of($params['wrapperClass'], Connection::class)) {
+                    if (class_exists(InvalidWrapperClass::class)) {
+                        throw InvalidWrapperClass::new($params['wrapperClass']);
+                    }
+
                     throw DBALException::invalidWrapperClass($params['wrapperClass']);
                 }
 
@@ -102,7 +108,11 @@ class ConnectionFactory
             $connection = DriverManager::getConnection($params, $config, $eventManager);
             $params     = $this->addDatabaseSuffix(array_merge($connection->getParams(), $overriddenOptions));
             $driver     = $connection->getDriver();
-            $platform   = $driver->getDatabasePlatform();
+            if (class_exists(Connection\StaticServerVersionProvider::class)) {
+                $platform = $driver->getDatabasePlatform(new Connection\StaticServerVersionProvider($params['serverVersion'] ?? ''));
+            } else {
+                $platform = $driver->getDatabasePlatform();
+            }
 
             if (! isset($params['charset'])) {
                 if ($platform instanceof AbstractMySQLPlatform) {
@@ -162,7 +172,10 @@ class ConnectionFactory
         try {
             return $connection->getDatabasePlatform();
         } catch (DriverException $driverException) {
-            throw new DBALException(
+            //TODO: what more specific exception class should we throw with DBAL 4?
+            $class = class_exists(DBALException::class) ? DBALException::class : RuntimeException::class;
+
+            throw new $class(
                 'An exception occurred while establishing a connection to figure out your platform version.' . PHP_EOL .
                 "You can circumvent this by setting a 'server_version' configuration value" . PHP_EOL . PHP_EOL .
                 'For further information have a look at:' . PHP_EOL .
@@ -226,7 +239,7 @@ class ConnectionFactory
      *                 URL extracted into individual parameter parts.
      * @psalm-return Params
      *
-     * @throws Exception
+     * @throws DBALException
      */
     private function parseDatabaseUrl(array $params): array
     {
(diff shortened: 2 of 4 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -46,6 +46,12 @@ jobs:
             dependencies: "highest"
             php-version: "8.2"
 
+          # DBAL 4
+          - php-version: "8.2"
+            dependencies: "highest"
+            stability: "dev"
+            remove-orm: true
```
stof, line 61: you should put `remove-orm` in the definition of the matrix (with `false` as value). Otherwise, this job will be collapsed with other matrix jobs matching the other settings.

stof: And adding such parameter in the CI matrix requires flagging tests that require the ORM to skip them when it is not installed.

dmaicher: @stof see https://github.com/doctrine/DoctrineBundle/pull/1708

ConnectionFactory.php
```diff
@@ -102,7 +108,11 @@ public function createConnection(array $params, ?Configuration $config = null, ?
             $connection = DriverManager::getConnection($params, $config, $eventManager);
             $params     = $this->addDatabaseSuffix(array_merge($connection->getParams(), $overriddenOptions));
             $driver     = $connection->getDriver();
-            $platform   = $driver->getDatabasePlatform();
```
dmaicher, line 105: so previously this skipped any server version and for mysql for example just returned `MySQLPlatform`.

We could use `$connection->getDatabasePlatform()` but it would mean we auto-detect the server version which could mean we actually connect to the database. We should avoid that here I guess.

Is there any other way to do this with DBAL 4 except for using a `StaticServerVersionProvider` with a fallback to an empty server version? :see_no_evil:  :thinking: 

@derrabus maybe you have an idea

ConnectionFactory.php
```diff
@@ -226,19 +239,15 @@ private function addDatabaseSuffix(array $params): array
...
     private function parseDatabaseUrl(array $params): array
     {
         if (! isset($params['url'])) {
             return $params;
         }
 
-        try {
-            $parsedParams = $this->dsnParser->parse($params['url']);
-        } catch (MalformedDsnException $e) {
-            throw new Exception('Malformed parameter "url".', 0, $e);
```
dmaicher, line 240: I removed this since `MalformedDsnException` is an instance of `Doctrine\DBAL\Exception`

ConnectionFactory.php
```diff
@@ -102,7 +108,11 @@ public function createConnection(array $params, ?Configuration $config = null, ?
             $connection = DriverManager::getConnection($params, $config, $eventManager);
```
dmaicher: todo: the `$eventManager` argument was removed on DBAL 4

ConnectionFactory.php
derrabus: > so previously this skipped any server version and for mysql for example just returned `MySQLPlatform`.

Why though? For MySQL drivers especially, this sounds like a bad idea because we need the server version to determine whether we talk to a MySQL or a MariaDB.

dmaicher: We check if its a "Mysql like" platform (`AbstractMySQLPlatform`) to set the `collation` option :thinking: 

See https://github.com/doctrine/DoctrineBundle/blob/b1ff93403daacc835a653db39431f2a492fa87b6/ConnectionFactory.php#L105-L127

Maybe this isn't even needed anymore on DBAL 4? See https://github.com/doctrine/dbal/issues/5214, https://github.com/doctrine/dbal/pull/5246 

I believe `AbstractMySQLPlatform` only handles `collation` on 4.0, or?

dmaicher: Hm but we also set a default charset `utf8mb4` for those platforms :thinking: We have to keep this I think?

ConnectionFactory.php
stof: IIRC, this was added to have a better error message for the user

ConnectionFactory.php
stof: Why not using `$connection->getDatabasePlatform()` instead, which would automatically use the version provider based on the configuration ?

stof: Or better, using `$this->getDatabasePlatform()` to have the better error message telling you how to do if you want to be able to instantiate the connection in environments not able to connect yet.

ConnectionFactory.php
```diff
@@ -248,10 +257,14 @@ private function parseDatabaseUrl(array $params): array
 
         $params = array_merge($params, $parsedParams);
 
-        // If a schemeless connection URL is given, we require a default driver or default custom driver
+        // If a schemaless connection URL is given, we require a default driver or default custom driver
```
stof: this change is wrong. We are talking about a URL without scheme here. this is not the same thing than schemaless.

ConnectionFactory.php
```diff
@@ -162,7 +172,10 @@ private function getDatabasePlatform(Connection $connection): AbstractPlatform
         try {
             return $connection->getDatabasePlatform();
         } catch (DriverException $driverException) {
-            throw new DBALException(
+            //TODO: what more specific exception class should we throw with DBAL 4?
```
stof: This should probably use ConnectionException IMO.

ConnectionFactory.php
dmaicher: ok I see. Then we need to check how we can have a more specific exception as we cannot instantiate an arbitrary `Doctrine\DBAL\Exception` anymore on DBAL 4.

stof: We could actually use a MalformedDsnException as the wrapper exception too.

ConnectionFactory.php
dmaicher: I tried this but it would mean we might connect to the database to determine the platform version (if not specified via config). For example all over the testsuite we would try to connect to some mysql db.

Since this is the initial connection that is replaced further down with changed parameters: we should avoid connecting, or?

ConnectionFactory.php
dmaicher: yeah you are right :+1:

DependencyInjection/DoctrineExtension.php
```diff
@@ -272,7 +273,8 @@ protected function loadDbalConnection($name, array $connection, ContainerBuilder
             ->setArguments([
                 $options,
                 new Reference(sprintf('doctrine.dbal.%s_connection.configuration', $name)),
-                new Reference(sprintf('doctrine.dbal.%s_connection.event_manager', $name)),
+                // event manager is only supported on DBAL < 4
+                class_exists(DBALException::class) ? new Reference(sprintf('doctrine.dbal.%s_connection.event_manager', $name)) : null,
```
derrabus: A more accurate feature detection might be checking if `Connection::getEventManager()` exists.

Tests/Dbal/Logging/BacktraceLoggerTest.php
```diff
@@ -3,15 +3,21 @@
...
 
+use function class_exists;
 use function current;
 
 /** @group legacy */
 class BacktraceLoggerTest extends TestCase
 {
     public function testBacktraceLogged(): void
     {
+        if (! class_exists(DebugStack::class)) {
```
ostrolucky: Hmm I guess we should deprecate it

derrabus: Yes.

stof: The BacktraceLogger class should have been defined as internal. Anyway, it is not used used anymore (except by this test) when the middleware system is available and the logging middleware is available in symfony/doctrine-bridge as we use the middleware in those cases.

ostrolucky: I'm fine with removing it in 2.11. My belief is that doing loud BC break is better than silent one. If somebody is using this class, it's better they get compile error rather than software keep working and discovering that it's not doing what they intended after a long time.

stof:
@derrabus this one needs rebasing to fix conflicts.

Also, I'd like to release 2.11.0 to have the other new features available. Should I wait for this PR or delay it to 2.12 ?

derrabus (approved):

derrabus:
@stof I'd like to see if we can get ORM 3 support done for 2.11 as well. Can you hold back the release for one more day?

stof:
yes

PR #1711, opened 2023-10-22
fix tests with Symfony 7 due to Annotation support removal

dmaicher:


Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -489,7 +491,7 @@ public function testSingleEntityManagerMultipleMappingBundleDefinitions(): void
         $definition = $container->getDefinition('doctrine.orm.default_metadata_driver');
 
         $this->assertDICDefinitionMethodCallAt(0, $definition, 'addDriver', [
-            new Reference('doctrine.orm.default_annotation_metadata_driver'),
+            new Reference(class_exists(AnnotationLoader::class) ? 'doctrine.orm.default_annotation_metadata_driver' : 'doctrine.orm.default_attribute_metadata_driver'),
```
dmaicher, line 494: not so happy with this detection for Symfony 7 and removed annotations support.

Anyone got a better idea?

I tried `$container->has('annotations.reader')` but it fails since we are not loading the  `FrameworkExtension` in tests.

ostrolucky (approved):

Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
stof: Inside a DI extension, the ContainerBuilder won't have any of the services defined by other extensions anyway.

PR #1714, opened 2023-10-23
Add support for disabling type comments in the schema

stof:
This adds a new `disable_type_comments` boolean setting in the DBAL configuration to expose the DBAL feature added in https://github.com/doctrine/dbal/pull/6150

DBAL 3.2 implemented the platform-aware schema comparison APIs which don't rely on the DC2Type comments on columns anymore and has deprecated the old schema comparison APIs. DBAL 4.0 removes type comments entirely (In 4.0, this setter would become a no-op in the Configuration class and it will be deprecated in 4.1).
For projects that use the new APIs (and doctrine/migrations uses them), the type comment is pure noise in the DB. And it would generate a DB migration when migrating to DBAL 4 to remove those comments (which are added for any `date*_immutable` fields for instance). The new setting allows to perform the cleanup earlier than the DBAL 4 release (and avoiding to add new type comments for further schema changes until then).

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -250,6 +250,12 @@ class DoctrineExtension extends AbstractDoctrineExtension
 
         unset($connection['auto_commit']);
 
+        if (isset($connection['disable_type_comments'])) {
+            $configuration->addMethodCall('setDisableTypeComments', [$connection['disable_type_comments']]);
+        }
+
+        unset($connection['disable_type_comments']);
+
         if (isset($connection['schema_filter']) && $connection['schema_filter']) {
             $definition = new Definition(RegexSchemaAssetFilter::class, [$connection['schema_filter']]);
             $definition->addTag('doctrine.dbal.schema_filter', ['connection' => $name]);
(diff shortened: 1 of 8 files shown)


dmaicher (approved):

dmaicher:
Should it be documented?

ostrolucky:
Indeed. Please add docs and test, if it's easy enough

stof:
I added those. For the documentation, I'm adding that only in the configuration reference. I don't think it needs lots of documentation.

DependencyInjection/DoctrineExtension.php
```diff
@@ -250,6 +250,12 @@ protected function loadDbalConnection($name, array $connection, ContainerBuilder
 
         unset($connection['auto_commit']);
 
+        if (isset($connection['disable_type_comments'])) {
+            $configuration->addMethodCall('setDisableTypeComments', [$connection['disable_type_comments']]);
+        }
+
+        unset($connection['disable_type_comments']);
```
dmaicher, line 257: could also go inside the `if` statement above I guess?

dmaicher (approved):

DependencyInjection/DoctrineExtension.php
dmaicher: Ah nevermind. It can be null.

ostrolucky (approved):

DependencyInjection/DoctrineExtension.php
stof: I kept it consistent with other parameters than need to be removed from the config passed to the connection factory.

PR #1716, opened 2023-10-24
[DRAFT] Isolating bin dependencies.

oleg-andreyev:
Reduce conflicts with dev-packages by moving bin dependencies into separate composer.json


vendor-bin/phpunit/composer.json
```diff
@@ -0,0 +1,6 @@
+{
+    "require-dev": {
+        "phpunit/phpunit": "^9.5.26",
```
stof: PHPUnit runs in the same process than our code so it cannot have different dependencies.

.github/workflows/continuous-integration.yml
```diff
@@ -73,6 +73,11 @@ jobs:
           ini-values: "zend.assertions=1"
           extensions: "pdo_sqlite"
 
+      - name: Install dependencies
+        run: |
+          composer update
+          composer bin all install
```
stof: why installing all ? The psalm bin needs to be installed only for the psalm job.

.github/workflows/continuous-integration.yml
```diff
@@ -73,6 +73,11 @@ jobs:
           ini-values: "zend.assertions=1"
           extensions: "pdo_sqlite"
 
+      - name: Install dependencies
+        run: |
+          composer update
```
stof: this should not be done here.

Instead, you need to move the command running `composer bin *` later in the workflow, after the step `Install dependencies with Composer`

oleg-andreyev: Composer usually complains that composer.lock is missing and performs update internally

stof: and that's the whole issue: you perform the install or update before the steps that modify the composer.json to change what will get installed.

vendor-bin/phpunit/composer.json
oleg-andreyev: It will work fine, external tools and their dependencies should not affect application/lib dependencies. Had many occasions that phpunit dependency example symfony/process conflicts with app/lib dependencies and had to upgrade phpunit first.

oleg-andreyev:
This is a Draft or RFC is you wish

vendor-bin/phpunit/composer.json
stof: That's the reason they must be compatible: when running tests, PHPUnit will run the tested code inside its own process. And as PHP cannot load 2 definitions of a class, your code will run against the Process class of the PHPUnit dependency instead of using its own dependency, which will break.

Isolating tools is possible when the tool does not execute your own code in its process.

oleg-andreyev: Right , phpunit is probably an exception , other tools like psalm could be moved as they don't autoload code

stof: And that's why I reported the issue for PHPUnit, not for Psalm

oleg-andreyev: reverted phpunit.

ostrolucky:
We will consider doing this when a need arises. I'm well aware of composer-bin-plugin and it is quite useful in projects, but I am not aware of any libraries using it. I don't think we should spearhead usage of it, especially since current transition to Symfony 7 goes smooth I would say, which is a proof this isn't worth making our setup more complicated and non-standard for contributors. I have compared installed set of dependencies with and without your PR and I can't see important differences there either.

greg0ire:
> I am not aware of any libraries using it.

We do in sql-formatter since https://github.com/doctrine/sql-formatter/pull/55, but because the need arose.

PR #1719, opened 2023-10-27
add DSN schemes for mysqli, oci8 and sqlsrv

### LINKED ISSUE #1718: Missing default schemes
Hi,

I've noticed that not all official drivers provided by Doctrine can be used in this bundle with a DSN. Only some drivers have an associated scheme registered in the `ConnectionFactory` class.

https://github.com/doctrine/DoctrineBundle/blob/f549108cd9ee253c8f1a7c16ab866ff5e6cae81f/ConnectionFactory.php#L29-L39

Imagine that you want to use the [`sqlsrv`](https://github.com/doctrine/dbal/blob/45941c67dd0505dab2fb970786d93055b7cbd2b6/src/Driver/SQLSrv/Driver.php) driver. By analogy with the other available drivers, I would expect the DSN for this driver to look like `sqlsrv://user:password@host:port`.
With the current configuration, using a DSN like this does not work because no driver is mapped to the `sqlsrv` scheme. In this case the default `pdo_mysql` driver is used.

Since `doctrine/dbal:3.7.0` and `doctrine/doctrine-bundle:2.10.0` we can add custom scheme-to-driver mappings with the `doctrine.dbal.driver_schemes` configuration option (made possible by doctrine/dbal#6059 and doctrine/DoctrineBundle#1669).

However, if I try to register the `sqlsrv` scheme with the `doctrine.dbal.driver_schemes` option, the configuration fails because of the harcoded list of "official drivers" in the `Configuration` class.

https://github.com/doctrine/DoctrineBundle/blob/f549108cd9ee253c8f1a7c16ab866ff5e6cae81f/DependencyInjection/Configuration.php#L141-L166

The same limitation applies to the `oci8` and `mysqli` official drivers which are absent from the default schemes listed in `ConnectionFactory`  but forbidden in `Configuration`.

I would gladely prepare a PR to address this but I'd rather ask for the direction you would prefer to go beforehand

One way to fix this would be to add a few entries to `ConnectionFactory::DEFAULT_SCHEME_MAP` to enable default support for the missing drivers. This seems to me like the prefered option because it enables more official drivers out of the box. I find it strange that no driver is available to use with a DNS by default for SQL Server and OCI.

Alternatively, the affected drivers could be removed form the hardcoded list in the `Configuration` class, leaving the responsability to every project needing them to register the drivers manually.

Thank you for taking the time to read this.

MrMitch:
This PR adds 3 schemes to be able to use the `mysqli`, `oci8` and `sqlsrv` drivers in DSN configurations.

Adding `mysqli` provides another driver option for MySQL (in addition to `pdo_mysql` which can already be used with the `mysql` and `mysql2` schemes).

Adding `oci8` and `sqlsrv` improves developer experience because these drivers were previously not supported and the validation made in the `Configuration` class prevents these schemes from being overridden (even though they were not available !).

https://github.com/doctrine/DoctrineBundle/blob/f549108cd9ee253c8f1a7c16ab866ff5e6cae81f/DependencyInjection/Configuration.php#L141-L166

Fixes #1718

### DIFF
--- a/ConnectionFactory.php
+++ b/ConnectionFactory.php
@@ -30,12 +30,15 @@ class ConnectionFactory
         'db2'        => 'ibm_db2',
         'mssql'      => 'pdo_sqlsrv',
         'mysql'      => 'pdo_mysql',
+        'mysqli'     => 'mysqli',
         'mysql2'     => 'pdo_mysql', // Amazon RDS, for some weird reason
+        'oci8'       => 'oci8',
         'postgres'   => 'pdo_pgsql',
         'postgresql' => 'pdo_pgsql',
         'pgsql'      => 'pdo_pgsql',
         'sqlite'     => 'pdo_sqlite',
         'sqlite3'    => 'pdo_sqlite',
+        'sqlsrv'     => 'sqlsrv',
     ];
 
     /** @var mixed[][] */


ostrolucky:
This is just alias map. If driver is not found in this map, it will default to whatever you specify there. If `oci8` is missing from this array, it doesn't mean `oci8` driver is not supported. And mapping `mysqli` to `pdo_mysql` is especially nasty thing, because `mysqli` is separate driver from `pdo_mysql`

ConnectionFactory.php
```diff
@@ -30,12 +30,15 @@ class ConnectionFactory
...
         'mysql'      => 'pdo_mysql',
+        'mysqli'     => 'mysqli',
         'mysql2'     => 'pdo_mysql', // Amazon RDS, for some weird reason
+        'oci8'       => 'oci8',
         'postgres'   => 'pdo_pgsql',
         'postgresql' => 'pdo_pgsql',
         'pgsql'      => 'pdo_pgsql',
         'sqlite'     => 'pdo_sqlite',
         'sqlite3'    => 'pdo_sqlite',
+        'sqlsrv'     => 'sqlsrv',
```
stof, line 41: mapping a value to itself is totally useless as alias, because there is no aliasing taking place.

stof:
I'm closing this because the suggested patch does not make sense. Those schemes already work as they correspond to the name of a DBAL driver.

MrMitch:
@ostrolucky @stof you are correct, my issue was caused by something else.
Thank you for reviewing this and sorry for waisting your time.

Issue #114, opened 2012-09-10
add support for configuring a result cache for the dbal


lsmith77:
i briefly looked into this, but the AbstractDoctrineExtension cache related code is pretty bound to just the ORM. not sure if to just duplicate code .. kinda too late for 2.1 of course ..

oleg-andreyev:
I would be nice to have this functionality

githoober:
+1

kimhemsoe:
We welcome pull requests

cristobal151:
Hi, any news on this?? 

Thanks.

psren:
I hacked around for my project, and this seems to work.
I'm not experienced here, so take it at your own risk:

```diff
diff --git a/src/DependencyInjection/DBALResultCachePass.php b/src/DependencyInjection/DBALResultCachePass.php
new file mode 100644
index 000000000..724652385
--- /dev/null
+++ b/src/DependencyInjection/DBALResultCachePass.php
@@ -0,0 +1,20 @@
+<?php declare(strict_types=1);
+
+namespace App\DependencyInjection;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+final class DBALResultCachePass implements CompilerPassInterface
+{
+    /**
+     * @inheritdoc
+     */
+    public function process(ContainerBuilder $container): void
+    {
+        $connectionDefinition = $container->getDefinition('doctrine.dbal.connection.configuration');
+
+        $connectionDefinition->addMethodCall('setResultCacheImpl', [new Reference('doctrine.orm.default_result_cache')]);
+    }
+}
diff --git a/src/Kernel.php b/src/Kernel.php
index 789685300..8a203bef9 100644
--- a/src/Kernel.php
+++ b/src/Kernel.php
@@ -2,6 +2,7 @@

 namespace App;

+use App\DependencyInjection\DBALResultCachePass;
 use Symfony\Bundle\FrameworkBundle\Kernel\MicroKernelTrait;
 use Symfony\Component\Config\Loader\LoaderInterface;
 use Symfony\Component\Config\Resource\FileResource;
@@ -37,6 +38,8 @@ class Kernel extends BaseKernel

     protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader)
     {
+        $container->addCompilerPass(new DBALResultCachePass());
+
[...]

PR #1352, opened 2021-05-25
Fix compatibility layer for cache services

alcaeus:
Due to container isolation, we cannot access the definition of the given cache service, which breaks autodetection. Thus, we introduce a new `is_psr` configuration option to be used with cache services for people to indicate that they are using a PSR-6 cache service.

Includes the test from #1350 in addition to another functional test for a PSR-6 service.

PR #1721, opened 2023-10-31
Extend dbal config and extension for result cache

chr-hertel:
This is super embarrassing, what do I overlook, this can't be the solution for #114, right?

tested it successfully with symfony demo:

```yaml
# config/packages/cache.yaml
framework:
    cache:
        pools:
            cache.dbal:
                adapter: cache.adapter.filesystem
```
```yaml
# config/packages/doctrine.yaml
doctrine:
    dbal:
        url: '%env(resolve:DATABASE_URL)%'
        result_cache: 'cache.dbal'
```
```php
$conn->executeQuery(
    sql: 'SELECT * FROM symfony_demo_post WHERE id = :id',
    params: ['id' => 1],
    qcp: new QueryCacheProfile(
        lifetime: 3600,
        cacheKey: 'my_cache_key',
    ),
);
```

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -249,6 +249,7 @@ class Configuration implements ConfigurationInterface
                     ->cannotBeEmpty()
                     ->defaultValue($this->getDefaultSchemaManagerFactory())
                 ->end()
+                ->scalarNode('result_cache')->end()
             ->end();
 
         // dbal < 2.11
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -303,6 +303,10 @@ class DoctrineExtension extends AbstractDoctrineExtension
 
         $configuration->addMethodCall('setSchemaManagerFactory', [new Reference($connection['schema_manager_factory'])]);
 
+        if (isset($connection['result_cache'])) {
+            $configuration->addMethodCall('setResultCache', [new Reference($connection['result_cache'])]);
+        }
+
         if (class_exists(LegacySchemaManagerFactory::class)) {
             return;
         }
--- a/Resources/config/schema/doctrine-1.0.xsd
+++ b/Resources/config/schema/doctrine-1.0.xsd
@@ -39,6 +39,7 @@
         <xsd:attribute name="profiling-collect-schema-errors" type="xsd:string" default="true" />
         <xsd:attribute name="server-version" type="xsd:string" />
         <xsd:attribute name="schema-manager-factory" type="xsd:string" />
+        <xsd:attribute name="result-cache" type="xsd:string" />
         <xsd:attribute name="use-savepoints" type="xsd:boolean" />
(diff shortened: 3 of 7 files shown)


dmaicher (commented):
I guess this would be enough indeed :thinking:

ostrolucky:
Why not making it consistent with ORM? There, it's like this https://github.com/doctrine/DoctrineBundle/blob/0ad7b24c6c6cfa492cd1815d676739f84193000a/Resources/doc/configuration.rst#L702-L704

chr-hertel:
Yes, good point, @ostrolucky, consistency would be great.
Trying to understand the implementation here: https://github.com/doctrine/DoctrineBundle/blob/2.10.x/DependencyInjection/DoctrineExtension.php#L1008-L1024

meaning the difference would be that there is a default ArrayCache registered if type pool is used without identifier?

I guess that feature made more sense before #1352 - or do I miss sth?

Edit: not saying it doesn't make sense to bring in that behavior as well, just only if I miss sth to understand before I do the change

ostrolucky:
My comment was mostly aimed at the options, not internal workings. In this PR you accept only service ID, while in ORM version you can choose if you use service id or a pool. But perhaps you are right, since DBAL supports symfony pool only, perhaps we should bite the bullet and just accept that ORM and DBAL node will be different and accept just service id. Any opinions from other maintainers on this?

dmaicher:
I think its fine to just allow a service id.

Actually every cache pool is registered as a service with the pool name as service id anyway, right? So this will allow using cache pools or any custom service (as long as its a PSR cache).

The ORM config technically also still supports the deprecated `doctrine/cache` instances via `CacheCompatibilityPass`. But we should not support those for the DBAL result cache.

ostrolucky (approved):
Fair enough. Let's keep it as this then. Test wouldn't be bad either, but can't really go wrong with such simple change either.

dmaicher:
This should go into 2.11.x though. Could you rebase it @chr-hertel ?

ostrolucky:
Target changed. No need to rebase :)

chr-hertel:
Too late :D

DependencyInjection/Configuration.php
```diff
@@ -249,6 +249,7 @@ private function getDbalConnectionsNode(): ArrayNodeDefinition
                     ->cannotBeEmpty()
                     ->defaultValue($this->getDefaultSchemaManagerFactory())
                 ->end()
+                ->scalarNode('result_cache')->end()
```
dmaicher, line 252: we also need to add this to the XSD schema file, or?

dmaicher (commented):
let's add a very simple test case so at least the feature is also "documented" via tests

DependencyInjection/Configuration.php
chr-hertel: yea, think so too, good one :)

dmaicher (approved):

dmaicher:
Thanks @chr-hertel

Issue #1677, opened 2023-06-15
Avoid conditional classes?


ro0NL:
After bumping doctrine, our psalm analyse broke for all service entity repositories due #1599 

#1611 does not solve it AFAIK: https://psalm.dev/r/75d58de9af

in practice this means all our repositories got wiped from the baseline without any warning, and future issues arent spotted anymore

i suggest to remove `@internal` on `LazyServiceEntityRepository` so we can easily mass replace existing `extends ServiceEntityRepository`, rather than waiting for _to be renamed ServiceEntityRepository when PHP 8.1 / Symfony 6.2 becomes required_ and have more maintenance, again.

Please KISS :)

stof:
We don't want people to use the LazyServiceEntityRepository class directly, as that would mean we cannot actually remove it anymore.

To me, this should rather be fixed in Psalm to be able to analyze this code (this is something that should be doable given that phpstan is able to analyse it)

ro0NL:
yes, but psalm wasnt fixed:

https://github.com/doctrine/DoctrineBundle/blob/96ffb1a9fc9ef2b6385710456c5bf7e7e8b05da0/psalm.xml.dist#L72-L77

> We don't want people to use the LazyServiceEntityRepository class directly, as that would mean we cannot actually remove it anymore.

ServiceEntityRepository and LazyServiceEntityRepository are separate features IMHO, and would avoid all this :)

ro0NL:
i'll wait to see how this plays out

stof:
LazyServiceEntityRepository is meant to fix the compatibility of ServiceEntityRepository with cases where you mark your repository service as being a lazy service in a Symfony 6.2+ project (Symfony 6.2 changed the way lazy services are implemented), to actually keep things lazy (as the parent constructor performs some logic).
This is really meant to be an implementation detail of ServiceEntityRepository for newer versions.

ro0NL:
perhaps the trait should have been mocked instead,

personally im in the "bump your composer constraints and release a new version" camp, then it doesnt matter ServiceEntityRepository is changed or deprecated

stof:
Well, the goal was to avoid dropping support for the 5.4 LTS version of Symfony (which would force us to keep maintaining an older version of the bundle until the EOM of that LTS version).

Once again, the fact that Psalm breaks with this conditional class definition is an issue in Psalm (and phpstan being able to support it is a proof that static analysis tools are able to support it).

ro0NL:
[...]

PR #1722, opened 2023-11-05
Compatibility with ORM 3

derrabus:
This PR adds compatibility with ORM 3. I'm not very proud of the code that I had to write and I'm open to ideas. 🙈 

As a side effect, I'm deprecating not enabling the lazy ghost implementation, just as the ORM did. If the app did not specify which proxy implementation should be used, this is what the bundle decides:

* ORM 2 on PHP < 8.1: Legacy implementation
* ORM 2 on PHP >= 8.1: Legacy implementation + runtime deprecation
* ORM 3: Lazy Ghosts

If the app explicitly opts out of lazy ghosts despite having ORM 3 installed, the container won't compile.


Repository/ServiceEntityRepository.php
```diff
@@ -7,10 +7,11 @@
 use LogicException;
 use Symfony\Component\VarExporter\LazyGhostTrait;
 
+use function property_exists;
 use function sprintf;
 use function trait_exists;
 
-if (trait_exists(LazyGhostTrait::class)) {
+if (trait_exists(LazyGhostTrait::class) && property_exists(EntityRepository::class, '_entityName')) {
```
derrabus: This basically opts us out of the new `LazyServiceEntityRepository` for ORM 3. I need a little help to get this to work, @nicolas-grekas.

nicolas-grekas: Ideally, we would get rid of LazyServiceEntityRepository: I introduced it to keep BC with existing protected properties on EntityRepository. But since those are private in 3.0, we should now be able to deal with lazyness by decorating public/protected methods. As a reminder, the purpose of the lazy proxy is to move the ` $manager = $registry->getManagerForClass($entityClass);` call out of the construction step.

composer.json
```diff
@@ -41,6 +41,7 @@
         "symfony/deprecation-contracts": "^2.1 || ^3",
         "symfony/doctrine-bridge": "^5.4.19 || ^6.0.7 || ^7.0",
         "symfony/framework-bundle": "^5.4 || ^6.0 || ^7.0",
+        "symfony/polyfill-php80": "^1.15",
```
ostrolucky, line 44: I don't think it's worth adding this just for `get_debug_type` which is used in exception message only. Usually var_export or gettype is doing good enough job.

Repository/ServiceEntityRepository.php
stof: This PR still support ORM 2.x though.

composer.json
derrabus: Sure, but then again, the ORM depends on this package already, so why not use it?

Repository/ServiceEntityRepository.php
derrabus: I've pushed a version with a proxy for `EntityRepository`.

composer.json
ostrolucky: fair enough

derrabus:
Ugh, Psalm 4 is a mess. :-(

Tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -1,39 +0,0 @@
...
-    public function testGetConfigTreeBuilderDoNotUseDoctrineCommon(): void
-    {
-        if (extension_loaded('pcov') && PHP_VERSION_ID >= 80100) {
-            $this->markTestSkipped('Segfaults, see https://github.com/krakjoe/pcov/issues/84');
-        }
-
-        $configuration = new Configuration(true);
-        $configuration->getConfigTreeBuilder();
-
-        $this->assertFalse(class_exists('Doctrine\Common\Proxy\AbstractProxyFactory', false));
```
derrabus, line 37: This is a very old test and I'm not sure if we still need it. I cannot make it pass anymore because I need to feature-check if we can allow not enabling the lazy ghost implementation.

ostrolucky:
Can we move some of those psalm suppressions inline? We put it to config file only when that doesn't work.

derrabus:
> Can we move some of those psalm suppressions inline? We put it to config file only when that doesn't work.

Done for the `MethodSignatureMismatch` suppression.

Repository/ContainerRepositoryFactory.php
```diff
@@ -47,8 +55,16 @@ public function getRepository(EntityManagerInterface $entityManager, $entityName
             if ($this->container->has($customRepositoryName)) {
                 $repository = $this->container->get($customRepositoryName);
 
+                if (! $repository instanceof EntityRepository && $strictTypeCheck) {
+                    throw new RuntimeException(sprintf('The service "%s" must extends EntityRepository (e.g. by extending ServiceEntityRepository), "%s" given.', $repositoryServiceId, get_debug_type($repository)));
```
ostrolucky:
```suggestion
                    throw new RuntimeException(sprintf('The service "%s" must extend EntityRepository (e.g. by extending ServiceEntityRepository), "%s" given.', $repositoryServiceId, get_debug_type($repository)));
```

Repository/RepositoryFactoryCompatibility.php
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Doctrine\Bundle\DoctrineBundle\Repository;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\EntityRepository;
+use Doctrine\ORM\Repository\RepositoryFactory;
+use Doctrine\Persistence\ObjectRepository;
+use ReflectionMethod;
+
+if ((new ReflectionMethod(RepositoryFactory::class, 'getRepository'))->hasReturnType()) {
```
ostrolucky, line 11: can you please add comment indicating which branch is for which ORM version?

Repository/ServiceEntityRepository.php
```diff
@@ -3,57 +3,38 @@
...
 use function trait_exists;
 
-if (trait_exists(LazyGhostTrait::class)) {
-    /**
-     * @template T of object
-     * @template-extends LazyServiceEntityRepository<T>
-     */
-    class ServiceEntityRepository extends LazyServiceEntityRepository
-    {
+if (property_exists(EntityRepository::class, '_entityName')) {
```
ostrolucky, line 11: Oh boy, people were already complaining about previous logic being too difficult for SA tools to understand (https://github.com/doctrine/DoctrineBundle/issues/1677), now it's going to get worse. But I guess we can't do anything about it

Repository/ServiceEntityRepositoryProxy.php
```diff
@@ -0,0 +1,134 @@
...
+ * To use in your class, inject the "registry" service and call
+ * the parent constructor. For example:
+ *
+ * class YourEntityRepository extends ServiceEntityRepository
+ * {
+ *     public function __construct(ManagerRegistry $registry)
+ *     {
+ *         parent::__construct($registry, YourEntity::class);
+ *     }
+ * }
```
ostrolucky: I think this was c&p'd and shouldn't be here for this internal class

Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
```diff
@@ -927,6 +937,17 @@ public function testAddFilter(): void
         $this->assertCount(2, $entityManager->getFilters()->getEnabledFilters());
     }
 
+    public function testDisablingLazyGhostOnOrm3Throws(): void
+    {
+        if (! class_exists(ProxyFactory::class) || method_exists(ProxyFactory::class, 'resetUninitializedProxy')) {
```
ostrolucky: Can you split to two branches and indicate which branch applies for which ORM version via comment? I guess ProxyFactory does not exist for ORM 3.0 and resetUninitializedProxy method does not exist for certain combination of ORM and doctrine/common? We need this stuff to make it easier to remove these BC layers later.

Repository/ServiceEntityRepository.php
derrabus: Not without breaking stuff, I'm afraid. It might be worth doing a 3.0 soon'ish.

Tests/DependencyInjection/AbstractDoctrineExtensionTest.php
derrabus: `ProxyFactory` exists on all supported ORM versions. We have a CI job that uninstalls the ORM before running the tests, so the first check makes sure we actually have an ORM. The method I'm checking has been removed in ORM 3, so if it exists, we have ORM 2.

derrabus: I've done as you said. It should be clearer now.

Repository/RepositoryFactoryCompatibility.php
derrabus: Done.

Repository/ServiceEntityRepositoryProxy.php
derrabus: I've moved those doc blocks to the `ServiceEntityRepository` class because that's that class people should actually extend.

ostrolucky (approved):

Repository/ServiceEntityRepository.php
ro0NL: when soon?

PR #1725, opened 2023-11-10
Fix move of CodeExtension

chr-hertel:
not sure if that's the right way to handle this and which target branch.

### DIFF
--- a/Tests/ProfilerTest.php
+++ b/Tests/ProfilerTest.php
@@ -8,10 +8,11 @@ use Doctrine\Persistence\ManagerRegistry;
 use PHPUnit\Framework\TestCase as BaseTestCase;
 use Symfony\Bridge\Doctrine\Middleware\Debug\DebugDataHolder;
 use Symfony\Bridge\Doctrine\Middleware\Debug\Query;
-use Symfony\Bridge\Twig\Extension\CodeExtension;
+use Symfony\Bridge\Twig\Extension\CodeExtension as CodeExtensionLegacy;
 use Symfony\Bridge\Twig\Extension\HttpKernelExtension;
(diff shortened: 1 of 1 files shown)


ostrolucky:
What problem are you solving? Is there an issue somewhere you could reference?

dmaicher:
solving this build fail I guess?

https://github.com/doctrine/DoctrineBundle/actions/runs/6829814285/job/18576668292?pr=1721

Tests/ProfilerTest.php
```diff
@@ -62,7 +64,12 @@ public function setUp(): void
         $urlGenerator = $this->getMockBuilder(UrlGeneratorInterface::class)->getMock();
         $urlGenerator->method('generate')->willReturn('');
 
-        $this->twig->addExtension(new CodeExtension('', '', ''));
+        if (class_exists(CodeExtensionV7::class)) {
+            $this->twig->addExtension(new CodeExtensionV7('', '', ''));
+        } else {
+            $this->twig->addExtension(new CodeExtension('', '', ''));
+        }
```
derrabus: Let's alias the old extension as `CodeExtensionLegacy` and the new one as `CodeExtension`. This will keep the diff smaller when we drop support for Symfony 5 one fine day.

derrabus (commented):
Thank you. Just some minor nit-picking.

Tests/ProfilerTest.php
chr-hertel: :heavy_check_mark:

chr-hertel:
sorry, yes, i had the issue in the pipeline of #1721 with Symfony 7.

caused by https://github.com/symfony/symfony/pull/52483

derrabus (approved):

derrabus:
Thank you!

PR #1727, opened 2023-11-13
Don't rely on LazyProxyTrait in LazyServiceEntityRepository

nicolas-grekas:
Makes things simpler and unlocks https://github.com/symfony/symfony/pull/52568


nicolas-grekas:
Could anyone help me with the psalm failure?

nicolas-grekas:
> Error: Property \Doctrine\Bundle\DoctrineBundle\Repository\LazyServiceEntityRepository::$registry does not have native type hint for its value but it should be possible to add it based on @var annotation "ManagerRegistry".

~Wrong hint by phpcs since we should be compatible with PHP 7.4, isn't it?~

nvm, typed props are allowed in 7.4 :facepalm:

Repository/LazyServiceEntityRepository.php
```diff
@@ -18,41 +20,68 @@
...
+    }
 
-        if ($this instanceof LazyObjectInterface) {
-            $initializer($this, '_entityName');
+    /**
+     * @param string|int $name
+     *
+     * @return mixed
+     */
+    public function __get($name)
```
derrabus:
```suggestion
    /** @return mixed */
    public function __get(string $name)
```

I don't understand how `$name` can be an `int`. 🤔

Repository/LazyServiceEntityRepository.php
```diff
@@ -18,41 +20,68 @@
...
-            return;
+        $scope = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2)[1]['class'] ?? null;
+
+        return (function () use ($name) {
+            return $this->$name;
+        })->bindTo($this, $scope)();
+    }
+
+    /** @param string|int $name */
+    public function __isset($name): bool
```
derrabus:
```suggestion
    public function __isset(string $name): bool
```

derrabus:
> Could anyone help me with the psalm failure?

I have no idea. This ancient Psalm version on this repository has become a burden, tbh.

Repository/LazyServiceEntityRepository.php
nicolas-grekas: updated

Repository/LazyServiceEntityRepository.php
nicolas-grekas: nvm updated

stof:
@derrabus I suggest that we update the Psalm setup to use Psalm 5 instead. It is probably easier than debugging Psalm 4...

ostrolucky:
Yeah we should update it. I didn't even realize it's so outdated TBH. Psalm 5 supports PHP 7.4, so that shouldn't be an issue either.

derrabus:
> @derrabus I suggest that we update the Psalm setup to use Psalm 5 instead.

I've tried a couple of days ago. If you upgrade Psalm, it spits out seventy-something errors that should be reviewed. Whoever wants to give it a try: be my guest. 🙂

ostrolucky (approved):

derrabus:
Okay then, screw Psalm for now. This should not block us. Thank you @nicolas-grekas.

ostrolucky:
I'll work on updating to psalm 5 soon

PR #1728, opened 2023-11-14
psalm 4.x -> 5.x

ostrolucky:


ostrolucky:
Looks like no-ORM jobs fail because of missing Doctrine\ORM\Events and rest of the failing jobs fail because on warnings. I guess phpunit version has been updated. I will adjust warnings outside of this PR. But not sure what to do with no-orm jobs

psalm.xml.dist
```diff
@@ -16,8 +18,10 @@
             <file name="Command/ImportMappingDoctrineCommand.php"/>
             <file name="Command/Proxy/OrmProxyCommand.php"/>
             <file name="DependencyInjection/Compiler/WellKnownSchemaFilterPass.php"/>
-            <!-- We need Psalm 5 to analyze this file -->
+            <!-- We need to run Psalm under >= PHP 8.1 to analyze this file due to intersection types (ParseError) -->
```
derrabus: Let's do that? We usually run Psalm on the latest stable PHP version on other repositories because of problems like this one.

ostrolucky: If we do that, we might not notice when new code written cannot actually work under PHP 7.4 unless we have a test for it, this is why choice was made to run under lowest common denominator. Let's keep it for now, until list of such files keep growing

DataCollector/DoctrineDataCollector.php
```diff
@@ -55,7 +55,7 @@ class DoctrineDataCollector extends BaseCollector
 
     /**
      * @var mixed[][]|null
-     * @psalm-var ?array<string, list<QueryType&array{count: int, index: int, executionPercent: float}>>
```
stof, line 58: shouldn't this be `executionPercent?: float` instead, to be consistent with the getter ?

DependencyInjection/DoctrineExtension.php
```diff
@@ -181,6 +181,7 @@ protected function dbalLoad(array $config, ContainerBuilder $container)
         $connections = [];
 
         foreach (array_keys($config['connections']) as $name) {
+            /** @psalm-suppress InvalidArrayOffset */
```
stof: This should probably be reported to Psalm as a bug. I don't see way an array key would be an invalid array offset here.

ostrolucky: From my understanding it's because $config is `array<mixed>`, which might include `array<array<array<....>>`

stof: but we don't loop over the array. We loop over `array_keys($config['connections'])`. And in PHP, array keys are known to always be `int|string`, not `mixed`.

psalm.xml.dist
derrabus: A static analyzer is not a linter. What we do here is we run the static analysis on a subset of our classes and against outdated dependencies. We did this on the ORM and DBAL repositories and it turned out to be a bad idea.

ostrolucky: Alright, let's bump the version there

DependencyInjection/DoctrineExtension.php
ostrolucky: https://github.com/vimeo/psalm/issues/10382

PR #1729, opened 2023-11-14
Enable report_fields_where_declared for ORM 3

derrabus:
Having `report_fields_where_declared` default to `false` does not make sense if ORM 3 is installed because it will result in a unusable driver configuration.

This PR picks a smart default depending on the installed ORM version. On top of that, it disallows opting out of the `report_fields_where_declared` setting when ORM 3 is installed.


dmaicher:
Ah nice :+1: I actually wanted to look into it myself in the next days as I also encountered then issue in https://github.com/sonata-project/sonata-doctrine-extensions/pull/494#discussion_r1391659057

dmaicher:
Seems there are some related test failures :thinking: https://github.com/doctrine/DoctrineBundle/actions/runs/6870900826/job/18686699968?pr=1729

derrabus:
The test should be fixed now.

dmaicher (approved):

PR #1734, opened 2023-12-07
fix tests by adjusting switch for enabling annotations

dmaicher:


Tests/DependencyInjection/Fixtures/DbalTestKernel.php
```diff
@@ -4,13 +4,13 @@
...
 use Doctrine\Bundle\DoctrineBundle\DoctrineBundle;
 use Doctrine\Bundle\DoctrineBundle\Tests\TestCaseAllPublicCompilerPass;
-use Doctrine\Common\Annotations\Annotation;
 use Psr\Log\NullLogger;
 use Symfony\Bundle\FrameworkBundle\FrameworkBundle;
 use Symfony\Component\Config\Loader\LoaderInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 use Symfony\Component\HttpKernel\Kernel;
+use Symfony\Component\Validator\Mapping\Loader\AnnotationLoader;
```
dmaicher: any better idea? using a class from `Symfony\Component\Validator` seems a bit hacky

ostrolucky: Can't we use `\Symfony\Component\HttpKernel\Kernel::VERSION`?

dmaicher: Let me try it. Previously I had issues with this approach because the framework bundle version might be different than the kernel version

ostrolucky (approved):

Tests/DependencyInjection/Fixtures/DbalTestKernel.php
dmaicher: Ok it works now :+1:

Issue #1745, opened 2024-01-23
Missing `doctrine/common` in composer.json


Ahummeling:
This project has a dependency on `doctrine/common` in the Bundle class (`DoctrineBundle`). It uses the `ClassUtils`.
As far as I can see, because `doctrine/orm` required `doctrine/common` up until 3.0, it was always installed and it was never a problem. But, when upgrading `doctrine/orm` to ithe `3.0-beta.2` version (to try and find some bugs to work on :p) I found that my project no longer worked.

The offending line is at `DoctrineBundle:104`:
```php
$originalClassName = ClassUtils::getRealClass($class);
```
Which is found in DoctrineBundle::boot`. The context here is (I believe) auto-loading a class for which we want to generate a proxy class. 

With `ClassUtils::getRealClass` looking as follows:
```php
/**
     * Gets the real class name of a class name that could be a proxy.
     *
     * @param string $className
     * @psalm-param class-string<Proxy<T>>|class-string<T> $className
     *
     * @return string
     * @psalm-return class-string<T>
     *
     * @template T of object
     */
    public static function getRealClass($className)
    {
        $pos = strrpos($className, '\\' . Proxy::MARKER . '\\');

        if ($pos === false) {
            /** @psalm-var class-string<T> */
            return $className;
        }

        return substr($className, $pos + Proxy::MARKER_LENGTH + 2);
    }
```

So the naive solution would be to add `doctrine/common` to the `composer.json` file.
An alternative solution would be to extract the necessary feature from the `doctrine/common` package, and put it in this project.

I remember something about deprecating something in a doctrine common namespace a few years ago, and I am not sure if the common library is in the process of being phased out. If this is the case, the alternative solution would obviously take preference, as the method in question is quite simple.

I've made a PR (#1746 ) for the naive solution. If another solution is desired, I would also be happy to work on that. Please let me know.

PR #1746, opened 2024-01-23
fix: add doctrine common to dependency list in composer.json

Ahummeling:
As described in #1745 , the PR hopes to resolve a missing dependency on `doctrine/common` that is encountered when upgrading `doctrine/orm` to `3.0.x`.

Running `composer update --prefer-lowest` installed the `2.0.x` version of `doctrine/orm`.
Running `composer why doctrine/common` informed me that `doctrine/orm 2.14.0 requires doctrine/common (^3.0.3)`

Running `composer req doctrine/common ^3.0.3` added the dependency as resulted in the change for this PR

After implementing this change, and installing this branch into my project, the issue was resolved.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,7 @@
     "require": {
         "php": "^7.4 || ^8.0",
         "doctrine/cache": "^1.11 || ^2.0",
+        "doctrine/common": "^3.0.3",
         "doctrine/dbal": "^3.7.0 || ^4.0",
         "doctrine/persistence": "^2.2 || ^3",
         "doctrine/sql-formatter": "^1.0.1",


greg0ire:
I think the right way to address this would be to use the `DefaultProxyClassNameResolver` introduced in https://github.com/doctrine/orm/pull/11009

Ahummeling:
> I think the right way to address this would be to use the `DefaultProxyClassNameResolver` introduced in [doctrine/orm#11009](https://github.com/doctrine/orm/pull/11009)

I agree that this is probably a better solution. 
But `doctrine/orm` is currently a dev dependency (`require-dev`) of the bundle package. Please correct me if I'm wrong, but any application installing `doctrine/doctrine-bundle` without `doctrine/orm`, won't install `doctrine/orm` and the same error would occur. Right?

So in addition to using the `DefaultProxyClassNameResolver`, we'd also have to move `doctrine/orm` from `require-dev` to require and set the version to: `^2.17|^3.0`, right?

I created a new PR (considering the change is entirely different) #1747. If you agree that that approach is more suited, please feel free to close this PR and we can work on getting the new one merged, thanks.

greg0ire:
> the same error would occur

Sorry, the same error as what?

Ahummeling:
> > the same error would occur
> 
> Sorry, the same error as what?

The "class not found" error, but as discussed in the other PR, that will not be the case, because we have already confirmed doctrine/orm is installed at this point.

PR #1747, opened 2024-01-25
fix: replace classutils with doctrine/orm version

### LINKED ISSUE #1745: Missing `doctrine/common` in composer.json

Ahummeling:
Closes #1745 

As discussed in #1746

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,7 @@
         "doctrine/annotations": "^1 || ^2",
         "doctrine/coding-standard": "^12",
         "doctrine/deprecations": "^1.0",
-        "doctrine/orm": "^2.14 || ^3.0",
+        "doctrine/orm": "^2.17 || ^3.0",
         "friendsofphp/proxy-manager-lts": "^1.0",
         "phpunit/phpunit": "^9.5.26",
         "psalm/plugin-phpunit": "^0.18.4",
@@ -69,7 +69,7 @@
     },
     "conflict": {
         "doctrine/annotations": ">=3.0",
-        "doctrine/orm": "<2.14 || >=4.0",
+        "doctrine/orm": "<2.17 || >=4.0",
         "twig/twig": "<1.34 || >=2.0 <2.4"
     },
     "suggest": {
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -32,6 +32,7 @@
         "php": "^7.4 || ^8.0",
         "doctrine/cache": "^1.11 || ^2.0",
         "doctrine/dbal": "^3.7.0 || ^4.0",
+        "doctrine/orm": "^2.17 || ^3.0",
```
greg0ire: I don't think you should move this dependency here. When `$this->container->getParameter('doctrine.orm.auto_generate_proxy_classes')` is `true`, then it must mean `doctrine/orm` is installed, right?

Ahummeling: Oh, ofcourse, you're entirely correct. Moving this change back to require-dev, I do still think the version should be bumped, to avoid installing a version of ORM without the new class

composer.json
```diff
@@ -48,7 +48,7 @@
         "doctrine/annotations": "^1 || ^2",
         "doctrine/coding-standard": "^12",
         "doctrine/deprecations": "^1.0",
-        "doctrine/orm": "^2.14 || ^3.0",
+        "doctrine/orm": "^2.17 || ^3.0",
```
stof, line 51: Bumping the dev requirement will only affect our CI. You also need to update the conflict rule below.

stof (changes requested):

composer.json
Ahummeling: My bad, updated that now

greg0ire (approved):

dmaicher (approved):

PR #1748, opened 2024-02-05
Fix the JS code in the profiler panel for Symfony 7

stof:
In Symfony 7, the `Sfjs` object holding the code for the web debug toolbar is not present in the profiler pages anymore as the JS code has been split.
This replaces its usages by native browser APIs instead (even when running on older versions of Symfony).

### DIFF
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -453,7 +453,6 @@
                     }).then(async function (response) {
                         targetElement.innerHTML = await response.text()
                         targetElement.setAttribute('data-sfurl', link.href)
-                        targetElement.classList.remove('loading')
                     }, function () {
                         targetElement.innerHTML = 'An error occurred while loading the query explanation.';
                     })
@@ -498,8 +497,6 @@
             });
 
             for (i = 0; i < items.length; ++i) {
-                items[i].classList.remove(i % 2 ? 'even' : 'odd');
-                items[i].classList.add(i % 2 ? 'odd' : 'even');
                 target.appendChild(items[i]);
             }
         }
--- a/Resources/views/Collector/db.html.twig
+++ b/Resources/views/Collector/db.html.twig
@@ -447,9 +447,17 @@
             var targetElement = document.getElementById(targetId);
 
             if (targetElement.style.display != 'block') {
-                Sfjs.load(targetId, link.href, null, function(xhr, el) {
-                    el.innerHTML = 'An error occurred while loading the query explanation.';
-                });
+                if (targetElement.getAttribute('data-sfurl') !== link.href) {
+                    fetch(link.href, {
+                        headers: {'X-Requested-With': 'XMLHttpRequest'}
+                    }).then(async function (response) {
+                        targetElement.innerHTML = await response.text()
+                        targetElement.setAttribute('data-sfurl', link.href)
+                        targetElement.classList.remove('loading')
+                    }, function () {
+                        targetElement.innerHTML = 'An error occurred while loading the query explanation.';
+                    })
+                }
 
                 targetElement.style.display = 'block';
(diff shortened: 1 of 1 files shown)


stof:
Psalm errors are probably caused by the release of DBAL 4 and/or ORM 3.0, where the CI job running the static analysis uses the new major versions.

Resources/views/Collector/db.html.twig
```diff
@@ -447,9 +447,17 @@
             var targetElement = document.getElementById(targetId);
 
             if (targetElement.style.display != 'block') {
-                Sfjs.load(targetId, link.href, null, function(xhr, el) {
-                    el.innerHTML = 'An error occurred while loading the query explanation.';
-                });
+                if (targetElement.getAttribute('data-sfurl') !== link.href) {
+                    fetch(link.href, {
+                        headers: {'X-Requested-With': 'XMLHttpRequest'}
```
ostrolucky, line 452: what's this header needed for?

Resources/views/Collector/db.html.twig
```diff
@@ -447,9 +447,17 @@
             var targetElement = document.getElementById(targetId);
 
             if (targetElement.style.display != 'block') {
-                Sfjs.load(targetId, link.href, null, function(xhr, el) {
-                    el.innerHTML = 'An error occurred while loading the query explanation.';
-                });
+                if (targetElement.getAttribute('data-sfurl') !== link.href) {
```
ostrolucky, line 450: what is this conditional logic for?

Resources/views/Collector/db.html.twig
```diff
@@ -447,9 +447,17 @@
...
+                    fetch(link.href, {
+                        headers: {'X-Requested-With': 'XMLHttpRequest'}
+                    }).then(async function (response) {
+                        targetElement.innerHTML = await response.text()
+                        targetElement.setAttribute('data-sfurl', link.href)
+                        targetElement.classList.remove('loading')
```
ostrolucky: what is adding this class?

Resources/views/Collector/db.html.twig
stof: this conditional logic is the same one than used by the old Symfony code. Its effect is to avoid loading the content with Ajax again if you hide the explanation block and display it again (without reloading the page)

Resources/views/Collector/db.html.twig
stof: I'm not sure about it. The Symfony code we were using was removing it, but it seems to be a left-over from older versions.

Resources/views/Collector/db.html.twig
stof: Our code relies on `$request->isXmlHttpRequest()` to decide whether the rendered content should extend the ajax layout of the profiler (which only contains a `{% block panel %}` to avoid rendering the full page or no) or the full layout. This method is based on this header to detect that (which is a de-facto standard shared by many JS frameworks in the past, like jQuery, Prototype and Mootools to allow telling the backend that this is an ajax request).
the `Sfjs.load()` function available in older Symfony versions was also passing this header.

ostrolucky (approved):

dmaicher (approved):

Resources/views/Collector/db.html.twig
```diff
@@ -490,15 +498,15 @@
             });
 
             for (i = 0; i < items.length; ++i) {
-                Sfjs.removeClass(items[i], i % 2 ? 'even' : 'odd');
-                Sfjs.addClass(items[i], i % 2 ? 'odd' : 'even');
+                items[i].classList.remove(i % 2 ? 'even' : 'odd');
+                items[i].classList.add(i % 2 ? 'odd' : 'even');
```
Seb33300: I think we can just remove them, `even` and `odd` classes seem to be no longer used.

PR #1749, opened 2024-02-06
Fix sorting profiler columns

Seb33300:
While using the profiler, I just noticed that sorting columns by clicking on the header was not working.

This happens because `Sfjs` is no longer available on that page.
The new profiler design seems to not require even/odd css classes, we can just remove them.

The file contains 2 others occurrences of `Sfjs` that should also be fixed, if someone wants to investigate and create another PR for them.


Seb33300:
Just saw Stof PR...

PR #1752, opened 2024-02-07
Improve support doctrine/orm 3.0

olegpro:
After update doctrine/orm to 3.0 version:

```
PHP Fatal error:  Uncaught TypeError: Doctrine\ORM\Mapping\Driver\XmlDriver::__construct(): Argument #2 ($fileExtension) must be of type string, null given, called in /app/var/cache/dev/ContainerN1ivQrS/App_KernelDevDebugContainer.php on line 889 and defined in /app/vendor/doctrine/orm/src/Mapping/Driver/XmlDriver.php:47
Stack trace:
#0 /app/var/cache/dev/ContainerN1ivQrS/App_KernelDevDebugContainer.php(889): Doctrine\ORM\Mapping\Driver\XmlDriver->__construct(Object(Doctrine\Persistence\Mapping\Driver\SymfonyFileLocator), NULL, true)
#1 /app/var/cache/dev/ContainerN1ivQrS/App_KernelDevDebugContainer.php(855): ContainerN1ivQrS\App_KernelDevDebugContainer::getDoctrine_Orm_DefaultEntityManagerService(Object(ContainerN1ivQrS\App_KernelDevDebugContainer), Object(ContainerN1ivQrS\EntityManagerGhostEbeb667))
#2 /app/vendor/symfony/var-exporter/Internal/LazyObjectState.php(61): ContainerN1ivQrS\App_KernelDevDebugContainer::ContainerN1ivQrS\{closure}(Object(ContainerN1ivQrS\EntityManagerGhostEbeb667))
#3 /app/vendor/symfony/var-exporter/LazyGhostTrait.php(118): Symfony\Component\VarExporter\Internal\LazyObjectState->initialize(Object(ContainerN1ivQrS\EntityManagerGhostEbeb667), 'config', 'Doctrine\\ORM\\En...')
#4 /app/vendor/doctrine/orm/src/EntityManager.php(513): ContainerN1ivQrS\EntityManagerGhostEbeb667->__get('config')
#5 /app/vendor/symfony/doctrine-bridge/CacheWarmer/ProxyCacheWarmer.php(45): Doctrine\ORM\EntityManager->getConfiguration()
#6 /app/vendor/symfony/http-kernel/CacheWarmer/CacheWarmerAggregate.php(99): Symfony\Bridge\Doctrine\CacheWarmer\ProxyCacheWarmer->warmUp('/app/var/cache/...', '/app/var/cache/...')
```

doctrine/orm 3.0:

```php
class XmlDriver extends FileDriver
[...]

PR #1753, opened 2024-02-10
Fix the DoctrineOrmMappingsPass passing null as the file extension argument of the XmlDriver

HedicGuibert:
The `createXmlMappingDriver` method of the DoctrineOrmMappingsPass is passing null as the file extension argument, while the [XmlDriver](https://github.com/doctrine/orm/blob/3.0.x/src/Mapping/Driver/XmlDriver.php#L49) expects it to be a string, generating a TypeError.

The error was raised only by trying to upgrade to orm 3.0. I did not have any issue in 2.18.
This is however quite concerning since it prevents the usage of a bundle like [Messenger Monitor](https://github.com/zenstruck/messenger-monitor-bundle/blob/1.x/src/ZenstruckMessengerMonitorBundle.php#L28) on a project using doctrine/orm 3.0


ostrolucky:
Dupe with https://github.com/doctrine/DoctrineBundle/pull/1752

ostrolucky:
Well, but it's a better commit message at least. So I'll merge this

PR #1755, opened 2024-02-12
Fix usage of deprecated doctrine/orm StaticPHP driver factory

PedroTroller:
## The background

I'm using the latest version of this bundle and the `staticphp` driver for entity mapping. I'm trying to upgrade to doctrine/orm:3.0.0 and this version of the ORM removes the deprecated `Doctrine\ORMapping\Driver\StaticPHPDriver` class (replaced by `Doctrine\Persistence\Mapping\Driver\StaticPHPDriver`).

## The problem

In the default DIC settings containing the driver class name `staticphp` still points to the old class that was deleted by doctrine/orm:3.0.0, it crashes completely after migration

```
Internal error: Class "Doctrine\ORM\Mapping\Driver\StaticPHPDriver"
```

## The solution

Stop referring to the old class and switch to the new one.

### DIFF
--- a/Resources/config/orm.xml
+++ b/Resources/config/orm.xml
@@ -34,7 +34,7 @@
         <parameter key="doctrine.orm.metadata.xml.class">Doctrine\ORM\Mapping\Driver\SimplifiedXmlDriver</parameter>
         <parameter key="doctrine.orm.metadata.yml.class">Doctrine\ORM\Mapping\Driver\SimplifiedYamlDriver</parameter>
         <parameter key="doctrine.orm.metadata.php.class">Doctrine\ORM\Mapping\Driver\PHPDriver</parameter>
-        <parameter key="doctrine.orm.metadata.staticphp.class">Doctrine\ORM\Mapping\Driver\StaticPHPDriver</parameter>
+        <parameter key="doctrine.orm.metadata.staticphp.class">Doctrine\Persistence\Mapping\Driver\StaticPHPDriver</parameter>
         <parameter key="doctrine.orm.metadata.attribute.class">Doctrine\ORM\Mapping\Driver\AttributeDriver</parameter>
 
         <!-- cache warmer -->


ostrolucky:
My only concern here is that this requires doctrine/persistence 3, but this project supports doctrine/persistence ^2.2. I can't see any reason why we cannot bump this requirement, though. Anyone?

stof:
We should detect the existence of the new class in the DoctrineExtension to keep support for both persistence 2.x (by replacing the value of the parameter). This way, we would keep the same range of support.

ostrolucky:
But why do we still need persistence 2.x?

stof:
Based on https://packagist.org/packages/doctrine/persistence/stats, doctrine/persistence 2.x still has quite some installs. But maybe we could still drop it.

what is sure is that we cannot drop it in 2.11.x. Dropping support for a major version of one of our dependencies should be done in a minor version rather than a patch.

ostrolucky:
I also wonder what's even reading these .class parameters. I don't see anything in project referencing them. Perhaps it's some external bundle, or project itself. They shouldn't rely on these .class parameters, they were deprecated long time ago.

stof:
They are used in the base class of the DI extension.

ostrolucky:
In that case, fix should be in doctrine-bridge. It shouldn't read these params anymore. https://github.com/doctrine/DoctrineBundle/blob/2.9.x/UPGRADE-2.1.md#upgrade-from-20-to-21

ostrolucky:
closing as explained

clementvtrd:
@ostrolucky, apologies for the duplicate post 🙏.
Do you have any insight on resolving the issue with the symfony/doctrine-bridge package? I attempted to understand how parameters could be replaced, but I was unsuccessful in identifying the parameter name being referenced.

ostrolucky:
Search for `addDriver` calls in symfony repo

MatTheCat:
Don’t know if things changed since this PR was opened but it really seems it is DoctrineBundle’s `DoctrineExtension` using these parameters, like in https://github.com/doctrine/DoctrineBundle/blob/7bfb8e834aa1e82d107d5f7cfa139c0c5eb45581/src/DependencyInjection/DoctrineExtension.php#L1166

EDIT: seems like https://github.com/symfony/symfony/pull/33319 got no follow-up :sweat_smile:

PR #1756, opened 2024-02-14
Fix multiple definitions

avinmaster:


Repository/ServiceEntityRepositoryOrm2.php
```diff
@@ -0,0 +1,26 @@
...
+ *     public function __construct(ManagerRegistry $registry)
+ *     {
+ *         parent::__construct($registry, YourEntity::class);
+ *     }
+ * }
+ *
+ * @template T of object
+ * @template-extends LazyServiceEntityRepository<T>
+ */
+class ServiceEntityRepositoryOrm2 extends LazyServiceEntityRepository
```
stof, line 24: This class must be tagged as `@internal` so that it is not referenced directly.

Repository/ServiceEntityRepository.php
```diff
@@ -8,46 +8,8 @@
...
-     *     }
-     * }
-     *
-     * @template T of object
-     * @template-extends LazyServiceEntityRepository<T>
-     */
-    class ServiceEntityRepository extends LazyServiceEntityRepository
-    {
-    }
+    class_alias(ServiceEntityRepositoryOrm2::class, 'ServiceEntityRepository');
```
stof, line 11: Defining only class_alias in the file means that the Composer class loader won't discover the class name when generating an optimized class loader

stof, line 11: This call is broken. The second argument should be the class name, which must not be `ServiceEntityRepository` but `Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository` (which can be written as `ServiceEntityRepository::class`)

stof (commented):
What is this actually fixing ?

Repository/ServiceEntityRepository.php
avinmaster: Hi @stof, I will prepare everything in 2-3 days. I just created a PR for you to look at and give comments before I started working on it. The problem here is that some analyzers cannot process this class correctly and this issue needs to be corrected somehow. Psalm, PHPStan and in general the tools of PHPStorm itself do not process correctly (phpcs too). I need to understand whether the problem is really in the code or in the handlers.

ostrolucky:
We can reopen once there is an update

ro0NL:
https://github.com/doctrine/DoctrineBundle/issues/1677

it breaks psalm yes

PR #1759, opened 2024-02-18
Update priority of built-in middleware

l-vo:
With the default priority, middlewares for logging and profiler don't see the eventual query changes performed by userland middlewares (that have by default a priority of 0)


l-vo:
Replaced by #1760

PR #1760, opened 2024-02-18
Update priority of built-in middleware

l-vo:
With the default priority, middlewares for logging and profiler don't see the eventual query changes performed by userland middlewares (that have by default a priority of 0)

### DIFF
--- a/DependencyInjection/DoctrineExtension.php
+++ b/DependencyInjection/DoctrineExtension.php
@@ -1187,14 +1187,14 @@ class DoctrineExtension extends AbstractDoctrineExtension
 
         $loggingMiddlewareAbstractDef = $container->getDefinition('doctrine.dbal.logging_middleware');
         foreach ($connWithLogging as $connName) {
-            $loggingMiddlewareAbstractDef->addTag('doctrine.middleware', ['connection' => $connName]);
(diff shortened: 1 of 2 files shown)


ostrolucky:
Nature of this is dubious, plus potentially BC breaking. Targetting 2.11.x. Not sure it should have prio 100 either, that's quite high. Something like `10` would be better I reckon.

l-vo:
@ostrolucky yes it can be 10, I was not sure about the number.

Why do you think it's dubious ?

ostrolucky:
Because doctrine-bundle never claimed to log everything, including custom middlewares. For some people it might be intended that queries in custom middlewares are not logged.

l-vo:
@ostrolucky it makes sense 👍

So I close it, thank you for your feedback :)

ostrolucky:
Well I agree with the change, my only argument was to not classify this as a bugfix, which I did ;)

l-vo:
But there is also the BC break problem, isn't it ?

ostrolucky:
It's a potential break for someone. But middleware priorities is not something we promise not to change.

l-vo:
@ostrolucky ok I see 👍 I reopened it with the priority set to 10. Thank you for your feedbacks :)

PR #1761, opened 2024-02-20
Fix doctrine:database:create and doctrine:database:drop for Postgresql on DBAL 4.x

### LINKED ISSUE #1757: Create and drop PostgreSQL database fails
Hello,
With ORM 3, `doctrine:database:create` and `doctrine:database:drop` failed.

ostrolucky:
Ref https://github.com/doctrine/dbal/pull/5705
Fixes #1757

### DIFF
--- a/Command/CreateDatabaseDoctrineCommand.php
+++ b/Command/CreateDatabaseDoctrineCommand.php
@@ -3,6 +3,7 @@
 namespace Doctrine\Bundle\DoctrineBundle\Command;
 
 use Doctrine\DBAL\DriverManager;
+use Doctrine\DBAL\Platforms\PostgreSQLPlatform;
 use InvalidArgumentException;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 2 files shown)


Command/CreateDatabaseDoctrineCommand.php
```diff
@@ -64,6 +65,10 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         /** @psalm-suppress InvalidArrayOffset Need to be compatible with DBAL < 4, which still has `$params['url']` */
         unset($params['dbname'], $params['path'], $params['url']);
 
+        if ($connection->getDatabasePlatform() instanceof PostgreSQLPlatform) {
+            $params['dbname'] = 'postgres';
+        }
```
greg0ire, line 70: Why do we unset `dbname` on line 66, only to set it again here? Also, I don't understand the comment on line 64, should it read "as we have already extracted" instead of "and we have already extracted"?

ostrolucky: For the first point, it's because I assume unsetting dbname might still be needed for other db platforms. For the second point: I guess so.

greg0ire (approved):

PR #1764, opened 2024-03-07
Change type hint from `Registry` to `ManagerRegistry` in `ProfilerController`

priyadi:
The previous code breaks decorating `ManagerRegistry`. `ProfilerController` only uses `getConnection()` which is in the interface, so it is not a problem. And it is also the only remaining place in the package where `Registry` is used as the type hint instead of `ManagerRegistry`.

### DIFF
--- a/Controller/ProfilerController.php
+++ b/Controller/ProfilerController.php
@@ -6,7 +6,7 @@ use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Platforms\OraclePlatform;
 use Doctrine\DBAL\Platforms\SQLitePlatform;
 use Doctrine\DBAL\Platforms\SQLServerPlatform;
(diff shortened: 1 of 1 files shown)


dmaicher (approved):
LGTM

ostrolucky:
Ideally we should use `\Doctrine\Persistence\ConnectionRegistry` here, since that method is defined there

priyadi:
> Ideally we should use `\Doctrine\Persistence\ConnectionRegistry` here, since that method is defined there

changed `ManagerRegistry` to `ConnectionRegistry`

dmaicher (approved):

dmaicher:
@ostrolucky do we consider this a bugfix for 2.11.x? or an improvement for 2.12.x?

ostrolucky:
Let's do 2.12. I can't see how is this a bugfix

PR #1767, opened 2024-03-14
Use a modern directory structure

greg0ire:
The [recommended](https://symfony.com/doc/current/bundles.html#bundle-directory-structure) directory structure is more similar to what we do in non-bundle repositories.
Also, this means smaller configuration for tools such as PHPCSStandards, PHPUnit or Psalm.

Although this is no bugfix, I'm targeting 2.11.x so that merging up does not become a nightmare.


greg0ire:
The Psalm failures demonstrate another reason why this is better: the `Middleware` directory was forgotten in this list:
https://github.com/doctrine/DoctrineBundle/blob/7150c1de28bcb3b73c1dc1017eda37582228dc85/psalm.xml.dist#L23-L33

ostrolucky:
yes, this is welcome

greg0ire:
I'm unsure where the 2 svg files should go. Should they go into `public`, or should I keep them under `templates/Collector`? They are referenced in https://github.com/doctrine/DoctrineBundle/blob/7150c1de28bcb3b73c1dc1017eda37582228dc85/Resources/views/Collector/db.html.twig#L11-L15

ostrolucky:
Keep them in templates. And let's target next minor.

greg0ire:
Will address the conflicts once #1768 is merged and merged up.

composer.json
```diff
@@ -81,12 +81,13 @@
...
     "autoload": {
         "psr-4": {
-            "Doctrine\\Bundle\\DoctrineBundle\\": ""
+            "Doctrine\\Bundle\\DoctrineBundle\\": "src"
         }
     },
     "autoload-dev": {
         "psr-4": {
-            "": "Tests/DependencyInjection"
+            "": "tests/DependencyInjection",
```
stof, line 89: As this seems to be about autoloading the `Fixtures` namespace, I would suggest using `Fixtures\\` as the PSR-4 prefix (and going one level deeper for the path) instead of a fallback for any class.

stof (changes requested):
To support the modern structure, the DoctrineBundle class needs to be updated to override the `getPath()` method (or to extend AsbtractBundle which already does the override) as the bundle class is not at the root path of the bundle anymore.

composer.json
greg0ire: 👍 will do in a separate PR

ostrolucky (approved):

greg0ire:
> To support the modern structure, the DoctrineBundle class needs to be updated to override the getPath() method (or to extend AsbtractBundle which already does the override) as the bundle class is not at the root path of the bundle anymore.

@ostrolucky why did you merge? Is that not an issue?

greg0ire:
Addressed in https://github.com/doctrine/DoctrineBundle/pull/1770

PR #1768, opened 2024-03-14
Analyze Middleware directory

greg0ire:
Spotted while working on https://github.com/doctrine/DoctrineBundle/pull/1767

### DIFF
--- a/psalm.xml.dist
+++ b/psalm.xml.dist
@@ -28,6 +28,7 @@
         <directory name="DependencyInjection"/>
         <directory name="EventSubscriber"/>
         <directory name="Mapping"/>
+        <directory name="Middleware"/>
         <directory name="Repository"/>
         <directory name="Tests"/>
         <directory name="Twig"/>
--- a/Middleware/BacktraceDebugDataHolder.php
+++ b/Middleware/BacktraceDebugDataHolder.php
@@ -17,7 +17,7 @@ class BacktraceDebugDataHolder extends DebugDataHolder
     /** @var string[] */
     private array $connWithBacktraces;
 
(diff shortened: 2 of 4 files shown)


psalm.xml.dist
```diff
@@ -37,6 +38,18 @@
...
+                <!-- Random package should not call this class from the Doctrine bridge, but we are in the Doctrine bundle here -->
+                <referencedClass name="Symfony\Bridge\Doctrine\Middleware\Debug\Driver"/>
+            </errorLevel>
+        </InternalClass>
+        <InternalMethod>
+            <errorLevel type="suppress">
+                <!-- Random package should not call this method from the Doctrine bridge, but we are in the Doctrine bundle here -->
+                <referencedMethod name="Symfony\Bridge\Doctrine\Middleware\Debug\Driver::__construct"/>
+            </errorLevel>
+        </InternalMethod>
```
ostrolucky: do these not work as `@psalm-suppress`?

SenseException: We do have a lot `psalm-suppres` in the code and we do have a lot `suppress` in the psalm config. Is there a pattern in DoctrineBundle when to use one or the other?

ostrolucky: Pattern is that we put stuff into config that does not work with inline suppress. There are plenty of such cases when psalm-suppress doesn't have an effect but putting it to config does.

greg0ire: > do these not work as `@psalm-suppress`?

I'll try, it's just that I forgot that we were doing things like this in that repository.

greg0ire: It works.

ostrolucky (approved):

PR #1770, opened 2024-03-15
Override Bundle::getPath()

greg0ire:
This is required when the bundle class is not at the root path of the
bundle, which is the case since https://github.com/doctrine/DoctrineBundle/pull/1767

### DIFF
--- a/src/DoctrineBundle.php
+++ b/src/DoctrineBundle.php
@@ -30,6 +30,7 @@ use Symfony\Component\HttpKernel\Bundle\Bundle;
 use function assert;
 use function class_exists;
 use function clearstatcache;
+use function dirname;
 use function spl_autoload_unregister;
 
 /** @final since 2.9 */
@@ -167,5 +168,10 @@ class DoctrineBundle extends Bundle
     /** @return void */
     public function registerCommands(Application $application)
     {
+    }
+
+    public function getPath(): string
+    {
+        return dirname(__DIR__);
     }
 }


stof (approved):

src/DoctrineBundle.php
```diff
@@ -168,4 +170,15 @@ public function shutdown()
     public function registerCommands(Application $application)
     {
     }
+
+    public function getPath(): string
+    {
+        /** @psalm-suppress RedundantPropertyInitializationCheck */
+        if (! isset($this->path)) {
+            $reflected  = new ReflectionObject($this);
+            $this->path = dirname($reflected->getFileName(), 2);
```
greg0ire: @stof I initially copy/pasted the wrong method, I believe the correct thing is to have this `2` here, since we are one directory deeper now.

stof: we could simplify the implementation a lot with `return dirname(__DIR__)` here instead of using Reflection to find the source of the class.

ostrolucky: not a fan of `dirname()`, since it's a function call right? Why not just `__DIR__.'/..'`?

And let's do one better and remove caching like so
```php
public function getPath(): string
{
  return `__DIR__.'/..';
}
```

since caching such a simple logic to `$path` with no method calls is pointless

src/DoctrineBundle.php
```diff
@@ -168,4 +170,14 @@ public function shutdown()
     public function registerCommands(Application $application)
     {
     }
+
+    public function getPath(): string
+    {
+        /** @psalm-suppress RedundantPropertyInitializationCheck */
+        if (! isset($this->path)) {
```
stof: I would even not care about assigning the `path` property. This is not needed (the base class only uses it as a cache for its `getPath` implementation that relies on Reflection)

src/DoctrineBundle.php
stof: @ostrolucky `dirname` with a known path (like `__DIR__`) gets computed at compile-time by OPCache, so no overhead.

`__DIR__.'/..'` will not be the absolute path, which might cause issues if some places are checking for paths starting with this path.

ostrolucky (approved):

src/DoctrineBundle.php
ostrolucky: I see. Thanks for explanation. Sorry @greg0ire !

ostrolucky (changes requested):
let's re-add dirname(), like stof explained

ostrolucky (approved):

Issue #1758, opened 2024-02-15
DBAL 4 `setNestTransactionsWithSavepoints` deprecation


dmaicher:
With the latest [recipe changes ](https://github.com/symfony/recipes/pull/1290) we are always explicitly enabling savepoints. 

However on DBAL 4 this leads to 

>    44x: Doctrine\DBAL\Connection::setNestTransactionsWithSavepoints is deprecated and will be removed in 5.0 (Connection.php:948 called by App_KernelTestDebugContainer.php:611, https://github.com/doctrine/dbal/pull/5383, package doctrine/dbal)

How should we handle this? Revert the recipe change? Somehow fix it here?

Also see https://github.com/symfony/demo/pull/1494 where I removed `use_savepoints: true` for now.

ostrolucky:
Yeah I would say it should just be removed from recipe. Recipes are mostly installed first time when the bundle is installed, which is usually when project is new, in that case newest orm and dbal would be installed.

PR #1773, opened 2024-03-17
make "use_savepoints = true" no-op with DBAL 4

dmaicher:
While working on some bundles that support DBAL 3 and DBAL 4 I noticed that its quite a struggle to activate savepoints in a deprecation free way :confused: 

So I agree with @derrabus (see https://github.com/symfony/recipes/pull/1290#issuecomment-1947418192) and propose to make this no-op with DBAL 4.

Fixes https://github.com/doctrine/DoctrineBundle/issues/1758

### DIFF
--- a/src/DependencyInjection/DoctrineExtension.php
+++ b/src/DependencyInjection/DoctrineExtension.php
@@ -307,7 +307,7 @@ class DoctrineExtension extends AbstractDoctrineExtension
                     $def->addMethodCall('setNestTransactionsWithSavepoints', [$connection['use_savepoints']]);
                 }
             } elseif (! $connection['use_savepoints']) {
-                throw new LogicException('The "use_savepoints" option can only be set to "true" when using DBAL >= 4');
+                throw new LogicException('The "use_savepoints" option can only be set to "true" and should ideally not be set when using DBAL >= 4');
             }
         }
 
--- a/src/DependencyInjection/DoctrineExtension.php
+++ b/src/DependencyInjection/DoctrineExtension.php
@@ -300,8 +300,15 @@ class DoctrineExtension extends AbstractDoctrineExtension
             $def->setClass($options['wrapperClass']);
         }
 
-        if (! empty($connection['use_savepoints'])) {
-            $def->addMethodCall('setNestTransactionsWithSavepoints', [$connection['use_savepoints']]);
+        if (isset($connection['use_savepoints'])) {
+            // DBAL >= 4 always has savepoints enabled. So we only need to call "setNestTransactionsWithSavepoints" for DBAL < 4
+            if (method_exists(Connection::class, 'getEventManager')) {
+                if ($connection['use_savepoints']) {
+                    $def->addMethodCall('setNestTransactionsWithSavepoints', [$connection['use_savepoints']]);
+                }
+            } elseif (! $connection['use_savepoints']) {
+                throw new LogicException('The "use_savepoints" option can only be set to "true" when using DBAL >= 4');
+            }
         }
 
         $container->setDefinition(
--- a/tests/DependencyInjection/AbstractDoctrineExtensionTest.php
+++ b/tests/DependencyInjection/AbstractDoctrineExtensionTest.php
@@ -240,6 +240,10 @@ abstract class AbstractDoctrineExtensionTest extends TestCase
 
     public function testDbalLoadSavepointsForNestedTransactions(): void
     {
+        if (!method_exists(Connection::class, 'getEventManager')) {
+            self::markTestSkipped('This test requires DBAL < 4');
+        }
+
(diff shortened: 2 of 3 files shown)


ostrolucky:
i'm fine with this

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -300,8 +300,15 @@ protected function loadDbalConnection($name, array $connection, ContainerBuilder
             $def->setClass($options['wrapperClass']);
         }
 
-        if (! empty($connection['use_savepoints'])) {
-            $def->addMethodCall('setNestTransactionsWithSavepoints', [$connection['use_savepoints']]);
+        if (isset($connection['use_savepoints'])) {
+            // DBAL >= 4 always has savepoints enabled. So we only need to call "setNestTransactionsWithSavepoints" for DBAL < 4
+            if (method_exists(Connection::class, 'getEventManager')) {
+                if ($connection['use_savepoints']) {
```
dmaicher, line 306: due to `empty` it was actually never calling `setNestTransactionsWithSavepoints(false)`. So I kept this behavior the same

greg0ire (commented):
Maybe you should also conditionally deprecate the `use_savepoints` configuration node.

dmaicher:
> Maybe you should also conditionally deprecate the `use_savepoints` configuration node.

Hm but then we still have deprecations? The idea was to just have `use_savepoins = true` without any deprecations

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -300,8 +300,15 @@ protected function loadDbalConnection($name, array $connection, ContainerBuilder
...
+                    $def->addMethodCall('setNestTransactionsWithSavepoints', [$connection['use_savepoints']]);
+                }
+            } elseif (! $connection['use_savepoints']) {
+                throw new LogicException('The "use_savepoints" option can only be set to "true" when using DBAL >= 4');
```
dmaicher: Note that this is technically a breaking change for people on DBAL 4 using `use_savepoints = false`. As `setNestTransactionsWithSavepoints` was never called in those cases it did not result in any error. But this seems odd to me so I think having an error in those cases is the correct behavior.

greg0ire:
> Hm but then we still have deprecations? The idea was to just have use_savepoins = true without any deprecations

Do you mean with DBAL 3 or 4? If not now, when do you plan to deprecate that configuration node?

ostrolucky:
@greg0ire Has a point. There is no point having use_savepoins (neither true nor false) in DBAL 4, because it's not configurable. When would we be able to remove this config node if people are fine with using it with dbal 4?

What's your struggle in configuring bundles supporting dbal3+4? With PHP config or TestKernel it should be quite easy to have dynamic switch.

src/DependencyInjection/DoctrineExtension.php
greg0ire:
```suggestion
                throw new LogicException('The "use_savepoints" option can only be set to "true" and should ideally not be set when using DBAL >= 4');
```

dmaicher:
I was thinking to deprecate it once we drop DBAL 3 support :thinking: 

> With PHP config or TestKernel it should be quite easy to have dynamic switch.

Yeah its doable but still a bit of a hassle :cry:

ostrolucky:
I'm just thinking out loud: why don't we turn on savepoints by default? Do you think there would be some side-effects?

dmaicher:
> I'm just thinking out loud: why don't we turn on savepoints by default? Do you think there would be some side-effects?

Possibly there could be some behavioral changes when doing nested transactions with rollbacks or so :thinking: 

Checking

https://github.com/doctrine/dbal/blob/db922ba9436b7b18a23d1653a0b41ff2369ca41c/src/Connection.php#L1380
https://github.com/doctrine/dbal/blob/db922ba9436b7b18a23d1653a0b41ff2369ca41c/src/Connection.php#L1541

@greg0ire do you think it would be fine to enable savepoints by default in a minor update?

greg0ire:
I don't think we should do that in behalf of users, no, otherwise we would have made `setNestTransactionsWithSavepoints` a no-op in DBAL 3 directly

dmaicher:
It would be nice if new projects (using DBAL 4) will not immediately see a deprecation caused by https://github.com/symfony/recipes/pull/1290

ostrolucky (approved):

greg0ire:
> It would be nice if new projects (using DBAL 4) will not immediately see a deprecation caused by https://github.com/symfony/recipes/pull/1290

Maybe it's time to create a new major version of the bundle, that would drop DBAL 3 & 4, while still maintaining this one for a while, just like we do for ORM and DBAL? That way you could have a dedicated recipe.

greg0ire (approved):

PR #1776, opened 2024-03-19
update metadata for new 2.12 release

dmaicher:


.symfony.bundle.yaml
```diff
@@ -7,9 +7,10 @@ branches:
     - "2.10.x"
     - "2.11.x"
     - "2.12.x"
+    - "2.13.x"
 maintained_branches:
-    - "2.11.x"
     - "2.12.x"
+    - "2.13.x"
 doc_dir: { "2.11.x": "Resources/doc/", "2.12.x": "docs/" }
```
greg0ire, line 14: Shouldn't this be simplified back to just `docs`? Since 2.11.x is no longer maintained? Cc @javiereguiluz

PR #1777, opened 2024-03-19
adapt .gitattributes for the new dir structure

dmaicher:


.gitattributes
```diff
@@ -5,5 +5,6 @@
 phpunit.xml.dist       export-ignore
 phpcs.xml.dist         export-ignore
 psalm.xml.dist         export-ignore
-/Resources/doc         export-ignore
-/Tests                 export-ignore
+/docs                  export-ignore
+/tests                 export-ignore
+.symfony.bundle.yaml   export-ignore
```
dmaicher, line 10: is it fine to exclude this? probably its not used with releases, or?

greg0ire: cc @javiereguiluz

javiereguiluz: Yes, we can exclude it. In symfony.com we use GitHub API to get the file contents directly from the default branch of the repository. Thanks.

greg0ire (approved):

PR #1778, opened 2024-03-19
Resolve controller resolver auto_mapping test deprecations

bobvandevijver:
Fixes the introduced test deprecations caused by the changed default of the controller resolver auto mapping value. Lines that have been changed to reflect the new default value have been annotated with a todo so they can be remove when preparing the 3.0 release.

As requested by @dmaicher in https://github.com/doctrine/DoctrineBundle/pull/1762#issuecomment-2005883237.

### DIFF
--- a/tests/Builder/BundleConfigurationBuilder.php
+++ b/tests/Builder/BundleConfigurationBuilder.php
@@ -5,7 +5,11 @@ namespace Doctrine\Bundle\DoctrineBundle\Tests\Builder;
 class BundleConfigurationBuilder
 {
     /** @var array<string, mixed> */
-    private array $configuration = [];
+    private array $configuration = [ // todo: revert to empty array with 3.x release
+        'orm' => [
+            'controller_resolver' => ['auto_mapping' => false],
+        ],
+    ];
 
     public static function createBuilder(): self
     {
@@ -71,7 +75,9 @@ class BundleConfigurationBuilder
     /** @param array<string, mixed> $config */
     public function addEntityManager(array $config): self
     {
-        $this->configuration['orm'] = $config;
+        foreach ($config as $key => $value) {
+            $this->configuration['orm'][$key] = $value;
+        }
 
         return $this;
     }
--- a/tests/Builder/BundleConfigurationBuilder.php
+++ b/tests/Builder/BundleConfigurationBuilder.php
@@ -2,14 +2,24 @@
 
 namespace Doctrine\Bundle\DoctrineBundle\Tests\Builder;
 
+use Doctrine\ORM\EntityManager;
+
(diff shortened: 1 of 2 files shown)


bobvandevijver:
@dmaicher I am not sure if this is the right approach. I thought there were only deprecations in the [DoctrineExtensionTest](https://github.com/doctrine/DoctrineBundle/pull/1778/files#diff-95cb7e648b59313419d71246ec5fa0a5c4b6c917bd286b3c71f13d826c9e5b2a), but they are all over the place, also in tests that do not touch the configuration at all (for example the [ContainerTest](https://github.com/doctrine/DoctrineBundle/blob/5418e811a14724068e95e0ba43353b903ada530f/tests/ContainerTest.php)).

All these deprecation messages disappear automatically when the default value is being changed in 3.0. Would it be acceptable to keep the deprecation notices, and only keep the single real test change (https://github.com/doctrine/DoctrineBundle/pull/1778/files#diff-95cb7e648b59313419d71246ec5fa0a5c4b6c917bd286b3c71f13d826c9e5b2aR635) so that the tests stay green after the default has changed?

dmaicher:
Or if its too much of a hassle to fix all of them maybe we can also add a [baseline](https://symfony.com/doc/current/components/phpunit_bridge.html#baseline-deprecations) and ignore them? @ostrolucky wdyt?

bobvandevijver:
That would require a different baseline for every different test run, which is not something I would prefer. Maybe a better option would be to update the `DoctrineExtension` to detect whether PHP unit is running, and then apply the new default already for the tests?

Something like:
```php
                if (defined('PHPUNIT_DOCTRINE_BUNDLE_TESTSUITE')) {
                    $config['controller_resolver']['auto_mapping'] = false;
                } else {
                    trigger_deprecation('doctrine/doctrine-bundle', '2.12', 'The default value of "doctrine.orm.controller_resolver.auto_mapping" will be changed from `true` to `false`. Explicitly configure `true` to keep existing behaviour.');
                    $config['controller_resolver']['auto_mapping'] = true;
                }
```

When combined with the following in the phpunit configuration, that would apply the new default already during tests:

```xml
    <php>
        <const name="PHPUNIT_DOCTRINE_BUNDLE_TESTSUITE" value="true"/>
    </php>
```

ostrolucky:
Thanks for the help. But indeed, looks like we would have to change more places than I'm comfortable. So I've taken the liberty to go with a baseline in the end. It shouldn't require us to have different baseline every different test run either.

bobvandevijver:
Alright, works for me! I will submit another PR with just the change for the test where setting auto mapping to true would actually be relevant for the test to succeed.

PR #1790, opened 2024-04-11
Wire up type aliases support in EntityValueResolver

NanoSector:
This is an addendum to PR https://github.com/symfony/symfony/pull/54545 targeted at issue https://github.com/symfony/symfony/issues/51765 in the Symfony Doctrine Bridge, which adds type alias support to EntityValueResolver.

This code injects the doctrine.orm.resolve_target_entities configuration into the EntityValueResolver class.

### DIFF
--- a/src/DependencyInjection/DoctrineExtension.php
+++ b/src/DependencyInjection/DoctrineExtension.php
@@ -549,19 +549,22 @@ class DoctrineExtension extends AbstractDoctrineExtension
                 $controllerResolverDefaults['evict_cache'] = true;
             }
 
-            if ($controllerResolverDefaults) {
-                $container->getDefinition('doctrine.orm.entity_value_resolver')->setArgument(2, (new Definition(MapEntity::class))->setArguments([
-                    null,
-                    null,
-                    null,
-                    $controllerResolverDefaults['mapping'] ?? null,
-                    null,
-                    null,
-                    null,
-                    $controllerResolverDefaults['evict_cache'] ?? null,
-                    $controllerResolverDefaults['disabled'] ?? false,
-                ]));
-            }
+            $valueResolverDefinition = $container->getDefinition('doctrine.orm.entity_value_resolver');
+
+            $valueResolverDefinition->setArgument(2, (new Definition(MapEntity::class))->setArguments([
+                null,
+                null,
+                null,
+                $controllerResolverDefaults['mapping'] ?? null,
+                null,
+                null,
+                null,
+                $controllerResolverDefaults['evict_cache'] ?? null,
+                $controllerResolverDefaults['disabled'] ?? false,
+            ]));
+
+            // Symfony 7.3 and higher expose type alias support in the EntityValueResolver
+            $valueResolverDefinition->setArgument(3, $config['resolve_target_entities']);
         }
 
         // not available in Doctrine ORM 3.0 and higher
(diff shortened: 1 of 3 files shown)


src/DependencyInjection/DoctrineExtension.php
```diff
@@ -625,6 +626,22 @@ protected function ormLoad(array $config, ContainerBuilder $container)
             $def
                 ->addTag('doctrine.event_listener', ['event' => Events::loadClassMetadata])
                 ->addTag('doctrine.event_listener', ['event' => Events::onClassMetadataNotFound]);
+
+            // Symfony 7.1 and higher expose type alias support in the EntityValueResolver
+            if (class_exists(EntityValueResolver::class)) {
+                $valueResolverReflection = new ReflectionClass(EntityValueResolver::class);
+
+                if ($valueResolverReflection->hasMethod('addTypeAlias')) {
```
nicolas-grekas: I think using a constructor argument instead of an adder can save us from using reflection at all. WDYT?

NanoSector: Initially that is what I tried to do, but I figured explicitly checking support without depending on constructor argument count and order was more desirable. This also seems more in line with other parts of the extension which also call `add*` or `set*` methods. I do agree that reflection is generally undesirable.
Having this as a method also allows for easier extension of the ValueResolver beyond the bundle, since the property holding this mapping is otherwise private; although this can also be achieved with explicit service configuration in the framework.

Also, setting the argument on its position by key doesn't seem to work. Replacing the reflection code in this PR with the following snippet:
```php
$valueResolverDefinition->setArgument(3, $config['resolve_target_entities']);
```

Results in an error: `TypeError: Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver::__construct(): Argument #3 ($defaults) must be of type Symfony\Bridge\Doctrine\Attribute\MapEntity, array given`
Even when using 4 in the above code snippet. Using explicit argument names (`->setArgument('$typeAliases', ...)`) doesn't work as this causes a container compilation error on older bridge versions because the argument does not exist; this would break backwards compatibility (or still need reflection to check the existance of the property/argument).

eltharin: for add an argument, I think you can remove the condition ``if ($controllerResolverDefaults) {`` at l.527 for always declare the mapentity (with default arguments) as arg 2 and then you can decalre a argument 3.

nicolas-grekas: @NanoSector up to give @eltharin's suggestion a try?

NanoSector: @nicolas-grekas Sorry, these PRs must have fallen off my radar. I've rebased them on the latest development branches and applied your and @eltharin 's suggestions.

NanoSector:
@ostrolucky Sorry to bother you, I am wondering if any action is required on my part, considering the PR still has 'on hold' status? Its Symfony counterpart is marked as ready.

ostrolucky:
It's onhold because upstream PR is not merged yet

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -541,19 +541,22 @@ protected function ormLoad(array $config, ContainerBuilder $container)
...
+                null,
+                $controllerResolverDefaults['mapping'] ?? null,
+                null,
+                null,
+                null,
+                $controllerResolverDefaults['evict_cache'] ?? null,
+                $controllerResolverDefaults['disabled'] ?? false,
+            ]));
+
+            // Symfony 7.2 and higher expose type alias support in the EntityValueResolver
```
NanoSector: This is now targeting 7.3, so this should be updated.

NanoSector:
@ostrolucky The [upstream PR](https://github.com/symfony/symfony/pull/54545) has been merged; is there anything left to do on my side?

PR #1803, opened 2024-05-17
[docs] Change `annotation` to `attribute`

JeroenMoonen:


ostrolucky (approved):

greg0ire (commented):
I think there other occurrences of `annotation` that could be replaced as well.

SenseException (changes requested):
> I think there other occurrences of annotation that could be replaced as well.

3 occurrences in the docs:
https://github.com/search?q=repo%3Adoctrine%2FDoctrineBundle+annotation&type=code&p=1
(including the already changed one)

PR #1804, opened 2024-05-30
Fully deprecate controller resolver auto mapping

bobvandevijver:
Symfony 7.1 has deprecated the use of the controller resolver auto mapper functionality in favour of the new mapped route parameters, which have the benefit of being explicit and concise (see https://github.com/symfony/symfony/pull/54455, https://github.com/symfony/symfony/pull/54720 and https://symfony.com/blog/new-in-symfony-7-1-mapped-route-parameters for the discussion and more details).

Note that we already deprecated not setting a value with https://github.com/doctrine/DoctrineBundle/pull/1762 to be able to change its default to false with 3.0, but now I propose to fully remove this configuration option with 3.0 instead, with both not setting and true being deprecated. Depending on when 3.0 would be available, removed the option could be postponed to 4.0 if desired.

/cc @nicolas-grekas - https://github.com/symfony/recipes/pull/1316

### DIFF
--- a/UPGRADE-2.13.md
+++ b/UPGRADE-2.13.md
@@ -1,0 +1,13 @@
+UPGRADE FROM 2.12 to 2.13
+========================
+
+Configuration
+-------------
+
+### Controller resolver auto mapping deprecated
+
+The controller resolver auto mapping functionality has been deprecated with Symfony 7.1, and is replaced with explicit mapped route parameters. Enabling the auto mapper by default using this bundle is now deprecated as well.
+
+Auto mapping uses any route parameter that matches with a field name of the Entity to resolve as criteria in a find by query.
+
+If you are relying on this functionality, you can update your code to use explicit mapped route parameters instead.
--- a/UPGRADE-3.0.md
+++ b/UPGRADE-3.0.md
@@ -4,13 +4,13 @@ UPGRADE FROM 2.x to 3.0
 Configuration
 -------------
 
-### Controller resolver auto mapping default configuration changed
+### Controller resolver auto mapping can no longer be configured
 
-The default value of `doctrine.orm.controller_resolver.auto_mapping` has changed from `true` to `false`.
+The `doctrine.orm.controller_resolver.auto_mapping` option now only accepts `false` as value, to disallow the usage of the controller resolver auto mapping feature by default. The configuration option will be fully removed in 4.0.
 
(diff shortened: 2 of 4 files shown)


UPGRADE-2.13.md
```diff
@@ -0,0 +1,13 @@
+UPGRADE FROM 2.12 to 2.13
+========================
+
+Configuration
+-------------
+
+### Controller resolver auto mapping deprecated
+
+The controller resolver auto mapping functionality has been deprecated with Symfony 7.1, and is replaced with explicit mapped route parameters. Enabling the auto mapper using this bundle is now deprecated as well. The configuration option will be fully removed in 3.0.
```
nicolas-grekas: > The configuration option will be fully removed in 3.0.

Actually, the option should remain in 3.0, but only `false` should be accepted (then in 3.1 we can deprecate it, and remove in 4.0)

bobvandevijver: Fair enough, has been adjusted.

ostrolucky (approved):

ostrolucky (changes requested):
Ah sorry it took a while to review, but tests need to be adjusted

```
Remaining self deprecation notices (100)

  100x: Since doctrine/doctrine-bundle 2.13: Enabling the controller resolver automapping feature has been deprecated. Symfony Mapped Route Parameters should be used as replacement.
    3x in DoctrineExtensionTest::testAutomapping from Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection
    2x in DoctrineExtensionTest::testDependencyInjectionConfigurationDefaults from Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection
    2x in DoctrineExtensionTest::testSingleEntityManagerWithEmptyConfiguration from Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection
    2x in DoctrineExtensionTest::testControllerResolver from Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection
    ...
```

bobvandevijver:
@ostrolucky No worries. I added the message to the `baseline-ignore` only, as I believe that the tests shouldn't be changed at this moment. The warning comes from the current configuration default, so it should be automatically solved when changing that in the future.

edit: ah, you're quick, great :+1:

greg0ire:
@bobvandevijver hi! I just created the 3.0.x branch, and I see you created an `UPGRADE-3.0.md` file :thinking: So, I guess you want to contribute some changes on 3.0.x then?

PR #1816, opened 2024-08-21
Clean LazyServiceEntityRepository

lyrixx:


src/Repository/LazyServiceEntityRepository.php
```diff
@@ -34,13 +34,7 @@ public function __construct(ManagerRegistry $registry, string $entityClass)
 
         if ($this instanceof LazyObjectInterface) {
             $this->initialize();
-
-            return;
         }
-
-        unset($this->_em);
-        unset($this->_class);
-        unset($this->_entityName);
```
lyrixx, line 43: Hmm, I think I got it. It exists with ORM 2, but not 3!

So we must keep it

PR #1826, opened 2024-10-06
Add a "run query" option in the profiler

Louis-Proffit:
Inside the "view runnable query" panel, a run button enables running the query and checking the results just like "explain query" does

### DIFF
--- a/src/Controller/ProfilerController.php
+++ b/src/Controller/ProfilerController.php
@@ -83,7 +83,6 @@ class ProfilerController
         ]));
     }
 
-
     /**
      * Renders the profiler panel for the given token.
      *
--- a/src/Controller/ProfilerController.php
+++ b/src/Controller/ProfilerController.php
@@ -87,9 +87,9 @@ class ProfilerController
     /**
      * Renders the profiler panel for the given token.
      *
-     * @param string $token The profiler token
+     * @param string $token          The profiler token
      * @param string $connectionName
-     * @param int $query
+     * @param int    $query
      *
      * @return Response A Response instance
      */
@@ -97,14 +97,14 @@ class ProfilerController
     {
         $this->profiler->disable();
 
-        $profile = $this->profiler->loadProfile($token);
+        $profile   = $this->profiler->loadProfile($token);
         $collector = $profile->getCollector('db');
 
         assert($collector instanceof DoctrineDataCollector);
 
         $queries = $collector->getQueries();
 
-        if (!isset($queries[$connectionName][$query])) {
+        if (! isset($queries[$connectionName][$query])) {
             return new Response('This query does not exist.');
         }
 
@@ -112,9 +112,9 @@ class ProfilerController
 
         $connection = $this->registry->getConnection($connectionName);
 
-        $result = $connection->executeQuery($query["sql"]);
+        $result = $connection->executeQuery($query['sql']);
 
-        return new Response($this->twig->render('@Doctrine/Collector/run.html.twig', ["result" => $result->fetchAllAssociative()]));
+        return new Response($this->twig->render('@Doctrine/Collector/run.html.twig', ['result' => $result->fetchAllAssociative()]));
     }
 
     /**
--- a/src/Controller/ProfilerController.php
+++ b/src/Controller/ProfilerController.php
@@ -83,6 +83,40 @@ class ProfilerController
         ]));
     }
 
+
+    /**
+     * Renders the profiler panel for the given token.
+     *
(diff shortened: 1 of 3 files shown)


templates/Collector/db.html.twig
```diff
@@ -90,6 +90,13 @@
             connectionName: request.query.get('connection'),
             query: request.query.get('query')
         })) }}
+        {% elseif 'run' == page %}
```
ostrolucky, line 93: I don't like that most of the lines are duplicated now. Something like this could be done instead:
```php
{% if page in ['explain', 'page'] %}
  {{ render(controller('Doctrine\\Bundle\\DoctrineBundle\\Controller\\ProfilerController::' ~ page ~ 'Action', {
            token: token,
            panel: 'db',
            connectionName: request.query.get('connection'),
            query: request.query.get('query')
        })) }}
```

src/Controller/ProfilerController.php
```diff
@@ -83,6 +83,39 @@ public function explainAction($token, $connectionName, $query)
...
+    /**
+     * Renders the profiler panel for the given token.
+     *
+     * @param string $token          The profiler token
+     * @param string $connectionName
+     * @param int    $query
+     *
+     * @return Response A Response instance
+     */
+    public function runAction($token, $connectionName, $query)
```
ostrolucky, line 95: This is old school PHP 5 style code, we don't use it for new stuff
```suggestion
    public function runAction(string $token, string $connectionName, int $query): Response
```

templates/Collector/db.html.twig
```diff
@@ -100,8 +107,8 @@
         .time-container { position: relative; }
         .time-container .nowrap { position: relative; z-index: 1; text-shadow: 0 0 2px #fff; }
         .time-bar { display: block; position: absolute; top: 0; left: 0; bottom: 0; background: #e0e0e0; }
-        .sql-runnable.sf-toggle-content.sf-toggle-visible { display: flex; flex-direction: column; }
-        .sql-runnable button { align-self: end; }
+        .sql-runnable.sf-toggle-content.sf-toggle-visible { display: flex; flex-direction: row; align-items: center}
+        .sql-runnable pre { flex-grow: 1}
```
ostrolucky, line 111: I'm not a fan of this design change. Let's have buttons next to each other, instead on top of each other

templates/Collector/db.html.twig
```diff
@@ -252,17 +259,22 @@
...
+                                                        <a class="btn btn-sm" href="{{ path('_profiler', { panel: 'db', token: token, page: 'run', connection: connection, query: i }) }}" onclick="return run(this);" data-target-id="run-{{ i }}-{{ loop.parent.loop.index }}">Run</a>
+                                                    </div>
                                                 </div>
                                             {% endif %}
 
                                             {% if query.explainable %}
                                                 <div id="explain-{{ i }}-{{ loop.parent.loop.index }}" class="sql-explain"></div>
                                             {% endif %}
 
+                                            <div id="run-{{ i }}-{{ loop.parent.loop.index }}" class="sql-explain"></div>
```
ostrolucky, line 276: shouldn't this be wrapped in `{% if query.runnable %}` condition? Or you believe the previous condition is not necessary?

templates/Collector/db.html.twig
```diff
@@ -468,6 +480,34 @@
...
+
+                targetElement.style.display = 'block';
+                link.innerHTML = 'Hide query result';
+            } else {
+                targetElement.style.display = 'none';
+                link.innerHTML = 'Run query';
+            }
+
+            return false;
+        }
```
ostrolucky, line 509: Please parametrize the function that was copied instead of copying whole thing and just changing few lines

templates/Collector/run.html.twig
```diff
@@ -0,0 +1,23 @@
...
+                {% for key, value in row %}
+                    <td>{{ value }}</td>
+                {% endfor %}
+            </tr>
+        {% endfor %}
+        </tbody>
+    </table>
+{% else %}
+    Result is empty
+{% endif %}
```
ostrolucky, line 23: this was also copied (explain.html.twig), please modularize through block/embed it if you believe some stuff have to be different

ostrolucky (changes requested):
Doesn't seem to be working for me

image

greg0ire:
I'm retargeting this to 3.1.x now that 3.0.0 is released.

greg0ire:
Well… no activity. Let's close this I guess.

PR #1829, opened 2024-10-13
feat: add support for ORM `fetch_mode_subselect_batch_size` configuration

simPod:


src/DependencyInjection/Configuration.php
```diff
@@ -663,6 +663,7 @@ private function getOrmEntityManagersNode(): ArrayNodeDefinition
                     ->scalarNode('quote_strategy')->defaultValue('doctrine.orm.quote_strategy.default')->end()
                     ->scalarNode('typed_field_mapper')->defaultValue('doctrine.orm.typed_field_mapper.default')->end()
                     ->scalarNode('entity_listener_resolver')->defaultNull()->end()
+                    ->scalarNode('fetch_mode_subselect_batch_size')->end()
```
ostrolucky, line 666: I'm not a fan of naming this differently than ORM method name

simPod: Should I prefer the method name before the array key name? IDK TBH

ostrolucky: Oh I see now. Current name is fine then

ostrolucky (approved):

PR #1832, opened 2024-11-04
Fix wiring of PHP metadata drivers on doctrine/orm 3.x

### LINKED ISSUE #58738: doctrine type 'php' and 'staticphp' use wrong doctrine class name
### Symfony version(s) affected

7.2

### Description

I've set up my doctrine configuration to use php (and also staticphp), but the underlying code point to the wrong class.

```yaml
        mappings:
            App:
                type: php # also fails with staticphp
                is_bundle: false
                prefix: 'App\Entity'
                alias: App
```

The cache generates this code:
```php

$b->addDriver(($container->privates['doctrine.orm.default_php_metadata_driver'] ??= 
new \Doctrine\ORM\Mapping\Driver\PHPDriver([(\dirname(__DIR__, 4).'/src/Entity')])), 'App\\Entity');        
```

But the actual name of the the class is not \Doctrine\ORM\Mapping\Driver\PHPDriver but should be

     not \Doctrine\ORM\Mapping\Driver\PHPDriver

     use \Doctrine\Persistence\Mapping\Driver\PHPDriver;

### How to reproduce

Create a new project using --version=next, run bin/console make:entity, set the driver in doctrine.yaml as above, create this entity and then run bin/console d:sch:update --dump-sql

```php
namespace App\Entity;

use App\Repository\TestRepository;
use Doctrine\ORM\Mapping\ClassMetadata;
use Doctrine\ORM\Mapping\Builder\ClassMetadataBuilder;
#[ORM\Entity(repositoryClass: TestRepository::class)]
class Test
{

    public static function loadMetadata(ClassMetadata $metadata)
    {
        $builder = new ClassMetadataBuilder($metadata);
        $builder->createField('id', 'integer')->isPrimaryKey()->generatedValue()->build();
        $builder->addField('username', 'string');
    }
}
```

### Possible Solution

_No response_

### Additional Context

_No response_

MatTheCat:
Follows https://github.com/symfony/symfony/pull/33319, takes over https://github.com/doctrine/DoctrineBundle/pull/1755 and fixes https://github.com/symfony/symfony/issues/58738.

Some metadata drivers were removed in doctrine/orm 3 and must then be used from doctrine/persistence.

Since the parameters used to infer their class name are deprecated, this PR rewrites `getMetadataDriverClass` so that it selects the right one.

Note that some class parameters are still used in the `DoctrineExtension` but I considered them out of this PR’s scope.

### DIFF
--- a/src/DependencyInjection/DoctrineExtension.php
+++ b/src/DependencyInjection/DoctrineExtension.php
@@ -22,13 +22,21 @@ use Doctrine\ORM\Configuration as OrmConfiguration;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Events;
 use Doctrine\ORM\Id\AbstractIdGenerator;
+use Doctrine\ORM\Mapping\Driver\AnnotationDriver;
+use Doctrine\ORM\Mapping\Driver\AttributeDriver;
+use Doctrine\ORM\Mapping\Driver\PHPDriver as LegacyPHPDriver;
 use Doctrine\ORM\Mapping\Driver\SimplifiedXmlDriver;
+use Doctrine\ORM\Mapping\Driver\SimplifiedYamlDriver;
+use Doctrine\ORM\Mapping\Driver\StaticPHPDriver as LegacyStaticPHPDriver;
 use Doctrine\ORM\Proxy\Autoloader;
 use Doctrine\ORM\Proxy\ProxyFactory;
 use Doctrine\ORM\Tools\Console\Command\ConvertMappingCommand;
 use Doctrine\ORM\Tools\Console\Command\EnsureProductionSettingsCommand;
 use Doctrine\ORM\Tools\Export\ClassMetadataExporter;
 use Doctrine\ORM\UnitOfWork;
+use Doctrine\Persistence\Mapping\Driver\MappingDriverChain;
+use Doctrine\Persistence\Mapping\Driver\PHPDriver;
+use Doctrine\Persistence\Mapping\Driver\StaticPHPDriver;
 use Doctrine\Persistence\Reflection\RuntimeReflectionProperty;
 use InvalidArgumentException;
 use LogicException;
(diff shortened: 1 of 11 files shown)


src/DependencyInjection/DoctrineExtension.php
```diff
@@ -1163,7 +1171,31 @@ public function getConfiguration(array $config, ContainerBuilder $container): Co
 
     protected function getMetadataDriverClass(string $driverType): string
     {
-        return '%' . $this->getObjectManagerElementName('metadata.' . $driverType . '.class') . '%';
+        switch ($driverType) {
+            case 'driver_chain':
+                return MappingDriverChain::class;
+
+            case 'annotation':
+                return AnnotationDriver::class;
```
stof, line 1183: this should probably check for existence, as this driver does not exist in doctrine/orm 3.x

MatTheCat: Since the `doctrine.orm.metadata.annotation.class` parameter still exists this should somewhat keep the current behavior?

Or should I throw an exception if the class does not exist?

ostrolucky: yes please throw exception informing the user this driver is only available for doctrine/orm 2.x

ostrolucky:
Looks like tests will need fixing. And for SA issues you can suppress them.

MatTheCat:
Hm should I split tests into `AnnotationDriver`-exists-or-not? Or sprinkle them with `class_exists`? :thinking:

ostrolucky:
My preference would be to upgrade current tests to use attribute instead of annotation and skip them for orm < 3.x

stof:
We don't need to skip them for ORM 2.x AFAIK. Attributes are supported there as well (on PHP 8+)

MatTheCat:
Do I keep the `DoctrineExtensionTest::testAnnotationsBundleMappingDetection`?

ostrolucky:
you can remove it

MatTheCat:
Okay, done. Not sure why some tests are still failing though :thinking:

ostrolucky:
I think those test failures are a symptom of this branch not being triggered now :( https://github.com/symfony/symfony/blob/c905bb4db3eec98e156199341891f62247f83fb6/src/Symfony/Bridge/Doctrine/DependencyInjection/AbstractDoctrineExtension.php#L224

MatTheCat:
Waat so the Doctrine Bridge does need to be updated 😅

ostrolucky:
I don't know, just saying that's the cause why tests fail. Proper course of action would be to find out why was that code branch added and find if same reasoning still applies nowadays. Then we know if we need to preserve the behavior somehow, or adjust tests

MatTheCat:
See https://github.com/symfony/symfony/pull/53681; seems that we cannot be compatible with Symfony < 6.4 by returning XML and YAML drivers’ class name instead of the parameter.

MatTheCat:
Maybe this PR should focus to simply returning the correct PHP driver’s class name? That would fix https://github.com/symfony/symfony/issues/58738, hopefully without any side effect.

ostrolucky:
I believe we just need to backport https://github.com/symfony/symfony/pull/53681 to 5.4 branch. It's either that, or we cut off symfony/doctrine-bridge 5.4.x support here

MatTheCat:
Okay opened https://github.com/symfony/symfony/pull/58772

MatTheCat:
Woops this got closed with https://github.com/symfony/symfony/pull/58772 and I cannot reopen it :sweat: 

Can someone do it for me or should I open another PR?

EDIT: thanks @nicolas-grekas :pray:

ostrolucky:
Last thing needed here I think is to adjust symfony/doctrine-bridge requirement to ensure we don't install < 5.4.46 version

MatTheCat:
Shouldn’t the `^6.0.7` constraint be bumped to `^6.4.3` and `^7.0` to `^7.0.3` so that we’re sure we got the fix?

ostrolucky:
indeed

MatTheCat:
CI is green :tada:

theofidry:
I know a bundle wiring is not under BC guarantee but is it really desirable to introduce such BC breaks in a patch version?

If you had a custom XML driver registered for instance, this entirely breaks the app and not necessarily in an obvious way (as commonly found with wiring issues).

It's also not really clear what is the upgrade path in this case.

PR #1843, opened 2024-11-26
Use `kernel.build_dir` by default for proxy dir and metadata cache

alexndlm:


ostrolucky:
why?

alexndlm:
> why?

https://github.com/symfony/symfony/blob/d867b1273343a4def8ce259d8aad966339faab4c/src/Symfony/Component/HttpKernel/KernelInterface.php#L116

stof:
Given that the proxies and the metadata cache are directly derived from the source code and can be fully warmed up, I think this makes sense.
It makes things better for project wanting to deploy with a readonly file system, separating things that get generated during the build from things written in the writable cache folder (potentially treated as ephemeral).
Note that by default, Symfony does not separate the build directory from the cache directory. It requires an explicit opt-in.

ostrolucky (approved):

stof:
However, seeing no changes in cache warmers looks a bit suspicious to me. I thought that cache warmers needed to use their second argument to properly warm up the build directory during the `cache:warmup` command.

ostrolucky:
Change of `$phpArrayFile` in `DoctrineExtension` modifies this argument that's passed to `DoctrineMetadataCacheWarmer`. Do you think there is something else to do?

PR #1852, opened 2025-01-13
Drop Psalm in favor of PHPStan

greg0ire:
As per the decision we made during the hackathon.
I have left many `psalm-suppress` annotations that seem informative in the codebase.


SenseException (approved):

ostrolucky:
No baseline please. Pick the level project follows. I don't want people getting errors they have to fix every time they touch random file.

SenseException:
They would have to write code for the current level while the existing errors in baseline can be fixed over time. It would get harder to reach a higher level when new code bring new errors.

phpstan.neon.dist
```diff
@@ -0,0 +1,13 @@
+parameters:
+   level: 8
+   phpVersion: 80402
+   paths:
+      - src
+      - tests
+
+   excludePaths:
+      - src/Command/Proxy/ConvertMappingDoctrineCommand.php
+      - src/Command/Proxy/EnsureProductionSettingsDoctrineCommand.php
```
ostrolucky, line 9: Why?

phpstan.neon.dist
```diff
@@ -0,0 +1,13 @@
+parameters:
+   level: 8
+   phpVersion: 80402
```
ostrolucky: I would prefer phpstan to infer the correct version

ostrolucky:
We could argue about pros and cons of baselines, but let's have that discussion separately. We didn't use baseline with psalm, so let's not add it in PR which claims to just replace psalm.

phpstan.neon.dist
greg0ire: Apparently, the errors in these 2 files cannot be baselined

greg0ire:
@ostrolucky I agree with @SenseException , but I want you to be happy with how this repo looks like since you're so active in it.

I pushed a version with what you asked… it means we are now at PHPStan level 1 :grimacing:

ostrolucky (approved):

phpstan.neon.dist
```diff
@@ -0,0 +1,9 @@
+parameters:
+   level: 1
```
derrabus, line 2: This is way too low compared to on what level Psalm ran.

ostrolucky: Yeah :( I will try to increase it slowly. Psalm was running on high level because of carefully crafted suppression rules, somebody has to put in the work to make them for phpstan. I've already run into https://github.com/phpstan/phpstan/issues/12414 (this is the kind of things baselines hide)

I'll make it to lvl2 today. Let me know if you would be interested in help to update it to some other lvl.

derrabus: I'll have a look, but I won't be able to start right away. A temporary baseline would've been good in order to not lose coverage and track the progress.

Guuzen:
> As per the decision we made during the hackathon

Could you (or anyone else) explain the reasons why you decided to make such a replacement?

stof:
@Guuzen see https://github.com/doctrine/.github/pull/50#issuecomment-2401552589

stof:
Note that previously, most other Doctrine packages were running both tools, not just Psalm. This means that the first reason does not apply to this repo specifically, but consistency in tooling in the organization also helps.

PR #1854, opened 2025-01-14
 phpstan: lvl 1 -> 2 

ostrolucky:


src/ConnectionFactory.php
```diff
@@ -119,7 +119,7 @@ public function createConnection(array $params, ?Configuration $config = null, ?
             $connection = DriverManager::getConnection(...array_merge([$params, $config], $eventManager ? [$eventManager] : []));
             $params     = $this->addDatabaseSuffix(array_merge($connection->getParams(), $overriddenOptions));
             $driver     = $connection->getDriver();
-            /** @psalm-suppress InvalidScalarArgument Bogus error, StaticServerVersionProvider implements Doctrine\DBAL\ServerVersionProvider  */
+            /** @phpstan-ignore arguments.count (DBAL 2.x doesn't accept an argument) */
```
derrabus:
```suggestion
            /** @phpstan-ignore arguments.count (DBAL 3.x doesn't accept an argument) */
```

I'm on a phone, so please double-check. But I think we've made that change going from 3 to 4.

phpstan.neon.dist
```diff
@@ -1,9 +1,20 @@
...
+    # Available in ORM < 3 only
+    - '#Doctrine\\ORM\\Tools\\EntityGenerator.#'
+    - '#Doctrine\\ORM\\Tools\\DisconnectedClassMetadataFactory.#'
+    - '#Doctrine\\ORM\\Tools\\Export\\ClassMetadataExporter.#'
+    # phpstan has no array shape intersection support https://github.com/phpstan/phpstan/issues/12414
+    - message: '#unresolvable type.#'
+      path: src/DataCollector/DoctrineDataCollector.php
+    # Probably needs Symfony plugin
+    - message: '#Call to an undefined method Symfony\\Component\\Config\\Definition\\Builder\\Node#'
+      path: src/DependencyInjection/Configuration.php
```
greg0ire, line 20: How do you make the choice to user `ignoreErrors` vs `phpstan-ignore`? Now that `phpstan-ignore` comes with error identifier, I think it's good.

ostrolucky: This rule suppresses 10 violations. It's hard to make a hard rule out of it, but 10 violations of same kind I would say are sufficient. Once we solve underlying issue (like here, _maybe_ it will be solved by symfony plugin), we can remove it from one place instead of hunting it all over. If it was like 3 violations, I would be more inclined to do it inline. But anything in between is subjective.

greg0ire (approved):

phpstan.neon.dist
derrabus: > Now that `phpstan-ignore` comes with error identifier

We can use those identifiers in the ignore rules of our neon file as well and we should probably do that.

greg0ire: To elaborate, in the past I thought the baseline approach was superior, because using `@phpstan-ignore-next-line` could hide more than the intended errors. Now that it's possible to do that directly in the code, I think it might be nicer to have the ignore rules close to the code, because sometimes I'm wondering why PHPStan "isn't working". Also, if you see an existing ignore rule in the code, you are more likely to challenge it, if you know how to properly fix it.

ostrolucky: Yes normally I prefer inline suppressions (and ask people to do it). There are bunch of issues that prevent us to do it efficiently unfortunately, for example it's not enough for phpstan to suppress missing class when instantiating it, phpstan is not smart enough to understand that if you ignore missing class it's also ok to ignore method calls on such class. So now you have to sprinkle code with suppressions each time such class is used as well. I would expect I can just suppress instantiating class that can be missing, instead of having to suppress dozens of other issues cascading from that as well.

PR #1855, opened 2025-01-14
phpstan: lvl 2 -> 3

ostrolucky:


phansys (approved):

src/Registry.php
```diff
@@ -23,7 +23,7 @@ class Registry extends ManagerRegistry implements ResetInterface
      * @param string[] $connections
      * @param string[] $entityManagers
      */
-    public function __construct(ContainerInterface $container, array $connections, array $entityManagers, string $defaultConnection, string $defaultEntityManager)
+    public function __construct(Container $container, array $connections, array $entityManagers, string $defaultConnection, string $defaultEntityManager)
```
SenseException, line 26: I'm not sure about this change for 2.13.x. Technically it could have been replaced in a project with another implementation of a container, but, on the other side, I hardly believe that someone would do such a step.

ostrolucky: Parent that's called below requires Container. If somebody overwrites and calls our constructor with instance that's not a Container (but ContainerInterface), it would already result in an error

dmaicher (approved):

src/Command/DoctrineCommand.php
```diff
@@ -59,6 +61,8 @@ protected function getEntityManager($name, $shardId = null)
             throw new InvalidArgumentException('Shards are not supported anymore using doctrine/dbal >= 3');
         }
 
+        assert($manager instanceof EntityManager);
```
stof, line 64: I think this assertion will cause issue for projects decorating the EntityManager. We should probably use EntityManagerInterface as type instead.

ostrolucky: Thanks for feedback. Addressed in https://github.com/doctrine/DoctrineBundle/commit/2363c43d9815a11657e452625cd64172d5587486

PR #1859, opened 2025-01-26
bump to phpstan level 4

dmaicher:


src/DependencyInjection/Compiler/WellKnownSchemaFilterPass.php
```diff
@@ -21,35 +17,5 @@ class WellKnownSchemaFilterPass implements CompilerPassInterface
...
-                continue;
-            }
-
-            if ($definition->getClass() !== PdoSessionHandler::class) {
-                continue;
-            }
-
-            $table = $definition->getArguments()[1]['db_table'] ?? 'sessions';
-
-            if (! method_exists($definition->getClass(), 'configureSchema')) {
```
dmaicher, line 37: since `method_exists($definition->getClass(), 'configureSchema')` is alyways true now this class is no-op, right? :eyes:

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -697,10 +696,6 @@ protected function loadOrmEntityManager(array $entityManager, ContainerBuilder $
             $methods['setEagerFetchBatchSize'] = $entityManager['fetch_mode_subselect_batch_size'];
         }
 
-        if (! method_exists(OrmConfiguration::class, 'setLazyGhostObjectEnabled')) {
```
dmaicher, line 700: that method should now always exist with ORM 2.17+

src/Command/CreateDatabaseDoctrineCommand.php
```diff
@@ -67,7 +67,11 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         unset($params['dbname'], $params['path'], $params['url']);
 
         if ($connection->getDatabasePlatform() instanceof PostgreSQLPlatform) {
-            /** @psalm-suppress InvalidArrayOffset It's still available in DBAL 3.x that we need to support */
+            /**
+             * for DBAL < 4
+             *
+             * @phpstan-ignore nullCoalesce.offset
+             */
```
ostrolucky: you should be able to add comments like this
```suggestion
            /** @phpstan-ignore nullCoalesce.offset (needed for DBAL < 4) */
```

pls apply to all of these instances

src/ConnectionFactory.php
```diff
@@ -76,7 +77,7 @@ public function createConnection(array $params, Configuration|null $config = nul
         }
 
         $overriddenOptions = [];
-        /** @psalm-suppress InvalidArrayOffset We should adjust when https://github.com/vimeo/psalm/issues/8984 is fixed */
+        /** @phpstan-ignore isset.offset */
```
ostrolucky:
```suggestion
        /** @phpstan-ignore isset.offset (We should adjust when https://github.com/phpstan/phpstan/issues/12414 is fixed) */
```

src/ConnectionFactory.php
```diff
@@ -96,7 +97,7 @@ public function createConnection(array $params, Configuration|null $config = nul
             }
         }
 
-        /** @psalm-suppress InvalidArrayOffset We should adjust when https://github.com/vimeo/psalm/issues/8984 is fixed */
+        /** @phpstan-ignore booleanOr.rightAlwaysFalse, isset.offset, isset.offset */
```
ostrolucky:
```suggestion
        /** @phpstan-ignore-next-line We should adjust when https://github.com/phpstan/phpstan/issues/12414 is fixed */
```

src/DependencyInjection/Compiler/WellKnownSchemaFilterPass.php
ostrolucky: Indeed. Let's remove class too, since you already removed compiler pass definition anyways

phpstan.neon.dist
```diff
@@ -18,3 +18,5 @@ parameters:
     # Probably needs Symfony plugin
     - message: '#Call to an undefined method Symfony\\Component\\Config\\Definition\\Builder\\Node#'
       path: src/DependencyInjection/Configuration.php
+    - '#Entity::\$id \(int\|null\) is never assigned int so it can be removed from the property type#'
+    - '#Entity::\$id is never read, only written#'
```
ostrolucky: Can we pls inline these instead? These errors are not something that's going away with some upgrade, plugin or so

src/DependencyInjection/Compiler/WellKnownSchemaFilterPass.php
dmaicher: Yeah although its a BC break I would also tend to remove it as its completely useless now

ostrolucky (approved):
Thanks for the help with this grunt work!

PR #1868, opened 2025-02-06
Ensure classes using the `#[Entity]` attribute are not declared as service

GromNaN:
In order to enable discovery of entity classes by the Symfony DIC, we need to exclude them from being registered as actual services.

- We can remove the exclusion of the `src/Entity/` directory from the [`symfony/framework-bundle` recipe](https://github.com/symfony/recipes/blob/2d8f656159ca9132c55b6e848171a7dabc1d8554/symfony/framework-bundle/7.2/config/services.yaml#L20).
- Entity classes using the `#[ApiResource]` attribute can be discovered using DI instead of using a custom-made mechanism of directory scanning and cache: https://github.com/api-platform/core/pull/6943
- This is not using the new upcoming helpers for compatiblity with Symfony 6.4: https://github.com/symfony/symfony/pull/59704
- The same was done in DoctrineMongoDBODMBundle: https://github.com/doctrine/DoctrineMongoDBBundle/pull/876

### DIFF
--- a/tests/DependencyInjection/DoctrineExtensionTest.php
+++ b/tests/DependencyInjection/DoctrineExtensionTest.php
@@ -32,6 +32,9 @@ use Doctrine\ORM\Mapping\Driver\AnnotationDriver;
 use Doctrine\ORM\Mapping\Driver\AttributeDriver;
 use Doctrine\ORM\Mapping\Driver\SimplifiedXmlDriver;
 use Doctrine\ORM\Mapping\Driver\SimplifiedYamlDriver;
+use Doctrine\ORM\Mapping\Embeddable;
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\MappedSuperclass;
 use Doctrine\Persistence\Mapping\Driver\MappingDriverChain;
 use InvalidArgumentException;
 use LogicException;
@@ -1163,6 +1166,43 @@ class DoctrineExtensionTest extends TestCase
                 'cacheConfig' => ['type' => 'service', 'id' => 'service_target_result'],
             ],
         ];
+    }
+
+    /** @return array<array{0: class-string}> */
+    public static function provideAttributeExcludedFromContainer(): array
+    {
+        return [
+            'Embeddable' => [Embeddable::class],
+            'Entity' => [Entity::class],
+            'MappedSuperclass' => [MappedSuperclass::class],
+        ];
+    }
+
+    /** @dataProvider provideAttributeExcludedFromContainer */
+    public function testEntityAttributeExcludesFromContainer(string $class)
+    {
+        if (! interface_exists(EntityManagerInterface::class)) {
+            self::markTestSkipped('This test requires ORM');
+        }
+
+        $container = $this->getContainer();
+        $extension = new DoctrineExtension();
+
+        $config = BundleConfigurationBuilder::createBuilder()
+            ->addBaseConnection()
+            ->addBaseEntityManager()
+            ->build();
+
+        $extension->load([$config], $container);
+
+        $attributes = $container->getAutoconfiguredAttributes();
+        $this->assertInstanceOf(Closure::class, $attributes[$class]);
+
+        $definition = new ChildDefinition('');
+        $attributes[$class]($definition);
+
+        $this->assertSame([['source' => sprintf('with #[%s] attribute', $class)]], $definition->getTag('container.excluded'));
+        $this->assertTrue($definition->isAbstract());
     }
 
     public function testAsEntityListenerAttribute()
--- a/src/DependencyInjection/DoctrineExtension.php
+++ b/src/DependencyInjection/DoctrineExtension.php
@@ -27,6 +27,9 @@ use Doctrine\ORM\Mapping\Driver\PHPDriver as LegacyPHPDriver;
 use Doctrine\ORM\Mapping\Driver\SimplifiedXmlDriver;
 use Doctrine\ORM\Mapping\Driver\SimplifiedYamlDriver;
 use Doctrine\ORM\Mapping\Driver\StaticPHPDriver as LegacyStaticPHPDriver;
+use Doctrine\ORM\Mapping\Embeddable;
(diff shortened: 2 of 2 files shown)


SenseException (approved):

tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -1165,6 +1167,35 @@ public static function cacheConfigurationProvider(): array
         ];
     }
 
+    public function testEntityAttributeExcludesFromContainer()
+    {
+        if (! interface_exists(Entity::class)) {
```
ostrolucky:
```suggestion
        if (! interface_exists(EntityManagerInterface::class)) {
```

tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -1165,6 +1167,35 @@ public static function cacheConfigurationProvider(): array
...
+        $extension->load([$config], $container);
+
+        $attributes = $container->getAutoconfiguredAttributes();
+        $this->assertInstanceOf(Closure::class, $attributes[Entity::class]);
+
+        $reflector  = new ReflectionClass(TestCustomIdGeneratorEntity::class);
+        $definition = new ChildDefinition('');
+        $attribute  = $reflector->getAttributes(Entity::class)[0]->newInstance();
+
+        $attributes[Entity::class]($definition, $attribute);
```
ostrolucky: Not sure if this isn't unnecessarily convoluted, is there a harm in just doing
```suggestion
        $definition = new ChildDefinition('');
        $attributes[Entity::class]($definition, new Entity());
```
?

GromNaN: Thanks for the review, I cleaned up the test.

stof:
this should also exclude MappedSuperclass and Embeddable

dunglas (approved):

GromNaN:
It's being done in the `FrameworkBundle` https://github.com/symfony/symfony/pull/59987, but I think that's still valuable to add this in `DoctrineBundle` as that ease maintenance (if new attributes are added). At some point in the future, they can be removed from the `FrameworkBundle`: when all the versions of the DoctrineBundle that are supported by FrameworkBundle will have this exclusion.

nicolas-grekas (approved):

ostrolucky:
Can you check errors here https://github.com/doctrine/DoctrineBundle/actions/runs/13908353031/job/38916534370 ? They seem related

GromNaN:
@ostrolucky see https://github.com/symfony/symfony/pull/60001

janklan:
Hello all. I did a minor version upgrade of DoctrineBundle, and my app broke because of this change. The entities could have been part of the DI container for ages (assuming the `../src/Entity` path was removed from the default exclusion list in the `services.yaml` file) and now they're not. 

Wouldn't that be considered a BC break?

GromNaN:
I assume you are using [`findTaggedServiceIds`](https://github.com/symfony/symfony/blob/f047c9a3cd81745c16833f98001fe1c8d9e5f9e1/src/Symfony/Component/DependencyInjection/ContainerBuilder.php#L1337C21-L1337C41).

By adding the `container.excluded` tag, this definitions are no longer returned. You need to loop over all the service definitions as done for the `#[JsonStreamable]` attribute with `json_streamer.streamable` tag: https://github.com/symfony/symfony/blob/f047c9a3cd81745c16833f98001fe1c8d9e5f9e1/src/Symfony/Component/JsonStreamer/DependencyInjection/StreamablePass.php#L33

ostrolucky:
> Wouldn't that be considered a BC break?

Not the kind where we guarantee a BC for. Fixes break applications too, if you rely on buggy behaviour.

janklan:
> > Wouldn't that be considered a BC break?
> 
> 
> 
> Not the kind where we guarantee a BC for. Fixes break applications too, if you rely on buggy behaviour.

Hmm. I don't think we should just call a long-standing practice a "buggy behaviour" and change it without any warning. Doing so undermines perceived reliability of the BC promise, I think.

It's a behaviour, buggy or not. It is now considered obsolete. I think it should be marked deprecated and then phased out like any other obsolete behaviour.

ostrolucky:
> I don't think we should just call a long-standing practice a "buggy behaviour" and change it without any warning. 

Long standing practice is to have Entities excluded from service container, I don't think you should ignore that either. That really smells like an oversight in your application.

> Doing so undermines perceived reliability of the BC promise

Service definitions are not covered by Symfony BC promise. And DoctrineBundle does not officially follow Symfony BC promise anyways.

> it should be marked deprecated and then phased out like any other obsolete behaviour.

I don't see any meaningful way to deprecate this without disrupting large portion of the ecosystem, do you?

janklan:
> Long standing practice is to have Entities excluded from service container

I agree with that. Another part of the long-standing practice was that you could change it, and now you can't.

> smells like an oversight in your application

Including entities in the container might be a bad call if it was used to turn Entities into classes with things autowired into them. I didn't include them for that purpose and explained my reasons [here](https://github.com/symfony/symfony/pull/59987#issuecomment-2781869433). 

> Service definitions are not covered by Symfony BC promise

I didn't know that

> And DoctrineBundle does not officially follow Symfony BC promise anyways.

Fair enough. 

> I don't see any meaningful way to deprecate this without disrupting large portion of the ecosystem, do you?

TL;DR: Can't you at least throw a warning while compiling the container that the entities are now forcefully excluded?

Full reply: I found this because our CI/CD pipeline started failing: one of the fixtures generators was inserting a null value into a not-null column. The value is auto-generated by an insert trigger. For some reason, the trigger was not firing. Looking closer, the trigger wasn't there at all. 

After some stumbling around, I found the custom queries were not executed during the migration process, and the reason for that was that the migration listener responsible for finding the entities and executing their custom queries was no longer receiving their list from the container.

Things were still working fine on previously deployed systems because all of the custom stuff was already set up. I think that's a pretty stealthy bug.

If there was any sort of warning, I'd probably say "Oh, yeah, no worries".

PR #1872, opened 2025-02-26
Add loaded entities profiler section

mamazu:
## The Feature
Showing a list of entity classes and the amount of entities that are managed:
image

## The reasoning
Doctrine has a nice lazy loading feature of entities however I don't think many people realize that this can create a huge object graph in the entity manager. As flushing will go in and check all the entities for updates.

### DIFF
--- a/src/DataCollector/DoctrineDataCollector.php
+++ b/src/DataCollector/DoctrineDataCollector.php
@@ -43,12 +43,15 @@ use function usort;
  *    errors: array<string, array<class-string, list<string>>>,
  *    managers: list<string>,
  *    queries: array<string, list<QueryType>>,
+ *    entityCounts: array<string, array<class-string, int>>
  * }
  * @psalm-property DataType $data
  */
 class DoctrineDataCollector extends BaseCollector
 {
     private int|null $invalidEntityCount = null;
+
+    private int|null $managedEntityCount = null;
 
     /**
      * @var mixed[][]|null
@@ -117,7 +120,7 @@ class DoctrineDataCollector extends BaseCollector
             }
 
             foreach ($em->getUnitOfWork()->getIdentityMap() as $className => $entityList) {
-                $entityCounts[$className] = ($entityCounts[$className] ?? 0) + count($entityList);
+                $entityCounts[$name][$className] = ($entityCounts[$name][$className] ?? 0) + count($entityList);
             }
 
             $emConfig   = $em->getConfiguration();
@@ -240,7 +243,16 @@ class DoctrineDataCollector extends BaseCollector
 
     public function getManagedEntityCount(): int
     {
-        return array_sum($this->data['entityCounts']);
+        if ($this->managedEntityCount === null) {
+            $total = 0;
+            foreach ($this->data['entityCounts'] as $entities) {
+                $total += array_sum($entities);
+            }
+
+            $this->managedEntityCount = $total;
+        }
+
+        return $this->managedEntityCount;
     }
 
     /** @return array<class-string, int> */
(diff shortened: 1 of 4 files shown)


dmaicher (approved):
:+1: from my side. Also tested it on one of my projects.

Should target 2.14.x though as its a new feature

dmaicher:
One observation: I have an app with 2 different databases and 2 entity managers. 
Currently the managed entity counts are displayed together in one table for both entity managers. 

Should it maybe be split into different tables per entity manager instead as we do for the "Entities Mapping" tab?

stof:
it should indeed be per entity manager

mamazu:
Rebased and split up between entity managers:
image

templates/Collector/db.html.twig
```diff
@@ -384,6 +393,22 @@
             </div>
         </div>
 
+        <div class="tab">
+            <h3 class="tab-title">Managed Entities</h3>
+            <div class="tab-content">
+                {% if not collector.managedEntityCountByClass %}
+                    <div class="empty">
+                        <p>No mapped entities.</p>
```
dmaicher:
```suggestion
                        <p>No managed entities.</p>
```

src/DataCollector/DoctrineDataCollector.php
```diff
@@ -114,6 +119,10 @@ public function collect(Request $request, Response $response, Throwable|null $ex
                 }
             }
 
+            foreach ($em->getUnitOfWork()->getIdentityMap() as $className => $entityList) {
+                $entityCounts[$name][$className] = ($entityCounts[$name][$className] ?? 0) + count($entityList);
```
stof:
```suggestion
                $entityCounts[$name][$className] = count($entityList);
```

We won't have the same key twice for each entity manager.

src/DataCollector/DoctrineDataCollector.php
```diff
@@ -165,10 +174,14 @@ public function collect(Request $request, Response $response, Throwable|null $ex
             }
         }
 
-        $this->data['entities'] = $entities;
-        $this->data['errors']   = $errors;
-        $this->data['caches']   = $caches;
-        $this->groupedQueries   = null;
+        // Sort entities by count (in descending order)
+        arsort($entityCounts);
```
stof: this does not make sense as this variable does not contain a map of entities to count anymore, but a map of manager names to such map. You need to sort `$entityCounts[$name]` after creating it for each entity manager instead (after line 125).

src/DataCollector/DoctrineDataCollector.php
```diff
@@ -228,6 +241,26 @@ public function getInvalidEntityCount()
...
+                $total += array_sum($entities);
+            }
+
+            $this->managedEntityCount = $total;
+        }
+
+        return $this->managedEntityCount;
+    }
+
+    /** @return array<class-string, int> */
```
stof: this actually return `array<string, array<class-string, int>>`

templates/Collector/db.html.twig
```diff
@@ -384,6 +393,22 @@
...
+                    </div>
+                {% else %}
+                    {% for manager, entityCounts in collector.managedEntityCountByClass %}
+                        <h3>{{ manager }} <small>entity manager</small></h3>
```
stof: should be a `<h4>` as the tab title is already a `h3`

mamazu: I've updated it but the entity mapping tables are also `h3`s. Should I also update them to be h4?

stof: I would say yes

mamazu: Done

dmaicher (approved):

stof (approved):

SenseException (approved):

mamazu:
Should I also add a little text explaining that this only shows the amount of managed entities at the end of the program when the collector runs?

ostrolucky:
No need I would say, that's how collector stats are usually collected. But I would welcome if there was as an assertion for this in ProfilerTest, that's our integration test for profiler.

PR #1877, opened 2025-04-02
Fix autowiring alias target name

nicolas-grekas:
The service id is not meant to be used as a target. The name is.


nicolas-grekas:
NVM

PR #1878, opened 2025-04-02
Improve named autowiring targets

nicolas-grekas:
This change provides better hints for `#[Target]` attributes, without changing anything else.

When running debug:autowiring:

image

And in the code, this allows:

```php
function __construct(#[Target('default.entity_manager') EntityManagerInterface $em)
```

### DIFF
--- a/src/DependencyInjection/DoctrineExtension.php
+++ b/src/DependencyInjection/DoctrineExtension.php
@@ -309,7 +309,7 @@ class DoctrineExtension extends AbstractDoctrineExtension
             ]);
 
         $container
-            ->registerAliasForArgument($connectionId, Connection::class, sprintf('%sConnection', $name))
+            ->registerAliasForArgument($connectionId, Connection::class, sprintf('%s.connection', $name))
             ->setPublic(false);
 
         // Set class in case "wrapper_class" option was used to assist IDEs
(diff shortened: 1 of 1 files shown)


stof:
isn't it a BC break to remove the old target in case projects were already using the `#[Target]` attribute ?

Btw, I think we should add a named target using *only* the configured named, without any suffix. This will be even better for usages in the `#[Target]` attribute.

nicolas-grekas:
All targets are going to be normalized into their camel-case version so this is going to end up being the same, that's what I meant by "without changing anything else".
We cannot remove the suffix without creating a new named autowiring alias (`EntityManagerInterface $theName`). This would pollute the `debug:autowiring` output to me, so I decided not adding the suffix-less target.

stof:
@nicolas-grekas are you sure they are normalized into the camel cased version for usage in the Target attribute ?
I thought this normalization happened only for the detection based on argument names.

nicolas-grekas:
Pretty much yes, in the end, we always check the named autowiring alias:
https://github.com/symfony/symfony/blob/7.3/src/Symfony/Component/DependencyInjection/Compiler/AutowirePass.php#L472
See also https://github.com/symfony/symfony/blob/7.3/src/Symfony/Component/DependencyInjection/Attribute/Target.php

stof (approved):

ostrolucky (approved):

PR #1885, opened 2025-06-05
Fix erroneous order of events for multi events per listener

### LINKED ISSUE #1884: [EntityListener] with multiple listeners per class only the first priority is used
### Bug Report

| Version                                     | 2.14.0

#### Summary

Using the `AsEntityListener` attribute to listen to entity events one can set a priority for each event listener. The documentation suggests that then all listeners for an entity event are sorted by priority and strictly executed in that order.

But when having multiple event listeners in a single class with differing priorities this actually produces unexpected results. Because only the priority of the first defined listener is actually used and also applied to the other listeners.

This is due to  [`EntityListenerPass::process()`](https://github.com/doctrine/DoctrineBundle/blob/ca6a7350b421baf7fbdefbf9f4993292ed18effb/src/DependencyInjection/Compiler/EntityListenerPass.php#L33) using `PriorityTaggedServiceTrait::findAndSortTaggedServices()` which effectively sorts all tagged services [by the priority of the _first_ attribute](https://github.com/symfony/symfony/blob/3fe77f0b5fd9ac22875ded3fa84e950da56d7977/src/Symfony/Component/DependencyInjection/Compiler/PriorityTaggedServiceTrait.php#L81-L82).

#### Current behavior

The defined entity listeners are not correctly executed in order of their individual priorities.

#### Expected behavior

Each defined entity listener should be executed according to its specific priority.

#### How to reproduce

Given these two classes with entity listeners

~~~php
#[AsEntityListener(entity: User::class, event: Events::prePersist, priority: 0)]
#[AsEntityListener(entity: User::class, event: Events::postPersist, priority: 0)]
final class FirstListener {
	public function prePersist(User $user, PrePersistEventArgs $event) {
		dump('this is ' . __METHOD__);
	}

	public function postPersist(User $user, PostPersistEventArgs $event) {
		dump('this is ' . __METHOD__);
	}
}
~~~

~~~php
#[AsEntityListener(entity: User::class, event: Events::prePersist, priority: 1)]
#[AsEntityListener(entity: User::class, event: Events::postPersist, priority: -1)]
final class SecondListener {
	public function prePersist(User $user, PrePersistEventArgs $event) {
		dump('this is ' . __METHOD__);
	}

	public function postPersist(User $user, PostPersistEventArgs $event) {
		dump('this is ' . __METHOD__);
	}
}
~~~

The expected output order of the `dump()` statements would be
> "this is App\EventListener\SecondListener::prePersist"
> "this is App\EventListener\FirstListener::prePersist"
> "this is App\EventListener\FirstListener::postPersist"
> "this is App\EventListener\SecondListener::postPersist"

but actually is

> "this is App\EventListener\SecondListener::prePersist"
> "this is App\EventListener\FirstListener::prePersist"
> "this is App\EventListener\SecondListener::postPersist"
> "this is App\EventListener\FirstListener::postPersist"

Switching just the order of the attributes on the second class

~~~php
#[AsEntityListener(entity: User::class, event: Events::postPersist, priority: -1)]
#[AsEntityListener(entity: User::class, event: Events::prePersist, priority: 1)]
final class SecondListener {
	public function prePersist(User $user, PrePersistEventArgs $event) {
		dump('this is ' . __METHOD__);
	}

	public function postPersist(User $user, PostPersistEventArgs $event) {
		dump('this is ' . __METHOD__);
	}
}
~~~

the output becomes

> "this is App\EventListener\FirstListener::prePersist"
> "this is App\EventListener\SecondListener::prePersist"
> "this is App\EventListener\FirstListener::postPersist"
> "this is App\EventListener\SecondListener::postPersist"

ostrolucky:
Fixes #1884

### DIFF
--- a/src/DependencyInjection/Compiler/EntityListenerPass.php
+++ b/src/DependencyInjection/Compiler/EntityListenerPass.php
@@ -6,7 +6,6 @@ use Doctrine\Bundle\DoctrineBundle\Mapping\ContainerEntityListenerResolver;
 use Doctrine\Bundle\DoctrineBundle\Mapping\EntityListenerServiceResolver;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
-use Symfony\Component\DependencyInjection\Compiler\PriorityTaggedServiceTrait;
 use Symfony\Component\DependencyInjection\Compiler\ServiceLocatorTagPass;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
@@ -17,6 +16,7 @@ use function is_a;
 use function method_exists;
 use function sprintf;
 use function substr;
+use function usort;
 
 /**
  * Class for Symfony bundles to register entity listeners
@@ -25,18 +25,15 @@ use function substr;
  */
 class EntityListenerPass implements CompilerPassInterface
 {
-    use PriorityTaggedServiceTrait;
-
     /** @return void */
     public function process(ContainerBuilder $container)
     {
-        $resolvers = $this->findAndSortTaggedServices('doctrine.orm.entity_listener', $container);
-
         $lazyServiceReferencesByResolver = [];
 
-        foreach ($resolvers as $reference) {
-            $id = $reference->__toString();
-            foreach ($container->getDefinition($id)->getTag('doctrine.orm.entity_listener') as $attributes) {
+        foreach ($container->findTaggedServiceIds('doctrine.orm.entity_listener') as $id => $tags) {
+            usort($tags, static fn (array $a, array $b) => ($a['priority'] ?? 0) <=> ($b['priority'] ?? 0));
+
+            foreach ($tags as $attributes) {
                 $name          = $attributes['entity_manager'] ?? $container->getParameter('doctrine.default_entity_manager');
                 $entityManager = sprintf('doctrine.orm.%s_entity_manager', $name);
 
--- a/tests/DependencyInjection/Compiler/EntityListenerPassTest.php
+++ b/tests/DependencyInjection/Compiler/EntityListenerPassTest.php
@@ -64,6 +64,43 @@ class EntityListenerPassTest extends TestCase
         yield 'With event and custom method' => [Events::postLoad, 'postLoadHandler', 'postLoadHandler'];
         yield 'With event and no matching method' => [Events::postLoad, null, '__invoke'];
     }
+
+    public function testMultipleAttributesOnSameClassKeepTheCorrectOrder(): void
+    {
+        $container = new ContainerBuilder();
+        $container->addCompilerPass(new EntityListenerPass());
+
+        $container->setParameter('doctrine.default_entity_manager', 'default');
+        $container->register('doctrine.orm.default_entity_manager', EntityManager::class);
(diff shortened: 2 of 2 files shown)


src/DependencyInjection/Compiler/EntityListenerPass.php
```diff
@@ -25,18 +25,15 @@
...
     {
-        $resolvers = $this->findAndSortTaggedServices('doctrine.orm.entity_listener', $container);
-
         $lazyServiceReferencesByResolver = [];
 
-        foreach ($resolvers as $reference) {
-            $id = $reference->__toString();
-            foreach ($container->getDefinition($id)->getTag('doctrine.orm.entity_listener') as $attributes) {
+        foreach ($container->findTaggedServiceIds('doctrine.orm.entity_listener') as $id => $tags) {
+            usort($tags, static fn (array $a, array $b) => ($a['priority'] ?? 0) <=> ($b['priority'] ?? 0));
```
acran, line 34: should be
```suggestion
            usort($tags, static fn (array $a, array $b) => ($b['priority'] ?? 0) <=> ($a['priority'] ?? 0));
```

i.e. tags with a larger number as priority come first!?

see [`PriorityTaggedServiceTrait#L113`](https://github.com/symfony/symfony/blob/3fe77f0b5fd9ac22875ded3fa84e950da56d7977/src/Symfony/Component/DependencyInjection/Compiler/PriorityTaggedServiceTrait.php#L113)

src/DependencyInjection/Compiler/EntityListenerPass.php
```diff
@@ -25,18 +25,15 @@
...
     public function process(ContainerBuilder $container)
     {
-        $resolvers = $this->findAndSortTaggedServices('doctrine.orm.entity_listener', $container);
-
         $lazyServiceReferencesByResolver = [];
 
-        foreach ($resolvers as $reference) {
-            $id = $reference->__toString();
-            foreach ($container->getDefinition($id)->getTag('doctrine.orm.entity_listener') as $attributes) {
+        foreach ($container->findTaggedServiceIds('doctrine.orm.entity_listener') as $id => $tags) {
```
acran, line 33: The next line will sort the tags _within the service_ by priority but the services are now simply in the order they are returned by `$container->findTaggedServiceIds()`, i.e., in the order they were defined.
But for the listeners to be registered in the correct order the tags of _all services_ need to be ordered together.

This probably needs to restructure the code to a single `foreach` over all tags of all returned services?

acran: So something like this:
```php
        $serviceTags = [];
        foreach ($container->findTaggedServiceIds('doctrine.orm.entity_listener') as $id => $tags) {
            foreach ($tags as $attributes) {
                $serviceTags[] = [
                    'serviceId' => $id,
                    'attributes' => $attributes,
                ];
            }
        }

        usort($serviceTags, static fn (array $a, array $b) => ($b['attributes']['priority'] ?? 0) <=> ($a['attributes']['priority'] ?? 0));

        foreach ($serviceTags as $tag) {
            $id            = $tag['serviceId'];
            $attributes    = $tag['attributes'];
```

acran (changes requested):

src/DependencyInjection/Compiler/EntityListenerPass.php
ostrolucky: Test I have added doesn't pass with such a change. Mind raising a PR?

acran: See #2029

PR #1888, opened 2025-06-15
add configuration for php native lazy object

eltharin:
add config option for enable php lazy object (disable by default in 3.4)


docs/en/configuration.rst
```diff
@@ -254,6 +254,8 @@ Configuration Reference
                         class_metadata_factory_name:  Doctrine\ORM\Mapping\ClassMetadataFactory
                         default_repository_class:     Doctrine\ORM\EntityRepository
                         auto_mapping:                 false
+						# Opt-in to PHP native lazy objects
+						enable_native_lazy_objects:   false
```
greg0ire: The indention is weird here

src/DependencyInjection/Configuration.php
```diff
@@ -652,6 +652,7 @@ private function getOrmEntityManagersNode(): ArrayNodeDefinition
                     ->scalarNode('class_metadata_factory_name')->defaultValue(ClassMetadataFactory::class)->end()
                     ->scalarNode('default_repository_class')->defaultValue(EntityRepository::class)->end()
                     ->scalarNode('auto_mapping')->defaultFalse()->end()
+                    ->booleanNode('enable_native_lazy_objects')->defaultFalse()->end()
```
greg0ire: Please add a call to `info` describing what is happening when this is set to `false`.

greg0ire: BTW, this node should probably not exist on PHP 8.4 :thinking:

derrabus:
Can we please add a test for this change?

What would happen, if I enabled this setting with a version of ORM that does not support the native proxies yet?

src/DependencyInjection/Configuration.php
```diff
@@ -652,6 +652,10 @@ private function getOrmEntityManagersNode(): ArrayNodeDefinition
                     ->scalarNode('class_metadata_factory_name')->defaultValue(ClassMetadataFactory::class)->end()
                     ->scalarNode('default_repository_class')->defaultValue(EntityRepository::class)->end()
                     ->scalarNode('auto_mapping')->defaultFalse()->end()
+                    ->booleanNode('enable_native_lazy_objects')
+                        ->defaultFalse()
+                        ->info('Enables the new implementation of PHP lazy objects instead proxies based on lazy ghosts')
```
greg0ire:
```suggestion
                        ->info('Enables the new implementation of PHP lazy objects instead of proxies based on lazy ghosts')
```

eltharin:
> Can we please add a test for this change?

I will try


> What would happen, if I enabled this setting with a version of ORM that does not support the native proxies yet?

Is it better to throw an error / alert in case of php < 8 or orm < 3.4 or say nothing and don't handle this option ?

src/DependencyInjection/Configuration.php
```diff
@@ -652,6 +652,10 @@ private function getOrmEntityManagersNode(): ArrayNodeDefinition
                     ->scalarNode('class_metadata_factory_name')->defaultValue(ClassMetadataFactory::class)->end()
                     ->scalarNode('default_repository_class')->defaultValue(EntityRepository::class)->end()
                     ->scalarNode('auto_mapping')->defaultFalse()->end()
+                    ->booleanNode('enable_native_lazy_objects')
+                        ->defaultFalse()
+                        ->info('Enables the new implementation of PHP lazy objects instead of proxies based on lazy ghosts')
```
ruudk: The message is confusing, this might improve it:

```suggestion
                        ->info('Enables the new native implementation of PHP lazy objects instead of generated proxies')
```

stof (approved):

greg0ire (approved):

greg0ire:
Thanks @eltharin !

ruudk:
Thanks everyone! Would be great to have it tagged so that we can start using this amazing new feature 🤩

greg0ire:
Ask and you shall receive: https://github.com/doctrine/DoctrineBundle/releases/tag/2.15.0

acasademont:
I'm wondering if this new option should have been moved "up" one level, and instead of being a configuration of the `entity_manager` node being a global `orm` one, same as the `enable_lazy_ghost_objects` or the `auto_generate_proxy_classes` one.

It would make supressing the deprecations now being shown when generating the old proxy classes easier.

PR #1890, opened 2025-06-17
Update Symfony bundle metadata

javiereguiluz:
The `doc_dir` section must configure each version explicitly. We didn't include 2.15 and 2.16 branches, so the doc build was failing and the docs at https://symfony.com/bundles/DoctrineBundle/current/index.html were not visible anymore. Thanks.

### DIFF
--- a/.symfony.bundle.yaml
+++ b/.symfony.bundle.yaml
@@ -14,6 +14,11 @@ branches:
 maintained_branches:
     - "2.15.x"
     - "2.16.x"
-doc_dir: { "2.12.x": "docs/", "2.13.x": "docs/en/" }
+doc_dir:
+    "2.12.x": docs/
+    "2.13.x": docs/en/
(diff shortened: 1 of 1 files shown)


.symfony.bundle.yaml
```diff
@@ -14,6 +14,11 @@ branches:
 maintained_branches:
     - "2.15.x"
     - "2.16.x"
-doc_dir: { "2.12.x": "docs/", "2.13.x": "docs/en/" }
+doc_dir:
+    "2.12.x": docs/
```
stof: Is this relevant only for maintained branches ? Old branches are also missing here.

If only maintained branch are necessary, I suggest switching back to the string config as all maintained branches have the new doc dir structure.

javiereguiluz: Yes, that's better. Done. Thanks.

stof (approved):

PR #1891, opened 2025-06-18
Sort middleware without maintaining index association

ruudk:
While debugging an issue with a custom DBAL Driver with Xdebug, I noticed that the order of the middleware was not with what I was expecting.

It turns out this was an issue with Xdebug, because I had the option "Sort Variables Alphabetically" enabled.

But it lead me to discover this.

The MiddlewarePass sorts the middleware based on priority. And it works as expected.

But it sorts it while maintaining index association. There is no point in keeping that. It only confuses others.


ostrolucky:
lgtm, but fix cs pls

ruudk:
Sorry it's fixed now

stof (approved):

Issue #1883, opened 2025-06-04
Wire up orm:mapping:describe command


stlgaits:
### Feature Request / RFC

Not sure whether this belongs here in DoctrineBundle or in ORM repo but I was wondering whether people were interested in adding a simple debug / info console command which prints out Doctrine Entities field mapping.

As far as I'm aware, the only related debug commands currently available are : 

- `bin/console doctrine:mapping:info` which outputs a basic list of mapped entities
-  `bin/console orm:mapping:describe <EntityClassName>` (https://github.com/doctrine/orm/blob/3.3.x/src/Tools/Console/Command/MappingDescribeCommand.php) which provides complete metadata (including field mapping) for a specific entity. However, this command iis not shipped by default with Symfony DoctrineBundle and must be manually registered in the project’s service container.

### Proposal

I would like to add an 'in-between' command which would return a simplified list of entity field mappings - i.e. field names and their types, as I find the `orm:mapping:describe` command way too verbose for my use case and would prefer a lighter alternative.

### Motivation

I originally came up with this idea while frequently switching between files in my project just to check an entity's fields when writing Doctrine queries (especially with QueryBuilder), which becomes cumbersome when working with value objects, embedded fields, etc.
Having a lightweight console command would be a quick win to enhance DX and save time.

### Implementation

I've implemented a prototype for this command (with some inspiration from the 2 commands mentioned above) in a work project. It allows you to pass a specific entity class name as an argument, or list all entities if no argument is given. Here is my current implementation : 

```php
use Doctrine\ORM\EntityManagerInterface;
use Symfony\Component\Console\Attribute\Argument;
use Symfony\Component\Console\Attribute\AsCommand;
use Symfony\Component\Console\Command\Command;
use Symfony\Component\Console\Style\SymfonyStyle;

#[AsCommand(
    name: 'app:debug-doctrine-mapping',
    description: 'Displays mapped fields of a Doctrine entity.',
)]
final class DebugDoctrineEntitiesCommand extends Command
{
    public function __construct(
        private readonly EntityManagerInterface $entityManager
    ) {
        parent::__construct();
    }

    public function __invoke(
        SymfonyStyle $io,
        #[Argument(description: 'The fully qualified class name of the entity', name: 'entity')]
        string|null $entityClass = null,
    ): int {
        $io->title('Debug Doctrine Entity Field Mapping');

        if (null !== $entityClass) {
            if (!class_exists($entityClass)) {
[...]

PR #1892, opened 2025-06-23
Register doctrine:mapping:describe command

stlgaits:
Registers the orm:mapping:describe command as `doctrine:mapping:describe <entity>` for consistency with other debug commands.

Relates to #1883

### DIFF
--- a/config/orm.xml
+++ b/config/orm.xml
@@ -241,6 +241,11 @@
             <tag name="console.command" command="doctrine:mapping:info" />
         </service>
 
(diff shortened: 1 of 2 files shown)


stof:
This should target the 2.16.x branch IMO. It is more a feature than a bugfix.

stlgaits:
> This should target the 2.16.x branch IMO. It is more a feature than a bugfix.

Ah ok thanks I wasn't sure!

greg0ire (approved):

SenseException (approved):

Issue #1895, opened 2025-07-02
"Doctrine\ORM\Proxy\Autoloader" is deprecated


Chris53897:
After upgrading doctrine/orm (3.4.3 => 3.5.0) the following deprecation message is new.

User Deprecated: Class "Doctrine\ORM\Proxy\Autoloader" is deprecated. Use native lazy objects instead. (Autoloader.php:74 called by DoctrineBundle.php:130, https://github.com/doctrine/orm/pull/12005, package doctrine/orm)

https://github.com/doctrine/DoctrineBundle/blob/2.15.x/src/DoctrineBundle.php#L130

I guess the solution is to only load it for PHP < 8.4

acasademont:
There are also some others

```
Calling Doctrine\ORM\Configuration::setProxyDir is deprecated and will not be possible in Doctrine ORM 4.0. (Configuration.php:68 called by getDoctrine_Orm_DefaultConfigurationService.php:40, https://github.com/doctrine/orm/pull/12005, package doctrine/orm)
```

```
Calling Doctrine\ORM\Configuration::setProxyNamespace is deprecated and will not be possible in Doctrine ORM 4.0. (Configuration.php:157 called by getDoctrine_Orm_DefaultConfigurationService.php:41, https://github.com/doctrine/orm/pull/12005, package doctrine/orm)
```

```
Calling Doctrine\ORM\Configuration::setAutoGenerateProxyClasses is deprecated and will not be possible in Doctrine ORM 4.0. (Configuration.php:123 called by getDoctrine_Orm_DefaultConfigurationService.php:42, https://github.com/doctrine/orm/pull/12005, package doctrine/orm)
```

All are related with some recently added deprecations in https://github.com/doctrine/orm/pull/12020 and https://github.com/doctrine/orm/pull/12036

garak:
I bumped into the same deprecation, and I see in the code that the relevant code part is executed only when the `auto_generate_proxy_classes` option is true. Now I'm wondering if the option still makes sense with PHP 8.4, and also if the `enable_lazy_ghost_objects` still makes sense (considering that the changelog for ORM 3.5.0 says it's deprecated)

mbabker:
https://github.com/doctrine/DoctrineBundle/pull/1898 should help with the deprecated `Configuration` class calls.  The proxy autoloader is a lot more complicated though so that'll need another PR.

miqrogroove:
[...]

PR #1898, opened 2025-07-03
Do not configure deprecated proxy options when native lazy ghost objects are enabled

mbabker:
Partially addresses #1895

This will stop calling the ORM Configuration class' deprecated `setAutoGenerateProxyClasses()`, `setProxyDir()`, and `setProxyNamespace()` methods when the native lazy objects feature is enabled and soft-deprecates the relevant config options by noting they'll be ignored in this case.  This purposely only stops calling those methods with ORM 3.5 since 3.4 doesn't have the deprecation notices and issues like https://github.com/doctrine/orm/issues/11997 meant with 3.4.0 specifically there were a couple of cases where not configuring them led to errors.


ostrolucky (approved):

greg0ire (approved):
Looks good. Is the plan to hard deprecate it in 2.17.x?

stof (approved):

Gemorroj:
what about 2.15.x branch?

ostrolucky:
That one is for bug fixes only

Gemorroj:
php 8.4
2.16.x branch
config:
```yaml
    orm:
        auto_generate_proxy_classes: true
        enable_lazy_ghost_objects: true
        report_fields_where_declared: true
        validate_xml_mapping: true
        naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware
        identity_generation_preferences:
            Doctrine\DBAL\Platforms\PostgreSQLPlatform: identity
        auto_mapping: true
        enable_native_lazy_objects: true
        mappings:
            App:
                type: attribute
                is_bundle: false
                prefix: 'App\Entity'
                alias: App
        controller_resolver:
            auto_mapping: false
```
```sh
composer update --with-all-dependencies

Executing script cache:clear [KO]
 [KO]
Script cache:clear returned with error code 1
!!
!!  In ProxyCacheWarmer.php line 49:
!!                                                       
!!    Unable to create the Doctrine Proxy directory "".  
!!                                                       
!!
!!
Script @auto-scripts was called via post-update-cmd
```

ostrolucky:
Hm maybe we should unset these settings only if enable_lazy_ghost_objects = false?

mbabker:
I'm heading out of town so can't help more right now, but lazy ghosts probably shouldn't be enabled alongside native ghosts.  But that's also a little weird because the settings in this PR and that lazy ghost option are in the "don't configure this for each EM" list, but native ghosts can be configured on each EM separately.

mbabker:
The cache warmer issue looks to be a bug in `symfony/doctrine-bridge`, [this block](https://github.com/symfony/symfony/blob/f54db1d4292175b51cc5347e4377b83d00034ad4/src/Symfony/Bridge/Doctrine/CacheWarmer/ProxyCacheWarmer.php#L46-L53) always tries to create the proxy directory and doesn't handle a null return from the object manager configuration's `getProxyDir()` method (both the ORM and MongoDB ODM have nullable returns, the PHPCR ODM defaults to the system temp folder so it won't return null).

Fixing some of this would be easier if the `enable_native_lazy_objects` option were in the [list of keys that isn't configured on each entity manager separately](https://github.com/doctrine/DoctrineBundle/blob/34673de5f6cba99f69454d79cc99951f95f83f84/src/DependencyInjection/Configuration.php#L426-L436), but that ship has sailed (because that'd also give a way to approach not enabling the deprecated proxy autoloader as well).  So right now, if a Symfony application running ORM 3.4 or later on PHP 8.4 has multiple entity managers, and at least one of them doesn't have the native lazy objects option enabled, the cache warmer and proxy autoloader need to be active.

acasademont:
is it too late to chang that option to make it global? it would be a bc-break but i don't think many people have used it yet...As you say it would make it easier to also fix the other deprecation notice

PR #1899, opened 2025-07-07
Document that dbname_suffix has no effects on sqlite

alexander-schranz:
/cc @stof what you think about this?

Related: https://github.com/symfony/recipes/pull/1404


stof:
Please also update the configuration.rst file in the docs.

stof (approved):

PR #1903, opened 2025-07-22
Fix Entity manager is closed on worker mode and PHP 8.4

### LINKED ISSUE #1896: Entity manager is closed on worker mode and PHP 8.4
### Bug Report

| Version                                     | 2.15.0

#### Summary

When using worker mode of FrankenPHP and PHP 8.4, the entity manager is closed after a few requests. Originally posted in https://github.com/symfony/symfony/issues/61005, it turns out that there's a slight bug in the bundle.

#### Current behavior

First request when using the reproducer works fine:

```http
POST /api/foos 
{ "title": "foo" }
```

But next requests are failing because the entity manager is closed.

#### Expected behavior

The entity manager should be reset, but not closed.

#### How to reproduce

Original reporter created a reproducer repo: https://github.com/fvozar/frankenphp-symfony-error. @kraz did a great investigation and turns out that `Registry::resetOrClearManager` is the culprit. I think native lazy objects are causing the problem: I am able to "fix" this by adding ` && \PHP_VERSION_ID < 80400` in the first condition:

https://github.com/doctrine/DoctrineBundle/blob/84f166beb72711b40419a9c0c62aa88729273d51/src/Registry.php#L82

On PHP < 8.4, it's easy to know if the manager is lazy (with interfaces). But I didn't find how to ensure the manager is a native lazy object programmatically.

I'd be happy to propose a fix. I create this issue to gather comments because I doubt that the right fix would be just a check on PHP version. But who knows 🙂 

Thanks!

kraz:
Fixes #1896
related to [symfony/symfony/#61005](https://github.com/symfony/symfony/issues/61005)

### DIFF
--- a/src/Registry.php
+++ b/src/Registry.php
@@ -6,6 +6,7 @@ use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\ORMException;
 use Doctrine\Persistence\Proxy;
 use ProxyManager\Proxy\LazyLoadingInterface;
+use ReflectionClass;
 use Symfony\Bridge\Doctrine\ManagerRegistry;
 use Symfony\Component\DependencyInjection\Container;
 use Symfony\Component\VarExporter\LazyObjectInterface;
@@ -13,6 +14,9 @@ use Symfony\Contracts\Service\ResetInterface;
 
 use function array_keys;
 use function assert;
+use function method_exists;
+
+use const PHP_VERSION_ID;
 
 /**
  * References all Doctrine connections and entity managers in a given Container.
@@ -79,10 +83,27 @@ class Registry extends ManagerRegistry implements ResetInterface
 
         assert($manager instanceof EntityManagerInterface);
 
-        if ((! $manager instanceof LazyLoadingInterface && ! $manager instanceof LazyObjectInterface) || $manager->isOpen()) {
-            $manager->clear();
+        // Determine if the version of symfony/dependency-injection is >= 7.3
+        /** @phpstan-ignore function.alreadyNarrowedType */
+        $sfNativeLazyObjects = method_exists('Symfony\Component\DependencyInjection\ContainerBuilder', 'findTaggedResourceIds');
 
-            return;
+        if (PHP_VERSION_ID < 80400 || ! $sfNativeLazyObjects) {
+            if ((! $manager instanceof LazyLoadingInterface && ! $manager instanceof LazyObjectInterface) || $manager->isOpen()) {
+                $manager->clear();
+
+                return;
+            }
+        } else {
+            $r = new ReflectionClass($manager);
+            if ($r->isUninitializedLazyObject($manager)) {
+                return;
+            }
+
+            if ($manager->isOpen()) {
+                $manager->clear();
+
+                return;
+            }
         }
 
         $this->resetManager($managerName);


greg0ire:
Maybe instead of checking with 3 different ways whether the service is lazy, we could obtain that information from the service definition?

kraz:
Unfortunately there is no way to detect if object itself is natively lazy. It's not possible to detect it at service container compile time by reading the container definition either. But, we can use the version of the Symfony DI bundle to determine if the lazy services are represented as lazy native objects, because since version 7.3 that's exactly the case. Prior version 7.3 the lazy services are `LazyObjectInterface`.

I reconsidered the usage of `$config->isNativeLazyObjectsEnabled()` because it's not meant for that purpose.

After analyzing the changes between 7.2 and 7.3 for:

[symfony/dependency-injection 7.2-7.3 diff](https://github.com/symfony/dependency-injection/compare/7.2...7.3)

I think it's best to check if the following method exists in order to determine if symfony version is >= 7.3:

`Symfony\Component\DependencyInjection\ContainerBuilder->findTaggedResourceIds`

Then:

For PHP < 8.4 or symfony < 7.3 the logic remains as is:
```php
      if ((! $manager instanceof LazyLoadingInterface && ! $manager instanceof LazyObjectInterface) || $manager->isOpen()) {
          $manager->clear();

          return;
      }

      $this->resetManager($managerName);
```

Else:

```php
    $r = new ReflectionClass($manager);
    if ($r->isUninitializedLazyObject($manager)) {
        return;
    }

    if ($manager->isOpen()) {
        $manager->clear();

        return;
    }

    $this->resetManager($managerName);
```

greg0ire:
> Unfortunately there is no way to detect if object itself is natively lazy.

I'm confused… isn't that exactly what https://www.php.net/manual/de/reflectionclass.isuninitializedlazyobject.php is supposed to achieve?

ostrolucky:
IIRC it only checks if it's lazy _uninitialized_ object. If it's initialized then you are out of luck.

kraz:
Yes, that's correct.

> After initialization, the object is indistinguishable from an object that was never lazy.
>
> -- [PHP RFC: Lazy Objects](https://wiki.php.net/rfc/lazy-objects)

src/Registry.php
```diff
@@ -79,10 +83,27 @@ private function resetOrClearManager(string $managerName, string $serviceId): vo
...
+        if (PHP_VERSION_ID < 80400 || ! $sfNativeLazyObjects) {
+            if ((! $manager instanceof LazyLoadingInterface && ! $manager instanceof LazyObjectInterface) || $manager->isOpen()) {
+                $manager->clear();
+
+                return;
+            }
+        } else {
+            $r = new ReflectionClass($manager);
+            if ($r->isUninitializedLazyObject($manager)) {
+                return;
```
ostrolucky, line 99: Doesn't this change the intended flow? Previous logic was calling $managed->clear() regardless if proxy is lazy or not, as long as $manager->isOpen() returns true

kraz: There is no reason to call anything on the object if it's initializer function was not even called yet, i.e. if the lazy object was not yet initialized and we remove this check, then the manager will get initialized just to check if it's open. That's why i think it's safe to leave it as uninitialized. It's what the "symfony-bridge" is already doing.

ostrolucky: Was the previous logic wrong then? Should we adjust it in the other branch too, where it's needlessly calling isOpen() regardless if it's a lazy object or not?

kraz: My reason for not adjusting the previous logic is because I wanted to address just the affected environment the issue arise (only for `>= PHP 8.4` in combination with `symfony/dependency-injection >= 7.3`). In my opinion the old logic is not wrong, but still it initializes the manager if it's not already been initialized, so calling `LazyObjectInterface::isLazyObjectInitialized()` or `LazyLoadingInterface::isProxyInitialized()` is a better approach in my opinion. But again, i wanted the change to affect only the environment in which the issue arise, after all, the old logic has been working for long, long time, so it ok to remain as is, I guess.

ostrolucky (approved):

aleho (approved):
Tested on a bigger project with multiple entity managers and FrankenPHP.

Consecutive requests to the same worker now don't result in "The EntityManager is closed." exceptions any more.

PR #1906, opened 2025-08-05
Do not enable the proxy autoloader or cache warmer when using native lazy objects

### LINKED ISSUE #1895: "Doctrine\ORM\Proxy\Autoloader" is deprecated

mbabker:
(Builds on top of https://github.com/doctrine/DoctrineBundle/pull/1905 and only the second commit is relevant to this PR).

This will fully resolve https://github.com/doctrine/DoctrineBundle/issues/1895 by no longer enabling the deprecated proxy autoloader when native lazy objects are in use.

As well, the proxy cache warmer will no longer be registered when native lazy objects are in use as the warmer doesn't really have a use when proxies aren't in use and it errors out when not configuring the deprecated proxy options (see https://github.com/doctrine/DoctrineBundle/pull/1898#issuecomment-3068208293 and follow-on comments for more here).

Lastly, some of the tests are updated to run with native lazy objects enabled where the environment supports it.

Closes #1895

### DIFF
--- a/src/DependencyInjection/DoctrineExtension.php
+++ b/src/DependencyInjection/DoctrineExtension.php
@@ -18,6 +18,7 @@ use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection;
 use Doctrine\DBAL\Driver\Middleware as MiddlewareInterface;
 use Doctrine\DBAL\Schema\LegacySchemaManagerFactory;
+use Doctrine\ORM\Configuration as ORMConfiguration;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Events;
 use Doctrine\ORM\Id\AbstractIdGenerator;
@@ -71,8 +72,10 @@ use Symfony\Component\VarExporter\ProxyHelper;
 use function array_intersect_key;
 use function array_keys;
 use function array_merge;
+use function assert;
 use function class_exists;
 use function interface_exists;
+use function is_bool;
 use function is_dir;
 use function method_exists;
 use function reset;
@@ -577,7 +580,19 @@ class DoctrineExtension extends AbstractDoctrineExtension
             trigger_deprecation('doctrine/doctrine-bundle', '2.11', 'Not setting "doctrine.orm.enable_lazy_ghost_objects" to true is deprecated.');
         }
 
-        $options = ['auto_generate_proxy_classes', 'enable_lazy_ghost_objects', 'proxy_dir', 'proxy_namespace'];
+        if ($config['enable_native_lazy_objects'] ?? false) {
+            /** @phpstan-ignore function.alreadyNarrowedType */
+            if (! method_exists(ORMConfiguration::class, 'enableNativeLazyObjects')) {
+                throw new LogicException(
+                    'Native lazy objects are not supported with your installed version of the ORM. Please upgrade to "doctrine/orm:^3.4".',
+                );
+            }
+        } elseif (! class_exists(AnnotationDriver::class)) {
+            // Only emit the deprecation notice for ORM 3 users
+            trigger_deprecation('doctrine/doctrine-bundle', '2.16', 'Not setting "doctrine.orm.enable_native_lazy_objects" to true is deprecated.');
+        }
+
+        $options = ['auto_generate_proxy_classes', 'enable_lazy_ghost_objects', 'enable_native_lazy_objects', 'proxy_dir', 'proxy_namespace'];
         foreach ($options as $key) {
             $container->setParameter('doctrine.orm.' . $key, $config[$key]);
         }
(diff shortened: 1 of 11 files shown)


src/DependencyInjection/DoctrineExtension.php
```diff
@@ -577,7 +580,21 @@ protected function ormLoad(array $config, ContainerBuilder $container)
...
         }
 
-        $options = ['auto_generate_proxy_classes', 'enable_lazy_ghost_objects', 'proxy_dir', 'proxy_namespace'];
+        if ($config['enable_native_lazy_objects'] ?? false) {
+            /** @phpstan-ignore function.alreadyNarrowedType */
+            if (! method_exists(ORMConfiguration::class, 'enableNativeLazyObjects')) {
+                throw new LogicException(
+                    'Native lazy objects are not supported with your installed version of the ORM. Please upgrade to "doctrine/orm:^3.4".',
+                );
+            }
```
beberlei, line 589: This is also missing a validation that PHP 8.4 is used, otherwise error because native lazy is not supported then. This would throw the error earlier than doctrine on configuration level, otherwise it would be delayed to runtime when ORM is used.

heiglandreas: Why would one set `enable_native_lazy_objects` to `true` when they are not at PHP8.4?

Besides that: It checks whether the ORM config has the necessary method to handle the config-value. 

I'd rather check that setting `enable_native_lazy_object` to `true` is only possible starting with PHP8.4. So even if someone sets it ti `true` but runs the code on 8.3 it should get set to `false` automatically.

WDYT?

greg0ire: I don't think it should be set to `false` silently. Maybe it's even possible to just not define the configuration node for PHP < 8.4

stof: Making configuration classes conditional on the PHP version is a bad idea: it makes it a mess to document the structure of the configuration and provides non-friendly error messages to the user ("configuration setting `enable_native_lazy_object` does not exist.", while they will see it in the documentation).
Note also that making conditional configuration in Yaml files is not possible, so this will make it impossible to write a recipe for the bundle that turns on the feature.

heiglandreas: Silently setting it to false (which is the same as just ignoring it) would allow people to configure it but it would not break when running on some not-supported system.

I can think of i.e. staging running on 8.4 but prod is still on 8.3 - silently setting it to false will allow one setting for both and it would run on both without issues.

mbabker: > Silently setting it to false (which is the same as just ignoring it) would allow people to configure it but it would not break when running on some not-supported system.

The option is already effectively no-op'ed for PHP 8.3 and earlier in https://github.com/doctrine/DoctrineBundle/pull/1906/files#diff-e2aeb4a501e79bc8ebab5ea93069b5bbcd3eb9168d37bd78ddd43075cddb8cc4R722 (the `Configuration::enableNativeLazyObjects()` method call is only made on PHP 8.4+).  So the "silently setting to false" thing is already happening in the tagged 2.15 releases.

greg0ire: > I can think of i.e. staging running on 8.4 but prod is still on 8.3

Is that something people actually do? Pushing the same code on 2 machines with different versions of PHP? It seems risky to me, for instance if you use a php 8.4-specific feature, this setup will cause an issue that surfaces only in production. Plus it's not like they could wait until production also uses 8.4 to enable native lazy objects.

beberlei: the ORM throws an exception if you are < 8.4 and use this setting. 

https://github.com/doctrine/orm/blob/3.5.x/src/Configuration.php#L668

So that is the behavior that exists and we should not "change" this behavior to be something different, because thats confusing. What we can do is surface this error earlier, which we can do by doing the check here again, moving this to a "compile time" error instead of waiting for it to happen for runtime.

erseco:
Thanks for this! We’re running PHP 8.4 with ORM 3.4+ and switching to `enable_native_lazy_objects: true`. Happy to help test a 2.16.x release/RC.

src/DependencyInjection/DoctrineExtension.php
mbabker: https://github.com/doctrine/DoctrineBundle/pull/1906/commits/67754db9a369e0de3cf26523ef1723701367af44 adds a throw at compile time for this.

PR #1907, opened 2025-08-07
feat: merge URL driverOption rather than just replace defaults

le-yak:
### Current behaviour
If the connection URL contains `driverOptions`, they completely replace the defaults.

```dotenv
# .env.local
DATABASE_URL=sqlsrv://db?driverOptions[TrustServerCertificate]=1
```
```yaml
# doctrine.yaml
doctrine:
    dbal:
        options:
            FormatDecimals: true
```

Result: `FormatDecimals` is discarded.

###  Proposed change
`driverOptions` from the URL are merged into the defaults, allowing environment-independant settings to be defined (e.g., I always want my decimals formatted, regardless of the environment) as well as allowing _additional_ settings, and even overrides, per environment (e.g., only in my dev env do I not have a proper certificate)

Desired result: both `FormatDecimals` and `TrustServerCertificate` are applied.

### DIFF
--- a/src/ConnectionFactory.php
+++ b/src/ConnectionFactory.php
@@ -275,6 +275,13 @@ class ConnectionFactory
             unset($params['driverClass']);
         }
 
+        // If both the URL and the default parameters define driverOptions,
+        // then merge them rather than discard all defaults.
+        // Options from the URL take precedence.
+        if (isset($params['driverOptions']) && isset($parsedParams['driverOptions'])) {
+            $parsedParams['driverOptions'] = array_merge($params['driverOptions'], $parsedParams['driverOptions']);
+        }
+
         $params = array_merge($params, $parsedParams);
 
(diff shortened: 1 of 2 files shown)


ostrolucky:
I think we are going to have to reject this, because this is against the intended way of working. DSN in Symfony is meant to always override the config defined options, see this comment https://github.com/doctrine/DoctrineBundle/pull/1328#issuecomment-824112731 (and entire thread). Furthermore, this kind of change will cause some sudden BC breaks.

le-yak:
Thank you for your reply. I agree that introducing a BC is problematic, I could disable this by default and introduce a new feature flag.

Just one last argument: in my proposal DSN still always wins. The intention is to avoid silently (and surprisingly IMO) discarding hard coded configurations.

Nonetheless consistency is paramount, I understand if the proposal is rejected. There are workarounds.

PR #1914, opened 2025-08-25
Convert config from XML to PHP

nicolas-grekas:
PHP is nicer than XML and https://github.com/symfony/symfony/pull/60568 might make this required at some point.


nicolas-grekas:
All green \o/

config/dbal.php
```diff
@@ -0,0 +1,146 @@
...
+
+        ->set('doctrine.dbal.connection_factory', param('doctrine.dbal.connection_factory.class'))
+            ->args([
+                param('doctrine.dbal.connection_factory.types'),
+                service('doctrine.dbal.connection_factory.dsn_parser'),
+            ])
+
+        ->set('doctrine.dbal.connection_factory.dsn_parser', DsnParser::class)
+            ->args([
+                [],
```
ostrolucky, line 64: I'm not sure why are there some definitions that are setting empty args like this?

nicolas-grekas: It maps to `<argument type="collection" />`

ostrolucky (approved):

SenseException (approved):

config/orm.php
```diff
@@ -0,0 +1,367 @@
...
+            ])
+            ->tag('doctrine.event_listener', ['event' => 'postGenerateSchema'])
+
+        ->set('doctrine.orm.naming_strategy.default', param('doctrine.orm.naming_strategy.default.class'))
+
+        ->set('doctrine.orm.naming_strategy.underscore', param('doctrine.orm.naming_strategy.underscore.class'))
+
+        ->set('doctrine.orm.naming_strategy.underscore_number_aware', param('doctrine.orm.naming_strategy.underscore.class'))
+            ->args([
+                'CASE_LOWER',
```
thePanz, line 233: @nicolas-grekas this should be an integer (the PHP constant), according to the UnderscoreNamingStrategy class

Just noticed in some deprecations :)

nicolas-grekas: arf, can you please send a PR?

thePanz: Done, already merged by @stof :)

config/dbal.php
```diff
@@ -0,0 +1,146 @@
...
+    $container->parameters()
+        ->set('doctrine.dbal.configuration.class', Configuration::class)
+        ->set('doctrine.data_collector.class', DoctrineDataCollector::class)
+        ->set('doctrine.dbal.connection.event_manager.class', ContainerAwareEventManager::class)
+        ->set('doctrine.dbal.connection_factory.class', ConnectionFactory::class)
+        ->set('doctrine.dbal.events.mysql_session_init.class', MysqlSessionInit::class)
+        ->set('doctrine.dbal.events.oracle_session_init.class', OracleSessionInit::class)
+        ->set('doctrine.class', Registry::class)
+        ->set('doctrine.entity_managers', [])
+        ->set('doctrine.default_entity_manager', null)
```
jdreesen, line 39: I just use DBAL without the ORM and this change breaks my container:

> TypeError: Doctrine\Bundle\DoctrineBundle\Registry::__construct(): Argument #5 ($defaultEntityManager) must be of type string, null given, called in /var/www/html/tests/app/var/cache/test/ContainerVbFK5Uk/getDoctrineService.php on line 27

It was `<parameter key="doctrine.default_entity_manager"></parameter>` in the XML config, which seems to result in an empty `string` instead of `null`.

jdreesen: I see that there's already [an issue](https://github.com/doctrine/DoctrineBundle/issues/1925) and [a PR](https://github.com/doctrine/DoctrineBundle/pull/1969).

PR #1916, opened 2025-08-27
Shift query backtrace so that class->method corresponds to correct file

### LINKED ISSUE #1915: [Profiling] profiling_collect_backtrace generating wrong line number
### Bug Report

Image
Image

| Version                                     | 2.15.1

#### Summary

`profiling_collect_backtrace` generating wrong line numbers

#### How to reproduce

`profiling_collect_backtrace: true`

ostrolucky:
Fixes #1915

Before:
image
After:
image

### DIFF
--- a/templates/Collector/db.html.twig
+++ b/templates/Collector/db.html.twig
@@ -287,12 +287,12 @@
                                                                 <td>{{ loop.index }}</td>
                                                                 <td>
                                                                             <span class="text-small">
-                                                                                {% set line_number = trace.line|default(1) %}
+                                                                                {% set line_number = trace.line|default(query.backtrace[loop.index].line|default(1)) %}
                                                                                 {% if trace.file is defined %}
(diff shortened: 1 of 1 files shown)


templates/Collector/db.html.twig
```diff
@@ -440,8 +440,8 @@
                                     {% set contains_errors = collector.mappingErrors[manager] is defined and collector.mappingErrors[manager][class.class] is defined %}
                                     <tr class="{{ contains_errors ? 'status-error' }}">
                                         <td>
-                                <a href="{{ class.file|file_link(class.line) }}">{{ class. class}}</a>
-                            </td>
+                                            <a href="{{ class.file|file_link(class.line) }}">{{ class. class}}</a>
+                                        </td>
```
thePanz, line 444: Looks like this is not relevant for the PR :)

ostrolucky: Right, but I didn't think it's worth it to make it a separate commit :\

PR #1918, opened 2025-09-01
Replace string with constant for CASE_LOWER

thePanz:
UnderscoreNamingStrategy signature is `public function __construct(private int $case = CASE_LOWER)`, moving this to use the constant.

ping @nicolas-grekas as he worked on those changes too :)


stof (approved):
That's a mistake in the conversion from XML to PHP. the XML file was using `<argument type="constant">`.

thePanz:
Thanks @stof for merging :+1:

PR #1921, opened 2025-09-02
Update branch metadata

greg0ire:
- 3.0.x is the new latest branch.
- 2.17.x is the new minor branch.
- 2.16.x is the new patch branch.
- 2.15.x is no longer maintained.

### DIFF
--- a/.doctrine-project.json
+++ b/.doctrine-project.json
@@ -1,4 +1,5 @@
 {
+  "active": true,
   "name": "Doctrine Bundle",
   "shortName": "DoctrineBundle",
   "slug": "doctrine-bundle",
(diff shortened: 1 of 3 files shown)


.doctrine-project.json
```diff
@@ -1,19 +1,32 @@
 {
+  "active": true,
```
greg0ire, line 2: @SenseException I had to add this property to have the new workflow pass. Maybe it shouldn't be mandatory, since it was working without it? Or was something, in fact, broken? Or maybe it will allow cleaning up some code?

ostrolucky (approved):

derrabus (approved):

.doctrine-project.json
SenseException: I don't know about that one. It could also be fetched from the repository itself. `true` is also the default value in the website-build code.

PR #1922, opened 2025-09-02
Remove "collate" default table option

greg0ire:
It has been deprecated in favor of "collation".

### DIFF
--- a/UPGRADE-3.0.md
+++ b/UPGRADE-3.0.md
@@ -4,6 +4,10 @@ UPGRADE FROM 2.x to 3.0
 Configuration
 -------------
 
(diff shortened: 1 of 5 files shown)


UPGRADE-3.0.md
```diff
@@ -4,6 +4,10 @@ UPGRADE FROM 2.x to 3.0
 Configuration
 -------------
 
+### The `collate` default table option is removed
```
ostrolucky: Can you expand the full path of the options please. Here I think it's `doctrine.dbal.default_table_options.collate`

ostrolucky (approved):

PR #1923, opened 2025-09-02
Drop support for doctrine/cache

greg0ire:


UPGRADE-3.0.md
```diff
@@ -1,6 +1,11 @@
 UPGRADE FROM 2.x to 3.0
 =======================
 
+Compatibility
+-------------
+
+Using `doctrine/cache` is no longer possible. Migrate to PSR-6 instead.
```
ostrolucky: I think this isn't specific enough. What isn't possible? What should user do? Surely users can continue using doctrine/cache, they just cannot configure doctrine caching options to use services binding to it directly.

ostrolucky:
Should we perhaps remove doctrine/cache from the conflict section in composer.json too?

ostrolucky (approved):

PR #1924, opened 2025-09-02
Remove controller resolver auto mapping configuration

bobvandevijver:
Remove controller resolver auto mapping configuration as discussed in #1804.

Already added upgrade files for 3.1 and 4.0 as additional actions need to be taken for those versions.


greg0ire:
The breaking changes should be documented in `UPGRADE-3.0.md`.

UPGRADE-3.1.md
greg0ire, line 1: This file should IMO be contributed when 3.1.x is made available.

UPGRADE-4.0.md
greg0ire, line 1: Same here.

bobvandevijver:
> The breaking changes should be documented in `UPGRADE-3.0.md`.

It already has been added to that file in the previous PR.

greg0ire:
Oh right!

UPGRADE-3.1.md
bobvandevijver: This wasn't the case with the previous PR, which is why I also it here. It also serves as a reminder that it should be deprecated/removed in that version.

But just say the word, can obviously remove it 👍🏻

greg0ire: Yeah, we don't usually do that in any other Doctrine repository, and sometimes, what actually gets done can differ from the original plan, so I would refrain to commit such "prophecies" to the repository. Maybe the best reminder for this would be an issue, I'm not sure.

ostrolucky (approved):

UPGRADE-3.1.md
bobvandevijver: I've removed the documents, and created #1926 and #1927 instead. There are no milestones yet, but it might be nice to create those already and link the issues. I cannot assign myself, but please give me a heads-up when the relative branches are opened, happy to contribute the required changes!

greg0ire (approved):

greg0ire:
Thanks @bobvandevijver !

PR #1929, opened 2025-09-03
Drop old major versions

greg0ire:
This should cause a drop in code coverage, producing a lot of dead code. Subsequent changes removing said dead code should not impact the coverage in terms of number of lines covered, and would be way too substantial to be reviewed in one go.

Support for Symfony LTS is kept.
 I propose that we work on removing things like support for YML mapping separately. I'm taking this example because it's just one part of dropping support for ORM 2, and there should IMO be several PRs to drop support for ORM 2 completely.

### DIFF
--- a/UPGRADE-3.0.md
+++ b/UPGRADE-3.0.md
@@ -7,6 +7,14 @@ Compatibility
 Configuring caching options to use services backed by `doctrine/cache` is no
 longer supported. Migrate to PSR-6 services instead.
 
+Support for the following major versions of the following packages has been dropped:
+
+- `doctrine/dbal` 3
+- `doctrine/persistence` 3
+- `doctrine/orm` 2
+- `psr/log` 1 and 2
+- `twig/twig` 2
+
 Configuration
 -------------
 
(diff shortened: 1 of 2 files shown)


UPGRADE-3.0.md
```diff
@@ -7,6 +7,14 @@ Compatibility
 Configuring caching options to use services backed by `doctrine/cache` is no
 longer supported. Migrate to PSR-6 services instead.
 
+Support for the following major versions of the following packages has been dropped:
+
+- `doctrine/dbal` 3
+- `doctrine/persistence` 3
+- `doctrine/orm` 2
+- `psr/log` 1 and 2
+- `twig/twig` 2
```
greg0ire, line 16: Note that I'm not listing `doctrine/annotations` here, since ultimately, it won't appear in our `composer.json` anymore.

greg0ire:
-3.82% coverage drop… I was expecting more, and I was expecting in particular more YML-related stuff in what Codecov lists… but it does list stuff that should be removed.

SenseException (approved):

PR #1930, opened 2025-09-04
Remove deprecated Symfony commands

greg0ire:


greg0ire:
These commands were mostly not covered => +6.79% in code coverage

dmaicher:
Should we also remove this service definition? 

https://github.com/doctrine/DoctrineBundle/blob/a6e0baa5df7f4415f4d6ff5008dccda400b99b58/config/orm.php#L292-L296

I think this command was removed on ORM 3

src/Command/ImportMappingDoctrineCommand.php
```diff
@@ -1,166 +0,0 @@
...
-        ManagerRegistry $doctrine,
-        private readonly array $bundles,
-    ) {
-        parent::__construct($doctrine);
-    }
-
-    protected function configure(): void
-    {
-        $this
-            ->setName('doctrine:mapping:import')
```
dmaicher, line 46: we also need to remove the service definition then

https://github.com/doctrine/DoctrineBundle/blob/5ed195e326f4a5a35fa507968f8b197e24db07b3/config/orm.php#L326-L331

dmaicher (approved):

PR #1938, opened 2025-09-04
Remove branchName from unmaintained branches

greg0ire:
Since doctrine/doctrine-website#372, they are no
longer necessary, it's possible to rely on tags. Once this is merged, the branches can be removed.


greg0ire:
Note: before removing the branch, make sure that open PRs target branches that won't be removed.

ostrolucky (approved):

PR #1943, opened 2025-09-04
Remove code related to doctrine/dbal 3

greg0ire:


docs/en/configuration.rst
```diff
@@ -676,32 +676,6 @@ the ORM resolves to:
...
-If the environment format configured in oracle does not meet doctrine requirements,
-you need to use the OracleSessionInit listener so that doctrine is aware of the format used by Oracle DB.
-
-You can do so easily with
-
-.. code-block:: yaml
-
-    services:
-        oracle.listener:
-            class: Doctrine\DBAL\Event\Listeners\OracleSessionInit
```
greg0ire, line 699: Since https://github.com/doctrine/dbal/pull/5786/files#diff-ea2eb9d59d26cb97d2f5ef4691e42b53c6c896dbf77c45dcef39cd1cdfc438afR10, this class is deprecated in favor of using a middleware.

We should probably show both ways in our docs.

greg0ire: https://github.com/doctrine/DoctrineBundle/pull/1944

greg0ire:
Technically there are breaking changes in there, but I don't think they are worth documenting.

derrabus (approved):

PR #1944, opened 2025-09-04
Document up-to-date way to deal with Oracle

greg0ire:
The documented way only works for DBAL 2

### DIFF
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -684,13 +684,21 @@ Oracle DB
 ~~~~~~~~~
 
 If the environment format configured in oracle does not meet doctrine requirements,
-you need to use the OracleSessionInit listener so that doctrine is aware of the format used by Oracle DB.
+you need to use a middleware so that doctrine is aware of the format
+used by Oracle DB. With ``doctrine/dbal``, the same could be done with
+an event listener.
 
 You can do so easily with
 
 .. code-block:: yaml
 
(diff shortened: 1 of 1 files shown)


docs/en/configuration.rst
```diff
@@ -684,13 +684,21 @@ Oracle DB
...
 You can do so easily with
 
 .. code-block:: yaml
 
     services:
+        # DBAL 3
+        oracle.middleware:
+            class: Doctrine\DBAL\Driver\OCI8\Middleware\InitializeSession
+            tags:
+                - { name: doctrine.middleware, connection: default }
```
greg0ire, line 700: Note that I have no idea if this actually works. Based on this piece of code, I think it could: https://github.com/doctrine/DoctrineBundle/blob/437a977d05373c7e06091ff7256aa6ae30a755c3/src/DependencyInjection/Compiler/MiddlewaresPass.php#L30-L44

It it does not hopefully somebody will send a PR to fix it, but it's IMO better to push people towards something that has better chances to work if they use DBAL 3.

greg0ire:
Cc @l-vo IIRC you worked on this topic.

docs/en/configuration.rst
derrabus: We should probably find a better way to enable the middleware. This is very… technical.

derrabus (approved):

PR #1957, opened 2025-09-04
Rely even less on YML in tests

greg0ire:
While [working on removing YML on the next major branch](https://github.com/doctrine/DoctrineBundle/pull/1958), I found a lot more places where I do not think it needs to be used.

### DIFF
--- a/tests/DependencyInjection/AbstractDoctrineExtensionTestCase.php
+++ b/tests/DependencyInjection/AbstractDoctrineExtensionTestCase.php
@@ -549,8 +549,8 @@ abstract class AbstractDoctrineExtensionTestCase extends TestCase
         ]);
 
         $this->assertDICDefinitionMethodCallAt(1, $definition, 'addDriver', [
-            new Reference('doctrine.orm.default_yml_metadata_driver'),
-            'Fixtures\Bundles\YamlBundle\Entity',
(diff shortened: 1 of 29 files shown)


tests/DependencyInjection/AbstractDoctrineExtensionTestCase.php
```diff
@@ -566,11 +566,6 @@ public function testSingleEntityManagerMultipleMappingBundleDefinitions(): void
             ! class_exists(AnnotationDriver::class),
         ]);
 
-        $ymlDef = $container->getDefinition('doctrine.orm.default_yml_metadata_driver');
-        $this->assertDICConstructorArguments($ymlDef, [
-            [__DIR__ . DIRECTORY_SEPARATOR . 'Fixtures' . DIRECTORY_SEPARATOR . 'Bundles' . DIRECTORY_SEPARATOR . 'YamlBundle' . DIRECTORY_SEPARATOR . 'Resources' . DIRECTORY_SEPARATOR . 'config' . DIRECTORY_SEPARATOR . 'doctrine' => 'Fixtures\Bundles\YamlBundle\Entity'],
-        ]);
```
greg0ire, line 572: This is the part that I'm not so sure about 🤔

ostrolucky (approved):

PR #1958, opened 2025-09-04
Remove support for YML

greg0ire:


ostrolucky (changes requested):
You removed way more than was supposed to, especially yaml files in tests/DependencyInjection/Fixtures/config/yml/ dir. Those are doctrine-bundle configs, not orm mappings

greg0ire:
Oh right, that makes sense :sweat_smile:

tests/DependencyInjection/YamlDoctrineExtensionTest.php
```diff
@@ -1,16 +0,0 @@
-<?php
-
-namespace Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection;
-
-use Symfony\Component\Config\FileLocator;
-use Symfony\Component\DependencyInjection\ContainerBuilder;
-use Symfony\Component\DependencyInjection\Loader\YamlFileLoader;
-
-class YamlDoctrineExtensionTest extends AbstractDoctrineExtensionTestCase
```
ostrolucky: should be kept

ostrolucky (approved):

PR #1969, opened 2025-09-05
Fix type error in Registry when no entity managers are configured

### LINKED ISSUE #1925: Registry causes Exception on constructor when default-entity-manager is not configured/defined
### Bug Report

| Version                                     | 2.16.0

#### Summary

`doctrine`-service can not instantiate when no repository is defined.

> I'm not using the ORM but using the connection for direct DBAL access

#### Current behavior

When trying to execute `doctrine:migrations:migrate` it will try to instatiate `doctrine` (instance of `Doctrine\Bundle\DoctrineBundle\Registry`) which will crash on the 5th parameter ($defaultEntityManager) expecting a `string` but setting `NULL`.

#### Expected behavior

Service should instantiate normally (with limitations or exceptions being thrown on interaction?).

I am not sure how it worked before `2.16.0` but the service caused no problems at `2.15.1`.

#### How to reproduce

Bringing the system to try and instantiate the `doctrine`-service.

e.g.:
- install doctrine migrations
- remove definition of any entity managers
- generate empty migration (e.g. with `doctrine:migrations:generate`)
- execute migration -> will crash kernel

Bosswell:
Closes #1925

### DIFF
--- a/config/dbal.php
+++ b/config/dbal.php
@@ -36,7 +36,7 @@ return static function (ContainerConfigurator $container): void {
         ->set('doctrine.dbal.events.oracle_session_init.class', OracleSessionInit::class)
         ->set('doctrine.class', Registry::class)
         ->set('doctrine.entity_managers', [])
-        ->set('doctrine.default_entity_manager', null);
+        ->set('doctrine.default_entity_manager', '');
 
     $container->services()
 


greg0ire:
Please associate your email address with your Github account, or change the
email in your commits to an address already associated with it. If you do not
want to expose your personal email address, you may use
`$userName@users.noreply.github.com`, that way we can still reach you through Github.
Additionally, you will get credit for this contribution on your Github profile.

greg0ire (approved):
Not a blocker, but if you could implement https://github.com/doctrine/DoctrineBundle/issues/1925#issuecomment-3253228000, that would be nice.

ostrolucky (approved):

PR #1970, opened 2025-09-05
Fix `doctrine.default_entity_manager` to be an empty `string` by default

jdreesen:
See: https://github.com/doctrine/DoctrineBundle/pull/1914#discussion_r2325298090


xabbuh:
There already is an open PR: #1969

jdreesen:
Oh, I didn't see that. Thx for the hint!

Issue #1968, opened 2025-09-05
TypeError: Registry::__construct() fails with null default entity manager in DBAL-only setup


jbatko-pg:
# Bug Report

| Q            | A
|--------------|-----------------
| Version      | 2.16.x

## Summary

When using DoctrineBundle **without the ORM component** (DBAL-only setup), container compilation fails because the parameter `doctrine.default_entity_manager` is set to `null`, but `Registry::__construct()` expects a `string`.

## Current Behavior

In `config/dbal.php, the following parameter is set:

```
$container->parameters()
    ->set('doctrine.dbal.configuration.class', Configuration::class)
    ->set('doctrine.data_collector.class', DoctrineDataCollector::class)
    ->set('doctrine.dbal.connection.event_manager.class', ContainerAwareEventManager::class)
    ->set('doctrine.dbal.connection_factory.class', ConnectionFactory::class)
    ->set('doctrine.dbal.events.mysql_session_init.class', MysqlSessionInit::class)
    ->set('doctrine.dbal.events.oracle_session_init.class', OracleSessionInit::class)
    ->set('doctrine.class', Registry::class)
    ->set('doctrine.entity_managers', [])
    ->set('doctrine.default_entity_manager', null); // <-- problematic line
```

This results in:

```
Doctrine\Bundle\DoctrineBundle\Registry::__construct():
Argument #5 ($defaultEntityManager) must be of type string, null given
```

## Expected Behavior

Container compilation should succeed in DBAL-only projects.  
`doctrine.default_entity_manager` should always be a string (empty string `''` if ORM is not used).

## How to Reproduce

1. Install `doctrine/doctrine-bundle:^2.16` and `doctrine/dbal`, **without** `doctrine/orm`.
2. Configure only DBAL in `config/packages/doctrine.yaml`:

```
doctrine:
  dbal:
    url: '%env(resolve:DATABASE_URL)%'
```

3. Run:

```
php bin/console cache:clear
```

4. Observe the `TypeError` during container compilation.

## Regression

Introduced by [PR #1914](https://github.com/doctrine/DoctrineBundle/pull/1914/files).

## Impact

This bug prevents Symfony projects that use DoctrineBundle solely for DBAL and migrations (without ORM) from compiling and running.

## Proposed Fix

Set `doctrine.default_entity_manager` to `''` (empty string) instead of `null` when no entity managers are configured. This satisfies the `Registry::__construct()` type hint and resolves the error.

[...]

PR #1983, opened 2025-09-06
Enable strict types

greg0ire:
It is in our coding standard, and doing so is not a breaking change.

### DIFF
--- a/.doctrine-project.json
+++ b/.doctrine-project.json
@@ -6,17 +6,19 @@
   "versions": [
     {
       "name": "3.0",
+      "branchName": "3.0.x",
       "slug": "latest",
       "upcoming": true
     },
     {
       "name": "2.17",
+      "branchName": "2.17.x",
       "slug": "2.17",
       "upcoming": true
     },
     {
       "name": "2.16",
-      "slug": "latest",
+      "slug": "2.16",
       "current": true
     },
     {
--- a/config/dbal.php
+++ b/config/dbal.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Symfony\Component\DependencyInjection\Loader\Configurator;
 
 use Doctrine\Bundle\DoctrineBundle\Command\CreateDatabaseDoctrineCommand;
@@ -44,7 +46,7 @@ return static function (ContainerConfigurator $container): void {
         ->alias(ManagerRegistry::class, 'doctrine')
         ->alias(LegacyManagerRegistry::class, 'doctrine')
 
-        ->set('data_collector.doctrine', param('doctrine.data_collector.class'))
+        ->set('data_collector.doctrine', (string) param('doctrine.data_collector.class'))
             ->args([
                 service('doctrine'),
                 true,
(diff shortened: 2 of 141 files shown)


ostrolucky:
Does it at least reduce phpstan errors? I'm not much of a fan of this change, as it conflicts with Symfony CS and doesn't really bring us much. On the other hand it makes code more brittle and gives us more work of having to juggle things between types, while PHP could automatically for us.

greg0ire:
It does not reduce PHPStan errors, even at max level.

> it conflicts with Symfony CS

OK, but so what? This repository is in the doctrine organization and is checked with PHPCodeSniffer, using the Doctrine coding standard. I don't think we should strive to be close to Symfony CS here, which full disclosure, I don't like [for](https://dev.to/greg0ire/why-using-yoda-conditions-you-should-probably-not) [several](https://dev.to/greg0ire/why-i-keep-pestering-everyone-about-long-lines) [reasons](https://dev.to/greg0ire/the-case-for-multiline-signatures-dl6).

>  On the other hand it makes code more brittle and gives us more work of having to juggle things between types, while PHP could automatically for us.

The only thing I've had to change is to do some string casts. Benefiting from PHP automatic type casts is something I think is OK in a project, but in a library, I think it's more important to avoid silly type errors, even if it means slightly more ugly code.

Finally since it's in our standard, that's what we do in all our other projects, including other Symfony bundles like the migrations or fixtures bundle. I don't see why this bundle should get a special treatment.

ostrolucky (approved):
I'm approving this, because I don't think it's worth it to make a fuss about such change. I am not a fan personally but at the same time it's not such a big deal that I want to block this. However, I want to remind certain things:

1. This is a major symfony bundle. It's sitting between symfony and doctrine. Even supported PHP versions are mirrored to what Symfony LTS supports. Compromises from both sides are important here. It's not like you implied that repo is in doctrine organization so we do whatever doctrine wants. By the way this is why I didn't want to eagerly start 3.x branch. There are dependencies in Symfony to things we deprecated long time ago (eg. `%.class` params) that we cannot drop before Symfony stops using them. Such change is unlikely to happen in Symfony LTS and we need to support it. So it will take years until we can release 3.x (so that we can continue supporting LTS), unless we maintain both 2.x and 3.x like last time, which I would really like to avoid. On the other hand, maybe existence of 3.x having those things removed will push Symfony to stop using them?
2. This is a risky change and our CI is not perfect, like we saw in https://github.com/doctrine/DoctrineBundle/issues/1968. Whole point of strict types is that it will blow up more often than non-strict.

greg0ire:
> There are dependencies in Symfony to things we deprecated long time ago (eg. %.class params) that we cannot drop before Symfony stops using them

Ok, so either we drop support for 6.4 on 3.x, or we keep the `%.class`, is that it? I think both are fine, the important part for this 3.x release IMO is more to be aligned with ORM 4, and for ORM 4, the important thing is to require PHP 8.4 so that native lazy objects are available. We can totally postpone the removal of `%.class` to 4.x if you feel it's better.

SenseException:
Then the target branch is supposed to be `3.0.x`, right?

greg0ire:
:thinking: no… I mean, there is no breaking change here, is there?

>  Strict typing applies to function calls made from within the file with strict typing enabled, not to the functions declared within that file. If a file without strict typing enabled makes a call to a function that was defined in a file with strict typing, the caller's preference (coercive typing) will be respected, and the value will be coerced. 

(from https://www.php.net/manual/en/language.types.declarations.php)

SenseException (approved):

PR #1984, opened 2025-09-06
Use native final modifier over annotation

greg0ire:


ostrolucky (changes requested):
I'm going to 👎 this one, but let's see if there are more votes. We use `@final` to wipe our hands clean in case we cause a BC break for someone. At the same time, we don't take away possibility to extend for people who still need to do it, BC break possibility be damned. Common use case when mocking for example. 

`final` instead of `@final` is going to prevent the latter, while not giving us anything in return. I've yet to seen any single issue created by someone complaining library maintainer broke something while they extended something marked `@final`. However, I've seen many issues created by people wanting to extend something but couldn't, because it was marked `final`.

greg0ire:
Interesting. I assumed these were introduced with the goal to make them hard final in mind. Also, I thought it was unlikely that anybody extended anything from the bundle, but I did not thought of mocking. Let's see what other maintainers think.

SenseException:
> Interesting. I assumed these were introduced with the goal to make them hard final in mind. 

That was also my assumption. And that `@final` is used in case there were mocks in tests and it wasn't possible to use `final` at the current time. If we also assume that the final annotation is supposed to be handled like `final` itself, we would introduce small changes (only breaks when extended) in method signatures, since no one is supposed to extend such a class. If we have to keep BC of `@final` classes, then that final annotation would have not much meaning.

ostrolucky:
> I assumed these were introduced with the goal to make them hard final in mind.

No, this is (or was) unofficial Symfony standard. Symfony didn't used to make things hard final. It was unofficial policy to make everything soft final. Loud minority from community started to insist on self-sabotaging and make things hard final because it's prettier to use native language construct than annotation (I would perhaps have some understanding if it was library maintainers themselves, but we are talking about people using the library starting to contribute such changes into library they are using..), so more and more code started to be added with native final. I'm not sure where symfony maintainers stand now on this, I think they gave up since then because they were hassled on this.

> I thought it was unlikely that anybody extended anything from the bundle

Yes it's unlikely, but not impossible. I don't like playing god and hard removing the possibilities just because I personally don't see the use cases. If we don't see the use case despite one existing and have classes `final`, people will have to fork the library and perhaps open up a PR. Because unlike `private` methods, there is no reflection hack there that can remove it. Only way is to open another can of worms with hacking composer autoloader which modifies those files on read. Anyways that PR? That PR is very unlikely to be merged because maintainers are very hesitant to remove final keywords. At first maintainers claimed they don't see the use case. Once someone has a use case, maintainers tend to minimize the importance of such use case and just say to c&p those classes, as if they have forgotten that flexibility used to be an indicator of a good code, unable to swallow their pride and admit they were wrong about making it final - it's easier to tell consumer to copy it and shift the burden of backporting all the future updates for that class to him. If they hear back that it's not possible to do so (eg. because that final class is type hinted, or referenced somewhere), then you tend to not hear back from the maintainer. Hence consumer is stuck on a fork for indefinite time. If we don't see the use case despite one existing and have classes `@final`, people can still carry on and just ask us to remove the annotation. But even if we don't, they can still carry on, on their own risk and not have to maintain their fork, which is usually maintained very sloppily. We probably have all been there at some point, where we needed to extend something that was final and in the end having to do ugly workarounds. And from the other side, maintainers? People can be very convinced on making things final, as long as they don't personally need to extend it. Once it stops being convenient because they need to extend it themselves (easiest and most common example is mocking), suddenly their need for final goes out the window and they happily remove that keyword. Similar case is with readonly classes BTW. So codebase seemingly follows some rules, eg. making VOs etc readonly, final, but then you start noticing exceptions that were historically created for whatever reasons. Except those reasons might no longer be there. Perhaps test that needed to mock the class no longer exists. Noticing the mistake of forgetting making class final is easy to notice in review, but forgetting re-adding final back to class when the reason it was removed is gone, not so much. End result is poorly followed rules with bunch of exceptions that might not be making sense anymore. I'm not sure what's the pros of that other than having fuzzy feeling of using native syntactic sugar instead docblock.

>  If we also assume that the final annotation is supposed to be handled like final itself, we would introduce small changes (only breaks when extended) in method signatures, since no one is supposed to extend such a class

I believe `since` keyword comes into play here. Because it's common we make something `@final` in minor versions, but we avoid doing BC breaking changes in minor versions. Once there is a new major version, `@final since x.x`  can be turned into `@final` and that assumption now stands that we are treating it as a final class.

Welp, this comment went way longer than expected. It's more like an expression of bitterness I have on this topic in general, not really specific to this PR. I should have probably better written a blog post or so on this.

greg0ire:
I think there is some truth in what you said. I've definitely seen discussions being open about removing final where you no longer hear about the maintainer because the user is in a hard to understand situation. The good thing with native `final` is that you can have this discussions where flaws of the api are exposed: users can show that an extension point is missing. That requires however an available maintainer to find another solution (can be decoration, can be contributing a new extension point). Sometimes, the discussion drags for months / does not even finish. I'll close this then. As I said, I originally assumed it was the intent, but obviously that's not the case.

PR #1992, opened 2025-09-07
Remove dependency on var-exporter

greg0ire:
It is unused since #1990


greg0ire:
Looks like it is still needed, because that's how `doctrine.orm.entity_manager.abstract` is made lazy, and there is a bug if we don't select a high enough version of `symfony/var-exporter`.

PR #1999, opened 2025-09-07
Test without symfony/doctrine-messenger

greg0ire:
We have code that checks the presence of that library, and a test that covers the case when it is missing.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -48,6 +48,8 @@ jobs:
           - ""
         remove-orm:
           - false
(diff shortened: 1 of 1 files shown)


greg0ire:
The coverage information says less code is covered :thinking:

greg0ire:
Should be fixed by #2000

greg0ire:
+0.07%, that's more like it.

greg0ire:
The newly covered lines are the expected ones: Screenshot 2025-09-07 at 19-53-49 Codecov

ostrolucky (approved):

PR #2005, opened 2025-09-07
Phpunit 10

greg0ire:
The latest major version of PHPUnit is 12 :grimacing: 
I think the issue was that we relied on `symfony/phpunit-bridge`. I've switched to PHPUnit's native deprecation detection system. Now it should be possible to upgrade to 11 and 12.


ostrolucky (approved):

tests/Command/ImportMappingDoctrineCommandTest.php
```diff
@@ -66,6 +66,7 @@ protected function tearDown(): void
         unset($this->kernel, $this->commandTester);
     }
 
+    #[WithoutErrorHandler]
```
stof, line 69: The proper way to write tests for legacy APIs is not to disable the error handler (which will also not report notices and warnings) but to add the `IgnoreDeprecations` attribute, which disables only the reporting of deprecations.

greg0ire: TIL :bulb: 

That's way more satisfying :)

PR #2024, opened 2025-09-09
Use stubs when possible

greg0ire:
It clarifies the intent. Also, it forces us to remove expressions that
do nothing such as `->expects($this->any())`


ostrolucky (approved):

tests/Command/CreateDatabaseDoctrineTest.php
```diff
@@ -62,37 +62,22 @@ public function testExecute(): void
     private function getMockContainer(string $connectionName, array|null $params = null): MockObject
     {
         // Mock the container and everything you'll need here
-        $mockDoctrine = $this->getMockBuilder(ManagerRegistry::class)
-            ->getMock();
+        $mockDoctrine = $this->createStub(ManagerRegistry::class);
```
stof, line 65: I would use `self::createStub` as it is a static method

PR #2025, opened 2025-09-09
Use IgnoreDeprecations instead of WithoutErrorHandler

greg0ire:
This is more explicit and also better because it only ignores deprecations.


greg0ire:
Looks like `WithoutErrorHandler` was hiding a lot of legitimate deprecations :grimacing: 
I will have to address those.

greg0ire:
I don't understand why the remaining PHP deprecation crop up now, in a file I didn't even touch :thinking: 

Addressed by 62d85fa65f7bbdaa385ec6562c7dcf1aa18cd2cc :shrug:

greg0ire:
Finally green :tada:

dmaicher (approved):

tests/MinimalConfig.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        $config ??= [];
+
+        return $config + [
+            'enable_lazy_ghost_objects' => true,
+            'controller_resolver' => ['auto_mapping' => false],
+            /** @phpstan-ignore function.alreadyNarrowedType */
+            'enable_native_lazy_objects' => PHP_VERSION_ID >= 80400 && method_exists(Configuration::class, 'enableNativeLazyObjects'),
+        ];
+    }
```
ostrolucky: Not sure why are there 2 of these methods if both of them append same configuration. Can't we just have completeOrmConfig and whatever is using minimal one will just do`['orm' => $this-completeOrmConfig()]`?

tests/DependencyInjection/DeprecationFreeExtension.php
```diff
@@ -0,0 +1,30 @@
...
+
+/** @phpstan-ignore class.extendsFinalByPhpDoc */
+final class DeprecationFreeExtension extends DoctrineExtension
+{
+    use MinimalConfig;
+
+    /** @param list<array<string, mixed>> $configs */
+    public function load(array $configs, ContainerBuilder $container): void
+    {
+        array_unshift($configs, $this->getMinimalOrmConfig());
```
ostrolucky: I don't understand need for this extension. Isn't it just removing the config and passing empty one to parent? Instead, why not construct DoctrineExtension with empty config already?

tests/MinimalConfig.php
greg0ire: You're right, and it's duplicated another time with `minimal_orm.php`. I will address this.

tests/DependencyInjection/DeprecationFreeExtension.php
greg0ire: No, that's not what `unshift` means, you're thinking of `shift`. I didn't use `array_push` because in this case, it's important to add `getMinimalOrmConfig()` to the beginning, so that it gets overriden by subsequent configurations. I'll add a comment to clarify that.

tests/DependencyInjection/DeprecationFreeExtension.php
```diff
@@ -0,0 +1,29 @@
...
+namespace Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection;
+
+use Doctrine\Bundle\DoctrineBundle\DependencyInjection\DoctrineExtension;
+use Doctrine\Bundle\DoctrineBundle\Tests\DeprecationFreeConfig;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+use function array_unshift;
+
+/** @phpstan-ignore class.extendsFinalByPhpDoc */
+final class DeprecationFreeExtension extends DoctrineExtension
```
stof, line 14: I don't think extending DoctrineExtension is the right implementation there, as it means that the tested code is not actually DoctrineExtension (and this will break in next major when making the class final instead of `@final` as phpstan won't be the only one to complain).

DoctrineExtensionTest should prepend the basic config allowing to get rid of deprecations before calling the unmodified DoctrineExtension that it wants to test IMO.

stof: Tests would look like that:

```diff
 $extension->load([
+    ['orm' => DeprecationFreeConfig::completeOrmConfig()],
     [
         // existing-loaded config
     ],
 ]);
```

Btw, the method name `completeOrmConfig` looks confusing to me. It is not about providing a complete config, which is how this name can be interpreted.

tests/DependencyInjection/AbstractDoctrineExtensionTestCase.php
```diff
@@ -62,11 +64,21 @@
 
 abstract class AbstractDoctrineExtensionTestCase extends TestCase
 {
-    abstract protected function loadFromFile(ContainerBuilder $container, string $file): void;
+    abstract protected function loadFromFile(
+        ContainerBuilder $container,
+        string $file,
+        bool $withMinimalOrmConfig = true,
```
stof: Why forcing all child classes to implement conditional loading of a minimal ORM config in a way that is _not_ specific to each format ? You could call `loadMinimalOrmConfig` in `AbstractDoctrineExtensionTestCase` before the call to `loadFromFile` in cases where you need it.

tests/DependencyInjection/AbstractDoctrineExtensionTestCase.php
```diff
@@ -62,11 +64,21 @@
...
+    ): void;
+
+    final protected function loadMinimalOrmConfig(ContainerBuilder $container): void
+    {
+        $loadPhp = new PhpFileLoader($container, new FileLocator(__DIR__ . '/Fixtures/config/php'));
+        $loadPhp->import('minimal_orm.php');
```
stof, line 75: No need to duplicate the minimal config. This could be implemented using the low-level API directly, while reusing the shared helper (maintaining it only once):
```php
$containerBuilder->loadFromExtension('doctrine', ['orm' => DeprecationFreeConfig::completeOrmConfig()]);
```

tests/DependencyInjection/AbstractDoctrineExtensionTestCase.php
```diff
@@ -1499,11 +1519,12 @@ private function loadContainer(
         string $fixture,
         array $bundles = ['XmlBundle'],
         CompilerPassInterface|null $compilerPass = null,
+        bool $withMinimalOrmConfig = true,
     ): ContainerBuilder {
         $container = $this->getContainer($bundles);
-        $container->registerExtension(new DoctrineExtension());
+        $container->registerExtension($withMinimalOrmConfig ? new DeprecationFreeExtension() : new DoctrineExtension());
```
stof: We should keep using DoctrineExtension that we want to test.

`$withMinimalOrmConfig` should control a call to `$this->loadMinimalOrmConfig($container)` instead.

tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -1482,6 +1500,7 @@ public function testControllerResolver(bool $simpleEntityManagerConfig): void
 
         $config['orm']['controller_resolver']     = ['auto_mapping' => true];
         $config['orm']['resolve_target_entities'] = ['Throwable' => 'stdClass'];
+        $config['orm']                            = DeprecationFreeConfig::completeOrmConfig($config['orm']);
```
stof: I'd rather keep the deprecation free config as a _separate_ config in the `$extension->load()` call below, making it clearer what is the config being tested at that point.

tests/DependencyInjection/DeprecationFreeExtension.php
greg0ire: > and this will break in next major when making the class final instead of `@final` as phpstan won't be the only one to complain

That's not the plan actually:

https://github.com/doctrine/DoctrineBundle/pull/1984

I agree with your other points and will rework this, however.

Regarding the naming, I would have liked to use `__invoke` since the class name seems more than enough, but that wasn't an option. Maybe just `getConfig()` then?

greg0ire: Or just `get()` since `Config` is mentioned in the class name.

tests/DependencyInjection/AbstractDoctrineExtensionTestCase.php
greg0ire: I think the reasoning was that there was more cases where it was needed than the opposite, but you're right, it's confusing. I'll do as you suggest.

tests/DependencyInjection/DeprecationFreeExtension.php
greg0ire: I ended up applying the pattern of using a separate config everywhere I could and then simplified this method and made it return a valid top-level configuration array. I think it's less confusing now.

tests/Builder/BundleConfigurationBuilder.php
```diff
@@ -69,9 +73,12 @@ public function addConnection(array $config): self
     }
 
     /** @param array<string, mixed> $config */
-    public function addEntityManager(array $config): self
+    public function addEntityManager(array $config, bool $withMinimalOrmConfig = true): self
     {
         $this->configuration['orm'] = $config;
+        if ($withMinimalOrmConfig) {
+            $this->configuration = array_merge_recursive(DeprecationFreeConfig::get(), $this->configuration);
```
greg0ire, line 80: This is the only place where something a bit special happens now.

SenseException (approved):

tests/DependencyInjection/DeprecationFreeExtension.php
greg0ire, line 1: 😧 I forgot to remove this class. Will send another PR.

PR #2027, opened 2025-09-10
Remove useless test

greg0ire:
On the 3.0.x branch I removed the part about auto_generate_proxy_classes (because it's deprecated), and inadvertenly removed some of the sense this test had (testing precedence of imports, regardless of whether the setting is deprecated or not).

In 01c714eef7d5c5515d664b1763af186ee7abd97b, another part of the meaning was lost (also because of a deprecation).

Now I should ideally find non-deprecated settings and test precedence again. However, I do not see how we are testing the extension itself here. To me, it looks like we are testing a Symfony feature. Let us see if there is any drop in code coverage.


greg0ire:
Coverage stays the same (which is not a guarantee that the test is useless, but at least does not prove it was useful).

ostrolucky (approved):

stof:
> To me, it looks like we are testing a Symfony feature.

The Symfony merging feature is actually powered by the Configuration class. Changing the configuration definition will impact the merged config. This is what this test is covering (which won't be reflected in code coverage btw)

greg0ire:
:thinking: but here, the important part seems to be the `import`. Whether an import overrides or is override is surely something that Symfony controls? Or is there something in the configuration class, that specifies how to behave in that case? Or was the test not really about this?

stof:
indeed.

And looking at the history of this test, I got the answer. This test was added in https://github.com/doctrine/DoctrineBundle/commit/0d5f544334a29e6dfabfd09223970f9d14ce83c8, before the creation of the `Configuration` system to merge configs.

greg0ire:
Ah thanks for looking that up, it makes a lot more sense now 👍

PR #2028, opened 2025-09-10
Remove redundant assertions

greg0ire:
Looks like the latest PHPStan just got a lot smarter about this.

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -22,7 +22,6 @@ use function array_intersect_key;
 use function array_key_exists;
 use function array_keys;
 use function array_pop;
-use function assert;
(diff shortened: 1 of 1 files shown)


derrabus (approved):

stof:
what became better is the phpdoc of `TreeBuilder` in Symfony, where we added conditional return types to give a more precise type to some methods (including `->getRootNode()`

stof:
And as the phpstan jobs installs dev versions of packages, it benefits from those upcoming improvements of Symfony 7.4 even though they are not released yet.

greg0ire:
Ah, thanks for the insight 👍 I must admit I was surprised that PHPStan had Symfony knowledge natively, but as you mentioned, that's not the case.

PR #2040, opened 2025-09-12
Phpunit 11

greg0ire:


tests/Command/Proxy/InfoDoctrineCommandTest.php
```diff
@@ -36,5 +37,7 @@ public function testExecute(): void
             'Found 3 mapped entities',
             $commandTester->getDisplay(),
         );
+
+        restore_exception_handler();
```
derrabus: Is this still necessary? I thought we've fixed that. 😕

greg0ire: It's sufficient to avoid the message in the commit message. If you have a better way, I'm interested.

ostrolucky: Let's go with this instead https://github.com/symfony/symfony/issues/53812#issuecomment-1962311843

stof: IIRC, the http-kernel component changes its behavior based on the availability of symfony/runtime (as it knows symfony/runtime manages the error handler already anyway for non-test usage).

greg0ire: @ostrolucky last time I tried that, it didn't work, but now it does. I must have messed up somehow.

ostrolucky (approved):

PR #2043, opened 2025-09-12
Upgrade to PHPUnit 12

greg0ire:
No new issue. I wasn't expecting that :exploding_head:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,7 @@
         "phpstan/phpstan": "2.1.1",
         "phpstan/phpstan-phpunit": "2.0.3",
         "phpstan/phpstan-strict-rules": "^2",
-        "phpunit/phpunit": "^10.5.53 || ^11.5.38",
+        "phpunit/phpunit": "^10.5.53 || ^12.3.10",
         "psr/log": "^1.1.4 || ^2.0 || ^3.0",
         "symfony/doctrine-messenger": "^6.4 || ^7.0",
         "symfony/expression-language": "^6.4 || ^7.0",


composer.json
```diff
@@ -52,7 +52,7 @@
         "phpstan/phpstan": "2.1.1",
         "phpstan/phpstan-phpunit": "2.0.3",
         "phpstan/phpstan-strict-rules": "^2",
-        "phpunit/phpunit": "^10.5.53 || ^11.5.38",
+        "phpunit/phpunit": "^10.5.53 || ^12.3.10",
```
ostrolucky, line 55: why skip 11.x? This means 10.x is forced on PHP 8.2

greg0ire: I do not see a clear benefit in doing that.
On 8.4 the benefit is that it allows to reduce differences with 3.0.x, by forcing us to address deprecations on 2.x

ostrolucky: I'm not sure I follow. You get those same benefits by allowing 11.x. Because CI will still run on PHP 8.4, where 12.x is going to be picked up.

greg0ire: I mean that if we had only one job running PHPUnit 12, that would be enough to get the benefit I see (forcing us to address deprecations). However if there is a bug that only exists when using PHPUnit 11.x, I'm not super interested in fixing it.

ostrolucky (approved):

composer.json
ostrolucky: meh, doesn't make much difference anyways so approving

PR #2047, opened 2025-09-12
Fix detection of errors

greg0ire:
I just noticed that PHPUnit was fine with me removing `#[IgnoreDeprecation]`. This was caused by an earlier commit, let's fix that.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         bootstrap="tests/bootstrap.php"
          colors="true"
          xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
          displayDetailsOnTestsThatTriggerDeprecations="true"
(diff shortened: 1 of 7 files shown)


tests/Command/ImportMappingDoctrineCommandTest.php
```diff
@@ -66,7 +65,6 @@ protected function tearDown(): void
         unset($this->kernel, $this->commandTester);
     }
 
-    #[IgnoreDeprecations]
```
greg0ire: I don't understand yet why those became unnecessary all of a sudden 🤔

greg0ire: Ah apparently, it has to do with https://github.com/doctrine/DoctrineBundle/blob/525b20333a06b8b4e8f5e84a85cb6addd0605ecc/tests/bootstrap.php#L7

greg0ire: @derrabus @ostrolucky that move was wrong it seems. I will revert to calling `restore_exception_handler()`, because it is masking a deprecation:

```
1) /home/greg/dev/DoctrineBundle/vendor/symfony/deprecation-contracts/function.php:25
Since symfony/var-exporter 7.3: The "Symfony\Component\VarExporter\LazyGhostTrait" trait is deprecated, use native lazy objects instead.

Triggered by:

* Doctrine\Bundle\DoctrineBundle\Tests\Repository\ServiceEntityRepositoryTest::testConstructInitializesWhenImplementingLazyObjectInterface
  /home/greg/dev/DoctrineBundle/tests/Repository/ServiceEntityRepositoryTest.php:43
```

ostrolucky (approved):

PR #2065, opened 2025-09-15
Display coverage for default branch

greg0ire:
Currently, the coverage badge displays a value of 76%, because of at least 2 reasons:

- It was directly referring to master.
- Even if we were referring to the default branch, the default branch was set to master.

master no longer exists, so I've set 2.16.x as the default branch in Codecov.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,8 @@
 
 Doctrine DBAL & ORM Bundle for the Symfony Framework.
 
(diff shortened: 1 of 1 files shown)


greg0ire:
The result can be seen by clicking on the "rich diff" icon.

ostrolucky (approved):

PR #2070, opened 2025-09-28
fix phpunit deprecations

dmaicher:
```
There were 2 PHPUnit test runner deprecations:

1) Test method Doctrine\Bundle\DoctrineBundle\Tests\RegistryTest::testResetLazyObject has attribute with version constraint string argument without explicit version comparison operator ("8.4")

2) Test method Doctrine\Bundle\DoctrineBundle\Tests\Repository\ServiceEntityRepositoryTest::testConstructInitializesWhenImplementingLazyObjectInterface has attribute with version constraint string argument without explicit version comparison operator ("8.4")

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -3,6 +3,7 @@
          colors="true"
          xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
          displayDetailsOnTestsThatTriggerDeprecations="true"
+         displayDetailsOnPhpunitDeprecations="true"
          failOnDeprecation="true"
 >
     <testsuites>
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: phpunit.xml.dist [hunk 1]
```diff
@@ -3,6 +3,7 @@
          colors="true"
          xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
          displayDetailsOnTestsThatTriggerDeprecations="true"
+         displayDetailsOnPhpunitDeprecations="true"
```
greg0ire, line 6: Is there maybe a `failOnPhpunitDeprecation` setting we could use here?

dmaicher: indeed there is one :+1:

greg0ire (approved):

PR #2073, opened 2025-10-02
Fix escaped placeholder and interpolate paramters with non sequential numeric keys

FlashBlack:
Starting with PHP 7.4 PDO began supporting parameter placeholder escaping - https://wiki.php.net/rfc/pdo_escape_placeholders
Made changes to support this adaptation.

Fix of interpolation parameters with non sequential numeric keys has also been improved.

Should I leave the PR in the 3.0.x branch or switch to the current one for compatibility?


ostrolucky:
Branch 2.17.x please, since it's not really breaking a BC, but it's a new feature.

FlashBlack:
Got it, done @ostrolucky

PR #2076, opened 2025-10-07
Inline parent class code

greg0ire:
In order to be make it easier to make changes in the metadata discovery mechanism, let us avoid having to change code in 2 different repositories and inline the code of the `AbstractDoctrineExtension` of the Symfony bridge here. This class will be deprecated and ultimately, removed.

One of the methods was protected and not used, and I did not copy it:
`loadObjectManagerCacheDriver()`.

No type hints have been added for the sake of backward compatibility, but there is still a breaking change: the extension no longer pass an `instanceof AbstractDoctrineExtension` assertion, and I think that's fine.

### DIFF
--- a/src/DependencyInjection/DoctrineExtension.php
+++ b/src/DependencyInjection/DoctrineExtension.php
@@ -47,8 +47,8 @@ use Doctrine\Persistence\Mapping\Driver\PHPDriver;
 use Doctrine\Persistence\Mapping\Driver\StaticPHPDriver;
 use InvalidArgumentException;
 use LogicException;
+use ReflectionClass;
 use Symfony\Bridge\Doctrine\Attribute\MapEntity;
-use Symfony\Bridge\Doctrine\DependencyInjection\AbstractDoctrineExtension;
 use Symfony\Bridge\Doctrine\IdGenerator\UlidGenerator;
 use Symfony\Bridge\Doctrine\IdGenerator\UuidGenerator;
 use Symfony\Bridge\Doctrine\Middleware\IdleConnection\Listener;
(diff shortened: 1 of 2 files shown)


greg0ire:
Note that I didn't copy any tests; the copied methods seem to be covered by existing tests already.

stof (approved):

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -98,8 +108,330 @@
...
+     */
+    protected function detectMetadataDriver(string $dir, ContainerBuilder $container): string|null
+    {
+        $configPath = $this->getMappingResourceConfigDirectory($dir);
+        $extension  = $this->getMappingResourceExtension();
+
+        if (glob($dir . '/' . $configPath . '/*.' . $extension . '.xml', GLOB_NOSORT)) {
+            $driver = 'xml';
+        } elseif (glob($dir . '/' . $configPath . '/*.' . $extension . '.yml', GLOB_NOSORT)) {
+            $driver = 'yml';
```
greg0ire, line 352: to be deprecated and removed

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -98,8 +108,330 @@
...
+        if ($objectManager['auto_mapping']) {
+            // automatically register bundle mappings
+            foreach (array_keys($container->getParameter('kernel.bundles')) as $bundle) {
+                if (isset($objectManager['mappings'][$bundle])) {
+                    continue;
+                }
+
+                $objectManager['mappings'][$bundle] = [
+                    'mapping' => true,
+                    'is_bundle' => true,
```
GromNaN, line 143: `is_bundle` should be deprecated in an other PR.

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -98,8 +108,330 @@
...
+            } else {
+                $mappingDriverDef = new Definition($this->getMetadataDriverClass($driverType), [
+                    array_values($driverPaths),
+                ]);
+            }
+
+            if (
+                str_contains($mappingDriverDef->getClass(), 'yml') || str_contains($mappingDriverDef->getClass(), 'xml')
+                || str_contains($mappingDriverDef->getClass(), 'Yaml') || str_contains($mappingDriverDef->getClass(), 'Xml')
+            ) {
```
GromNaN, line 304: The class is the actual class name, you can remove the check on the old parameter names (related to https://github.com/symfony/symfony/pull/53681)
```suggestion
            if (str_contains($mappingDriverDef->getClass(), 'Yaml') || str_contains($mappingDriverDef->getClass(), 'Xml')) {
```

GromNaN (approved):

PR #2077, opened 2025-10-07
Make DependencyInjection namespace internal

greg0ire:
Nobody is supposed to use types inside that namespace (see #1984), so let us make it internal, it should make changes easier to do.

I also marked classes inside the namespace as hard final, despite my previous PR making many types final being rejected, because I think it makes sense for types that are marked with `@internal`.

### DIFF
--- a/UPGRADE-3.0.md
+++ b/UPGRADE-3.0.md
@@ -155,6 +155,13 @@ You should use stop passing an event manager argument.
 + $connectionFactory->createConnection($params, $config, $mappingTypes)
 ```
 
+DependencyInjection namespace becomes internal
+----------------------------------------------
+
+The `Doctrine\Bundle\DoctrineBundle\DependencyInjection` namespace is now
+considered internal, and all classes inside it are marked as final.
+Don't reference any classes from this namespace directly.
+
 `Doctrine\Bundle\DoctrineBundle\EventSubscriber\EventListenerInterface` has been removed
 ----------------------------------------------------------------------------------------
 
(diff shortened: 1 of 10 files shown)


UPGRADE-3.0.md
```diff
@@ -155,6 +155,13 @@ You should use stop passing an event manager argument.
 + $connectionFactory->createConnection($params, $config, $mappingTypes)
 ```
 
+DependencyInjection namespace becomes internal
+----------------------------------------------
+
+The `Doctrine\Bundle\DoctrineBundle\DependencyInjection` namespace is now
+considered internal, and all classes inside it are marked as final with a PHP keyword.
```
GromNaN: This precision adds confusion.
```suggestion
considered internal, and all classes inside it are marked as final.
```

UPGRADE-3.0.md
```diff
@@ -155,6 +155,13 @@ You should use stop passing an event manager argument.
...
+Avoid referencing any classes from this namespace directly.
```
GromNaN: You're too kind.
```suggestion
Don't reference any classes from this namespace directly.
```

src/DependencyInjection/Compiler/CacheSchemaSubscriberPass.php
```diff
@@ -14,9 +14,9 @@
  *
  * Must be run later after ResolveChildDefinitionsPass.
  *
- * @final since 2.9
```
GromNaN, line 17: This annotation needs to be updated in previous version, from `@final` to `@internal`
```
@internal since 2.99
```

derrabus (approved):

PR #2080, opened 2025-10-08
Simplify detection code

greg0ire:


stof (approved):

SenseException (approved):

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -286,10 +286,7 @@ protected function registerMappingDrivers(array $objectManager, ContainerBuilder
                 ]);
             }
 
-            if (
-                str_contains($mappingDriverDef->getClass(), 'yml') || str_contains($mappingDriverDef->getClass(), 'xml')
-                || str_contains($mappingDriverDef->getClass(), 'Yaml') || str_contains($mappingDriverDef->getClass(), 'Xml')
-            ) {
+            if (str_contains($mappingDriverDef->getClass(), 'Xml')) {
```
GromNaN: I would change it to a hard comparison with `SimplifiedXmlDriver::class`.
And only if `$container->hasDefinition($mappingService)` was `false` and the service was not set outside of the DoctrineExtension.

GromNaN (approved):

PR #2083, opened 2025-10-09
return 'annotation' type support to Doctrine mapping configuration

### LINKED ISSUE #2082: annotation as a mapping type in config is no longer supported
### Bug Report

| Version                                     | 2.17.0
| Previous Version if the bug is a regression | -

#### Summary

After upgrading to `v2.17.0`, using annotation as a mapping type in the Doctrine configuration is no longer supported. The reason is [this commit](https://github.com/doctrine/DoctrineBundle/commit/8667b001582f5479029e5e530b21fe19a82e9863#diff-e2aeb4a501e79bc8ebab5ea93069b5bbcd3eb9168d37bd78ddd43075cddb8cc4) - code from `AbstractDoctrineExtension` was moved directly to `DoctrineExtension`, however, [this version](https://github.com/symfony/doctrine-bridge/blame/7.3/DependencyInjection/AbstractDoctrineExtension.php) was probably used. 

The problem is, the bundle should [support](https://github.com/doctrine/DoctrineBundle/blob/2.17.x/composer.json#L41) the older Symfony version where the annotation type is supported yet, see https://github.com/symfony/doctrine-bridge/blame/6.4/DependencyInjection/AbstractDoctrineExtension.php#L252

IMHO, this change was not intended, as it is a BC break that should rather be introduced in a major version, not a minor.

#### Current behavior
With the following configuration:
```
doctrine:
    orm:
        mappings:
            MyBundle:
                type: annotation
```

I got an error with the message:
```
Can only configure "xml", "yml", "php", "staticphp" or "attribute" through  the DoctrineBundle. Use your own bundle to configure other metadata drivers. You can register them by adding a new driver to the "doctrine.orm.default _metadata_driver" service definition. 
```

#### Expected behavior

using `annotation` as mapping type is allowed

vitek-rostislav:
fixes #2082

### DIFF
--- a/src/DependencyInjection/DoctrineExtension.php
+++ b/src/DependencyInjection/DoctrineExtension.php
@@ -333,8 +333,8 @@ class DoctrineExtension extends Extension
             throw new InvalidArgumentException(sprintf('Specified non-existing directory "%s" as Doctrine mapping source.', $mappingConfig['dir']));
         }
 
-        if (! in_array($mappingConfig['type'], ['xml', 'yml', 'php', 'staticphp', 'attribute'])) {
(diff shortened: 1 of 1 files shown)


greg0ire:
Thanks for the attempt, appreciate it!

PR #2089, opened 2025-10-09
Deprecate methods related to annotations and yaml

greg0ire:
These methods are removed in the next major version.


stof (approved):

stof:
> Technically, these classes are not internal yet,

it won't become internal either.

greg0ire:
Oh right!

PR #2107, opened 2025-10-13
Conditionally deprecate proxy-related-settings

greg0ire:
There is no indication that they should be removed, even after enabling native lazy objects.


src/DependencyInjection/Configuration.php
greg0ire, line 502: Note: because there is a default value, I don't think I will be able to tell from the extension whether this setting was explicitly set or not. For that reason, I did not implement this deprecation in the extension, where it would have been possible to add a condition so that the deprecation is only thrown when native lazy objects are enabled.

SenseException (approved):

PR #2109, opened 2025-10-13
Address PHPUnit notices

greg0ire:
They were all about using stubs rather than mocks when not adding any expectations.

### DIFF
--- a/tests/Command/CreateDatabaseDoctrineTest.php
+++ b/tests/Command/CreateDatabaseDoctrineTest.php
@@ -11,7 +11,7 @@ use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DriverManager;
 use Doctrine\DBAL\Schema\SchemaManagerFactory;
 use Doctrine\Persistence\ManagerRegistry;
-use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\MockObject\Stub;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\Console\Tester\CommandTester;
 use Symfony\Component\DependencyInjection\Container;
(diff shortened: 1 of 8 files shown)


greg0ire:
I don't have much experience with PHPUnit notices right now, so I didn't add `failOnPHPUnitNotices` yet because I'm not sure how easy to address they usually are.

SenseException (approved):

tests/Command/CreateDatabaseDoctrineTest.php
```diff
@@ -59,25 +59,23 @@ public function testExecute(): void
...
 
         $mockConnection = $this->createStub(Connection::class);
         $mockConnection->method('getConfiguration')->willReturn($config);
         $mockConnection->method('getParams')->willReturn($params);
         $mockDoctrine->method('getConnection')->willReturn($mockConnection);
 
-        $mockContainer = $this->getMockBuilder(Container::class)
-            ->onlyMethods(['get'])
-            ->getMock();
+        $mockContainer = $this->createStub(Container::class);
```
stof, line 78: This was a *partial* mock before. I'm not sure how much it mattered for the test, but your new code is not equivalent.

tests/Command/CreateDatabaseDoctrineTest.php
```diff
@@ -59,25 +59,23 @@ public function testExecute(): void
...
-    private function getMockContainer(string $connectionName, array|null $params = null): MockObject
+    private function getMockContainer(string $connectionName, array|null $params = null): Stub
     {
         // Mock the container and everything you'll need here
         $mockDoctrine = $this->createStub(ManagerRegistry::class);
 
         $mockDoctrine->method('getDefaultConnectionName')->willReturn($connectionName);
 
-        $config = (new Configuration())->setSchemaManagerFactory($this->createMock(SchemaManagerFactory::class));
+        $config = (new Configuration())->setSchemaManagerFactory($this->createStub(SchemaManagerFactory::class));
```
stof, line 71: should probably be `self::createStub` given that it is a static method (unlike `createMock`)

tests/Command/CreateDatabaseDoctrineTest.php
greg0ire: Indeed, but I don't see how it would matter either.

tests/Command/CreateDatabaseDoctrineTest.php
greg0ire: Yeah, probably… I might change that later.

PR #2069, opened 2025-09-25
Allow Symfony 8

ajgarlag:
This will allow to run the test suite against Symfony 8. Once it's merged, I'll open a new PR to revert this one that should be merged before releasing a new minor version. See https://github.com/doctrine/orm/pull/12110

Fix #1910

PR #2113, opened 2025-10-15
Revert "Allow Symfony 8"

ajgarlag:
This reverts commit https://github.com/doctrine/DoctrineBundle/commit/81cc5b3b1e7c403363953160d50b370a79f5f308 of https://github.com/doctrine/DoctrineBundle/pull/2069.

As long as Symfony 8 is not final yet, we should not indicate compatibility. This should be added to the 3.1.0 milestone, so we don't forget to merge it before we release a stable 3.1.0.


ostrolucky:
Hmm I believe it's too late now. Even if we merge and release this, composer will just pick up older version of the bundle when installing it with Symfony 8. Intention was to do this revert before a release, but it was forgotten. Not sure if we can do something about it now.

nicolas-grekas:
There's nothing wrong advertising Symfony 8 as supported. The release is close, in feature freeze: the scope is known.

stof:
> Even if we merge and release this, composer will just pick up older version of the bundle when installing it with Symfony 8. Intention was to do this revert before a release, but it was forgotten.

the merge commit of the previous PR is not part of any tag. It was merged *after* the last release.

ostrolucky:
Ah ok somehow I though it was merged into 3.x. We are all good then, we should ideally merge this pr before release 3.1 though

VincentLanglet:
> Ah ok somehow I though it was merged into 3.x. We are all good then, we should ideally merge this pr before release 3.1 though

Why ? Symfony already has multiple beta version and won't have any BC break.
If Symfony 8 works fine with this bundle as if, it's better to allow people installing beta version if they want to.

A 3.1.0 release would be even better, like it was done for the migration bundle 
https://github.com/doctrine/DoctrineMigrationsBundle/releases/tag/3.6.0

ostrolucky:
> Why ? Symfony already has multiple beta version and won't have any BC break.

I don't share your optimism. I remember some BC breaks in Symfony happened even in RC versions. That said, I wrote my comment when beta version didn't exist yet. We can consider keeping the support for Symfony 8 in next doctrine-bundle release.

greg0ire:
Symfony 8 is released, so this can be closed.

PR #2115, opened 2025-10-16
Fix detectMappingType if the entity contain a package attribute

### LINKED ISSUE #2111: `The annotation driver is only available in doctrine/orm v2` if the class is using the @package attribute
### Bug Report

| Version                                     | 2.17.1
| Previous Version if the bug is a regression | 2.17.0

#### Summary

Close to #2088

If an interface or a class which is not an entity contain any annotation with Entity before any attribute, the file is considered as using annotation and the error `The annotation driver is only available in doctrine/orm v2` is thrown

Removing the package annotation fix the issue, replacing the regex `@.*Entity` by `@[^a-z]*Entity` also help.

Note: We will remove the `@package` annotation on our side, so this is not an issue for our project.

#### Current behavior

The entity is detected as using annotation instead of attribute and an error is thrown.

#### Expected behavior

The package attribute is ignored.

#### How to reproduce

Add an interface starting like that : 

```
/**
 * Interface UserInterface.
 *
 * @package UserBundle\Entity
 */
interface UserInterface
{
    public function getId(): int;
    public function getUsername(): string;
    // ...
```

jtattevin:
Update the regex used to discover if the class is using annotations or attribute to fix the case 

Fix #2111


src/DependencyInjection/DoctrineExtension.php
```diff
@@ -418,8 +418,8 @@ private function detectMappingType(string $directory, ContainerBuilder $containe
             }
 
             if (
-                preg_match('/^(?: \*|\/\*\*) @.*' . $quotedMappingObjectName . '\b/m', $content)
-                || preg_match('/^(?: \*|\/\*\*) @.*Embeddable\b/m', $content)
+                preg_match('/^(?: \*|\/\*\*) @[a-zA-Z\\\\]*' . $quotedMappingObjectName . '\b/m', $content)
```
stof: `a-zA-Z` is not the range of characters allowed in PHP namespaces. Your current regex will reject some valid class names.

jtattevin: Yes, my bad and thank for the feedback

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -418,8 +418,8 @@ private function detectMappingType(string $directory, ContainerBuilder $containe
             }
 
             if (
-                preg_match('/^(?: \*|\/\*\*) @[a-zA-Z\\\\]*' . $quotedMappingObjectName . '\b/m', $content)
-                || preg_match('/^(?: \*|\/\*\*) @[a-zA-Z\\\\]*Embeddable\b/m', $content)
+                preg_match('/^(?: \*|\/\*\*) @\\\\?([a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*\\\\?)*' . $quotedMappingObjectName . '\b/m', $content)
+                || preg_match('/^(?: \*|\/\*\*) @\\\\?([a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*\\\\?)*Embeddable\b/m', $content)
```
greg0ire: This regex is becoming hard to read. Please use [the x modifier](https://www.php.net/manual/en/reference.pcre.pattern.modifiers.php#:~:text=x%20(-,PCRE_EXTENDED,-)) and add comments.

greg0ire (commented):
Please add tests.

src/DependencyInjection/DoctrineExtension.php
jtattevin: I'm not familiar with the x modifier, so I'll take a look and update the code.

Note: I'm away for a few day, so I'm happy to update the pr but it will only be tuesday.

jtattevin:
> Please add tests.

Currently I need to create a bundle for each positive test case (`@ORM\Entity`, `@\Doctrine\ORM\Mapping\Entity`, `@ORMEntity`), it may lead to a lot of bundle, is that ok for you ? (That said, i'm not sure if there is many more case except those three that need to be checked ?)
All the cases that should not return annotations can be together, but the cases that match should be isolated to be certain that the right line is matched.

src/DependencyInjection/DoctrineExtension.php
greg0ire: Don't worry, I'm not in a hurry :wink:

greg0ire:
> it may lead to a lot of bundle, is that ok for you ?

That seems overkill when what you truly want to do is test a regex against a string.
Maybe you should extract the regex logic inside a public static method marked as `@internal`, so that it's possible to test it without making it part of the public API of the bundle?

jtattevin:
> Maybe you should extract the regex logic inside a public static method

Ok for me, was thinking about that but afraid to change too much the structure :)

greg0ire:
The regex-based detection logic is removed in 3.0.x so don't worry too much about the structure, it will eventually be OK.

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -429,6 +429,21 @@ private function detectMappingType(string $directory, ContainerBuilder $containe
...
+    /**
+     * Check if the file content contains a class-like annotation
+     *
+     * @internal
+     */
+    public static function textContainsAnnotation(string $quotedMappingObjectName, string $content): bool
+    {
+        return preg_match('/^(?:[ ]\*|\/\*\*)[ ]@                # Match phpdoc start or line with an at
+            \\\\?                                               # Can start with antislash
+            ([a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*\\\\?)*    # Match 0-n namespace, the antislash is optionnal as it may be a prefix if an alias is used
```
greg0ire:
```suggestion
            ([a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*\\\\?)*    # Match 0-n namespace, the antislash is optional as it may be a prefix if an alias is used
```

greg0ire: Also, what do you mean by "0-n namespace"? It seems you're saying that in `Doctrine\DoctrineBundle`, `DoctruneBundle` is a namespace, which it isn't, only `Doctrine` and `Doctrine\DoctrineBundle` are.

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -429,6 +429,21 @@ private function detectMappingType(string $directory, ContainerBuilder $containe
...
 
+    /**
+     * Check if the file content contains a class-like annotation
+     *
+     * @internal
+     */
+    public static function textContainsAnnotation(string $quotedMappingObjectName, string $content): bool
+    {
+        return preg_match('/^(?:[ ]\*|\/\*\*)[ ]@                # Match phpdoc start or line with an at
+            \\\\?                                               # Can start with antislash
```
greg0ire: The regex comments are misaligned

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -429,6 +429,21 @@ private function detectMappingType(string $directory, ContainerBuilder $containe
...
+            ([a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*\\\\?)*    # Match 0-n namespace, the antislash is optionnal as it may be a prefix if an alias is used
+            ' . $quotedMappingObjectName . '                    # The target class
+            [a-zA-Z0-9_\x80-\xff]*                              # Match a suffix if the class is aliased
```
stof: this does not make sense. Aliasing a class does not restrict it to apply a suffix anyway, and it would match unrelated things. This part should be reverted.

src/DependencyInjection/DoctrineExtension.php
stof: -1 for the part about the prefix thing. This leads to matching *anyway* annotation that has `Entity` somewhere in its name).

We should keep checking for the `Entity` word

jtattevin: Yes, that it, to say that it can be of namespace component like `Doctrine` or `DoctrineBundle`.
I can replace by `Match namespace, the antislash is optional as it may be a prefix if an alias is used` if you prefer ?

src/DependencyInjection/DoctrineExtension.php
jtattevin: The idea was to match also if you have : 
```
use Doctrine\ORM\Mapping\Entity as ORMEntity;
use Doctrine\ORM\Mapping\Entity as EntityORM;

/** @EntityORM */
/** @ORMEntity */
class Something { ... }
```

But yes, if you alias to `DatabaseObject` it would not match like the previous version.
Do you still want me to revert this part ?

stof: if we want to matching these, we would have to match it only when there is no namespace part (we cannot have a namespace reference before an aliased name). And making it reliable would require inspecting use statements to detect such aliases, which is overkill IMO (if the detection fails, the dev can always provide a type explicitly)

src/DependencyInjection/DoctrineExtension.php
greg0ire: Or "Match namespace components", just like you wrote above. That's fine.

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -438,6 +429,21 @@ private function detectMappingType(string $directory, ContainerBuilder $containe
...
 
+    /**
+     * Check if the file content contains a class-like annotation
+     *
+     * @internal
+     */
+    public static function textContainsAnnotation(string $quotedMappingObjectName, string $content): bool
+    {
+        return preg_match('/^(?:[ ]\*|\/\*\*)[ ]@               # Match phpdoc start or line with an at
+            \\\\?                                               # Can start with antislash
```
greg0ire: In English it's called a backslash

greg0ire:
I've fixed the build in #2125 , you may want to rebase.

jtattevin:
Thank for the fixes, the rebase is done

greg0ire (approved):

ostrolucky (approved):

greg0ire:
Thanks @jtattevin !

PR #2120, opened 2025-10-22
Fix attribute driver mapping pass compatibility with ORM 3

### LINKED ISSUE #1844: Exception when using DoctrineOrmMappingsPass::createAttributeMappingDriver together with doctrine/orm > v3
### Bug Report

Not sure if i should put it here or into doctrine/orm as it seems like compatibility problem.

|    Q                                              |   A
|-------------------------------| ------
| doctrine\doctrine-bundle         | 2.13.1
| doctrine\orm                              | 3.3.0

#### Summary

We have a private bundle, that we're reusing on multiple projects. This bundle provides some entities with mapping. Within this bundle we're calling DoctrineOrmMappingsPass to register bundle's entities mapping:  

```php 
$container->addCompilerPass(DoctrineOrmMappingsPass::createAttributeMappingDriver($namespaces, $directories));
```

#### Current behavior

Currently app crashes hard with  \InvalidArgumentException 

`The $reportFieldsWhereDeclared argument is no longer supported, make sure to omit it when calling Doctrine\ORM\Mapping\Driver\AttributeDriver::__construct.`

#### Expected behavior

It probably should not crash)
As soon as its going to be omitted by doctrine\orm we would like to avoid pass reportFieldsWhereDeclared from bundle.

#### How to reproduce

Its quite simple: 
1)  doctrine/orm version > 3
2) latest doctrine/doctrine-bundle
3) call DoctrineOrmMappingsPass to add mapping by attributes - for example like following:

```php 
namespace My\CustomBundle;

use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\DoctrineOrmMappingsPass;
use Symfony\Component\HttpKernel\Bundle\Bundle;

class MyCustomBundle extends Bundle
{
    public function build(ContainerBuilder $container): void
    {
        $namespaces = [
            'My\CustomBundle\Model',
        ];

        $directories = [
            __DIR__ . '/src/Model',
        ];
        
        $container->addCompilerPass(DoctrineOrmMappingsPass::createAttributeMappingDriver($namespaces, $directories));
     }
}
```

bobvandevijver:
I'm not completely happy with my method of detecting ORM v2 (it uses the presence of a removed method in v3), but it works. Fixes #1844.

### DIFF
--- a/src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
+++ b/src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
@@ -180,6 +180,8 @@ class DoctrineOrmMappingsPass extends RegisterMappingsPass
     public static function createAttributeMappingDriver(array $namespaces, array $directories, array $managerParameters = [], $enabledParameter = false, array $aliasMap = [], bool $reportFieldsWhereDeclared = false)
     {
         $driverArgs = [$directories];
+
+        // Add additional args for ORM <3.0
         if (method_exists(AttributeDriver::class, 'getReader')) {
             $driverArgs[] = $reportFieldsWhereDeclared;
         }
--- a/src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
+++ b/src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
@@ -16,6 +16,8 @@ use Symfony\Bridge\Doctrine\DependencyInjection\CompilerPass\RegisterMappingsPas
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Reference;
 
+use function method_exists;
+
 /**
  * Class for Symfony bundles to configure mappings for model classes not in the
  * auto-mapped folder.
@@ -177,7 +179,12 @@ class DoctrineOrmMappingsPass extends RegisterMappingsPass
      */
     public static function createAttributeMappingDriver(array $namespaces, array $directories, array $managerParameters = [], $enabledParameter = false, array $aliasMap = [], bool $reportFieldsWhereDeclared = false)
     {
-        $driver = new Definition(AttributeDriver::class, [$directories, $reportFieldsWhereDeclared]);
+        $driverArgs = [$directories];
+        if (method_exists(AttributeDriver::class, 'getReader')) {
+            $driverArgs[] = $reportFieldsWhereDeclared;
+        }
+
+        $driver = new Definition(AttributeDriver::class, $driverArgs);
 
         return new DoctrineOrmMappingsPass($driver, $namespaces, $managerParameters, $enabledParameter, $aliasMap);
     }
--- a/tests/DependencyInjection/Compiler/DoctrineOrmMappingsPassTest.php
+++ b/tests/DependencyInjection/Compiler/DoctrineOrmMappingsPassTest.php
@@ -7,12 +7,14 @@ namespace Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection\Compiler;
 use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\DoctrineOrmMappingsPass;
 use Doctrine\Bundle\DoctrineBundle\Tests\TestCase;
 use Doctrine\Deprecations\PHPUnit\VerifyDeprecations;
+use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\Driver\AttributeDriver;
 use Doctrine\Persistence\Mapping\Driver\MappingDriverChain;
 use PHPUnit\Framework\Attributes\IgnoreDeprecations;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 
 use function assert;
+use function interface_exists;
 use function realpath;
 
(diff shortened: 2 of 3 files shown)


src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
```diff
@@ -177,7 +179,12 @@ public static function createAnnotationMappingDriver(array $namespaces, array $d
      */
     public static function createAttributeMappingDriver(array $namespaces, array $directories, array $managerParameters = [], $enabledParameter = false, array $aliasMap = [], bool $reportFieldsWhereDeclared = false)
     {
-        $driver = new Definition(AttributeDriver::class, [$directories, $reportFieldsWhereDeclared]);
+        $driverArgs = [$directories];
+        if (method_exists(AttributeDriver::class, 'getReader')) {
+            $driverArgs[] = $reportFieldsWhereDeclared;
+        }
+
+        $driver = new Definition(AttributeDriver::class, $driverArgs);
```
greg0ire, line 189: It would be great to cover this method with a test that creates a container, adds the compiler pass returned by this method, compiles it and then attempts to retrieve the driver service from it.

greg0ire:
@bobvandevijver do you still plan to work on this?

bobvandevijver:
@greg0ire Thanks for the ping, this completely slipped my mind!

I've added the requested test, I hope it aligns with what you meant!

tests/DependencyInjection/Compiler/DoctrineOrmMappingsPassTest.php
```diff
@@ -31,4 +37,24 @@ public function testCreateAnnotationMappingDriverIsDeprecated(): void
             ['/path/to/entities'],
         );
     }
+
+    public function testAttributeDriverIsRegistered(): void
+    {
+        $driverNamespace = 'DoctrineBundle\Entity';
+        $container       = $this->createXmlBundleTestContainer(
```
bobvandevijver, line 50: I re-used this method as the compiler pass requires almost everything to be configured, such as the default entity manager.

bobvandevijver:
@greg0ire Quick ping to get this PR from my todo list 😃

greg0ire (approved):

tests/TestCase.php
```diff
@@ -22,7 +22,7 @@
 
 class TestCase extends BaseTestCase
 {
-    public function createXmlBundleTestContainer(): ContainerBuilder
+    public function createXmlBundleTestContainer(callable|null $func = null): ContainerBuilder
```
GromNaN, line 25: Where is this method called? Why do you need this parameter?

bobvandevijver: It is used in the test case to be able to configure the compiler pass, see also my [comment on this](https://github.com/doctrine/DoctrineBundle/pull/2120#discussion_r2639328396).

GromNaN (approved):

src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
```diff
@@ -177,7 +179,12 @@ public static function createAnnotationMappingDriver(array $namespaces, array $d
      */
     public static function createAttributeMappingDriver(array $namespaces, array $directories, array $managerParameters = [], $enabledParameter = false, array $aliasMap = [], bool $reportFieldsWhereDeclared = false)
     {
-        $driver = new Definition(AttributeDriver::class, [$directories, $reportFieldsWhereDeclared]);
+        $driverArgs = [$directories];
+        if (method_exists(AttributeDriver::class, 'getReader')) {
```
ostrolucky, line 185: please add a comment to see for which ORM version is it adding compatibility for

SenseException (approved):

src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
bobvandevijver: Done 👍🏻

ostrolucky (approved):

tests/DependencyInjection/Compiler/DoctrineOrmMappingsPassTest.php
```diff
@@ -31,4 +39,28 @@ public function testCreateAnnotationMappingDriverIsDeprecated(): void
...
+        if (! interface_exists(EntityManagerInterface::class)) {
+            self::markTestSkipped('This test requires ORM');
+        }
+
+        $driverNamespace = 'DoctrineBundle\Entity';
+        $container       = $this->createXmlBundleTestContainer(
+            static function (ContainerBuilder $containerBuilder) use ($driverNamespace): void {
+                $containerBuilder->addCompilerPass(DoctrineOrmMappingsPass::createAttributeMappingDriver(
+                    [$driverNamespace],
+                    [realpath(__DIR__ . '/Entity')],
```
greg0ire, line 54: @bobvandevijver when I execute this test, this function returns `false`… is that expected?

bobvandevijver: No, I would not expect that, but it does not really matter in the context here. See https://github.com/doctrine/DoctrineBundle/pull/2201#discussion_r2797276678.

PR #2140, opened 2025-11-27
chore: test Symfony 8

alexisLefebvre:
Replace:

- #2139

PR #2142, opened 2025-11-28
chore: show parameters in name of CI jobs

alexisLefebvre:
Extracted from:

- GH-2140

The goal is to make the list more readable, before this PR it looked like this:

> image

It display from 3 to 5 parameters, the first one can be the Symfony version or the PHP version, which is not easy to read.

With this PR, it looks like this:

> image

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,5 +1,4 @@
-
-name: "Continuous Integration"
+name: "CI: PHPUnit"
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


greg0ire (commented):
I love the end result. Are there functions you could be using to make the code easier to maintain?

https://docs.github.com/en/actions/reference/workflows-and-actions/expressions

Maybe `format` and/or `join` could come in handy.

.github/workflows/continuous-integration.yml
```diff
@@ -27,7 +26,20 @@ on:
...
+            matrix.php-version || 'Ø',
+            matrix.symfony-require || 'Ø',
+            ${{ matrix.dependencies || 'Ø' }},
+            ${{ matrix.stability || 'Ø'
+          ),
+          matrix.remove-orm && 'remove ORM' || '',
+          matrix.remove-doctrine-messenger && 'remove Messenger' || ''
+        ],
+        ' - '
+      ) }}
```
greg0ire, line 35: Using join on an array of length 1 produces no effect

greg0ire (commented):
Looks good. Please squash your commits

greg0ire (approved):

derrabus (approved):
Love it. Can you please to this for Doctrine DBAL and ORM as well? 🥺

SenseException (approved):

greg0ire:
Thanks @alexislefebvre !

PR #2155, opened 2025-12-03
reset primary read replica connections when reseting registry

### LINKED ISSUE #2154: WorkerMode with Primary Replica Database will not switch back to replica connection
#### Summary

When using **FrankenPHP** in **Worker mode** with a Primary / Read Replica connection. If in a http request we do an write action in database, the connection will switch to primary connection and doctrine will switch back to replica on a new http request.

There is a closed issue in Franken PHP repo https://github.com/php/frankenphp/issues/1444 but IMO adding a ` $this->connection->ensureConnectedToReplica();` in the start or the end of each request is a workaround not a fix.

As a developer, I expect that, each new HTTP request will reset the connection and we will be on replica. Like when we are not in worker mode.

#### Current behavior

When cennection switch to primary, it stay connected to Primary on each new HTTP request

#### Expected behavior

Each new HTTP request will be connected to replica even if previous http request did a write action in database

#### How to reproduce

* Run a FrankenPHP server in Worker Mode with a Symfony Application.
* Configure a postgresql server with a read replica
* Add 2 route, one with a `SELECT` the other with a `INSERT` statement
* call the endpoint with the `INSERT`  statement
* call the endpoint with the `SELECT` statement
* look primary postgresql logs, you will see the select statement.

sad270:
fix: https://github.com/doctrine/DoctrineBundle/issues/2154

### DIFF
--- a/src/Registry.php
+++ b/src/Registry.php
@@ -36,6 +36,9 @@ class Registry extends ManagerRegistry implements ResetInterface
         foreach ($this->getManagerNames() as $managerName => $serviceId) {
             $this->resetOrClearManager($managerName, $serviceId);
         }
+        foreach ($this->getConnectionNames() as $connectionName => $serviceId) {
+            $this->resetPrimaryReadReplicaConnection($connectionName, $serviceId);
+        }
     }
 
     private function resetOrClearManager(string $managerName, string $serviceId): void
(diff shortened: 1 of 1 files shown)


stof:
Closing the connection is better than forcing to connect to the replica. If the next request starts with something that needs to use the primary, it will avoid connecting for nothing (same if the next request does not use the DB at all btw).

stof:
As discussed in the issue, this requires being careful with the case of messenger consumers using the doctrine transport. The transport will not like a connection getting closed between message as the transport has a longer lifecycle.

sad270:
There is a `--no-reset` option on symfony messenger, if we don't want to reset, but the option will not reset all the symfony resetable services. 🤔

Maybe in the configuration we should add a "close_on_reset" configuration (or something like). If it is `true`, we close the connection on kernel reset. And for the backward compatibility we set a default value to `false`

derrabus:
Sorry, I closed the PR by accident. If it's still relevant, please open a new PR.

PR #2160, opened 2025-12-04
Remove VarExporter from dependencies

derrabus:
In 3.x, native lazy objects are the only proxies that we support. It should not matter anymore if VarExporter is installed and in which version.


stof (approved):

nicolas-grekas (approved):
🚀

ostrolucky (approved):

PR #2163, opened 2025-12-12
feat: add explicit null check in XMLSchemaTest (PHPStan Level 8)

shakaran:
Improves the robustness of the `testValidateSchema` method in `XMLSchemaTest.php` by adding null checks before importing XML nodes. This prevents potential errors when the expected XML elements are not present.

Test robustness improvements:

* Added checks to ensure that `dbal` and `orm` elements are not null before attempting to import them with `importNode`, preventing possible exceptions if these elements are missing. 

```
 vendor/bin/phpstan analyze --level=8 tests/DependencyInjection/XMLSchemaTest.php
 ------ ------------------------------------------------------------------------------------------------ 
  Line   tests/DependencyInjection/XMLSchemaTest.php                                                     
 ------ ------------------------------------------------------------------------------------------------ 
  :45    Parameter #1 $node of method DOMDocument::importNode() expects DOMNode, DOMElement|null given.  
         🪪  argument.type                                                                               
  :58    Parameter #1 $node of method DOMDocument::importNode() expects DOMNode, DOMElement|null given.  
         🪪  argument.type                                                                               
 ------ ------------------------------------------------------------------------------------------------ 

```

### DIFF
--- a/tests/DependencyInjection/XMLSchemaTest.php
+++ b/tests/DependencyInjection/XMLSchemaTest.php
@@ -41,7 +41,7 @@ class XMLSchemaTest extends TestCase
 
         $dbalElements = $dom->getElementsByTagNameNS($xmlns, 'dbal');
         if ($dbalElements->length) {
-            $dbalDom    = new DOMDocument('1.0', 'UTF-8');
+            $dbalDom     = new DOMDocument('1.0', 'UTF-8');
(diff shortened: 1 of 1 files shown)


ostrolucky:
We are not going with PR by file approach. It's going to be a PR by raised phpstan level.

PR #2190, opened 2025-12-22
Fix position of `$enableXsdValidation` in `DoctrineOrmMappingsPass::createXmlMappingDriver`

### LINKED ISSUE #2189: [3.2] BC Break in DoctrineOrmMappingsPass::createXmlMappingDriver signature causing conflicts with ecosystem bundles
### Bug Report

| Version                                     | 3.2

#### Summary

I am upgrading my project to doctrine/doctrine-bundle ^3.0 and encountered a breaking change that causes compatibility issues with other Symfony bundles (specifically web-auth/webauthn-symfony-bundle).

In version 3.2, the $aliasMap argument was removed (https://github.com/doctrine/DoctrineBundle/commit/1b5a951989bacf01df4b8057ade4f9202d20aeca#diff-deac0dc9c56580dc9ccd479560e2fa41eb2c3b4705ba737858cf2b7862a349f2L64) from the createXmlMappingDriver method signature in Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\DoctrineOrmMappingsPass.

#### Current behavior

Existing bundles that support older Doctrine versions call this method with 5 arguments (passing [] for $aliasMap). This results in a fatal error in DoctrineBundle 3.2 because the 4th argument is now expected to be bool $enableXsdValidation, but receives an array, and the method receives too many arguments.

`TypeError: Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\DoctrineOrmMappingsPass::createXmlMappingDriver(): Argument #4 ($enableXsdValidation) must be of type bool, array given`

#### Expected behavior

Was this hard break intended without a compatibility layer? It makes upgrading difficult as we have to wait for the entire ecosystem to tag releases strictly compatible with DoctrineBundle 3.x. Would you consider restoring the argument (even if unused/deprecated) to maintain signature compatibility?

KDederichs:
Fix by allowing $enableXsdValidation to contain an array. If it does trigger a deprecation and grab the correct $enableXsdValidation parameter via func_get_arg

Fixes #2189

### DIFF
--- a/src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
+++ b/src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
@@ -70,7 +70,8 @@ final class DoctrineOrmMappingsPass extends RegisterMappingsPass
             Deprecation::trigger(
                 'doctrine/doctrine-bundle',
                 'https://github.com/doctrine/DoctrineBundle/pull/2190',
-                'Providing a $aliasMap argument to createXmlMappingDriver() is deprecated and has no effect.',
+                'Providing a $aliasMap argument to %s is deprecated and has no effect.',
+                __METHOD__,
             );
             $enableXsdValidation = false;
 
--- a/src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
+++ b/src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
@@ -13,10 +13,14 @@ use Doctrine\Persistence\Mapping\Driver\SymfonyFileLocator;
 use Symfony\Bridge\Doctrine\DependencyInjection\CompilerPass\RegisterMappingsPass;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Reference;
+use TypeError;
 
 use function func_get_arg;
 use function func_num_args;
+use function gettype;
 use function is_array;
+use function is_bool;
+use function sprintf;
 
 /**
  * Class for Symfony bundles to configure mappings for model classes not in the
@@ -76,7 +80,12 @@ final class DoctrineOrmMappingsPass extends RegisterMappingsPass
             $enableXsdValidation = false;
 
             if (func_num_args() === 5) {
-                $enableXsdValidation = func_get_arg(4);
+                $enableXsdValidationArg = func_get_arg(4);
+                if (! is_bool($enableXsdValidationArg)) {
+                    throw new TypeError(sprintf('$enableXsdValidation is expected to be boolean, %s provided', gettype($enableXsdValidationArg)));
+                }
+
+                $enableXsdValidation = $enableXsdValidationArg;
             }
         }
 
--- a/src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
+++ b/src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
@@ -62,7 +62,7 @@ final class DoctrineOrmMappingsPass extends RegisterMappingsPass
      *                                          optional.
      * @param bool         $enableXsdValidation
      *
-     * @phpstan-ignore     missingType.iterableValue
+     * @phpstan-ignore missingType.iterableValue
      */
     public static function createXmlMappingDriver(array $namespaces, array $managerParameters = [], string|false $enabledParameter = false, bool|array $enableXsdValidation = false): self
     {
--- a/src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
+++ b/src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
@@ -70,7 +70,7 @@ final class DoctrineOrmMappingsPass extends RegisterMappingsPass
             Deprecation::trigger(
                 'doctrine/doctrine-bundle',
                 'https://github.com/doctrine/DoctrineBundle/pull/2190',
-                'Providing a $aliasMap to createXmlMappingDriver is deprecated and has no effect.',
+                'Providing a $aliasMap argument to createXmlMappingDriver() is deprecated and has no effect.',
             );
             $enableXsdValidation = false;
 
--- a/src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
+++ b/src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
 
+use Doctrine\Deprecations\Deprecation;
 use Doctrine\ORM\Mapping\Driver\AttributeDriver;
 use Doctrine\ORM\Mapping\Driver\XmlDriver;
 use Doctrine\Persistence\Mapping\Driver\PHPDriver;
@@ -12,6 +13,10 @@ use Doctrine\Persistence\Mapping\Driver\SymfonyFileLocator;
 use Symfony\Bridge\Doctrine\DependencyInjection\CompilerPass\RegisterMappingsPass;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Reference;
+
+use function func_get_arg;
+use function func_num_args;
+use function is_array;
 
 /**
  * Class for Symfony bundles to configure mappings for model classes not in the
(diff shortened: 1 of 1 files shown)


src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
```diff
@@ -47,17 +52,27 @@ public function __construct(Definition|Reference $driver, array $namespaces, arr
...
-     *                                        optional.
+     * @param string[]      $namespaces          Hashmap of directory path to namespace.
+     * @param string[]      $managerParameters   List of parameters that could which object manager name
+     *                                           your bundle uses. This compiler pass will automatically
+     *                                           append the parameter name for the default entity manager
+     *                                           to this list.
+     * @param string|false  $enabledParameter    Service container parameter that must be present to
+     *                                           enable the mapping. Set to false to not do any check,
+     *                                           optional.
+     * @param string[]|bool $enableXsdValidation This should never be an array, but is is due to BC reasons.
```
GromNaN: Only the expected type should be documented, for static analysis.
```suggestion
     * @param bool $enableXsdValidation
```

src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
```diff
@@ -47,17 +52,27 @@ public function __construct(Definition|Reference $driver, array $namespaces, arr
...
      */
-    public static function createXmlMappingDriver(array $namespaces, array $managerParameters = [], string|false $enabledParameter = false, bool $enableXsdValidation = false): self
+    public static function createXmlMappingDriver(array $namespaces, array $managerParameters = [], string|false $enabledParameter = false, bool|array $enableXsdValidation = false): self
     {
+        if (is_array($enableXsdValidation)) {
+            trigger_deprecation('doctrine/doctrine-bundle', '3.2', 'Starting version 3.2 passing $aliasMap to createXmlMappingDriver is deprecated and has no effect.');
```
GromNaN: Doctrine has the `doctrine/deprecations` package. The function `trigger_deprecation` is provided by `symfony/deprecation-contracts`.


```suggestion
            \Doctrine\Deprecations\Deprecation::trigger(
                'doctrine/doctrine-bundle',
                'https://github.com/doctrine/DoctrineBundle/pull/2190',
                'Providing a $aliasMap to createXmlMappingDriver is deprecated and has no effect."
            );
```

GromNaN (commented):
Thanks for the PR.
A few changes to comply with doctrine standards, but it seems correct to me.

src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
KDederichs: PHPStan complains if I just use bool here, that's why I did it that way. (See the PHPStan error)
Not sure if you can just ignore that error somehow or if it should be PHPStan compliant with the message.

KDederichs: @GromNaN I added a phpstan ignore directive to suppress this now which seems to make the tests run fine. I hope that works with the coding standards because I don't see how that's solved otherwise

GromNaN (approved):

src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
```diff
@@ -47,17 +52,33 @@ public function __construct(Definition|Reference $driver, array $namespaces, arr
...
+        if (is_array($enableXsdValidation)) {
+            Deprecation::trigger(
+                'doctrine/doctrine-bundle',
+                'https://github.com/doctrine/DoctrineBundle/pull/2190',
+                'Providing a $aliasMap to createXmlMappingDriver is deprecated and has no effect.',
+            );
+            $enableXsdValidation = false;
+
+            if (func_num_args() === 5) {
+                $enableXsdValidation = func_get_arg(4);
```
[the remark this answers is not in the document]
greg0ire: This looks like sweeping the issue under the rug, to be frank.

src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
```diff
@@ -47,17 +52,33 @@ public function __construct(Definition|Reference $driver, array $namespaces, arr
...
+     * @phpstan-ignore     missingType.iterableValue
      */
-    public static function createXmlMappingDriver(array $namespaces, array $managerParameters = [], string|false $enabledParameter = false, bool $enableXsdValidation = false): self
+    public static function createXmlMappingDriver(array $namespaces, array $managerParameters = [], string|false $enabledParameter = false, bool|array $enableXsdValidation = false): self
     {
+        if (is_array($enableXsdValidation)) {
+            Deprecation::trigger(
+                'doctrine/doctrine-bundle',
+                'https://github.com/doctrine/DoctrineBundle/pull/2190',
+                'Providing a $aliasMap to createXmlMappingDriver is deprecated and has no effect.',
```
greg0ire:
```suggestion
                'Providing a $aliasMap argument to createXmlMappingDriver() is deprecated and has no effect.',
```

greg0ire: You could use `__METHOD__`

src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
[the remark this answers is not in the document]
KDederichs: You actually want to apply that one? IMO makes little sense if we're just trying to keep BC of existing bundles.
It should throw errors in peoples faces if they try to pass weird stuff there.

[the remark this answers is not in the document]
greg0ire: What in my message could possibly make you think I want to apply that one?

src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
```diff
@@ -47,17 +52,33 @@ public function __construct(Definition|Reference $driver, array $namespaces, arr
...
+     * @phpstan-ignore missingType.iterableValue
      */
-    public static function createXmlMappingDriver(array $namespaces, array $managerParameters = [], string|false $enabledParameter = false, bool $enableXsdValidation = false): self
+    public static function createXmlMappingDriver(array $namespaces, array $managerParameters = [], string|false $enabledParameter = false, bool|array $enableXsdValidation = false): self
     {
+        if (is_array($enableXsdValidation)) {
+            Deprecation::trigger(
+                'doctrine/doctrine-bundle',
+                'https://github.com/doctrine/DoctrineBundle/pull/2190',
+                'Providing a $aliasMap argument to createXmlMappingDriver() is deprecated and has no effect.',
```
greg0ire: Please use `__METHOD__`

KDederichs: Sorry not quite sure what you mean by that.
Is it like a placeholder for createXmlMappingDriver() ?

KDederichs: Ok got it after looking at the rest of the doctrine codebase

src/DependencyInjection/Compiler/DoctrineOrmMappingsPass.php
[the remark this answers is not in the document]
KDederichs: It was more of a question tbh.
But yeah just ignoring that then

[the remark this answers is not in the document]
GromNaN: @greg0ire I don't understand your message. Do you think there is a better way to resolve the issue?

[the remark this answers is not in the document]
greg0ire: What I am saying is that copilot is suggesting we silently ignore non booleans, and I think that's a terrible suggestion.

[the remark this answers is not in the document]
greg0ire: A better solution IMO would be to throw a type error when the type is not one of the expected types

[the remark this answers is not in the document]
KDederichs: @greg0ire I added a bool type check that throws an error if a non bool is found, could you please check if this/the error message is alright with you?

greg0ire (approved):

PR #2193, opened 2026-01-03
Wire the Console EntityValueResolver as a service

chalasr:
This PR enables the EntityValueResolver for Console commands that has been introduced in https://github.com/symfony/symfony/pull/62917. 

~~Needs https://github.com/symfony/symfony/pull/62917 to be merged first~~ ✅

Please let me know if I missed anything.

### DIFF
--- a/src/DependencyInjection/DoctrineExtension.php
+++ b/src/DependencyInjection/DoctrineExtension.php
@@ -44,6 +44,7 @@ use Doctrine\Persistence\Mapping\Driver\StaticPHPDriver;
 use InvalidArgumentException;
 use LogicException;
 use ReflectionClass;
+use Symfony\Bridge\Doctrine\ArgumentResolver\Console\EntityValueResolver;
 use Symfony\Bridge\Doctrine\Attribute\MapEntity;
 use Symfony\Bridge\Doctrine\IdGenerator\UlidGenerator;
 use Symfony\Bridge\Doctrine\IdGenerator\UuidGenerator;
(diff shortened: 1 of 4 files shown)


chalasr:
The Symfony PR has been merged so this PR is ready.

tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -1507,6 +1508,62 @@ public function testControllerResolver(bool $simpleEntityManagerConfig): void
         $this->assertEquals(new MapEntity(null, null, null, [], null, null, null, true, true), $container->get('controller_resolver_defaults'));
     }
 
+    #[IgnoreDeprecations]
```
stof: Why ignoring deprecations ? This test is meant to cover a new feature, not a deprecated feature. It has no reason to ignore deprecations.

W0rma:
@chalasr Should this PR target `3.2.x` instead of `2.18.x` since the `composer.json` of `2.18.x` does not allow installing  `symfony/doctrine-bridge:8.1`?

chalasr:
Retargeted to 3.3.x and rebased. Reviews welcome 🙏

tests/DependencyInjection/DoctrineExtensionTest.php
chalasr: removed

W0rma:
I tried the changes in this PR in one of my project and it worked like a charm :thumbsup:

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -817,7 +818,27 @@ private function ormLoad(array $config, ContainerBuilder $container): void
...
+        if (class_exists(EntityValueResolver::class)) {
+            $consoleValueResolverDefinition = $container->getDefinition('doctrine.orm.entity_value_resolver.console');
+            $consoleValueResolverDefinition->setArgument(2, (new Definition(MapEntity::class))->setArguments([
+                null,
+                null,
+                null,
+                null,
+                null,
+                null,
+                null,
```
greg0ire: This long sequence of `null` is not great…

Should we use the named argument syntax here? I've found other places where we do that but maybe it was not possible at the time, and maybe it's a habit we should break?

chalasr: From [Symfony's BC promise](https://symfony.com/doc/current/contributing/code/bc.html)

> [10] Parameter names are only covered by the compatibility promise for constructors of Attribute classes. Using PHP named arguments might break your code when upgrading to newer Symfony versions.

Worth the hassle? Happy to do the change if you think so.

xabbuh: We could use `setArgument()` (for each argument) with an index here.

chalasr: That would still require to set every previous argument as the definition is built on the fly, so same amount of `null` isn't it?

greg0ire: I'd say let's take the chance and if Symfony ends up changing argument names, let's revert to positional arguments

chalasr: 👍 done

greg0ire (approved):

GromNaN (approved):

PR #2197, opened 2026-01-30
Add `#[AsDbalType]` attribute

syl20b:
**Description**

This PR add a new `#[AsDbalType]` attribute to automatically register the related class as a DBAL type.

**Problem Solved**

Currently, to register a custom DBAL type, it must be manually declared in the Doctrine configuration file:

``` yaml
# config/packages/doctrine.yaml

doctrine:
    dbal:
        types:
            my_custom_type: App\Type\MyCustomType

```

This approach is verbose and requires the developer to manage configuration in addition to the type class itself.

**Feature Added**

The addition of the  `#[AsDbalType]`  attribute allows the application to automatically detect and register any class that uses this attribute.

-----

**Concrete Usage Example**

Here is how you can register a custom type directly via the attribute, without touching the YAML configuration:

**1. Creating the Custom Type**

``` php
<?php

namespace App\Type;

use Doctrine\DBAL\Types\Type;
use Doctrine\DBAL\Platforms\AbstractPlatform;
use Doctrine\Bundle\DoctrineBundle\Attribute\AsDoctrineType;

#[AsDbalType(name: 'my_custom_type')]
final class MyCustomType extends Type
{
    // ... implementation of methods
}

```

**2. Usage in an Entity**

You can then use this type directly in your entities:

``` php
<?php

namespace App\Entity;

class Product
{
    #[ORM\Column(type: 'my_custom_type', length: 255)]
    private ?string $name = null;
    
    // ...
}

```

**Advantages**

  * **Zero Configuration:** No need to modify `doctrine.yaml` for every new custom type.
  * **Discoverability:** The DBAL type and its name are declared in the same location as its implementation.

### DIFF
--- a/docs/en/dbal-type.rst
+++ b/docs/en/dbal-type.rst
@@ -1,0 +1,88 @@
+DBAL Types
+==========
+
+Custom DBAL types can be registered using the ``AsDbalType`` attribute. This
+attribute allows you to define a name for your custom type directly in the class
+definition. If the name is not provided, the class name will be used as the default.
+
+To register a custom DBAL type, create a class that extends
+``Doctrine\DBAL\Types\Type`` and add the ``#[AsDbalType]`` attribute to it:
+
+.. code-block:: php
+
+    namespace App\Doctrine\Type;
+
+    use Doctrine\Bundle\DoctrineBundle\Attribute\AsDbalType;
+    use Doctrine\DBAL\Platforms\AbstractPlatform;
+    use Doctrine\DBAL\Types\Type;
+
+    #[AsDbalType(name: 'money')]
+    class MoneyType extends Type
+    {
+        public function getSQLDeclaration(array $column, AbstractPlatform $platform): string
+        {
+            return $platform->getDecimalTypeDeclarationSQL($column);
+        }
+
+        public function convertToPHPValue(mixed $value, AbstractPlatform $platform): mixed
+        {
+            return $value;
+        }
+
+        public function convertToDatabaseValue(mixed $value, AbstractPlatform $platform): mixed
+        {
+            return $value;
+        }
+    }
+
+When using the ``AsDbalType`` attribute, the type will be automatically
+registered with Doctrine.
+
+Manual Registration
+-------------------
+
+Alternatively, you can register custom types in your configuration:
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        # config/packages/doctrine.yaml
+        doctrine:
+            dbal:
+                types:
+                    money: App\Doctrine\Type\MoneyType
+
+    .. code-block:: xml
+
+        <!-- config/packages/doctrine.xml -->
+        <container xmlns="http://symfony.com/schema/dic/services"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns:doctrine="http://symfony.com/schema/dic/doctrine"
+            xsi:schemaLocation="http://symfony.com/schema/dic/services
+                http://symfony.com/schema/dic/services/services-1.0.xsd
+                http://symfony.com/schema/dic/doctrine
+                http://symfony.com/schema/dic/doctrine/doctrine-1.0.xsd">
+
+            <doctrine:config>
+                <doctrine:dbal>
+                    <doctrine:type name="money">App\Doctrine\Type\MoneyType</doctrine:type>
+                </doctrine:dbal>
+            </doctrine:config>
+        </container>
+
+    .. code-block:: php
+
+        // config/packages/doctrine.php
+        use App\Doctrine\Type\MoneyType;
+        use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+
+        return static function (ContainerConfigurator $containerConfigurator): void {
+            $containerConfigurator->extension('doctrine', [
+                'dbal' => [
+                    'types' => [
+                        'money' => MoneyType::class,
+                    ],
+                ],
+            ]);
+        };
--- a/src/Attribute/AsDbalType.php
+++ b/src/Attribute/AsDbalType.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Bundle\DoctrineBundle\Attribute;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+final readonly class AsDbalType
+{
+    public function __construct(public string|null $name = null)
+    {
+    }
+}
--- a/src/DependencyInjection/Compiler/RegisterDbalTypePass.php
+++ b/src/DependencyInjection/Compiler/RegisterDbalTypePass.php
@@ -1,0 +1,102 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler;
+
+use Doctrine\Bundle\DoctrineBundle\Attribute\AsDbalType;
+use Doctrine\DBAL\Types\Type;
+use ReflectionClass;
+use Symfony\Component\DependencyInjection\ChildDefinition;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
+
+use function is_subclass_of;
+use function method_exists;
+use function sprintf;
+
+/** @internal */
+final class RegisterDbalTypePass implements CompilerPassInterface
+{
+    private const string TAG = 'doctrine.dbal.type';
+
+    /**
+     * @param ReflectionClass<T> $reflector
+     *
+     * @template T of ReflectionClass
+     */
+    public static function autoconfigureFromAttribute(ChildDefinition $definition, AsDbalType $type, ReflectionClass $reflector): void
+    {
+        $attributes = [
+            'type_name' => $type->name ?? $reflector->name,
+        ];
+
+        // Determine if the version of symfony/dependency-injection is >= 7.3
+        /** @phpstan-ignore function.alreadyNarrowedType */
+        if (method_exists($definition, 'addResourceTag')) {
+            $definition->addResourceTag(self::TAG, $attributes);
+        } else {
+            // Needed to keep compatibility with symfony/dependency-injection < 7.3
+            $definition->addTag(self::TAG, $attributes)
+                ->addTag('container.excluded', ['source' => sprintf('by tag "%s"', self::TAG)]);
+        }
+    }
+
+    public function process(ContainerBuilder $container): void
+    {
+        $types = $container->getParameter('doctrine.dbal.connection_factory.types');
+
+        foreach ($this->findTaggedResourceIds($container) as $id => $tags) {
+            foreach ($tags as $tag) {
+                $class = $container->getDefinition($id)->getClass();
+                if (! $class) {
+                    throw new InvalidArgumentException(sprintf('The definition of "%s" must define its class.', $id));
+                }
+
+                if (! is_subclass_of($class, Type::class)) {
+                    throw new InvalidArgumentException(sprintf('The "%s" class must extends "%s".', $class, Type::class));
+                }
+
+                $types[$tag['type_name'] ?? $id] = ['class' => $class];
+            }
+        }
+
+        $container->setParameter('doctrine.dbal.connection_factory.types', $types);
+    }
+
+    /** @return array<string, array<array{type_name?: string}>> */
+    private function findTaggedResourceIds(ContainerBuilder $container): array
+    {
+        // Determine if the version of symfony/dependency-injection is >= 7.3
+        /** @phpstan-ignore function.alreadyNarrowedType */
+        if (method_exists($container, 'findTaggedResourceIds')) {
+            return $container->findTaggedResourceIds(self::TAG);
+        }
+
+        // Needed to keep compatibility with symfony/dependency-injection < 7.3
+        $tags = [];
+        foreach ($container->getDefinitions() as $id => $definition) {
+            if (! $definition->hasTag(self::TAG)) {
+                continue;
+            }
+
+            if (! $definition->hasTag('container.excluded')) {
+                throw new InvalidArgumentException(sprintf('The resource "%s" tagged "%s" is missing the "container.excluded" tag.', $id, self::TAG));
+            }
+
+            $class = $container->getParameterBag()->resolveValue($definition->getClass());
+            if (! $class || $definition->isAbstract()) {
+                throw new InvalidArgumentException(sprintf('The resource "%s" tagged "%s" must have a class and not be abstract.', $id, self::TAG));
+            }
+
+            if ($definition->getClass() !== $class) {
+                $definition->setClass($class);
+            }
+
+            $tags[$id] = $definition->getTag(self::TAG);
+        }
+
+        return $tags;
+    }
+}
--- a/src/DependencyInjection/DoctrineExtension.php
+++ b/src/DependencyInjection/DoctrineExtension.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\Bundle\DoctrineBundle\DependencyInjection;
 
+use Doctrine\Bundle\DoctrineBundle\Attribute\AsDbalType;
 use Doctrine\Bundle\DoctrineBundle\Attribute\AsDoctrineListener;
 use Doctrine\Bundle\DoctrineBundle\Attribute\AsEntityListener;
 use Doctrine\Bundle\DoctrineBundle\Attribute\AsMiddleware;
@@ -11,6 +12,7 @@ use Doctrine\Bundle\DoctrineBundle\CacheWarmer\DoctrineMetadataCacheWarmer;
 use Doctrine\Bundle\DoctrineBundle\ConnectionFactory;
 use Doctrine\Bundle\DoctrineBundle\Dbal\RegexSchemaAssetFilter;
 use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\IdGeneratorPass;
+use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\RegisterDbalTypePass;
 use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\ServiceRepositoryCompilerPass;
 use Doctrine\Bundle\DoctrineBundle\Mapping\ContainerEntityListenerResolver;
 use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepositoryInterface;
@@ -549,6 +551,9 @@ final class DoctrineExtension extends Extension
 
             $this->loadDbalConnection($name, $connection, $container);
         }
+
+        /** @phpstan-ignore argument.type (Needed for the $reflector parameter) */
+        $container->registerAttributeForAutoconfiguration(AsDbalType::class, RegisterDbalTypePass::autoconfigureFromAttribute(...));
 
         $container->registerForAutoconfiguration(MiddlewareInterface::class)->addTag('doctrine.middleware');
 
--- a/tests/DependencyInjection/DoctrineExtensionTest.php
+++ b/tests/DependencyInjection/DoctrineExtensionTest.php
@@ -5,11 +5,14 @@ declare(strict_types=1);
 namespace Doctrine\Bundle\DoctrineBundle\Tests\DependencyInjection;
 
 use Closure;
+use Doctrine\Bundle\DoctrineBundle\Attribute\AsDbalType;
 use Doctrine\Bundle\DoctrineBundle\Attribute\AsDoctrineListener;
 use Doctrine\Bundle\DoctrineBundle\Attribute\AsEntityListener;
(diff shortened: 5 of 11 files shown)


src/DependencyInjection/Compiler/RegisterAsDoctrineTypeAttributePass.php
```diff
@@ -0,0 +1,23 @@
...
+
+/** @internal */
+final class RegisterAsDoctrineTypeAttributePass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
+    {
+        $types = $container->getParameter('doctrine.dbal.connection_factory.types');
+
+        foreach ($container->findTaggedServiceIds('doctrine.dbal.type') as $id => $tags) {
+            $types[$tags[0]['name']] = ['class' => $container->getDefinition($id)->getClass()];
```
Jean-Beru: Should we have to loop over every tags? Even if the attribute is not repeatable, the service can be tagged directly using `AutoconfigureTag`.

Jean-Beru: Maybe it could be interesting to check if the `name` property exists.

src/Attribute/AsDoctrineType.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Bundle\DoctrineBundle\Attribute;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+final readonly class AsDoctrineType
```
stof:
```suggestion
final readonly class AsDbalType
```

Other doctrine projects might have a concept of type as well.

src/DependencyInjection/Compiler/RegisterAsDoctrineTypeAttributePass.php
stof: We should indeed process all tags, and throw a proper exception when the name is not provided in the tag instead of producing a notice.

syl20b: :ok_hand: 

I would also check if the class extends `Doctrine\DBAL\Types\Type`

src/DependencyInjection/Compiler/RegisterAsDoctrineTypeAttributePass.php
```diff
@@ -0,0 +1,23 @@
...
+        }
+
+        $container->setParameter('doctrine.dbal.connection_factory.types', $types);
```
stof: I'm not sure this is the right approach currently. The proposed API would make people think they can use DI in their types by registering them as services, while the implementation won't use those services.

There is 2 possible ways there:
- implement proper support for using the services for the type instances, thanks to the DBAL 4.x feature allowing that
- register only the class as done here, but build that feature using resource tags (which helps a bit with the intent)

src/DependencyInjection/Compiler/RegisterAsDoctrineTypeAttributePass.php
syl20b: Done

syl20b: Done

src/Attribute/AsDoctrineType.php
syl20b: Done

src/DependencyInjection/Compiler/RegisterAsDoctrineTypeAttributePass.php
syl20b: Indeed, I missed it. 
I change the implementation to use resource tags.
Thanks

syl20b: @stof @Jean-Beru I changed the first implementation to use resource tags when available (e.g symfony/dependency-injection >= 7.3), and fallback to tags otherwise. 
Let me know if this is a good way.

tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -947,6 +949,35 @@ public static function cacheConfigurationProvider(): array
...
+
+        $config = BundleConfigurationBuilder::createBuilder()
+            ->addBaseConnection()
+            ->build();
+
+        $extension->load([$config], $container);
+
+        /** @phpstan-ignore function.alreadyNarrowedType */
+        $attributes = method_exists($container, 'getAttributeAutoconfigurators')
+            ? array_map(static fn (array $arr) => $arr[0], $container->getAttributeAutoconfigurators())
```
Jean-Beru, line 969:
```suggestion
            ? array_column($container->getAttributeAutoconfigurators(), 0)
```
?

syl20b: @Jean-Beru I used `array_map` to be consistent with the others similar tests. 

Plus, the return type of the `getAttributeAutoconfigurators()` method is an array with the attribute FQCN as a key and the closure as first entry of the sub array : 

```php
array:2 [
  "Doctrine\Bundle\DoctrineBundle\Attribute\AsDbalType" => array:1 [
    0 => Closure(ChildDefinition $definition, AsDbalType $type): void^ {#4186
      returnType: "void"
      class: "Doctrine\Bundle\DoctrineBundle\DependencyInjection\DoctrineExtension"
    }
  ]
  "Doctrine\Bundle\DoctrineBundle\Attribute\AsMiddleware" => array:1 [
    0 => Closure(ChildDefinition $definition, AsMiddleware $attribute): void^ {#7886
      returnType: "void"
      class: "Doctrine\Bundle\DoctrineBundle\DependencyInjection\DoctrineExtension"
    }
  ]
]
```

So calling `array_column` as is would not give us the expected behavior. 

WDYT?

Jean-Beru: You're right, my bad 🙂

Jean-Beru (approved):

syl20b:
@stof is it ok for you?

silasjoisten:
I think this PR belongs more into the symfony/doctrine-bridge and there is already a PR for this: https://github.com/symfony/symfony/pull/63774

syl20b:
@silasjoisten 

> I think this PR belongs more into the symfony/doctrine-bridge and there is already a PR for this: [symfony/symfony#63774](https://github.com/symfony/symfony/pull/63774)

AFAIK, the bridge aims to do a bridge between a Symfony component and Doctrine, but this  feature is not related to a Symfony component, so it should belongs to the bundle IMHO

silasjoisten:
I’m not fully convinced this should live in the bundle.

According to the Symfony docs, DoctrineBridge is meant to “provide integration for Doctrine with Symfony components”, while DoctrineBundle mainly handles configuration and wiring.

This change feels more like integration logic (how Symfony registers and discovers Doctrine types) rather than something that should be driven by bundle configuration.

We already keep things like Doctrine types and Symfony specific behavior in the Bridge, so moving the registration logic there would keep responsibilities more clearly separated and consistent with existing patterns.

Otherwise we risk spreading behavior across both layers, which makes it harder to reason about and reuse outside of a full DoctrineBundle setup.

syl20b:
@silasjoisten 

> According to the Symfony docs, DoctrineBridge is meant to “provide integration for Doctrine with Symfony components”, while DoctrineBundle mainly handles configuration and wiring.

I agree with that, but there is no component involved here

> This change feels more like integration logic (how Symfony registers and discovers Doctrine types) rather than something that should be driven by bundle configuration.
>  We already keep things like Doctrine types and Symfony specific behavior in the Bridge, so moving the registration logic there would keep responsibilities more clearly separated and consistent with existing patterns.

Types you mention are Symfony component related types (Uid, Clock,...)

syl20b:
@greg0ire What do you think about that ?

greg0ire:
I agree with you, and furthermore, there are other attributes already: https://github.com/doctrine/DoctrineBundle/tree/3.2.x/src/Attribute

greg0ire (commented):
Should this come with docs?

syl20b:
> Should this come with docs?

@greg0ire Documentation added ;)

docs/en/dbal-type.rst
```diff
@@ -0,0 +1,93 @@
...
+
+    #[AsDbalType(name: 'money')]
+    class MoneyType extends Type
+    {
+        public function getSQLDeclaration(array $column, AbstractPlatform $platform): string
+        {
+            return $platform->getDecimalTypeDeclarationSQL($column);
+        }
+
+        public function convertToPHPValue($value, AbstractPlatform $platform): mixed
```
greg0ire:
```suggestion
        public function convertToPHPValue(mixed $value, AbstractPlatform $platform): mixed
```

docs/en/dbal-type.rst
```diff
@@ -0,0 +1,93 @@
...
+        {
+            return $value;
+        }
+
+        public function convertToDatabaseValue($value, AbstractPlatform $platform): mixed
```
greg0ire:
```suggestion
        public function convertToDatabaseValue(mixed $value, AbstractPlatform $platform): mixed
```

greg0ire (approved):
To be squash-merged

syl20b:
@greg0ire commits squashed :+1:

GromNaN:
I'm working on this topic using a `TypeRegistry` per connection, so that we can leverage dependency injection.
- https://github.com/doctrine/dbal/pull/7342
- https://github.com/doctrine/orm/pull/12421

I want to give a more advanced review before we introduce a solution that will be updated.

In any case, I think it is useful to have an attribute like `#[AsType]` or `#[AsDatabaseType]` to specify the type name and the connection or entity manager to which it should be associated.

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -550,6 +552,23 @@ private function dbalLoad(array $config, ContainerBuilder $container): void
...
+
+            // Determine if the version of symfony/dependency-injection is >= 7.3
+            /** @phpstan-ignore function.alreadyNarrowedType */
+            if (method_exists($definition, 'addResourceTag')) {
+                $definition->addResourceTag($tag, $attributes);
+            } else {
+                // Needed to keep compatibility with symfony/dependency-injection < 7.3
+                $definition->addTag('doctrine.dbal.type', $attributes)
+                    ->addTag('container.excluded', ['source' => sprintf('by tag "%s"', $tag)]);
+            }
```
GromNaN: The type instance should be a service injected into the static `TypeRegistry`. Since https://github.com/doctrine/dbal/pull/6705, instance with dependency injection are supported.


```suggestion
            $definition->addTag($tag, $attributes);
```

src/Attribute/AsDbalType.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Bundle\DoctrineBundle\Attribute;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+final readonly class AsDbalType
```
GromNaN, line 10: `AsDbalType` is a cryptic name that only insiders will understand. I'd prefer a simpler name:
```suggestion
final readonly class AsDatabaseType
```

GromNaN (changes requested):

src/DependencyInjection/DoctrineExtension.php
GromNaN: I realize this isn't a simple request, because right now the code is modifying the configuration. But to inject a service, you need to call `Type::register` at boot time.

src/Attribute/AsDbalType.php
syl20b: @GromNaN Attribute renamed ;)

syl20b:
> I'm working on this topic using a `TypeRegistry` per connection, so that we can leverage dependency injection.
>
> * [Add `TypeRegistry` to `Configuration` and use it in all internal type resolution dbal#7342](https://github.com/doctrine/dbal/pull/7342)
> * [Replace static `Type::*` calls with instance-based `TypeRegistry` lookups orm#12421](https://github.com/doctrine/orm/pull/12421)
> 
> I want to give a more advanced review before we introduce a solution that will be updated.
> 
> In any case, I think it is useful to have an attribute like `#[AsType]` or `#[AsDatabaseType]` to specify the type name and the connection or entity manager to which it should be associated.

@GromNaN  I love the idea  of being able to configure types per connection and having them potentially registered as services. The migration path might be long and could be split into several small steps :

1. Add an `#[AsDatabaseType]` attribute to improve DX, based on the existing feature  (what this PR is supposed to do)
2. Modify DBAL to enable types per connection configuration (which seem to be implemented in the 2 PRs above)
3. Integrate this evolution into the bundle
4. Allow custom types to be defined/registered as services

WDYT ? 

I would be happy to help.

In the meantime, I reworked my first implementation to make the attribute name optional, and I renamed the attribute as proposed.

SenseException (approved):
I believe 3.2.0 isn't the correct branch anymore because this is a new feature.

src/Attribute/AsDbalType.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Bundle\DoctrineBundle\Attribute;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+final readonly class AsDatabaseType
```
GromNaN: Sorry to suggest another change, but I think the most appropriate name would be `AsColumnType`. This is because it’s used in the `ORM\Column` attribute. `AsDatabaseType` could be confused with mysql/postgres...

docs/en/dbal-type.rst
```diff
@@ -0,0 +1,93 @@
...
+
+        public function convertToDatabaseValue(mixed $value, AbstractPlatform $platform): mixed
+        {
+            return $value;
+        }
+
+        public function getName(): string
+        {
+            return 'money';
+        }
```
GromNaN: This method is not in the implementation, it must be removed from this doc.

src/DependencyInjection/Compiler/RegisterDbalTypePass.php
```diff
@@ -0,0 +1,82 @@
...
+{
+    public function process(ContainerBuilder $container): void
+    {
+        $types = $container->getParameter('doctrine.dbal.connection_factory.types');
+
+        foreach ($this->findTaggedResourceIds($container) as $id => $tags) {
+            foreach ($tags as $tag) {
+                if (! array_key_exists('name', $tag)) {
+                    throw new InvalidArgumentException(sprintf('The "name" attribute is mandatory for the "doctrine.dbal.type" tag on the "%s" type.', $id));
+                }
```
GromNaN: I would move the logic that get the service Id as type name in this compiler pass, so that it works for services declared in the config.
But that's not a blocker, it can be improved later.

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -550,6 +553,24 @@ private function dbalLoad(array $config, ContainerBuilder $container): void
             $this->loadDbalConnection($name, $connection, $container);
         }
 
+        $container->registerAttributeForAutoconfiguration(AsDatabaseType::class, static function (ChildDefinition $definition, AsDatabaseType $type, Reflector $reflector): void {
+            assert($reflector instanceof ReflectionClass);
```
GromNaN: Symfony uses the reflection on the callback to send only objects of the expected reflection type. You can be specific.
```suggestion
        $container->registerAttributeForAutoconfiguration(AsDatabaseType::class, static function (ChildDefinition $definition, AsDatabaseType $type, ReflectionClass $reflector): void {
```

docs/en/dbal-type.rst
syl20b: @GromNaN Done ;)

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -550,6 +553,24 @@ private function dbalLoad(array $config, ContainerBuilder $container): void
...
+            $tag        = 'doctrine.dbal.type';
```
GromNaN: I like to have the autoconfiguration callback collocated in the compiler pass class. The tag being a private constant of this class. 

Example: https://github.com/symfony/symfony/blob/3acb4ac5f495a51c87c009169e147a7d7e244b0f/src/Symfony/Bundle/TwigBundle/DependencyInjection/Compiler/AttributeExtensionPass.php
and https://github.com/symfony/symfony/blob/3acb4ac5f495a51c87c009169e147a7d7e244b0f/src/Symfony/Bundle/TwigBundle/DependencyInjection/TwigExtension.php#L202

src/Attribute/AsDbalType.php
syl20b: @GromNaN Ok, no worries. 
As we are dealing with DBAL Types, don't we have to stick with the vendor concept (e.g. AsDbalType)?
AsColumnType is more ORM oriented, don't you think?

GromNaN: `AsDbalType` is very cryptic, but we already have the `doctrine.dbal.types` config. You are right that `AsColumnType` is too specific to the ORM, even if I think is mostly used with the ORM. Let's stay with `AsDatabaseType`.

src/DependencyInjection/DoctrineExtension.php
syl20b: PHPStan is yelling with the `ReflectionClass` type in the signature. 
I can ignore this error, but I'm wondering if it's the best approach here. WDYT @GromNaN ?

GromNaN: Yes, you can add a `@phpstan-ignore` annotation.

syl20b: Done !

src/DependencyInjection/DoctrineExtension.php
syl20b: @GromNaN Much clearer that way ! Done ;)

src/DependencyInjection/Compiler/RegisterDbalTypePass.php
syl20b: @GromNaN You want me to not remove this exception but fill the name with the service id instead?

GromNaN: Yes, the service Id is a perfect default name (unique, it's the class name by convention, and it's already an identifier for this class).

syl20b: @GromNaN Done ;)

src/DependencyInjection/Compiler/RegisterDbalTypePass.php
```diff
@@ -0,0 +1,102 @@
...
+                $class = $container->getDefinition($id)->getClass();
+                if (! $class) {
+                    throw new InvalidArgumentException(sprintf('The definition of "%s" must define its class.', $id));
+                }
+
+                if (! is_subclass_of($class, Type::class)) {
+                    throw new InvalidArgumentException(sprintf('The "%s" class must extends "%s".', $class, Type::class));
+                }
+
+                $types[$tag['type_name'] ?? $class] = ['class' => $class];
```
GromNaN: Sorry, we misunderstood each other. I want the type name to be the definition Id from the service container.
Without auto-configuration, the explicit configuration looks like this:
```yaml
services:
    App\Doctrine\DBAL\Type\CustomType:
        tags: ["doctrine.dbal.type"]
```

```suggestion
                $types[$tag['type_name'] ?? $id] = ['class' => $class];
```

syl20b: My mistake , sorry.

I fixed it.

GromNaN (approved):
That sounds great to me. We'll have to tweak the mechanics for https://github.com/doctrine/DoctrineBundle/pull/2221, but without any breaking changes.

syl20b:
@GromNaN Is it good for you ? Can I squash my commits ?

syl20b:
@GromNaN Commits squashed. Many thanks ;)

src/Attribute/AsDbalType.php
GromNaN: Hey, just realized that `dbal` is actually the name of the package; so not as cryptic I was saying. `AsDbalType` is quite descriptive, and fine since it's an advanced feature that requires a good understanding of the ORM/DBAL layers.

syl20b: Ok, that make more sense to me too. Attribute renamed ;)

syl20b:
@greg0ire @SenseException is it still Ok for you?

GromNaN:
Thank you @syl20b

docs/en/dbal-type.rst
```diff
@@ -0,0 +1,88 @@
...
+
+        public function convertToDatabaseValue(mixed $value, AbstractPlatform $platform): mixed
+        {
+            return $value;
+        }
+    }
+
+When using the ``AsDbalType`` attribute, the type will be automatically
+registered with Doctrine.
```
garak, line 39: This is not true. The type also needs to be registered as a service; otherwise, the attribute won't work

uuf6429: I guess this is still relevant seeing as I haven't been able to make the attribute work? I'm wondering why else this PR would have been released if it doesn't work.. 🤔

syl20b: Cf https://github.com/doctrine/DoctrineBundle/issues/2272#issuecomment-5192929735

GromNaN: The key point is that you must declare them in the DI configuration with autoconfiguration enabled so that they are treated as "resource classes" and does not create a service instance.

PR #2198, opened 2026-02-10
Wire PostgreSqlNotifyOnIdleListener for doctrine-messenger

nicolas-grekas:
Sidekick of https://github.com/symfony/symfony/pull/47666
Adapted from https://github.com/d-ph/DoctrineBundle/commit/ae224f31f0dabe1ad77f6996364443ef946fcfb3 by @d-ph

PR #2201, opened 2026-02-11
Merge 2.18.x up into 3.2.x

greg0ire:


tests/DependencyInjection/Compiler/DoctrineOrmMappingsPassTest.php
```diff
@@ -144,4 +147,29 @@ public function testCreateStaticPhpMappingDriver(): void
...
+        if (! interface_exists(EntityManagerInterface::class)) {
+            self::markTestSkipped('This test requires ORM');
+        }
+
+        $driverNamespace = 'DoctrineBundle\Entity';
+        $container       = $this->createXmlBundleTestContainer(
+            static function (ContainerBuilder $containerBuilder) use ($driverNamespace): void {
+                $containerBuilder->addCompilerPass(DoctrineOrmMappingsPass::createAttributeMappingDriver(
+                    [$driverNamespace],
+                    /** @phpstan-ignore argument.type */
```
greg0ire: In the line below, `realpath` returns false. To be clarified by @bobvandevijver

tests/DependencyInjection/Compiler/DoctrineOrmMappingsPassTest.php
```diff
@@ -144,4 +147,29 @@ public function testCreateStaticPhpMappingDriver(): void
...
+                    [realpath(__DIR__ . '/Entity')],
+                ));
+            },
+        );
+
+        $metadataDriver = $container->get('doctrine.orm.default_metadata_driver');
+        /** @phpstan-ignore function.impossibleType, instanceof.alwaysFalse */
```
greg0ire: I don't understand why I had to add this. It's as if the service definition for this service is wrong 🤔 

At runtime, it works.

greg0ire:
Regarding the remaining failure, see https://github.com/doctrine/DoctrineBundle/pull/2198#issuecomment-3887441443

tests/DependencyInjection/Compiler/DoctrineOrmMappingsPassTest.php
greg0ire: I've understood the issue thanks to Claude and added an explanation.

tests/DependencyInjection/Compiler/DoctrineOrmMappingsPassTest.php
greg0ire: I ended up using `dummy/path`, since apparently it does not matter :shrug:

bobvandevijver: Yeah, the path doesn't matter, but the argument is required to supply. If I recall correctly I grabbed the `realpath` one from another test that uses the `createXmlBundleTestContainer` method, but using `dummy/path` should be fine. The test is about the driver being registered correctly, which is validated before the path is used anyways.

GromNaN (approved):

ostrolucky (approved):

PR #2203, opened 2026-02-27
Add proper target names to autowiring aliases

nicolas-grekas:
`#[Target('')]` should not need using fixed suffixes. This does it.

This also fixes a few services names that use FQCN ids, which is discouraged practice for bundles (it pollutes the list of autowireable services)

### DIFF
--- a/src/DependencyInjection/DoctrineExtension.php
+++ b/src/DependencyInjection/DoctrineExtension.php
@@ -632,8 +632,7 @@ final class DoctrineExtension extends Extension
             ]);
 
         $container
-            ->registerAliasForArgument($connectionId, Connection::class, sprintf('%s.connection', $name))
-            ->setPublic(false);
+            ->registerAliasForArgument($connectionId, Connection::class, sprintf('%s.connection', $name), $name);
 
         // Set class in case "wrapper_class" option was used to assist IDEs
         if (isset($options['wrapperClass'])) {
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
(failures unrelated)

ostrolucky:
Let's separate the changes needed to fix target names from changes to service IDs. Fix we can merge quickly, but changes in service IDs are potentially BC breaking and I would like input from others if they agree with those.

config/dbal.php
```diff
@@ -93,13 +92,13 @@
...
-        ->set(RunSqlCommand::class)
+        ->set('doctrine.run_sql_command', RunSqlCommand::class)
             ->args([
-                service(ManagerRegistryAwareConnectionProvider::class)->nullOnInvalid(),
+                service('doctrine.manager_registry_aware_connection_provider')->nullOnInvalid(),
             ])
             ->tag('console.command', ['command' => 'dbal:run-sql'])
 
-        ->set(ProfilerController::class)
+        ->set('doctrine.profiler_controller', ProfilerController::class)
```
stof: wouldn't this break the reference in the template at https://github.com/doctrine/DoctrineBundle/blob/10b9d817eb226fd568c7b3921a537f54815f3bb5/templates/Collector/db.html.twig#L91 if you don't update it ?

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -632,16 +632,15 @@ private function loadDbalConnection(string $name, array $connection, ContainerBu
...
 
         // Set class in case "wrapper_class" option was used to assist IDEs
         if (isset($options['wrapperClass'])) {
             $def->setClass($options['wrapperClass']);
         }
 
         $container->setDefinition(
-            ManagerRegistryAwareConnectionProvider::class,
+            'doctrine.manager_registry_aware_connection_provider',
             new Definition(ManagerRegistryAwareConnectionProvider::class, [$container->getDefinition('doctrine')]),
```
stof, line 644: This looks totally broken to me. It should *reference* the `doctrine` service, not copy its definition as an inline definition.

And it would be better to define this service fully in `dbal.php` as done for other services.

nicolas-grekas:
> Let's separate the changes

Agreed, split in #2207

ostrolucky (approved):

greg0ire (approved):

PR #2204, opened 2026-02-28
Wire the clock argument of PostgreSqlNotifyOnIdleListener

nicolas-grekas:
Added in https://github.com/symfony/symfony/pull/63402

### DIFF
--- a/config/messenger.php
+++ b/config/messenger.php
@@ -65,6 +65,7 @@ return static function (ContainerConfigurator $container): void {
         ->set('messenger.transport.doctrine.pg_notify_on_idle_listener', PostgreSqlNotifyOnIdleListener::class)
             ->args([
                 service('logger'),
+                service('clock')->ignoreOnInvalid(),
             ])
             ->tag('kernel.event_subscriber');
 };


config/messenger.php
```diff
@@ -65,6 +65,7 @@
         ->set('messenger.transport.doctrine.pg_notify_on_idle_listener', PostgreSqlNotifyOnIdleListener::class)
             ->args([
                 service('logger'),
+                service('clock')->nullOnInvalid(),
```
ostrolucky: I am not sure this is good enough, since that constructor argument is non-nullable

nicolas-grekas: good catch, fixed thanks

ostrolucky (approved):

greg0ire:
The failures look related

nicolas-grekas:
Failure is OK.
I went with `ignoreOnInvalid`, which will never happen because clock is a dep of messenger.
Yet because of the way FrameworkBundle interacts with the CI, the clock service is removed in this case. But that's only a CI side-effect (having fwb in require and messenger+clock in require-dev)

nicolas-grekas:
I made the clock argument nullable in https://github.com/symfony/symfony/pull/63566

greg0ire (approved):

greg0ire:
Thanks @nicolas-grekas !

PR #2209, opened 2026-03-18
Document the dbname_suffix configuration option

### LINKED ISSUE #1705: Missing explanation of 'dbname_suffix' config value meaning in documentation
Go to [configuration reference page ](https://www.doctrine-project.org/projects/doctrine-bundle/en/2.10/configuration.html) and search for `dbname_suffix` - you will see there is no comments about it anywhere.

The meaning is unclear (at least for me).
Is it adding suffix to database name? If so - what it designed for if we can just use `dbname`?

lacatoire:
Fixes #1705

### DIFF
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -946,6 +946,32 @@ can configure. The following block shows all possible configuration keys:
             </doctrine:config>
         </container>
 
+Database Name Suffix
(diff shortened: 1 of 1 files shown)


ostrolucky:
I think name of the configuration is self-explanatory, I don't want our docs bloated

Issue #1605, opened 2023-01-03
Documentation diverges between this repo and Symfony.com


mpdude:
The README points to https://symfony.com/doc/current/reference/configuration/doctrine.html for documentation, claiming that this information would be drawn from `Resources/docs`. This is not true.

`Resources/docs` ends up at https://www.doctrine-project.org/projects/doctrine-bundle.html, whereas Symfony documentation comes from https://github.com/symfony/symfony-docs/blob/master/reference/configuration/doctrine.rst (and a few more recommendations/mentions in other places of that repo).

In particular, the "configuration reference" seems to have diverged at some point, with some interesting details and tidbits being available only in either of the two.

How could we resolve that?

I see value in having the documentation here, close to the authoritative source code. 

On the other hand, there seems to be much more activity and awareness in https://github.com/symfony/symfony-docs, and symfony.com seems to be what search engines prefer and return for almost all Symfony + Doctrine related searches.

greg0ire:
Also, the link in the sidebar appears to be dead: https://symfony.com/bundles (links to https://symfony.com/bundles/DoctrineBundle/current/index.html)

greg0ire:
@mpdude I think https://symfony.com/bundles/DoctrineBundle/2.7.x/configuration.html should be enough,  https://symfony.com/doc/current/reference/configuration/doctrine.html should redirect to that, maybe (after merging what's interesting)?

ostrolucky:
Should this be fixed by https://github.com/doctrine/DoctrineBundle/pull/1606, or what else should be done here?

greg0ire:
https://github.com/symfony/symfony-docs/blob/6.2/reference/configuration/doctrine.rst good bits should be copy/pasted to https://github.com/doctrine/DoctrineBundle/blob/2.8.x/Resources/doc/configuration.rst by a kind soul, after what the former can be removed (or replaced with a link to the latter) I think.

PR #2210, opened 2026-03-18
Add server_version, autocommit and SSL documentation

lacatoire:
Port useful documentation sections from symfony/symfony-docs that were missing in DoctrineBundle's own configuration reference.

Relates to #1605


ostrolucky:
I don't think doctrine-bundle should become a place where we combine ORM and DBAL docs. I also don't think this is a place to explain how YAML parses numbers. We could make an exception for some very commonly needed things, but I don't think any of these qualify. For example, most people don't specify server_version, but configure connection via DATABASE_URL. And that one is sufficiently documented through generated .env via symfony flex. Any other opinions?

PR #2220, opened 2026-03-27
Fix region_lifetime not applied to explicitly declared SLC regions

### LINKED ISSUE #2216: `region_lifetime` is not applied to explicitly declared SLC regions
| Version                                     | 2.18.2
| Previous Version if the bug is a regression | unknown

#### Summary

When regions are explicitly declared under `second_level_cache.regions`, the `region_lifetime` default is not applied to them. Instead, `lifetime: 0` is passed to `DefaultRegion`, resulting in cache keys stored in Redis without TTL (`TTL = -1`).

#### Current behavior

When a region is explicitly declared without `lifetime`:
```yaml
second_level_cache:
    region_lifetime: 3600
    regions:
        my_region:
            cache_driver: { type: service, id: cache.app }
```

Keys stored in Redis have no TTL (`TTL = -1`), meaning they never expire.

#### Expected behavior

`region_lifetime` should be used as a fallback for explicitly declared regions that do not have their own `lifetime` set. Keys should be stored in Redis with `TTL = 3600`.

#### How to reproduce

1. Configure `second_level_cache` with `region_lifetime` and at least one explicit region without `lifetime`
2. Add `#[ORM\Cache(usage: 'NONSTRICT_READ_WRITE', region: 'my_region')]` to an entity
3. Perform a `find($id)` to trigger SLC
4. Check TTL in Redis:
```bash
redis-cli TTL "your_key"
# returns -1 instead of 3600
```

#### Root cause

In `DoctrineExtension.php` line 1446:
```php
$regionsDef->addMethodCall('setLifetime', [$name, $region['lifetime']]);
```

`$region['lifetime']` defaults to `0` for regions without explicit `lifetime`. This calls `RegionsConfiguration::setLifetime($name, 0)`, storing `0` in `$this->lifetimes[$regionName]`.

Then in `RegionsConfiguration::getLifetime()`:
```php
return $this->lifetimes[$regionName] ?? $this->defaultLifetime;
```

Since the key exists with value `0`, the `??` fallback to `$this->defaultLifetime` (which holds `region_lifetime`) never triggers. `DefaultRegion` receives `0` and skips `expiresAfter()`.

#### Workaround

Explicitly set `lifetime` for every declared region:
```yaml
second_level_cache:
    region_lifetime: 3600
    regions:
        my_region:
            cache_driver: { type: service, id: cache.app }
            lifetime: 3600  # must be set explicitly
```

victorshdev:
Fixes #2216

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -604,7 +604,7 @@ final class Configuration implements ConfigurationInterface
                                         ->scalarNode('lock_path')->defaultValue('%kernel.cache_dir%/doctrine/orm/slc/filelock')->end()
                                         ->scalarNode('lock_lifetime')->defaultValue(60)->end()
                                         ->scalarNode('type')->defaultValue('default')->end()
-                                        ->scalarNode('lifetime')->defaultValue(0)->end()
+                                        ->scalarNode('lifetime')->defaultNull()->end()
                                         ->scalarNode('service')->end()
                                         ->scalarNode('name')->end()
                                     ->end()
--- a/src/DependencyInjection/DoctrineExtension.php
+++ b/src/DependencyInjection/DoctrineExtension.php
@@ -1191,7 +1191,7 @@ final class DoctrineExtension extends Extension
 
                     $container
                         ->setDefinition($regionId, new Definition(DefaultRegion::class))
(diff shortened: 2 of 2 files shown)


ostrolucky:
target 3.2.x please

greg0ire:
Closing and reopening to trigger the CI

ostrolucky (approved):

PR #2221, opened 2026-04-03
Update `#[AsDbalType]` attribute to register DBAL types as Symfony services

GromNaN:
Update the `#[AsDbalType]` PHP attribute and a `RegisterDbalTypePass` compiler pass to register Doctrine DBAL types as Symfony services with a per-connection `TypeRegistry`, without touching the global static type registry.

The PR https://github.com/doctrine/DoctrineBundle/pull/2197 that added this attribute only supports injection of class names into the configuration. This PR requires DBAL and ORM changes to inject services.

## How it works

- A new `#[AsDbalType(name: 'money')]` attribute tags the service with `doctrine.dbal.type`
- `RegisterDbalTypePass` builds a `TypeRegistry` per connection, populated with:
  - Config-based types (`doctrine.dbal.types`) as inline definitions
  - Service-tagged types as service references, resolved lazily through a `ServiceLocator`
- The `TypeRegistry` is wired to both the DBAL `Configuration` and the ORM `Configuration` of each associated entity manager
- Types can be restricted to specific connections by repeating the attribute: `#[AsDbalType(name: 'money', connection: 'reporting')]`

Because types are registered as services and resolved lazily, they support full dependency injection: a type class can declare constructor arguments and receive autowired or explicitly configured dependencies.

## Example

```php
use Doctrine\Bundle\DoctrineBundle\Attribute\AsDbalType;
use Doctrine\DBAL\Platforms\AbstractPlatform;
use Doctrine\DBAL\Types\Type;

#[AsDbalType(name: self::class /* by default */)]
final class MoneyType extends Type
{
    public function getSQLDeclaration(array $column, AbstractPlatform $platform): string
    {
        return $platform->getDecimalTypeDeclarationSQL($column);
    }
}
```

```php
#[ORM\Column(type: MoneyType::class)]
private Money $price;
```

## Backward compatibility (DBAL < 4.5)

Per-connection `TypeRegistry` injection requires DBAL >= 4.5. On older versions, the compiler pass falls back to registering the types in the global type registry through `doctrine.dbal.connection_factory.types`. This fallback has two limitations, both enforced at compile time:

- Dependency injection is not available: DBAL instantiates types with `new $class()`, so a type whose constructor has mandatory arguments is rejected with a clear error message.
- Per-connection restriction is not supported; types apply to all connections.

In this fallback, the type service definition is marked `container.excluded` since the type is instantiated by DBAL rather than consumed as a service.

## Dependencies

This PR depends on changes in DBAL and ORM to support per-connection `TypeRegistry`:

- DBAL doctrine/dbal#7342
- ORM doctrine/orm#12421

## Demo

A working example using `#[AsDbalType]` to store emails as base64 in the database is available at: https://github.com/GromNaN/symfony-demo/pull/5

### DIFF
--- a/src/Attribute/AsDbalType.php
+++ b/src/Attribute/AsDbalType.php
@@ -9,8 +9,8 @@ use Attribute;
 /**
  * Registers the tagged class as a Doctrine DBAL type on the connection.
  *
- * With DBAL >= 4.5 and ORM >= 3.7 when the ORM are used, the type is registered
- * per connection and enables constructor dependency injection.
+ * With DBAL >= 4.5 and ORM >= 3.7, the type is registered per connection and
+ * enables constructor dependency injection.
  *
  * With DBAL < 4.5, the type is registered globally instead.
  */
--- a/src/Attribute/AsDbalType.php
+++ b/src/Attribute/AsDbalType.php
@@ -7,24 +7,21 @@ namespace Doctrine\Bundle\DoctrineBundle\Attribute;
 use Attribute;
 
 /**
- * Registers a DBAL type as a service, with full dependency injection support
- * when DBAL's TypeRegistry injection is available (DBAL >= 4.5).
+ * Registers the tagged class as a Doctrine DBAL type on the connection.
  *
- * When TypeRegistry injection is supported, the type is registered in the
- * per-connection TypeRegistry, enabling constructor dependency injection.
- * When not supported, the type falls back to the global type registry
- * via doctrine.dbal.connection_factory.types (no DI support).
+ * With DBAL >= 4.5 and ORM >= 3.7 when the ORM are used, the type is registered
+ * per connection and enables constructor dependency injection.
+ *
+ * With DBAL < 4.5, the type is registered globally instead.
  */
 #[Attribute(Attribute::TARGET_CLASS | Attribute::IS_REPEATABLE)]
 final readonly class AsDbalType
 {
     /**
      * @param string|null $name       The DBAL type name used in column mappings (e.g. in #[Column(type: ...)]).
-     *                                Defaults to the fully-qualified class name of the type when omitted,
-     *                                so that #[Column(type: MyType::class)] works without declaring a name.
+     *                                Defaults to the fully-qualified class name of the type when omitted.
      * @param string|null $connection Restrict the type to a specific named connection.
      *                                When null (default), the type is registered for all connections.
-     *                                Ignored when TypeRegistry injection is not supported by the installed DBAL version.
      */
     public function __construct(
         public string|null $name = null,
--- a/src/Attribute/AsDbalType.php
+++ b/src/Attribute/AsDbalType.php
@@ -6,10 +6,29 @@ namespace Doctrine\Bundle\DoctrineBundle\Attribute;
 
 use Attribute;
 
-#[Attribute(Attribute::TARGET_CLASS)]
+/**
+ * Registers a DBAL type as a service, with full dependency injection support
+ * when DBAL's TypeRegistry injection is available (DBAL >= 4.5).
+ *
+ * When TypeRegistry injection is supported, the type is registered in the
+ * per-connection TypeRegistry, enabling constructor dependency injection.
+ * When not supported, the type falls back to the global type registry
+ * via doctrine.dbal.connection_factory.types (no DI support).
+ */
+#[Attribute(Attribute::TARGET_CLASS | Attribute::IS_REPEATABLE)]
 final readonly class AsDbalType
 {
-    public function __construct(public string|null $name = null)
-    {
+    /**
+     * @param string|null $name       The DBAL type name used in column mappings (e.g. in #[Column(type: ...)]).
+     *                                Defaults to the fully-qualified class name of the type when omitted,
+     *                                so that #[Column(type: MyType::class)] works without declaring a name.
+     * @param string|null $connection Restrict the type to a specific named connection.
+     *                                When null (default), the type is registered for all connections.
+     *                                Ignored when TypeRegistry injection is not supported by the installed DBAL version.
+     */
+    public function __construct(
+        public string|null $name = null,
+        public string|null $connection = null,
+    ) {
     }
 }
--- a/src/DependencyInjection/Compiler/RegisterDbalTypePass.php
+++ b/src/DependencyInjection/Compiler/RegisterDbalTypePass.php
@@ -40,7 +40,7 @@ final class RegisterDbalTypePass implements CompilerPassInterface
 
     public function process(ContainerBuilder $container): void
     {
-        if (method_exists(DbalConfiguration::class, 'setTypeRegistry')) {
+        if (method_exists(DbalConfiguration::class, 'setTypeProvider')) {
             $this->registerInTypeRegistry($container);
         } else {
             $this->registerInConfig($container);
@@ -115,10 +115,10 @@ final class RegisterDbalTypePass implements CompilerPassInterface
 
             $container
                 ->getDefinition(sprintf('doctrine.dbal.%s_connection.configuration', $name))
-                ->addMethodCall('setTypeRegistry', [$registryRef]);
+                ->addMethodCall('setTypeProvider', [$registryRef]);
 
             foreach ($connectionToOrmConfigs[$name] ?? [] as $ormConfigId) {
-                $container->getDefinition($ormConfigId)->addMethodCall('setTypeRegistry', [$registryRef]);
+                $container->getDefinition($ormConfigId)->addMethodCall('setTypeProvider', [$registryRef]);
             }
         }
     }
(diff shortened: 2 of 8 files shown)


src/Attribute/AsDatabaseType.php
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Doctrine\Bundle\DoctrineBundle\Attribute;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_CLASS | Attribute::IS_REPEATABLE)]
+class AsDatabaseType
+{
+    public function __construct(
+        public readonly string $name,
```
GromNaN: The `$name` parameter is required and has no default. A default derived from the class name (e.g. stripping `Type` and snake-casing → `FooBarType` → `foo_bar`) is intentionally **not** provided.

Reasons:
- The type name appears in `#[ORM\Column(type: ...)]` on entity properties, which is static metadata — you cannot call a function or reference a computed value there. A `NAME` constant on the type class is the recommended pattern:
  ```php
  #[AsDatabaseType(name: EncryptedEmailType::NAME)]
  final class EncryptedEmailType extends Type
  {
      public const string NAME = 'encrypted_email';
  }

  // Entity:
  #[ORM\Column(type: EncryptedEmailType::NAME)]
  private ?string $email = null;
  ```
- A derived name is hard to search for in the codebase; an explicit constant is easy to trace.
- This follows the same convention used in Doctrine MongoDB ODM custom types.

GromNaN: Alternatively, we could use the class name as the type name, as is done for [`#[Autoconfigure]`](https://symfony.com/doc/current/service_container/tags.html#autoconfiguring-tags).
This would give us a name that matches the class name, as is done for default services.

```php
#[AsDatabaseType]
final class EncryptedEmailType extends Type
{
    // DBAL Type implementation
}

// Entity:
#[ORM\Column(type: EncryptedEmailType::class)]
private ?string $email = null;
```

stof: We should still allow an optional name, in case you want to use a different name than the class name (the compiler pass will need to support such case anyway to allow registering multiple services using the same class but different dependencies)

GromNaN: Done. The `$name` parameter is now optional (defaults to `null`). When omitted, the compiler pass uses the service id as the type name — which is the FQCN by default with autowiring, so `#[Column(type: MyType::class)]` works out of the box. An explicit name can still be provided when needed.

GromNaN:
To be rebased after https://github.com/doctrine/DoctrineBundle/pull/2197

derrabus:
@GromNaN Do you want to continue your work here?

GromNaN:
I'll finish the PRs on DBAL and ORM before polishing this PR on the Bundle. In August.

tests/DataCollector/DoctrineDataCollectorTest.php
```diff
@@ -56,6 +60,14 @@ public function testCollectEntities(): void
 
         $manager->method('getMetadataFactory')->willReturn($factory);
         $manager->method('getConfiguration')->willReturn($config);
+        if (method_exists(DBALConfiguration::class, 'getTypeRegistry')) {
+            $dbalConfig = $this->createStub(DBALConfiguration::class);
+            $dbalConfig->method('getTypeRegistry')->willReturn(new TypeRegistry());
+            $connection = $this->createStub(Connection::class);
+            $connection->method('getConfiguration')->willReturn($dbalConfig);
+            $manager->method('getConnection')->willReturn($connection);
+        }
```
GromNaN: ORM's `SchemaValidator` resolves the `TypeRegistry` from the connection's DBAL configuration. Since `TypeRegistry` is `final` it cannot be doubled, so PHPUnit fails to auto-generate a return value for `getTypeRegistry()`. We provide a real instance through a stubbed connection instead.

This is something projects injecting a mock of the `EntityManagerInterface` will have to do in some places.

uuf6429:
Hi! I was a bit confused by the PR description seeing as the attribute already exists from the [other PR](https://github.com/doctrine/DoctrineBundle/pull/2197) linked above - could we maybe update it to explain why this PR is still needed or what's it trying to solve?

PS: I saw some documentation in that other PR which is quite good, but interestingly the only relevant reference in a Google search for "AsDbalType" brings me to this PR. ~~Isn't that documentation published anywhere? AFAIK it's also [missing on symfony](https://symfony.com/bundles/DoctrineBundle/current/index.html).~~ Aah, found it: https://www.doctrine-project.org/projects/doctrine-bundle/en/3.3/dbal-type.html

GromNaN:
@uuf6429 Sorry for the confusion. This PR was created before the merge, but was not merged because it requires upstream work on DBAL and ORM. Thanks for pointing the inconsistence, I updated the title and description to reflect the current state.

src/DependencyInjection/Compiler/RegisterDbalTypePass.php
```diff
@@ -21,82 +29,130 @@ final class RegisterDbalTypePass implements CompilerPassInterface
...
+
+        $taggedServiceIds = $container->findTaggedServiceIds(self::TAG);
 
-                if (! is_subclass_of($class, Type::class)) {
-                    throw new InvalidArgumentException(sprintf('The "%s" class must extends "%s".', $class, Type::class));
+        if ($configTypes === [] && $taggedServiceIds === []) {
+            return;
+        }
+
+        // Map connection name → ORM configuration service IDs (when ORM is installed)
```
stof: this whole logic mapping to the ORM config should be removed as the ORM PR has been updated to make the EntityManager rely on the TypeRegistry of its Connection instead of taking it separately in its own config.

src/Attribute/AsDbalType.php
```diff
@@ -6,10 +6,29 @@
...
 
-#[Attribute(Attribute::TARGET_CLASS)]
+/**
+ * Registers a DBAL type as a service, with full dependency injection support
+ * when DBAL's TypeRegistry injection is available (DBAL >= 4.5).
+ *
+ * When TypeRegistry injection is supported, the type is registered in the
+ * per-connection TypeRegistry, enabling constructor dependency injection.
+ * When not supported, the type falls back to the global type registry
+ * via doctrine.dbal.connection_factory.types (no DI support).
```
stof: This phpdoc is partially documenting the internal processing of the attribute rather than the user-facing behavior. Documenting the processing belongs to the compiler pass IMO.

src/DependencyInjection/Compiler/RegisterDbalTypePass.php
```diff
@@ -21,82 +29,130 @@ final class RegisterDbalTypePass implements CompilerPassInterface
 {
     private const string TAG = 'doctrine.dbal.type';
 
-    /**
-     * @param ReflectionClass<T> $reflector
-     *
-     * @template T of ReflectionClass
-     */
+    /** @param ReflectionClass<object> $reflector */
```
stof: I would use `ReflectionClass<*>` as we don't care about the actual type.

phpstan makes the generic type invariant on PHP 8.4+ (due to the API related to lazy objects) rather than covariant, so `ReflectionClass<object>` is not a way to accept any ReflectionClass.

GromNaN: Wonderful! Thank you for that explanation.

src/Attribute/AsDbalType.php
GromNaN: Indeed, I removed internal details and focused on the feature from a user POV.

src/DependencyInjection/Compiler/RegisterDbalTypePass.php
GromNaN: Good catch. I've removed the unnecessary code.

src/DependencyInjection/Compiler/RegisterDbalTypePass.php
```diff
@@ -21,82 +28,111 @@ final class RegisterDbalTypePass implements CompilerPassInterface
...
+            }
+
+            // Service-tagged types: global (no connection restriction) or matching this connection
+            foreach ($taggedServiceIds as $id => $tags) {
+                foreach ($tags as $tag) {
+                    if ($name !== ($tag['connection'] ?? $name)) {
+                        continue;
+                    }
+
+                    $services[$tag['type_name'] ?? $tag['type'] ?? $id] = new Reference($id);
```
stof: why supporting 2 different attribute names for the type name ?

GromNaN: Good point. I liked `type` better for the explicit tag configuration (shorter, reads nicely), but `type_name` is already the key shipped in 3.3.0, so keeping it avoids a BC break for anyone tagging a service by hand. I dropped the extra  fallback and kept `type_name` as the only key.

PR #2224, opened 2026-04-10
Group large number of queries by default

HypeMC:
Profiling commands that execute many queries makes the non-grouped Doctrine panel hard to use and sometimes impractical to open.

This PR groups queries by default when their number exceeds a threshold (currently 50), preserving the current behavior for common cases while improving usability for large query sets.

### DIFF
--- a/templates/Collector/db.html.twig
+++ b/templates/Collector/db.html.twig
@@ -168,7 +168,7 @@
 
     <div class="sf-tabs" style="margin-top: 20px;">
         <div class="tab {{ collector.queries is empty ? 'disabled' }}">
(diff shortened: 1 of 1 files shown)


ostrolucky:
I'm not entirely sure what is this doing. Have you got some screenshots?

HypeMC:
@ostrolucky It just activates the existing group queries feature by default when there are large  number of queries.

image

ostrolucky:
Alright. But I would say let's bump the limit to 100

HypeMC:
@ostrolucky Done

PR #2225, opened 2026-04-24
Remove form related services if the Symfony Form component is not installed

xabbuh:
related to symfony/symfony#63671

### DIFF
--- a/src/DependencyInjection/DoctrineExtension.php
+++ b/src/DependencyInjection/DoctrineExtension.php
@@ -68,6 +68,7 @@ use Symfony\Component\DependencyInjection\Loader\PhpFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
 use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\FormTypeGuesserInterface;
 use Symfony\Component\Messenger\Bridge\Doctrine\Transport\DoctrineTransportFactory;
 use Symfony\Component\Messenger\MessageBusInterface;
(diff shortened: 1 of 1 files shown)


ostrolucky:
Please retarget to 3.2.x

xabbuh:
Why not 2.18? It still receives bugfixes, doesn't it?

ostrolucky:
It's questionable if this change is a bugfix and from what I see in linked issue, it's symfony 7.4.7+ specific issue. Actually I don't see why is the fix here with no explanation, since it was broken by changes on Symfony side.

nicolas-grekas (approved):

GromNaN (approved):

GromNaN:
> It's questionable if this change is a bugfix and from what I see in linked issue, it's symfony 7.4.7+ specific issue. Actually I don't see why is the fix here with no explanation, since it was broken by changes on Symfony side.

To address @ostrolucky's concern: this is not a Symfony 7.4.7-specific regression. The exact same crash was already reported in 2019 symfony/symfony#32396 when PHP 7.2 tightened interface resolution at class-load time, and re-surfaced now via symfony/symfony#63671.

The root cause lives in `symfony/doctrine-bridge`. [`DoctrineOrmTypeGuesser`](https://github.com/symfony/symfony/blob/7.4/src/Symfony/Bridge/Doctrine/Form/DoctrineOrmTypeGuesser.php) unconditionally `implements FormTypeGuesserInterface` with no guard against `symfony/form` being absent. That is where the real fix belongs.

This change is a valid **defensive programming**: DoctrineBundle should never register form-related services when `symfony/form` is not installed, regardless of Symfony version.

PR #2229, opened 2026-05-12
[Bug] Lower EntityValueResolver priority below Request/Session value resolvers

ousamabenyounes:
Follow-up to symfony/symfony#54337 and symfony/symfony#64189, where Nicolas Grekas suggested patching DoctrineBundle instead of FrameworkBundle.

`doctrine.orm.entity_value_resolver` is tagged with priority `110`, which places it above every FrameworkBundle resolver — including `RequestValueResolver` and `SessionValueResolver` at priority `50`. For every controller argument typed `Request` or `Session`, `EntityValueResolver::resolve()` runs first and calls `ManagerRegistry::getManagerForClass()`, bootstrapping the EM just to discover the argument is not an entity. The reporter measures a 40-50 ms hit per request on a fresh `symfony/skeleton` + `symfony/webapp-pack`.

Lowering the tag priority to `40` (below FrameworkBundle's `Request` and `Session` resolvers at 50, and below `BackedEnum`/`Uid`/`DateTime`/`RequestAttribute` at 100) lets those resolvers claim their narrow-typed arguments first; `EntityValueResolver` only runs for arguments none of them resolved, so the EM is not bootstrapped for `Request \$request` / `SessionInterface \$session` controllers.

Behavior for entity-typed arguments (`User \$user` etc.) is unchanged — none of the earlier resolvers claim user-land entity types, so the chain falls through to `EntityValueResolver` exactly as before.

Target: `2.18.x` (lowest maintained branch where the bug exists; the fix propagates upward via the merge-up branches).

Test asserts the tag priority is strictly less than `50`. Verified locally: reverting the priority change makes the test fail with "Failed asserting that 110 is less than 50."

### DIFF
--- a/config/orm.php
+++ b/config/orm.php
@@ -265,7 +265,7 @@ return static function (ContainerConfigurator $container): void {
                 service('doctrine'),
                 service('doctrine.orm.entity_value_resolver.expression_language')->ignoreOnInvalid(),
             ])
-            ->tag('controller.argument_value_resolver', ['priority' => 110, 'name' => EntityValueResolver::class])
+            ->tag('controller.argument_value_resolver', ['priority' => 40, 'name' => EntityValueResolver::class])
 
         ->set('doctrine.orm.entity_value_resolver.expression_language', ExpressionLanguage::class)
 
--- a/tests/DependencyInjection/DoctrineExtensionTest.php
+++ b/tests/DependencyInjection/DoctrineExtensionTest.php
@@ -1507,6 +1507,30 @@ class DoctrineExtensionTest extends TestCase
         $this->assertEquals(new MapEntity(null, null, null, [], null, null, null, true, true), $container->get('controller_resolver_defaults'));
     }
 
+    #[RequiresMethod(EntityValueResolver::class, '__construct')]
+    public function testEntityValueResolverPriorityIsBelowRequestAndSessionResolvers(): void
+    {
+        if (! interface_exists(EntityManagerInterface::class)) {
+            self::markTestSkipped('This test requires ORM');
+        }
+
+        $container = $this->getContainer();
+        $extension = new DoctrineExtension();
+        $config    = BundleConfigurationBuilder::createBuilderWithBaseValues()->build();
+
+        $extension->load([DeprecationFreeConfig::get(), $config], $container);
+
+        $tags = $container->getDefinition('doctrine.orm.entity_value_resolver')->getTag('controller.argument_value_resolver');
+
+        $this->assertCount(1, $tags);
+        $this->assertArrayHasKey('priority', $tags[0]);
+        $this->assertLessThan(
+            50,
+            $tags[0]['priority'],
+            'EntityValueResolver must be tagged below FrameworkBundle Request/Session value resolvers (priority 50) so it does not bootstrap an entity manager on controllers that only declare Request or Session arguments. See symfony/symfony#54337.',
+        );
+    }
+
     #[TestWith(['AnnotationsBundle', 'attribute', 'Vendor'], 'Bundle without anything')]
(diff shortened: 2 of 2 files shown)


stof:
Putting it below RequestValueResolver is a no-go to me. This would inject the unconverted request attribute instead of converting it to an entity (as the first resolver providing a value wins).

stof:
Ah sorry, I confused RequestValueResolver with RequestAttributeValueResolver.

As this is about improving performance, not about fixing a bug (i.e. something not working), I'd rather ship this in the next minor version than as a patch version. Changing the priority has a risk of breaking things for projects using other resolvers that might also treat those arguments as supported if they run in a different order.

stof:
However, RequestAttributeValueResolver runs at priority 100, so this change is still invalid.

ousamabenyounes:
Thanks for the review @stof — agreed. I believe this is also what #1574/#1578 surfaced — if I read those threads correctly, the same lowering was reverted within 24h in 2022 on the `/posts/{post}` + `Post $post` use case.

Pivoting to a decorator — still in DoctrineBundle, keeping priority `110`, short-circuits before the EM bootstrap when the type-hint cannot be a user-land entity:

```php
final class LazyEntityValueResolver implements ValueResolverInterface
{
    public function __construct(private EntityValueResolver $inner) {}

    public function resolve(Request $request, ArgumentMetadata $argument): array
    {
        if ($argument->getAttributes(MapEntity::class, ArgumentMetadata::IS_INSTANCEOF)) {
            return $this->inner->resolve($request, $argument);
        }
        $type = $argument->getType();
        if (!$type || (!class_exists($type) && !interface_exists($type))) {
            return [];
        }
        if (str_starts_with($type, 'Symfony\\') || str_starts_with($type, 'Psr\\')) {
            return [];
        }
        return $this->inner->resolve($request, $argument);
    }
}
```

Honors `#[MapEntity]` (explicit overrides always delegate), leaves `Post \$post` + `/posts/{post}` untouched (`App\Entity\Post` is neither `Symfony\` nor `Psr\` → delegates), no priority change so zero BC risk for third-party resolvers. Class marked `@internal`. Will retarget to `2.19.x` per your "ship in next minor" note and add `UPGRADE-2.19.md`. OK to push the rework before I invest the test/UPGRADE work?

stof:
@ousamabenyounes I don't think we should do special cases for the `Psr` and `Symfony` top-level namespaces in DoctrineBundle (this is making assumptions about how the `Symfony` namespace is used, while it is outside Doctrine's control). Making such assumption in your own project can be fine (you know which packages using the `Symfony` namespace you are using and so can verify that the assumption that they never ship an entity is valid), but that's not OK in an open-source package that has no way to validate the assumption.

The simple improvement for that would be to let FrameworkBundle bump the priority of many of its resolver that could win over EntityValueResolver (especially useful by separating `RequestAttributeValueResolver` from things like `BackedEnumResolver` to allow `EntityValueResolver` to come between them, which is impossible today).

Btw, a valuable contribution to DoctrineBundle would be to add a functional test for this `/posts/{post}` + `Post $post` use case, so that the testsuite catches regressions (it must be a functional test, not a unit test asserting the priority based on a copy of the current priority defined in FrameworkBundle, unless we also have tests asserting that the unit test is consistent with FrameworkBundle which would become insane)

ousamabenyounes:
Thanks for the review @stof — agreed on both points (perf optimisation rather than bug fix, and the priority change carries BC risk). Closing this one in favor of #2230, which keeps only the functional regression test you suggested: a `MicroKernel` exercising `/posts/{post}` + `Post $post` end-to-end, no runtime change. Verified locally that dropping the resolver tag priority below `RequestAttributeValueResolver` (priority 100) flips the response from `200` to `500`, so the test fails closed.

PR #2230, opened 2026-05-12
Add functional regression test for EntityValueResolver on primary-key routes

ousamabenyounes:
Focused follow-up to #2229: drop the priority-change approach (rejected as a perf optimisation rather than a bug fix), keep only the regression test stof requested.

> Btw, a valuable contribution to DoctrineBundle would be to add a functional test for this `/posts/{post}` + `Post $post` use case, so that the testsuite catches regressions
> — stof on #2229 (https://github.com/doctrine/DoctrineBundle/pull/2229#issuecomment-3548470770)

Past attempts at lowering `EntityValueResolver`'s tag priority (#1574, #1578) broke this exact scenario and had to be reverted within 24 hours. The `symfony/doctrine-bridge` unit tests do not catch it — it requires a real kernel, a real router and a real `EntityManager`.

## What the test does

- Boots a `MicroKernelTrait` kernel with `FrameworkBundle` + `DoctrineBundle` and an in-memory sqlite DBAL.
- Maps a `Post` entity, declares a `/posts/{post}` route bound to a controller with `__invoke(Post $post)`.
- Persists a `Post`, hits the route, asserts `200` + the JSON body matches the resolved entity.

## Regression check

Verified locally: dropping the `controller.argument_value_resolver` tag priority to `40` in `config/orm.php` (the change that was previously rejected on #2229) flips the response from `200` to `500` — the test fails closed.

## Scope

Test-only: 4 new files under `tests/ArgumentResolver/`, no runtime change. Targeted at `2.18.x` so the coverage propagates upward via the merge-up branches.

Once this lands, #2229 can be closed (already de-facto rejected).

### DIFF
--- a/tests/ArgumentResolver/EntityValueResolverFunctionalTest.php
+++ b/tests/ArgumentResolver/EntityValueResolverFunctionalTest.php
@@ -7,7 +7,6 @@ namespace Doctrine\Bundle\DoctrineBundle\Tests\ArgumentResolver;
 use Doctrine\Bundle\DoctrineBundle\Tests\ArgumentResolver\Fixtures\EntityValueResolverFunctionalKernel;
 use Doctrine\Bundle\DoctrineBundle\Tests\ArgumentResolver\Fixtures\Post;
 use Doctrine\ORM\EntityManagerInterface;
-use Doctrine\ORM\Tools\SchemaTool;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\HttpFoundation\Request;
 
@@ -38,7 +37,9 @@ class EntityValueResolverFunctionalTest extends TestCase
             $em        = $container->get('doctrine.orm.default_entity_manager');
             assert($em instanceof EntityManagerInterface);
 
-            (new SchemaTool($em))->createSchema([$em->getClassMetadata(Post::class)]);
+            $em->getConnection()->executeStatement(
+                'CREATE TABLE posts (id INTEGER NOT NULL, title VARCHAR(255) NOT NULL, PRIMARY KEY(id))',
+            );
 
             $post = new Post('Hello world');
             $em->persist($post);
--- a/tests/ArgumentResolver/EntityValueResolverFunctionalTest.php
+++ b/tests/ArgumentResolver/EntityValueResolverFunctionalTest.php
@@ -6,11 +6,6 @@ namespace Doctrine\Bundle\DoctrineBundle\Tests\ArgumentResolver;
 
 use Doctrine\Bundle\DoctrineBundle\Tests\ArgumentResolver\Fixtures\EntityValueResolverFunctionalKernel;
 use Doctrine\Bundle\DoctrineBundle\Tests\ArgumentResolver\Fixtures\Post;
-use Doctrine\DBAL\Schema\Column;
-use Doctrine\DBAL\Schema\Name\UnqualifiedName;
-use Doctrine\DBAL\Schema\PrimaryKeyConstraint;
-use Doctrine\DBAL\Schema\Table;
-use Doctrine\DBAL\Types\Types;
 use Doctrine\ORM\EntityManagerInterface;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\HttpFoundation\Request;
@@ -42,32 +37,15 @@ class EntityValueResolverFunctionalTest extends TestCase
             $em        = $container->get('doctrine.orm.default_entity_manager');
             assert($em instanceof EntityManagerInterface);
 
-            $table = Table::editor()
-                ->setUnquotedName('posts')
-                ->addColumn(
-                    Column::editor()
-                        ->setUnquotedName('id')
-                        ->setTypeName(Types::INTEGER)
-                        ->setNotNull(true)
-                        ->setAutoincrement(true)
-                        ->create(),
-                )
-                ->addColumn(
-                    Column::editor()
-                        ->setUnquotedName('title')
-                        ->setTypeName(Types::STRING)
-                        ->setNotNull(true)
-                        ->create(),
-                )
-                ->setPrimaryKeyConstraint(
-                    new PrimaryKeyConstraint(null, [UnqualifiedName::unquoted('id')], false),
-                )
-                ->create();
-
-            $connection = $em->getConnection();
-            foreach ($connection->getDatabasePlatform()->getCreateTablesSQL([$table]) as $sql) {
-                $connection->executeStatement($sql);
-            }
+            // SchemaTool::createSchema triggers Schema::createTable, deprecated in
+            // DBAL 4 (https://github.com/doctrine/dbal/pull/7373); the DBAL 4 fluent
+            // replacement (Table::editor() / Column::editor()) is unavailable on
+            // DBAL 3 which this branch must still support, so a raw CREATE TABLE is
+            // emitted directly through the connection. The test kernel is SQLite-
+            // only, so a SQLite literal is safe here.
+            $em->getConnection()->executeStatement(
+                'CREATE TABLE posts (id INTEGER NOT NULL, title VARCHAR(255) NOT NULL, PRIMARY KEY(id))',
+            );
 
             $post = new Post('Hello world');
             $em->persist($post);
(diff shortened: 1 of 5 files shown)


tests/ArgumentResolver/EntityValueResolverFunctionalTest.php
```diff
@@ -0,0 +1,64 @@
...
+ * Regression coverage for the /posts/{post} + Post $post happy path resolved by
+ * Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver. The tag priority
+ * configured in config/orm.php (and the priorities of FrameworkBundle's own
+ * value resolvers) must stay compatible so that the entity argument is still
+ * populated from the route placeholder.
+ */
+#[RequiresMethod(EntityValueResolver::class, '__construct')]
+class EntityValueResolverFunctionalTest extends TestCase
+{
+    #[IgnoreDeprecations]
```
stof: there is no reason to ignore deprecations in this test. This is not a test covering a deprecated feature. The test should be written in a way that does not trigger deprecations instead.

tests/ArgumentResolver/EntityValueResolverFunctionalTest.php
```diff
@@ -0,0 +1,64 @@
...
+use function restore_exception_handler;
+
+/**
+ * Regression coverage for the /posts/{post} + Post $post happy path resolved by
+ * Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver. The tag priority
+ * configured in config/orm.php (and the priorities of FrameworkBundle's own
+ * value resolvers) must stay compatible so that the entity argument is still
+ * populated from the route placeholder.
+ */
+#[RequiresMethod(EntityValueResolver::class, '__construct')]
```
stof: why this attribute ?

tests/ArgumentResolver/EntityValueResolverFunctionalTest.php
```diff
@@ -0,0 +1,64 @@
...
+        $response = $kernel->handle(Request::create('/posts/' . $post->id, 'GET'));
+
+        self::assertSame(200, $response->getStatusCode());
+        self::assertSame(
+            ['id' => $post->id, 'title' => 'Hello world'],
+            json_decode((string) $response->getContent(), true),
+        );
+
+        $kernel->shutdown();
+        restore_exception_handler();
```
stof: Those should be in a `finally` block (or in a teardown method) so that they are done even in case of test failures.

Regarding the exception handler, a better solution IMO is to avoid FrameworkBundle messing up with the exception handler, which can be done by adding `symfony/runtime` as a dev dependency due to the way FrameworkBundle chooses its behavior in that regard.

stof (changes requested):

ousamabenyounes:
Thanks for the review @stof — agreed on all three. Pushed 2653f677: dropped `#[IgnoreDeprecations]` and `#[RequiresMethod]`, moved `$kernel->shutdown()` and `restore_exception_handler()` into a `finally` block. On the `symfony/runtime` suggestion: kept `restore_exception_handler()` here since existing tests (`UrlOverrideTest`, `RegistryTest`, `InfoDoctrineCommandTest`, `CacheCompatibilityPassTest`) all use the same pattern — happy to handle that as a separate sweep across the suite if you'd prefer.

One open question on the deprecation point. The scenario covered (`/posts/{post}` + `Post $post` without an attribute) is `auto_mapping`, which the test was specifically built to protect — the two prior reverts (#1574, #1578) broke this exact case. Without `#[IgnoreDeprecations]` the run now reports deprecation #1804.

Two paths, would you prefer:

(a) switch kernel to `auto_mapping: false` + `#[MapEntity]` on the controller param — silent test, but loses the priority-chain regression coverage (the explicit attribute short-circuits the resolver chain, so the original bug would no longer reproduce here);

(b) keep the scenario and declare the expected deprecation via `expectDeprecationWithIdentifier('https://github.com/doctrine/DoctrineBundle/pull/1804')` (pattern already used in `CacheCompatibilityPassTest`, etc.).

stof:
@ousamabenyounes `auto_mapping: true` is *not* about the case where the route parameter named like the argument contains the primary key of the entity. It is about the case where we search the entity based on mapping *other* fields of the entity to route parameters of the same name (for instance a `slug` route parameter used to find the Post entity based on its `slug` field).

ousamabenyounes:
Thanks for the clarification @stof — you're right, I conflated the two paths. The `/posts/{post}` + `Post $post` case hits `EntityValueResolver::find()` via the primary-key short-circuit, which doesn't depend on `auto_mapping`. That config only governs the criteria path (`getCriteria()`) — i.e. `/posts/{slug}` mapping to the `slug` field, as you described.

Fixed in ac9f3a3: kernel now sets `auto_mapping: false`. Test stays green with no deprecation, and the regression check still trips closed — dropping the resolver tag priority to 40 in `config/orm.php` still flips the response from 200 to 500 (since #1574/#1578 broke tag ordering, not the auto-mapping defaults).

tests/ArgumentResolver/EntityValueResolverFunctionalTest.php
ousamabenyounes: Done — dropped in 2653f67. Underlying deprecation is now avoided entirely by `auto_mapping: false` in ac9f3a3, so the test no longer triggers any deprecation.

tests/ArgumentResolver/EntityValueResolverFunctionalTest.php
ousamabenyounes: Done in 2653f67 — dropped.

stof (approved):

tests/ArgumentResolver/EntityValueResolverFunctionalTest.php
```diff
@@ -0,0 +1,61 @@
...
+            $response = $kernel->handle(Request::create('/posts/' . $post->id, 'GET'));
+
+            self::assertSame(200, $response->getStatusCode());
+            self::assertSame(
+                ['id' => $post->id, 'title' => 'Hello world'],
+                json_decode((string) $response->getContent(), true),
+            );
+        } finally {
+            $kernel->shutdown();
+            restore_exception_handler();
```
ostrolucky: This is quite non-standard. Why?

stof: When `symfony/runtime` is not installed, FrameworkBundle registers the Symfony exception handler on boot (when the runtime component is installed, FrameworkBundle assumes that the runtime component is responsible for managing the exception handler registration). And that registration messes with the PHPUnit exception handler, which is reported by PHPUnit as an issue. See https://github.com/doctrine/DoctrineBundle/pull/2230#discussion_r3232485646 for my proposal about it.

ousamabenyounes:
Thanks @stof — applied your proposal in ce1fd84 + 635073a8.

- `symfony/runtime` added to `require-dev`; plugin disabled in `config.allow-plugins` (only the autoloaded `SymfonyRuntime` class matters for `FrameworkBundle::boot()`'s `class_exists()` gate).
- `restore_exception_handler()` + its `use function` import dropped.

@ostrolucky — that's why the call was there. Verified locally with `vendor/bin/phpunit tests/ArgumentResolver/EntityValueResolverFunctionalTest.php` (PHP 8.5.6, PHPUnit 12.5.27).

**Without the fix** (revert `symfony/runtime` from deps, no `restore_exception_handler()`):
```
There was 1 risky test:

1) Doctrine\Bundle\DoctrineBundle\Tests\ArgumentResolver\EntityValueResolverFunctionalTest::testEntityArgumentResolvedFromRoutePlaceholder
Test code or tested code did not remove its own exception handlers

/app/tests/ArgumentResolver/EntityValueResolverFunctionalTest.php:27

OK, but there were issues!
Tests: 1, Assertions: 2, Deprecations: 3, Risky: 1.
```

**With the fix** (`symfony/runtime` in deps, no `restore_exception_handler()`):
```
OK, but there were issues!
Tests: 1, Assertions: 2, Deprecations: 3.
```

`Risky: 1` gone — FrameworkBundle now sees `SymfonyRuntime::class` and delegates exception handler management to the runtime component as you described. The 3 remaining deprecations are pre-existing (event-dispatcher + dbal), unrelated to this PR.

ostrolucky (approved):

PR #2234, opened 2026-05-21
[RFC] Warn about user entities stored into session without implementing `__serialize`

MatTheCat:
Attempt to address https://github.com/symfony/symfony/issues/11155: Symfony makes it easy to use entities as security users, and those entities will always benefit from a `__serialize` and `__unserialize` methods when authenticated against a stateful firewall: it will limit the data stored into the session and avoid errors due to uninitialized proxies.

The issue is, you generally learn about this after opening an issue due to such error (like https://github.com/symfony/symfony/issues/64017) as there is no heads-up before that.

This PR tries to compensate by logging a warning before a stateful firewall triggers the serialization of a user entity which doesn’t implement `Serializable` or `__serialize` in debug mode.

Warning logs are highlighted in Symfony Web Debug Toolbar which should attract developers attention and helping them understand what’s wrong:

### DIFF
--- a/config/security_debug.php
+++ b/config/security_debug.php
@@ -1,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+use Doctrine\Bundle\DoctrineBundle\Security\Debug\UserSerializationChecker;
+
+return static function (ContainerConfigurator $container): void {
+    $container->services()
+
+        ->set('doctrine.security.debug.user_serialization_checker', UserSerializationChecker::class)
+            ->args([
+                service('security.firewall.map'),
+                service('security.untracked_token_storage'),
+                service('doctrine'),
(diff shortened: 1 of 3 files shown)


ostrolucky:
Sorry, but this doesn't belong here, given that user is not required to use doctrine-bundle.

MatTheCat:
The issue comes with user entities, which can only be achieved with the doctrine-bundle. If it is not the place to open such a PR where would that be?

ostrolucky:
No, issue comes from any object that is serialized to session and implements UserInterface. It's not dependent to doctrine nor entities. Correct place is SecurityBundle.

MatTheCat:
Pretty sure there are issues regarding entities; I mentioned some in the PR description.

Anyways this cannot be implemented in the SecurityBundle: we wouldn’t want to warn developers using POPOs as security user since they typically would be tailored for that exact use.

PR #2235, opened 2026-06-03
Fix dark theme time bar colors in the profiler

MatTheCat:
| Before | After |
| - | - |
|  |  |

### DIFF
--- a/templates/Collector/db.html.twig
+++ b/templates/Collector/db.html.twig
@@ -101,9 +101,15 @@
 
 {% block queries %}
     <style>
+        :root {
+            --query-time-bar-color: var(--gray-200);
+        }
+        .theme-dark {
+            --query-time-bar-color: var(--gray-700);
+        }
         .time-container { position: relative; }
-        .time-container .nowrap { position: relative; z-index: 1; text-shadow: 0 0 2px #fff; }
-        .time-bar { display: block; position: absolute; top: 0; left: 0; bottom: 0; background: #e0e0e0; }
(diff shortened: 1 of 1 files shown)


templates/Collector/db.html.twig
```diff
@@ -101,9 +101,15 @@
...
+            --query-time-bar-color: #E0E0E0;
+        }
+        .theme-dark {
+            --query-time-bar-color: #3A3A3A;
+        }
         .time-container { position: relative; }
-        .time-container .nowrap { position: relative; z-index: 1; text-shadow: 0 0 2px #fff; }
-        .time-bar { display: block; position: absolute; top: 0; left: 0; bottom: 0; background: #e0e0e0; }
+        .time-container .nowrap { position: relative; z-index: 1; text-shadow: 0 0 2px var(--page-background); }
+        .time-bar { display: block; position: absolute; top: 0; left: 0; bottom: 0; background: var(--query-time-bar-color); }
```
stof, line 112: Would it work to use the `--page-grid-dot-color` color defined in the WebProfilerBundle styles instead of introducing a separate color there ? The color is close to the one you use in this PR.

stof: or maybe one of the other colors of those styles, if there is a better fit.

MatTheCat: Semantic variables should not be used for unrelated purpose, but I can use the nearest `--gray-*`.

MatTheCat:
Opened #2248 on a correct branch..!

PR #2241, opened 2026-06-09
Avoid reporting error from unrelated call

### LINKED ISSUE #2239: [Symfony 8.1] Debug toolbar throws "Failed to replace query parameters: PCRE error 4" when query parameters contain binary strings (e.g. binary UUIDs)
### Bug Report

  | Version                                     | 3.2.3

#### Summary

 `DoctrineExtension::replaceQueryParameters()` in the Symfony debug toolbar polls `preg_last_error()` after running `preg_replace_callback()`. The callback uses `escapeFunction()`, which deliberately runs `preg_match('//u', $value)` as a
  UTF-8 probe to detect binary strings. For binary parameters that probe sets `PREG_BAD_UTF8_ERROR` (4) in the global PCRE error state. The outer wrapper then mistakes that stale state for its own failure and throws. In projects that bind
  binary UUIDs (`UuidV7::toBinary()` → `BINARY(16)`) as PKs/FKs, every rendered query crashes the Doctrine collector — the debug toolbar becomes unusable.

#### Current behavior

Opening the Symfony web profiler → **Doctrine** → **SQL Queries** crashes with:

  ```
  An exception has been thrown during the rendering of a template ("Failed to replace query parameters: PCRE error 4")
  in @Doctrine/Collector/db.html.twig at line 265.
  ```

  This happens for any query whose parameters include at least one binary string. The chain:

  1. `replaceQueryParameters()` runs `preg_replace_callback(...)` over the SQL.
  2. For each placeholder the callback invokes `escapeFunction($value)`.
  3. `escapeFunction()` line 72 runs `preg_match('//u', $binary)` to detect non-UTF-8 (this is the intended binary-detection path — it converts the value to `0x…` hex).
  4. That probe sets `preg_last_error() === PREG_BAD_UTF8_ERROR` (the documented behavior of `//u` against a non-UTF-8 subject).
  5. After `preg_replace_callback()` returns, `replaceQueryParameters()` reads `preg_last_error()` and throws, even though the outer call itself succeeded and `$result` is a perfectly valid string.

  Relevant code (`src/Twig/DoctrineExtension.php`):

  ```php
  // escapeFunction() — line 72: deliberate UTF-8 probe (sets PCRE error state)
  case is_string($result) && ! preg_match('//u', $result):
      $result = '0x' . strtoupper(bin2hex($result));
      break;
  ```

  ```php
  // replaceQueryParameters() — line 142: false positive
  $pregError = preg_last_error();
  if ($pregError !== PREG_NO_ERROR) {
      throw new RuntimeException(sprintf('Failed to replace query parameters: PCRE error %d', $pregError));
  }
  ```

#### Expected behavior

The runnable SQL is rendered without throwing — binary parameters appear as `0x…` hex literals (exactly what `escapeFunction()` was designed to produce). `preg_last_error()` reflecting state from inside the callback should not be treated
  as a failure of the outer `preg_replace_callback()`.

  `preg_replace_callback()` already returns `null` on actual failure, and the file has an `if ($result === null)` guard right below the broken polling. Removing the `preg_last_error()` check is sufficient:

  ```diff
  -        $pregError = preg_last_error();
  -        if ($pregError !== PREG_NO_ERROR) {
  -            throw new RuntimeException(sprintf('Failed to replace query parameters: PCRE error %d', $pregError));
  -        }
  -
           if ($result === null) {
               throw new RuntimeException('Failed to replace query parameters: unexpected null result');
           }
  ```

#### How to reproduce

```php
  <?php
  require 'vendor/autoload.php';

  use Doctrine\Bundle\DoctrineBundle\Twig\DoctrineExtension;

  $ext = new DoctrineExtension();
  echo $ext->replaceQueryParameters(
      'SELECT * FROM t WHERE id = ?',
      ["\xff\xfe\xfd"], // any non-UTF-8 byte sequence (e.g. UuidV7::toBinary() output)
  );
  ```

  Run it:

  ```
  $ composer require doctrine/doctrine-bundle:3.2.3
  $ php repro.php

 Fatal error: Uncaught RuntimeException: Failed to replace query parameters: PCRE error 4
    in vendor/doctrine/doctrine-bundle/src/Twig/DoctrineExtension.php:144
  ```

### PR DESCRIPTION
**Pull request by greg0ire**
preg_match('//u', …)  is not the most idiomatic way of checking the encoding of a string. The most idiomatic way would be to use… `mb_check_encoding()`.

Even then, checking `preg_last_error()` should only happen if the previous preg call reported a failure, so as to avoid leading to the same issue with another call somewhere else.

Hopefully, in the future, we get a version of these functions that throw exceptions.

Fixes #2239

### DIFF
--- a/src/Twig/DoctrineExtension.php
+++ b/src/Twig/DoctrineExtension.php
@@ -25,14 +25,12 @@ use function implode;
 use function is_array;
 use function is_bool;
 use function is_string;
+use function mb_check_encoding;
 use function preg_last_error;
-use function preg_match;
 use function preg_replace_callback;
 use function sprintf;
 use function strtoupper;
 use function substr;
-
-use const PREG_NO_ERROR;
 
 /**
  * This class contains the needed functions in order to do the query highlighting
@@ -68,8 +66,7 @@ class DoctrineExtension extends AbstractExtension
         $result = $parameter;
 
         switch (true) {
-            // Check if result is non-unicode string using PCRE_UTF8 modifier
-            case is_string($result) && ! preg_match('//u', $result):
+            case is_string($result) && ! mb_check_encoding($result, 'UTF-8'):
                 $result = '0x' . strtoupper(bin2hex($result));
                 break;
 
@@ -139,13 +136,11 @@ class DoctrineExtension extends AbstractExtension
             $query,
         );
 
-        $pregError = preg_last_error();
-        if ($pregError !== PREG_NO_ERROR) {
-            throw new RuntimeException(sprintf('Failed to replace query parameters: PCRE error %d', $pregError));
-        }
-
         if ($result === null) {
-            throw new RuntimeException('Failed to replace query parameters: unexpected null result');
+            throw new RuntimeException(sprintf(
+                'Failed to replace query parameters: PCRE error %d',
+                preg_last_error(),
+            ));
         }
 
         return $result;
--- a/tests/Twig/DoctrineExtensionTest.php
+++ b/tests/Twig/DoctrineExtensionTest.php
@@ -118,6 +118,16 @@ class DoctrineExtensionTest extends TestCase
     {
         $this->assertEquals(1.1, DoctrineExtension::escapeFunction(1.1));
     }
+
+    public function testReplaceBinaryParameterAfterEscaping(): void
+    {
+        $extension  = new DoctrineExtension();
+        $query      = 'column = ?';
+        $parameters = [pack('H*', '9d40b8c1417f42d099af4782ec4b20b6')];
+
+        $result = $extension->replaceQueryParameters($query, $parameters);
+        $this->assertEquals('column = 0x9D40B8C1417F42D099AF4782EC4B20B6', $result);
+    }
 }
 
 class DummyClass

### CODE REVIEW

**Review by dmaicher — APPROVED** [review]

File: src/Twig/DoctrineExtension.php [hunk 1]
```diff
@@ -68,8 +66,7 @@ public static function escapeFunction(mixed $parameter): string|int|float
         $result = $parameter;
 
         switch (true) {
-            // Check if result is non-unicode string using PCRE_UTF8 modifier
-            case is_string($result) && ! preg_match('//u', $result):
+            case is_string($result) && ! mb_check_encoding($result, 'UTF-8'):
```
stof, line 69: this should involve adding a dependency on `ext-mbstring` (whcih can be fulfilled by `symfony/polyfill-mbstring`) as this function comes from an extension that might not be enabled.

whataboutpereira:
I love it, a few hours of digging ("It was just working!") only to realize it has already been fixed. :)

Note to self: run `composer update` at the first sign of trouble.

greg0ire:
Yeah I've been there… so frustrating.

PR #2247, opened 2026-06-10
Add missing dependency

greg0ire:
In #2241, I added a call to `mb_check_encoding()`, making `mbstring` required.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,6 +30,7 @@
     "homepage": "https://www.doctrine-project.org",
     "require": {
         "php": "^8.4",
+        "ext-mbstring": "*",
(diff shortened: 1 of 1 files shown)


ostrolucky (approved):

dmaicher (approved):

derrabus:
Symfony components and bundles usually depend on `symfony/polyfill-mbstring` instead of depending on the extension directly.

stof:
@derrabus As `symfony/polyfill-mbstring` provides `ext-mbstring` in its composer metadata (which is possible thanks to being a full polyfill of the extension), requiring the extension works fine and is still compatible with projects requiring the polyfill and running without the extension.

PR #2252, opened 2026-06-25
Fix doctrine:database:create failing without server_version for some connections

ostrolucky:
Supersedes #2251

### DIFF
--- a/src/Command/CreateDatabaseDoctrineCommand.php
+++ b/src/Command/CreateDatabaseDoctrineCommand.php
@@ -4,8 +4,8 @@ declare(strict_types=1);
 
 namespace Doctrine\Bundle\DoctrineBundle\Command;
 
+use Doctrine\DBAL\Driver\AbstractPostgreSQLDriver;
 use Doctrine\DBAL\DriverManager;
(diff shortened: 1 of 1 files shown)


derrabus:
I'm going to revert this change because of #2261.

src/Command/CreateDatabaseDoctrineCommand.php
```diff
@@ -68,7 +68,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         /** @phpstan-ignore unset.offset */
         unset($params['dbname'], $params['path'], $params['url']);
 
-        if ($connection->getDatabasePlatform() instanceof PostgreSQLPlatform) {
+        if ($connection->getDriver() instanceof AbstractPostgreSQLDriver) {
```
derrabus, line 71: This is incorrect. We cannot make any assumptions based on the type of the driver.

`$connection->getDriver()` gives us the whole middleware stack and the actual database driver is at the very bottom of it. So only if the middleware stack of a Postgres connection is empty, we will have postgres driver here.

PR #2253, opened 2026-06-30
Add configuration for the console entity value resolver

### LINKED ISSUE #64760: [Console] MapEntity does not work
### Symfony version(s) affected

8.1.0

### Description

I just gave the console argument value resolvers introduced in https://github.com/symfony/symfony/pull/62917 a try and noticed that the `#[MapEntity]` attribute does not work as expected.

I'm creating the bug report here although it might be related to [`doctrine/doctrine-bundle`](https://github.com/doctrine/DoctrineBundle).

### How to reproduce

I took the command from https://symfony.com/doc/current/console/value_resolver.html#entityvalueresolver, used an entity class which exists in my app and got the following exception when running the command (`bin/console app:test 1`):

> `TypeError: App\Command\TestCommand::__invoke(): Argument #1 ($user) must be of type App\Entity\User, string given`

### Possible Solution

https://github.com/symfony/symfony/pull/62917 added the new class https://github.com/symfony/doctrine-bridge/blob/8.1/ArgumentResolver/Console/EntityValueResolver.php but the class does not seem in any symfony configuration file.

I think a configuration similar to https://github.com/doctrine/DoctrineBundle/blob/940ff32c627be92234362614f413fe041d83a5ce/config/orm.php#L165 (for `Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver`) is necessary for the new class.
It should be tagged with `console.argument_value_resolver`.

In my app I could fix the issue by adding the following configuration:

```php
    use Symfony\Bridge\Doctrine\ArgumentResolver\Console\EntityValueResolver as ConsoleEntityValueResolver;

    $container
        ->set('console.argument_resolver.entity', ConsoleEntityValueResolver::class)
        ->args([
            service('doctrine'),
            service('doctrine.orm.entity_value_resolver.expression_language')->ignoreOnInvalid(),
        ])
        ->tag('console.argument_value_resolver', ['priority' => 110, 'name' => ConsoleEntityValueResolver::class])
    ;
```

### Additional Context

_No response_

W0rma:
Fixes https://github.com/symfony/symfony/issues/64760

### DIFF
--- a/config/orm.php
+++ b/config/orm.php
@@ -31,6 +31,7 @@ use Doctrine\ORM\Tools\Console\Command\SchemaTool\DropCommand;
 use Doctrine\ORM\Tools\Console\Command\SchemaTool\UpdateCommand;
 use Doctrine\ORM\Tools\Console\Command\ValidateSchemaCommand;
 use Doctrine\ORM\Tools\ResolveTargetEntityListener;
+use Symfony\Bridge\Doctrine\ArgumentResolver\Console\EntityValueResolver as ConsoleEntityValueResolver;
 use Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver;
 use Symfony\Bridge\Doctrine\Form\DoctrineOrmTypeGuesser;
 use Symfony\Bridge\Doctrine\Form\Type\EntityType;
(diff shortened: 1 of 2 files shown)


xabbuh:
see #2193

W0rma:
> see #2193

Thank you for the heads up!

PR #2262, opened 2026-07-23
Twig 4 compatibility

### LINKED ISSUE #1879: Twig 4 compatilibity
Is there any downside to adding a return type of array to getFilters()?

```
Script cache:clear returned with error code 255
!!  PHP Fatal error:  Declaration of Doctrine\Bundle\DoctrineBundle\Twig\DoctrineExtension::getFilters() must be compatible with Twig\Extension\AbstractExtension::getFilters(): array in /home/tac/sites/pokemon/vendor/doctrine/doctrine-bundle/src/Twig/DoctrineExtension.php on line 44
!!  Symfony\Component\ErrorHandler\Error\FatalError {#5861
!!    #message: "Compile Error: Declaration of Doctrine\Bundle\DoctrineBundle\Twig\DoctrineExtension::getFilters() must be compatible with Twig\Extension\AbstractExtension::getFilters(): array"
!!    #code: 0
!!    #file: "./vendor/doctrine/doctrine-bundle/src/Twig/DoctrineExtension.php"
!!    #line: 44
!!    -error: array:4 [
!!      "type" => 64
!!      "message" => "Declaration of Doctrine\Bundle\DoctrineBundle\Twig\DoctrineExtension::getFilters() must be compatible with Twig\Extension\AbstractExtension::getFilters(): array"
!!      "file" => "/home/tac/sites/pokemon/vendor/doctrine/doctrine-bundle/src/Twig/DoctrineExtension.php"
!!      "line" => 44
!!    ]
!!  }
!!  
```

/vendor/doctrine/doctrine-bundle/src/Twig/DoctrineExtension.php

Or is this something that will need to wait for a point release?

derrabus:
Symfony started to stabilize Twig 4 which enables us to test against it. Our 3.x series is already compatible, but on 2.x we're missing one native return type on `DoctrineExtension::getFilters()`. Since we don't block the istallation of Twig 4, people will get a runtime error as soon as they try to use Twig 4 with DoctrineBundle 2.

Adding this return type would be a breaking change, which is why I'm adding it conditionally here: The method only has a return type if the parent has it. When merging up to 3.x, we can revert all of this.

Fixes #1879

### DIFF
--- a/src/Twig/DoctrineExtension.php
+++ b/src/Twig/DoctrineExtension.php
@@ -40,6 +40,8 @@ use function substr;
  */
 class DoctrineExtension extends AbstractExtension
 {
+    use AbstractExtensionCompatibility;
+
     private SqlFormatter $sqlFormatter;
 
     /**
@@ -47,7 +49,7 @@ class DoctrineExtension extends AbstractExtension
      *
      * @return TwigFilter[]
      */
-    public function getFilters()
+    private function doGetFilters(): array
     {
         $out     = [
             new TwigFilter('doctrine_prettify_sql', [$this, 'prettifySql'], ['is_safe' => ['html']]),
--- a/composer.json
+++ b/composer.json
@@ -68,7 +68,7 @@
         "symfony/var-exporter": "^6.4.1 || ^7.0.1",
         "symfony/web-profiler-bundle": "^6.4 || ^7.0",
         "symfony/yaml": "^6.4 || ^7.0",
-        "twig/twig": "^2.14.7 || ^3.0.4"
+        "twig/twig": "^2.14.7 || ^3.0.4 || ^4"
     },
     "conflict": {
         "doctrine/annotations": ">=3.0",
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -28,6 +28,10 @@
         <exclude-pattern>tests/*</exclude-pattern>
         <exclude-pattern>src/Repository/RepositoryFactoryCompatibility.php</exclude-pattern>
         <exclude-pattern>src/Repository/ServiceEntityRepository.php</exclude-pattern>
+        <exclude-pattern>src/Twig/AbstractExtensionCompatibility.php</exclude-pattern>
+    </rule>
+    <rule ref="SlevomatCodingStandard.TypeHints.ReturnTypeHint">
+        <exclude-pattern>src/Twig/AbstractExtensionCompatibility.php</exclude-pattern>
     </rule>
     <rule ref="Squiz.Classes.ClassFileName.NoMatch">
         <exclude-pattern>tests/*</exclude-pattern>
--- a/src/Twig/AbstractExtensionCompatibility.php
+++ b/src/Twig/AbstractExtensionCompatibility.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Bundle\DoctrineBundle\Twig;
+
+use ReflectionMethod;
+use Twig\Extension\AbstractExtension;
+
+if ((new ReflectionMethod(AbstractExtension::class, 'getFilters'))->hasReturnType()) {
+    /** @internal */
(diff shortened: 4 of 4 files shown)


ostrolucky:
On 2.18.x we should just block twig >= 4. This kind of change is not appropriate for this maintenance only branch, we don't add support for new major versions of libraries in there.

derrabus:
> On 2.18.x we should just block twig >= 4.

If we do this, we would only make compose pick an old 2.x release without a Twig 4 conflict rule. This won't achieve anything.

ostrolucky:
What it will is, that it will block twig 4 upgrade for people who upgrade to latest 2.18.x first. That's good enough and standard practice.

derrabus:
> What it will is, that it will block twig 4 upgrade for people who upgrade to latest 2.18.x first.

Yes, and it will block Twig 4 for projects that are on a Symfony LTS on PHP 8.2/8.3 while Symfony is actively pushing Twig 4 support to Symfony 6.4 and up. I think, we should align with Symfony's strategy here and not work against it.

nicolas-grekas (approved):
This is an easy patch with a high impact. Both alternatives (doing nothing or blocking) are going to create needless friction.

src/Twig/DoctrineExtension.php
```diff
@@ -40,14 +40,16 @@
...
+    use AbstractExtensionCompatibility;
+
     private SqlFormatter $sqlFormatter;
 
     /**
      * Define our functions
      *
      * @return TwigFilter[]
      */
-    public function getFilters()
```
ostrolucky, line 50: Why is it simply not 
```diff
- public function getFilters()
+ public function getFilters(): array
```
?

derrabus: Sure, if we're all fine with a small BC break on a class that nobody should extend anyway?

ostrolucky: Well that BC break would be caused by that trait under Twig 4 anyways. It's internal class so let's just do it this easier way.

derrabus: > Well that BC break would be caused by that trait under Twig 4 anyways.

Yes, but then it's technically Twig 4 that breaks BC. But I'm fine either way. Shall I target 2.19.x, given that's a documented BC break? We have a change sitting on that branch anyway, waiting for a release for quite some time. 🫣

stof: @ostrolucky `@internal since 2.11` (vs `@internal`) means we still need to preserve BC until the 3.0 version as projects that were using 2.10 might use it in ways where they expect BC.

> Shall I target 2.19.x, given that's a documented BC break?

Shipping this in a minor version rather than a patch version would make sense.

derrabus: see #2265

ostrolucky: As far as I'm aware, 2.18 is a last 2.x branch, same as 6.4 is a last symfony 6.x branch. There won't be 2.19. Either this is shipped as a bugfix, or only in 3.x

stof: than, why do we have a 2.19.x branch in this repo containing unreleased changes ?

I suggest we do the 2.19.0 release (and then avoid creating the 2.20.x branch if we don't want a new minor version in 2.x)

derrabus: The only other change on 2.19 is a deprecation regarding legacy proxy implementations. The purpose of that deprecation was to ease the upgrade to 3.x. And deprecations are usually not shipped in a bugfix release. So having a 2.19 is fine, as far as I can tell.

ostrolucky: IIRC there is a 2.19.x branch only because it's automatically created by a release bot. We should just delete it.

ostrolucky: But fine let's do 2.19.x and that's it for 2.x

derrabus:
Closing in favor of #2265.

PR #2276, opened 2026-08-14
Register the DBAL messenger middlewares from symfony/doctrine-bridge 8.2

nicolas-grekas:
symfony/symfony#65302 and symfony/symfony#65329 deprecate `DoctrineCloseConnectionMiddleware` and `DoctrineOpenTransactionLoggerMiddleware` in favor of new middlewares that target DBAL connections instead of entity managers, and symfony/symfony#65346 completes the set: it deprecates `DoctrinePingConnectionMiddleware` and adds `DoctrineDbalTransactionMiddleware` as the DBAL-level alternative to `DoctrineTransactionMiddleware` (which is not deprecated, since its flush has no DBAL equivalent).

This PR registers the four new middlewares so they can be referenced from messenger bus configuration:

- `doctrine_dbal_transaction`
- `doctrine_dbal_ping_connection`
- `doctrine_dbal_close_connection`
- `doctrine_dbal_open_transaction_logger`

All take the `doctrine` registry as first argument (`ManagerRegistry` extends `ConnectionRegistry`), and the logger one takes the `logger` service second, matching the new constructors where connection names come last; user-provided middleware arguments are appended after these by `ResolveChildDefinitionsPass`, so no argument remapping is needed. The definitions are removed when the classes are not available, i.e. with symfony/doctrine-bridge < 8.2, so this is a no-op until then. The PHPStan ignore entry uses `reportUnmatched: false` so it becomes inert once the bridge classes are visible to the analysis.

### DIFF
--- a/src/DependencyInjection/DoctrineExtension.php
+++ b/src/DependencyInjection/DoctrineExtension.php
@@ -48,6 +48,10 @@ use Symfony\Bridge\Doctrine\ArgumentResolver\Console\EntityValueResolver;
 use Symfony\Bridge\Doctrine\Attribute\MapEntity;
 use Symfony\Bridge\Doctrine\IdGenerator\UlidGenerator;
 use Symfony\Bridge\Doctrine\IdGenerator\UuidGenerator;
+use Symfony\Bridge\Doctrine\Messenger\DoctrineDbalCloseConnectionMiddleware;
+use Symfony\Bridge\Doctrine\Messenger\DoctrineDbalOpenTransactionLoggerMiddleware;
+use Symfony\Bridge\Doctrine\Messenger\DoctrineDbalPingConnectionMiddleware;
+use Symfony\Bridge\Doctrine\Messenger\DoctrineDbalTransactionMiddleware;
 use Symfony\Bridge\Doctrine\Middleware\IdleConnection\Listener;
 use Symfony\Bridge\Doctrine\PropertyInfo\DoctrineExtractor;
 use Symfony\Bridge\Doctrine\Validator\DoctrineLoader;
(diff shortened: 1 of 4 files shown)


MatTheCat:
I won’t be needing this. Ping @rey0bs

src/DependencyInjection/DoctrineExtension.php
```diff
@@ -1452,6 +1456,22 @@ private function loadMessengerServices(ContainerBuilder $container): void
...
+            'messenger.middleware.doctrine_dbal_open_transaction_logger' => DoctrineDbalOpenTransactionLoggerMiddleware::class,
+        ];
+
+        foreach ($dbalMiddlewares as $middlewareId => $middlewareClass) {
+            if (class_exists($middlewareClass)) {
+                continue;
+            }
+
+            $container->removeDefinition($middlewareId);
+        }
```
GromNaN, line 1473: Maybe read the class from the definition to avoid redundancy:
```suggestion
        // The DBAL middlewares exist since symfony/doctrine-bridge 8.2
        $dbalMiddlewares = [
            'messenger.middleware.doctrine_dbal_transaction',
            'messenger.middleware.doctrine_dbal_ping_connection',
            'messenger.middleware.doctrine_dbal_close_connection',
            'messenger.middleware.doctrine_dbal_open_transaction_logger',
        ];

        foreach ($dbalMiddlewares as $middlewareId) {
            if (class_exists($container->getDefinition($middlewareId)->getClass()) {
                continue;
            }

            $container->removeDefinition($middlewareId);
        }
```

tests/DependencyInjection/DoctrineExtensionTest.php
```diff
@@ -789,6 +793,21 @@ public function testMessengerIntegration(): void
...
+
+        $dbalMiddlewares = [
+            'messenger.middleware.doctrine_dbal_transaction' => [DoctrineDbalTransactionMiddleware::class, 1],
+            'messenger.middleware.doctrine_dbal_ping_connection' => [DoctrineDbalPingConnectionMiddleware::class, 1],
+            'messenger.middleware.doctrine_dbal_close_connection' => [DoctrineDbalCloseConnectionMiddleware::class, 1],
+            'messenger.middleware.doctrine_dbal_open_transaction_logger' => [DoctrineDbalOpenTransactionLoggerMiddleware::class, 2],
+        ];
+
+        foreach ($dbalMiddlewares as $middlewareId => [$middlewareClass, $argumentCount]) {
+            if (class_exists($middlewareClass)) {
```
GromNaN, line 805: The test should use an other way to detect the feature: `Composer\Installed` to check the `symfony/doctrine-bridge` version?

GromNaN (approved):
LGTM, with some suggestions.

Issue #2124, opened 2025-10-27
Auto-discover entities based on attribute


Nayte91:
#### What
Automatically detect and register entity mappings by scanning for classes with the `#[Entity]` attribute when no explicit mappings are configured, eliminating the need for repetitive manual configuration.

#### Why
Currently, every entity mapping must be explicitly declared in `doctrine.yaml`:

```yaml
doctrine:
    orm:
        mappings:
            User:
                dir: '%kernel.project_dir%/src/User/Entity'
                prefix: 'App\User\Entity'
            Event:
                dir: '%kernel.project_dir%/src/Event/Entity'
                prefix: 'App\Event\Entity'
            # ... repeated for each domain
```
This becomes particularly tedious in DDD-structured projects with multiple bounded contexts, where entities are distributed across many domain folders. Going for a auto discover offers multiple benefits:
- Drastically reduces boilerplate configuration
- Aligns with "convention over configuration" philosophy
- Complements Symfony 7.3's auto-exclusion of entities from service container (PR [#59987](https://github.com/symfony/symfony/pull/59987))
- Complements Symfony 7.4's controllers auto-registering (PR [#61492](https://github.com/symfony/symfony/pull/61492))
- Provides DX similar to other ORMs (Laravel Eloquent, TypeORM, Hibernate)
- Fully backwards compatible with below option.

#### How
The implementation could have zero configuration. For BC reason, it would add a new configuration option:
```yaml
doctrine:
    orm:
        auto_discover_entities: true //true by default on next major version
```

I suppose that the approach would be:
1. Detection by somehow reading composer PSR4 config,
2. Discovery at compile time by scanning recursively folders,
3. Automatically register discovered mappings with each namespace,
4. Caching results in compiled container to avoid runtime overhead (Only rescanned when container is rebuilt)

I guess main implementation would be in `DoctrineExtension::load()` method, similar to how the legacy `auto_mapping` feature worked?

Nayte91:
## Progress update

Remaining work, by dependency:
- [x] symfony/framework-bundle: `doctrine.orm.entity` tag → symfony/symfony#65437,
- [x] symfony/doctrine-bridge: class-list driver → symfony/symfony#65437,
[...]

PR #2277, opened 2026-08-19
Add `auto_discover_entities` option to register entities discovered by the container scan

Nayte91:
Implements the proposal from #2124. 

Companion of symfony/symfony#65437 (merged in 8.2), which tags #[Entity]/#[MappedSuperclass] classes with doctrine.orm.entity during the container scan.

This PR adds a per-entity-manager option:

```yaml
doctrine:
    orm:
        auto_discover_entities: true
```

When enabled, `DoctrineExtension` registers a persistence `ClassNames` locator feeding an ORM `AttributeDriver`, set as the entity manager's `MappingDriverChain` default driver; the new `EntityAutoDiscoveryPass` (`TYPE_BEFORE_REMOVING`) fills the locator with the tagged class names via `ContainerBuilder::findTaggedResourceIds('doctrine.orm.entity', false)` — the same pattern Symfony's `MessengerPass` uses. Class names are resolved through the definitions, since PHP-abstract scanned classes (typically mapped superclasses) are registered under `.abstract.`-prefixed ids.

For most applications this removes the need for the explicit `mappings` configuration block entirely.

Design notes:
- **Explicit mappings still win**: prefixed chain drivers are consulted first; the discovery driver is only the chain's default. Both can coexist on the same entity manager (`getAllClassNames()` de-duplicates by design).
- **Multi-EM semantics**: every entity manager with the flag enabled receives the full discovered class list; partitioning entities across managers still requires explicit mappings — the option targets the single-EM majority case.
- **Legacy exclude globs**: entities behind an old `exclude: '../src/Entity'` line in `services.yaml` are not discovered (those stubs never receive the tag) — documented with a prominent warning in the configuration reference, plus a test pinning the limitation.
- **No dependency bumps**: enabling the option throws a `LogicException` on Symfony < 8.2, and when `ClassNames` (doctrine/persistence 4.2) or `AttributeDriver ClassLocator` support (doctrine/orm 3.6) is missing. Uses the "cleaner home" noted in symfony/symfony#65437's review; the transitional bridge driver was removed upstream in symfony/symfony#65445 (merged).
- **Deprecation**: not setting the option is deprecated (same approach as #1762 for `controller_resolver.auto_mapping`), so it can default to `true` in a future major version. See `UPGRADE-3.4.md`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
         "ext-mbstring": "*",
         "doctrine/dbal": "^4.0",
         "doctrine/deprecations": "^1.0",
-        "doctrine/persistence": "^4",
+        "doctrine/persistence": "^4.2",
         "doctrine/sql-formatter": "^1.0.1",
         "symfony/cache": "^6.4 || ^7.0 || ^8.0",
         "symfony/config": "^6.4 || ^7.0 || ^8.0",
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -573,6 +573,7 @@ final class Configuration implements ConfigurationInterface
                     ->scalarNode('class_metadata_factory_name')->defaultValue(ClassMetadataFactory::class)->end()
                     ->scalarNode('default_repository_class')->defaultValue(EntityRepository::class)->end()
                     ->scalarNode('auto_mapping')->defaultFalse()->end()
+                    ->booleanNode('auto_discover_entities')->defaultNull()->info('Registers all entity classes discovered by the container scan, removing the need for explicit "mappings". When not set, enabled if neither "mappings" nor "auto_mapping" is configured and Symfony 8.2 and doctrine/orm 3.6 are available.')->end()
                     ->scalarNode('naming_strategy')->defaultValue('doctrine.orm.naming_strategy.default')->end()
                     ->scalarNode('quote_strategy')->defaultValue('doctrine.orm.quote_strategy.default')->end()
                     ->scalarNode('typed_field_mapper')->defaultValue('doctrine.orm.typed_field_mapper.default')->end()
--- a/config/schema/doctrine-1.0.xsd
+++ b/config/schema/doctrine-1.0.xsd
@@ -215,6 +215,7 @@
     </xsd:group>
 
     <xsd:attributeGroup name="entity-manager-config">
+        <xsd:attribute name="auto-discover-entities" type="xsd:boolean" />
         <xsd:attribute name="auto-mapping" type="xsd:string" />
         <xsd:attribute name="connection" type="xsd:string" />
         <xsd:attribute name="default-repository-class" type="xsd:string" />
--- a/docs/en/configuration.rst
+++ b/docs/en/configuration.rst
@@ -246,6 +246,11 @@ Configuration Reference
                         class_metadata_factory_name:  Doctrine\ORM\Mapping\ClassMetadataFactory
                         default_repository_class:     Doctrine\ORM\EntityRepository
                         auto_mapping:                 false
+                        # Registers all classes tagged with "doctrine.orm.entity" by the container scan,
+                        # removing the need for explicit "mappings". When not set, enabled if neither
+                        # "mappings" nor "auto_mapping" is configured and Symfony 8.2 and doctrine/orm 3.6
+                        # are available.
+                        auto_discover_entities:       ~
                         # 0pt-in to the new mapping driver mode as of Doctrine ORM 2.14. See https://github.com/doctrine/orm/pull/6728.
                         validate_xml_mapping: false
                         naming_strategy:              doctrine.orm.naming_strategy.default
@@ -704,6 +709,34 @@ The following example shows an overview of the caching configurations:
             pools:
                 doctrine.result_cache_pool:
                     adapter: cache.app
+
+Entity Auto-Discovery
+~~~~~~~~~~~~~~~~~~~~~
+
+Setting ``auto_discover_entities: true`` on an entity manager (requires Symfony
+8.2) registers every entity class discovered by the service container scan:
+classes carrying the ``#[Entity]`` or ``#[MappedSuperclass]`` attribute are
+tagged with ``doctrine.orm.entity`` while ``src/`` is scanned, removing the
+need for an explicit ``mappings`` block. Explicit mappings keep priority — the
+discovered classes only feed the metadata driver chain's default driver.
+
+Discovery relies on ``ContainerBuilder::findTaggedResourceIds()`` (Symfony
+8.2) and on the ORM's ``AttributeDriver`` accepting a ``ClassLocator`` (doctrine/orm
+3.6). When the option is not set, discovery is enabled automatically for entity
+managers that configure neither ``mappings`` nor ``auto_mapping``, provided both
+requirements are met. Set it to ``false`` to opt out, or to ``true`` to require
+it: a ``LogicException`` naming the missing requirement is then thrown.
+
+.. warning::
+
+    Only classes seen by the container scan are discovered. If your
+    ``services.yaml`` still excludes the entity directory (the legacy
+    ``exclude: '../src/{Entity,...}'`` line from older skeletons), those
+    classes are registered from their file path without attribute
+    autoconfiguration and will **not** be discovered. Remove the entity
+    directory from the ``exclude`` list: since Symfony 7.3, entities are
+    automatically excluded from the service container without being hidden
+    from discovery.
 
 Mapping Configuration
 ~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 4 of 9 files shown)


nicolas-grekas (changes requested):
Two findings after verifying the DI mechanics against symfony/symfony#65437 (now merged):

1. `EntityAutoDiscoveryPass` feeds service ids into the driver, but abstract scanned classes are registered under a `.abstract.`-prefixed id. A `#[MappedSuperclass]` is typically abstract, so the driver ends up with `.abstract.App\Entity\Foo` in its class list and metadata loading fails. Map each id through `$container->getDefinition($id)->getClass()` instead, like `MessengerPass` does for `messenger.message`.

2. Entities excluded by a legacy `exclude: '../src/Entity'` glob are not discovered: those stubs are registered from the file path by `FileLoader::addContainerExcludedTag()`, without reflection or autoconfiguration, so they never get the `doctrine.orm.entity` tag. The abstract stub in `EntityAutoDiscoveryPassTest` describes a state the container never produces; the real producer of abstract stubs is the scan itself. Passing `false` to `findTaggedResourceIds()` is still needed, but for abstract scanned classes. Apps that kept the old exclude line would silently discover zero entities, so this needs a loud failure or prominent docs.

Nayte91 (commented):
Thank you for review, commit amended and force-pushed according to your findings:
1. Class names are now resolved through `$definition->getClass()` with MessengerPass.
2. You're right about legacy exclude: I went with prominent docs (new "Entity Auto-Discovery" section with a warning) + a test.

Nayte91:
3. Following the "cleaner home" comment in symfony/symfony#65437, commit amended to drop the bridge's transitional `AttributeClassesDriver`:
    - minimums dependencies bumped to `doctrine/orm ^3.6` / `doctrine/persistence ^4.2`;
    - guard now throws a `LogicException` on `Symfony < 8.2` (previously a silent no-op);
    - behavior, docs and the review-driven tests are unchanged.

The now-unused bridge driver has been removed upstream (symfony/symfony#65445, merged); this PR's only Symfony-side dependency is the doctrine.orm.entity tag, already in 8.2.

Nayte91:
Previous comments was meant to arrive after the review request, but I messed up :x

src/DependencyInjection/Configuration.php
```diff
@@ -573,6 +573,7 @@ private function getOrmEntityManagersNode(): ArrayNodeDefinition
                     ->scalarNode('class_metadata_factory_name')->defaultValue(ClassMetadataFactory::class)->end()
                     ->scalarNode('default_repository_class')->defaultValue(EntityRepository::class)->end()
                     ->scalarNode('auto_mapping')->defaultFalse()->end()
+                    ->booleanNode('auto_discover_entities')->defaultFalse()->info('Set to "true" to register all entity classes discovered by the container scan, removing the need for explicit "mappings". Requires Symfony 8.2.')->end()
```
ostrolucky: Let's deprecate not setting it. I agree we should eventually default it to true, although I guess that happens in 5.x (4.x can deprecate setting to to false). Don't forget updating UPGRADE file and creating a PR for 4.x)

composer.json
```diff
@@ -45,7 +45,7 @@
     },
     "require-dev": {
         "doctrine/coding-standard": "^14",
-        "doctrine/orm": "^3.4.4",
+        "doctrine/orm": "^3.6",
```
ostrolucky: test suite should pass without this change. having to change this indicates chages are not BC

composer.json
```diff
@@ -33,7 +33,7 @@
         "ext-mbstring": "*",
         "doctrine/dbal": "^4.0",
         "doctrine/deprecations": "^1.0",
-        "doctrine/persistence": "^4",
+        "doctrine/persistence": "^4.2",
```
ostrolucky, line 36: why?

ostrolucky (changes requested):
none of the php code comments are useful enough to warrant their presence

src/DependencyInjection/Configuration.php
nicolas-grekas: what about defaulting to true immediately? what would be the drawback?

ostrolucky: same as in https://github.com/doctrine/DoctrineBundle/issues/2278#issuecomment-5515313962

composer.json
Nayte91: The wiring relied on ClassNames (persistence 4.2) and on AttributeDriver accepting a ClassLocator (ORM 3.6), so I raised. 
Agreed it doesn't belong in a minor: reverted.

Nayte91:
@ostrolucky I took all your comments in consideration, amended and force pushed.
- No comments other than phpstan's needs,
- No dependencies bump,
- Default null config.

nicolas-grekas:
Moving this out of the inline thread for visibility.

Context: I asked why not default `auto_discover_entities` to `true` right away, and @ostrolucky pointed at https://github.com/doctrine/DoctrineBundle/issues/2278#issuecomment-5515313962, i.e. we should not ship things users on Symfony < 8.2 cannot act on.

I agree with that constraint, and I think it also applies to the deprecation this PR currently adds. The bundle allows `^6.4 || ^7.0 || ^8.0`, so most users would get a deprecation whose only reachable resolution is writing `auto_discover_entities: false` into their config. The value they would actually want is unavailable on their stack, so we would be trading a deprecation notice for an opt-out line that everyone carries forever.

And an unconditional `true` really is not possible in a minor: `MappingDriverChain::getAllClassNames()` unions the default driver's classes on top of the prefixed drivers, so every entity manager with the flag on would receive every discovered entity. Multi-EM setups that partition by prefix would collapse into "everything everywhere", with `doctrine:schema:update --em=X` and migration diffs proposing every table on every connection. That is 4.0 material.

So here is a middle ground for 3.4. Keep `defaultNull()`, but give *unset* a useful meaning instead of "deprecated, treated as false":

- unset, the entity manager configures neither `mappings` nor `auto_mapping`, and symfony/dependency-injection >= 8.2 + doctrine/persistence >= 4.2 + doctrine/orm >= 3.6 are installed: register the discovery driver.
- unset, anything else: off, bit for bit today's behavior.
- explicit `true`: register it, and keep the current `LogicException` when the version floor is not met, since the user asked for it explicitly.
- explicit `false`: off.

Why this is safe: an entity manager that configures no `mappings` and no `auto_mapping` gets zero drivers today (`registerMappingDrivers()` only iterates over what `loadMappingInformation()` collected), so its chain maps nothing at all. Discovery can only add there, it cannot change an existing mapping. Multi-EM setups always configure `mappings`, so they are excluded by construction.

Why it is still useful rather than a no-op: `DoctrineOrmMappingsPass` adds its prefixed drivers straight to the chain in a compiler pass and never goes through the `mappings` config key. An app whose bundles self-register their mappings therefore keeps an empty `mappings` key and would get discovery for its own entities on top of the bundle drivers. That is exactly the combination this feature is after.

Net result: no deprecation, no UPGRADE entry, no `auto_discover_entities: false` boilerplate, and nothing for Symfony 6.4/7.x users to do. The recipe change listed in #2278 (dropping the `App` mappings block) is what turns it on for new applications, and 4.0 can then make it unconditional and decide what multi-EM should mean.

Nayte91:
Thank you @nicolas-grekas for the detailed review, amended and force-pushed:

- `defaultNull()` kept. Unset = discovery on, only if the EM has no `mappings`, no `auto_mapping`, and DI 8.2 + persistence 4.2 + ORM 3.6 are installed. Otherwise: off, as today,
- Explicit `true` still throws `LogicException` when the floor isn't met, explicit `false` = off,
- No more deprecation so UPGRADE file removed,
- covered by tests.

Note for when this leaves draft: with doctrine/DoctrineBundle#2277, explicit `auto_discover_entities: true` throws a `LogicException` below doctrine/persistence 4.2 / doctrine/orm 3.6. Adding both to the manifest's `conflict` (next to symfony/framework-bundle <8.2) would make Flex fall back to the 3.0 recipe on such projects instead. Two lines if you want.

ostrolucky:
alright. but only orm goes to conflict, persistence can be bumped again in require section. please document as well what logicexception is it throwing: what exactly from persistence 4.2 and orm 3.6 does it need?

Nayte91:
@ostrolucky done, amended and force-pushed:
- doctrine/persistence ^4.2 back in require (because brings `ClassNames`),
- ORM stays optional: `AttributeDriver` accepting a `ClassLocator` (3.6) is checked at compile time,
- The `LogicException` now names what's missing: symfony/dependency-injection 8.2 (`ContainerBuilder::findTaggedResourceIds()`) or doctrine/orm 3.6 (`ClassLocator`). Same in the docs and the option's `info()`.
- Recipe side: only doctrine/orm <3.6 goes to conflict.

PR #2279, opened 2026-08-20
Register the `EntityExists` constraint validator from symfony/doctrine-bridge 8.2

nicolas-grekas:
symfony/symfony#63483 added an `EntityExists` constraint to the Doctrine bridge (8.2). Its validator needs the `ManagerRegistry`, so like `UniqueEntityValidator` it has to be registered as a `validator.constraint_validator` service by this bundle; without it the constraint fails at `ConstraintValidatorFactory` on any full-stack app.

The tag carries no alias: `EntityExists::validatedBy()` resolves to the validator class name, which is what an alias-less tag indexes by.

The definition is removed when the class does not exist (bridge < 8.2), following the `UlidGenerator`/`UuidGenerator` pattern, and the container test asserts whichever side applies. Ran locally against `symfony/doctrine-bridge` 8.2.x-dev: the service instantiates from the real container; phpcs clean.


ostrolucky (changes requested):
please add comments to guards explaining in which symfony version were the classes added: important for clean ups in future when removing support for old symfony versions

PR #2282, opened 2026-09-08
Set default lock mode in find method

### LINKED ISSUE #2281: Deprecation on EntityManager::find(), use LockMode::NONE
The following deprecation is being reported when upgrading to Doctrine/ORM v3.7.0:
```
User Deprecated: Passing null as lock mode to Doctrine\ORM\EntityManager::find() is deprecated and
will not be possible in Doctrine ORM 4.0, pass LockMode::NONE instead. (EntityManager.php:279 called
by EntityRepository.php:87, https://github.com/doctrine/orm/pull/12548, package doctrine/orm)
```

This is being mentioned on the upgrade document: https://github.com/doctrine/orm/blob/3.7.x/UPGRADE.md#deprecated-passing-null-as-lock-mode

> ## Deprecated passing null as lock mode
> Since LockMode::NONE is now a no-op, passing null as lock mode to EntityManager::find(), EntityManager::refresh() or EntityRepository::find() is deprecated as it is equivalent to passing LockMode::NONE. Omit the argument or pass LockMode::NONE instead.

Should we check the current ORM version here: https://github.com/doctrine/DoctrineBundle/blob/6004c18e61f7755f1d3722a9b584bb4afb55e80b/src/Repository/ServiceEntityRepository.php#L66 and pass LockMode::NONE in case it is null? WDYT?

ruudk:
Fixes #2281

### DIFF
--- a/src/Repository/ServiceEntityRepositoryProxy.php
+++ b/src/Repository/ServiceEntityRepositoryProxy.php
@@ -54,11 +54,11 @@ class ServiceEntityRepositoryProxy extends EntityRepository implements ServiceEn
             ->createResultSetMappingBuilder($alias);
     }
 
-    public function find(mixed $id, LockMode|int|null $lockMode = null, int|null $lockVersion = null): object|null
+    public function find(mixed $id, LockMode|int|null $lockMode = LockMode::NONE, int|null $lockVersion = null): object|null
     {
         /** @psalm-suppress InvalidReturnStatement This proxy is used only in combination with newer parent class */
         return ($this->repository ??= $this->resolveRepository())
-            ->find($id, $lockMode, $lockVersion);
+            ->find($id, $lockMode ?? LockMode::NONE, $lockVersion);
     }
 
     /**


src/Repository/ServiceEntityRepository.php
```diff
@@ -59,11 +59,11 @@ public function createResultSetMappingBuilder(string $alias): ResultSetMappingBu
             ->createResultSetMappingBuilder($alias);
     }
 
-    public function find(mixed $id, LockMode|int|null $lockMode = null, int|null $lockVersion = null): object|null
+    public function find(mixed $id, LockMode|int|null $lockMode = LockMode::None, int|null $lockVersion = null): object|null
```
ruudk: I think we should trigger a deprecation when passing null, can we do that here too?

ruudk:
Then in the next major we can drop the nullability completely.

src/Repository/ServiceEntityRepository.php
greg0ire: Not in a patch release I'd say, but when this is merged up, I guess it's fine.

greg0ire:
```suggestion
    public function find(mixed $id, LockMode|int|null $lockMode = LockMode::NONE, int|null $lockVersion = null): object|null
```

ruudk: Once this is merged, I'll prep a PR for a deprecation in 3.4.x, and then a removal in 4.0

src/Repository/ServiceEntityRepository.php
```diff
@@ -59,11 +59,11 @@ public function createResultSetMappingBuilder(string $alias): ResultSetMappingBu
             ->createResultSetMappingBuilder($alias);
     }
 
-    public function find(mixed $id, LockMode|int|null $lockMode = null, int|null $lockVersion = null): object|null
+    public function find(mixed $id, LockMode|int|null $lockMode = LockMode::NONE, int|null $lockVersion = null): object|null
     {
         /** @psalm-suppress InvalidReturnStatement This proxy is used only in combination with newer parent class */
         return ($this->repository ??= $this->resolveRepository())
-            ->find($id, $lockMode, $lockVersion);
+            ->find($id, $lockMode ?? LockMode::NONE, $lockVersion);
```
ruudk: Should we bump the minimum doctrine/orm too?

greg0ire:
Should this target 2.x?

ruudk:
> Should this target 2.x?

You tell me. Either way, we should bump the minimum required orm also with it.

greg0ire:
You should target 2.x if the deprecation can happen there, then we will merge up. So my real question is: can the deprecation happen there. Bumping the minimum required version in a patch release sounds bad. Alternatively, if this is risky, we can mute the deprecation in a patch release and address it in a minor release.

greg0ire (approved):

dmaicher (approved):

greg0ire:
Thanks @ruudk , I probably won't have time to create the merge up PR until tomorrow, and I don't know how hard it will be. Feel free to beat me to it.

ruudk:
@greg0ire I'm a bit confused now what to do next. Should we deprecate passing null on 3.x and then remove the nullability on 4.x?

stof:
@ruudk yes

greg0ire:
On 3.4.x, yes

PR #2284, opened 2026-09-09
Deprecate passing null as lock mode to ServiceEntityRepository::find()

ruudk:
Follow-up to #2282, as discussed in https://github.com/doctrine/DoctrineBundle/pull/2282#issuecomment-5601843269.

ORM 3.7 [deprecated passing `null` as lock mode](https://github.com/doctrine/orm/pull/12548), and ORM 4.0 will not accept it anymore. #2282 made `ServiceEntityRepository::find()` default to `LockMode::NONE` and normalize an explicitly passed `null`, so that users stop getting an ORM deprecation they cannot act on.

Users who pass `null` themselves still need to hear about it, since the parameter loses its nullability in DoctrineBundle 4.0. This triggers our own deprecation for that case.

Note that the nullability cannot be dropped on `4.0.x` yet: as long as DoctrineBundle 4.0 requires ORM 3 (`"doctrine/orm": "<3.0 || >=4.0"` in `conflict`), the parent `Doctrine\ORM\EntityRepository::find()` still declares `LockMode|int|null $lockMode = null`, and narrowing that in a child class is a fatal error. The removal has to happen in the same change that requires ORM 4.

### DIFF
--- a/UPGRADE-3.4.md
+++ b/UPGRADE-3.4.md
@@ -1,0 +1,14 @@
+UPGRADE FROM 3.3 to 3.4
+=======================
+
+ServiceEntityRepository
+-----------------------
+
+Passing `null` as `$lockMode` to `ServiceEntityRepository::find()` is deprecated
+and will not be possible in DoctrineBundle 4.0. Omit the argument or pass
+`LockMode::NONE` instead.
+
+```diff
+-$repository->find($id, null);
++$repository->find($id);
+```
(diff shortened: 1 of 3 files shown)


greg0ire:
Nice, but let me do https://github.com/doctrine/DoctrineBundle/pull/2285 first, and then you rebase.

greg0ire:
Done, please rebase.

ruudk:
@greg0ire done

greg0ire:
It seems there are CI jobs failing. Please take a look at [this guide](https://www.doctrine-project.org/contribute/#dealing-with-checks-and-tools) for more on how to handle those.

